PR #87, opened 2019-06-19
Add ability to vary authorizers per/operation

mtdowling:
This commit updates the API Gateway traits to allow Lambda authorizers
to vary per/operation. Authorizers are now only attached to an operation
when the `authorizer` trait is used on a service, resource, or operation.
The OpenApiMapper interface was updated to make this simpler so that you
can now map over the security requirements object of an operation and
rename them from the underlying Smithy authentication scheme (e.g.,
aws.v4) to the custom authorizer security scheme.

Included with this change is an update to the
RemoveUnusedComponents mapper to ensure that any unused security
definitions are stripped from the resulting spec.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddAuthorizers.java
+++ b/aws/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddAuthorizers.java
@@ -15,25 +15,41 @@
 
 package software.amazon.smithy.aws.apigateway.openapi;
 
+import java.util.List;
+import java.util.Map;
 import java.util.logging.Logger;
 import software.amazon.smithy.aws.traits.apigateway.Authorizer;
+import software.amazon.smithy.aws.traits.apigateway.AuthorizerIndex;
+import software.amazon.smithy.aws.traits.apigateway.AuthorizerTrait;
 import software.amazon.smithy.aws.traits.apigateway.AuthorizersTrait;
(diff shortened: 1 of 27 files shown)


aws/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddAuthorizers.java
```diff
@@ -15,75 +15,120 @@
...
-                                    authorizer.getIdentityValidationExpression().map(Node::from))
-                .withOptionalMember("identitySource", authorizer.getIdentitySource().map(Node::from))
-                .withOptionalMember("authorizerResultTtlInSeconds", authorizer.getResultTtlInSeconds().map(Node::from))
-                .build();
+        builder.components(components.build());
+        return builder.build();
+    }
 
-        return builder.putExtension(EXTENSION_NAME, authorizerNode).build();
+    // TODO: should this also allow overrides via configuration properties?
```
kstich: Is this configuration override necessary for having the [`x-amazon-apigateway-authtype` end up as `oauth2`](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authtype.html)?

mtdowling: Yes, and any other potential values that we might need to support

PR #92, opened 2019-06-19
Update root package javadoc to aggregate

kstich:
Since the root Smithy package has no source, this commit overrides
the root package's javadoc task to aggregate the javadocs of each
subproject.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -183,3 +183,18 @@ subprojects {
         }
     }
 }
+
+/*
+ * Javadoc
+ * ====================================================
+ *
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -183,3 +183,20 @@ subprojects {
...
     }
 }
+
+/*
+ * Javadoc
+ * ====================================================
+ *
+ * Build an aggregate javadoc for all subprojects.
+ */
+if (plugins.hasPlugin("java")) {
```
mtdowling: Since this is the root project, we don't need this if statement like we do for subprojects

PR #111, opened 2019-07-08
Code writer sections and interpolation

mtdowling:
This commit is a major rehaul of `CodeWriter` that allows it to be extended, cleans up how blank line trimming works, adds a custom format for interpolating calls to `CodeFormatter#write`, and adds named sections that can be intercepted and extended.

Here are the rendered docs:

Helper class for generating code.

A CodeWriter can be used to write basically any kind of code, including whitespace sensitive and brace-based.

The following example generates some Python code:

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.write("def Foo(str):")
             .indent()
             .write("print str")
     String code = writer.toString();
     

Code interpolation
------------------

The `write(java.lang.Object, java.lang.Object...)`, `openBlock(java.lang.String, java.lang.Object...)`, and `closeBlock(java.lang.String, java.lang.Object...)` methods take a code expression and a variadic list of arguments that are interpolated into the expression. Consider the following call to `write`:

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.write("Hello, $L", "there!");
     String code = writer.toString();
     

In the above example, `$L` is interpolated and replaced with the relative argument `there!`.

A CodeWriter supports three kinds of interpolations: relative, positional, and named. Each of these kinds of interpolations pass a value to a _formatter_.

### Formatters

Formatters are named functions that accept an object as input, accepts a string that contains the current indentation (it can be ignored if not useful), and returns a string as output. The `CodeWriter` registers three built-in formatters:

*   `L`: Outputs a literal value of an `Object` using the following implementation: (1) A null value is formatted as "". (2) An empty `Optional` value is formatted as "". (3) A non-empty `Optional` value is recursively formatted using the value inside of the `Optional`. (3) All other valeus are formatted using the result of calling `String.valueOf(java.lang.Object)`.
*   `S`: Adds double quotes around the result of formatting a value first using the default literal "L" implementation described above and then wrapping the value in an escaped string safe for use in Java according to https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.6. This formatter can be overridden if needed to support other programming languages.

Custom formatters can be registered using `putFormatter(char, java.util.function.BiFunction)`.

### Relative parameters

Placeholders in the form of "$" followed by a formatter name are treated as relative parameters. The first instance of a relative parameter interpolates the first positional argument, the second the second, etc.

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.write("$L $L $L", "a", "b", "c");
     System.out.println(writer.toString());
     // Outputs: "a b c"
     

All relative arguments must be used as part of an expression and relative interpolation cannot be mixed with positional variables.

### Positional parameters

Placeholders in the form of "$" followed by a positive number, followed by a formatter name are treated as positional parameters. The number refers to the 1-based index of the argument to interpolate.

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.write("$1L $2L $3L, $3L $2L $1L", "a", "b", "c");
     System.out.println(writer.toString());
     // Outputs: "a b c c b a"
     

All positional arguments must be used as part of an expression and relative interpolation cannot be mixed with positional variables.

### Named parameters

Named parameters are parameters that take a value from the context of the current state. They take the following form `$:`, where is a string that starts with a lowercase letter, followed by any number of `[A-Za-z0-9_#$.]` characters, and is the name of a formatter.

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.putContext("foo", "a");
     writer.putContext("baz.bar", "b");
     writer.write("$foo:L $baz.bar:L");
     System.out.println(writer.toString());
     // Outputs: "a b"
     

### Escaping interpolation

You can escape the "$" character using two "$$".

    
     CodeWriter writer = new CodeWriter().write("$$L");
     System.out.println(writer.toString());
     // Outputs: "$L"
     

Opening and closing blocks
--------------------------

`CodeWriter` provides a short cut for opening code blocks that that have an opening an closing delimiter (for example, "{" and "}") and that require indentation inside of the delimiters. Calling `openBlock(java.lang.String, java.lang.Object...)` and providing the opening statement will write and format a line followed by indenting one level. Calling `closeBlock(java.lang.String, java.lang.Object...)` will print a formatted statement and dedent.

    
     CodeWriter writer = CodeWriter.createDefault()
     .openBlock("if ($L) {", someValue)
     .write("System.out.println($S);", "Hello!")
     .closeBlock("}");
     

The above example outputs (assuming someValue is equal to "foo"):

    
     if (foo) {
         System.out.println("Hello!");
     }
     

Pushing and popping state
-------------------------

The CodeWriter can maintain a stack of transformation states, including the text used to indent, a prefix to add before each line, the number of times to indent, a map of context values, and whether or not whitespace is trimmed from the end of newlines. State can be pushed onto the stack using `pushState()` which copies the current state. Mutations can then be made to the top-most state of the CodeWriter and do not affect previous states. The previous transformation state of the CodeWriter can later be restored using `popState()`.

The CodeWriter is stateful, and a prefix can be added before each line. This is useful for doing things like create Javadoc strings:

    
     CodeWriter writer = CodeWriter.createDefault();
     writer
             .pushState()
             .write("/**")
             .setNewlinePrefix(" * ")
             .write("This is some docs.")
             .write("And more docs.\n\n\n")
             .write("Foo.")
             .popState()
             .write(" *\/");
The above example outputs:

    
     /**
      * This is some docs.
      * And more docs.
      *
      * Foo.
      *\/
       ^ Minus this escape character

The CodeWriter maintains some global state that is not affected by `pushState()` and `popState()`:

*   The number of successive blank lines to trim.
*   Code formatters registered through `putFormatter(char, java.util.function.BiFunction)`
*   The character used for newlines
*   Whether or not a trailing newline is inserted or removed from the result of converting the `CodeWriter` to a string.

Limiting blank lines
--------------------

Many coding standards recommend limiting the number of successive blank lines. This can be handled automatically by `CodeWriter` by calling `trimBlankLines`. The removal of blank lines is handled when the `CodeWriter` is converted to a string. Lines that consist solely of spaces or tabs are considered blank. If the number of blank lines exceeds the allowed threshold, they are omitted from the result.

Trimming trailing spaces
------------------------

Trailing spaces can be automatically trimmed from each line by calling `trimTrailingSpaces()`.

In the the following example:

    
     CodeWriter writer = CodeWriter.createDefault();
     String result = writer.trimTrailingSpaces().write("hello ").toString();

The value of `result` contains `"hello"`

Extending CodeWriter
--------------------

`CodeWriter` can be extended to add functionality for specific programming languages. For example, Java specific code generator could be implemented that makes it easier to write Javadocs.

     class JavaCodeWriter extends CodeWriter {
         public JavaCodeWriter javadoc(Runnable runnable) {
             pushState()
             write("/**")
             setNewlinePrefix(" * ")
             runnable.run();
             popState()
             write(" *\/");
             return this;
         }
     }

     JavaCodeWriter writer = new JavaCodeWriter();
     writer.javadoc(() -> {
         writer.write("This is an example.");
     });

Code sections
-------------

Named sections can be marked in the code writer that can be intercepted and modified by _section interceptors_. This gives the `CodeWriter` a lightweight extension system for augmenting generated code.

A section of code can be captured using a block state or an inline section. Section names must match the following regular expression: `^[a-z]+[a-zA-Z0-9_.#$]*$`.

### Block states

A block section is created by passing a string to `pushState()`. This string gives the state a name and captures all of the output written inside of this state to an internal buffer. This buffer is then passed to each registered interceptor for that name. These interceptors can choose to use the default contents of the section or emit entirely different content. Interceptors are expected to make calls to the `CodeWriter` in order to emit content. Interceptors need to have a reference to the `CodeWriter` as one is not provided to them when they are invoked. Interceptors are invoked in the order in which they are added to the `CodeBuilder`.

    
     CodeWriter writer = CodeWriter.createDefault();
     writer.onSection("example", text -> writer.write("Intercepted: " + text"));
     writer.pushState("example");
     writer.write("Original contents");
     writer.popState();
     System.out.println(writer.toString());
     // Outputs: "Intercepted: example\n"

### Inline sections

An inline section is created using a special `CodeWriter` interpolation format that appends "@" followed by the section name. Inline sections are function just like block sections, but they can appear inline inside of other content passed in calls to `write(java.lang.Object, java.lang.Object...)`. An inline section that makes no calls to `write(java.lang.Object, java.lang.Object...)` expands to an empty string.

Inline sections are created in a format string inside of braced arguments after the formatter. For example, `${L@foo}` is an inline section that uses the literal "L" value of a relative argument as the default value of the section and allows interceptors registered for the "foo" section to make calls to the `CodeWriter` to modify the section.

     CodeWriter writer = CodeWriter.createDefault();
     writer.onSection("example", text -> writer.write("Intercepted: " + text"));
     writer.write("Leading text...${L@example}...Trailing text...", "foo");
     System.out.println(writer.toString());
     // Outputs: "Leading text...Intercepted: foo...Trailing text...\n"

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
@@ -18,6 +18,7 @@ package software.amazon.smithy.utils;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.regex.Pattern;
 
 /**
@@ -82,7 +83,7 @@ import java.util.regex.Pattern;
  * }
  * </pre>
  */
-public final class CodeWriter {
+public class CodeWriter {
     private final StringBuilder builder = new StringBuilder();
     private final Deque<State> states = new ArrayDeque<>();
     private State currentState;
@@ -144,7 +145,7 @@ public final class CodeWriter {
      * @return Returns the generated code.
      */
     @Override
-    public String toString() {
+    public final String toString() {
         String result = builder.toString();
         // Insert a new line if one is pending or if trailing new lines are to be
         // added and the content doesn't already end with a newline.
@@ -166,7 +167,7 @@ public final class CodeWriter {
      *
      * @return Returns the code writer.
      */
-    public CodeWriter pushState() {
+    public final CodeWriter pushState() {
         State copiedState = new State(currentState);
         states.push(copiedState);
         currentState = copiedState;
@@ -183,7 +184,7 @@ public final class CodeWriter {
      * @return Returns the CodeWriter.
      * @throws IllegalStateException if there a no states to pop.
      */
-    public CodeWriter popState() {
+    public final CodeWriter popState() {
         if (states.size() == 1) {
             throw new IllegalStateException("Cannot pop CodeWriter state because at the root state");
         }
@@ -199,7 +200,7 @@ public final class CodeWriter {
      * @param newline Newline character to use.
      * @return Returns the CodeWriter.
      */
-    public CodeWriter setNewline(String newline) {
+    public final CodeWriter setNewline(String newline) {
         currentState.newline = newline;
         currentState.newlineRegexQuoted = Pattern.quote(newline);
         return this;
@@ -212,7 +213,7 @@ public final class CodeWriter {
      * @param newlinePrefix Newline prefix to use.
      * @return Returns the CodeWriter.
      */
-    public CodeWriter setNewlinePrefix(String newlinePrefix) {
+    public final CodeWriter setNewlinePrefix(String newlinePrefix) {
         currentState.newlinePrefix = newlinePrefix;
         return this;
     }
@@ -223,7 +224,7 @@ public final class CodeWriter {
      * @param indentText Indentation text.
      * @return Returns the CodeWriter.
      */
-    public CodeWriter setIndentText(String indentText) {
+    public final CodeWriter setIndentText(String indentText) {
         currentState.indentText = indentText;
         return this;
     }
@@ -252,7 +253,7 @@ public final class CodeWriter {
      * @param formatter Formatter to use.s
      * @return Returns the CodeWriter.
      */
-    public CodeWriter setFormatter(Formatter formatter) {
+    public final CodeWriter setFormatter(Formatter formatter) {
         currentState.formatter = Objects.requireNonNull(formatter);
         return this;
     }
@@ -262,7 +263,7 @@ public final class CodeWriter {
      *
      * @return Returns the CodeWriter.
      */
-    public CodeWriter trimTrailingSpaces() {
+    public final CodeWriter trimTrailingSpaces() {
         return trimTrailingSpaces(true);
     }
 
@@ -272,7 +273,7 @@ public final class CodeWriter {
      * @param trimTrailingSpaces Set to true to trim trailing spaces.
      * @return Returns the CodeWriter.
      */
-    public CodeWriter trimTrailingSpaces(boolean trimTrailingSpaces) {
+    public final CodeWriter trimTrailingSpaces(boolean trimTrailingSpaces) {
         currentState.trimTrailingSpaces = trimTrailingSpaces;
         return this;
     }
@@ -282,7 +283,7 @@ public final class CodeWriter {
      *
      * @return Returns the CodeWriter.
      */
-    public CodeWriter trimBlankLines() {
+    public final CodeWriter trimBlankLines() {
         return trimBlankLines(1);
     }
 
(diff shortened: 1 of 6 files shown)


smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
```diff
@@ -79,17 +207,136 @@
...
+ * {@code CodeBuilder}.
+ *
+ * <pre>{@code
+ * CodeWriter writer = CodeWriter.createDefault();
+ * writer.onSection("example", text -> writer.write("Intercepted: " + text"));
+ * writer.pushState("example");
+ * writer.write("Original contents");
+ * writer.popState();
+ * System.out.println(writer.toString());
+ * // Outputs: "Intercepted: example\n"
```
kstich: s/example/Original contents

kstich:
It looks like a leftover import from #114 is causing Travis to fail.

PR #123, opened 2019-07-22
Update and document aws.api#data trait

kstich:

### DIFF
--- a/docs/source/spec/aws-core.rst
+++ b/docs/source/spec/aws-core.rst
@@ -592,6 +592,134 @@ previous example:
         }
 
 
+.. _aws.api#data-trait:
+
+``aws.api#data`` trait
+=================================
+
+Summary
+    Designates the target as containing data of a known classification level.
+Trait selector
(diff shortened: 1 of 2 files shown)


docs/source/spec/aws-core.rst
```diff
@@ -592,6 +592,107 @@ previous example:
...
+    information.
+
+.. tabs::
+
+    .. code-tab:: smithy
+
+        use aws.api#data
+
+        structure MyStructure {
+          members: {
```
mtdowling: There's no "memebrs: {" wrapper in the IDL

docs/source/spec/aws-core.rst
```diff
@@ -592,6 +592,107 @@ previous example:
         }
 
 
+.. _aws.api#data-trait:
+
+``aws.api#data`` trait
```
mtdowling: This should call out that data classifications are scoped. If you apply the trait to a structure, then all members inherit the data classification of the structure unless they override it. I don't recall if this property holds true in more high-level ways like applying it to an operation or service though.

PR #133, opened 2019-08-05
Add put and collectionOperations to resources

JordonPhillips:
*Issue #, if available:* N/a

*Description of changes:*

This removes the collection trait in favor of adding put and
collectionOperations attributes on resources. With this change, an
idempotent operation that creates a resource using all of its
identifiers will now go under `put`, and `create` is reserved for
operations that don't include all the identifiers. This creates a clear
distinciton between create operations that work on the resource instance
vs the collection.

Additionally, a `collectionOperations` attribute was added that will
contain all operations that apply to the resource collection. Operations
in the regular operations list must now bind all the resources
identifiers. This makes the determination of which operations are
instance operations vs which are collection operations more explicit,
and allows for better validation.


docs/source/spec/core.rst
```diff
@@ -1507,20 +1513,18 @@ of an operation provide values for the identifiers of a resource.
...
 :dfn:`Collection operations` are used when an operation is meant to operate on
 a collection of resources rather than a specific resource. Collection
-operations are formed when an operation bound to a resource is marked with the
-``collectionTrait`` and one or more of the child identifiers of a resource
-are not bound to the input structure of an operation. The
-:ref:`list <list-lifecycle>` lifecycle operation is an example of a collection
-operation.
+operations are formed when an operation is bound to a resource with `collectionOperations`,
+or when bound to the :ref:`list <list-lifecycle>` or :ref:`create <create-lifecycle>`
+lifecycle operations.
```
mtdowling: I think this needs to define how to create a valid collection operation. You have to omit one or more identifiers of the resource it's bound to (though all parent identifiers must be bound to the input of the operation).

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborVisitor.java
```diff
@@ -91,11 +91,22 @@ private void addBound(List<Relationship> result, Shape container, ShapeId bindin
...
-        shape.getCreate().ifPresent(id -> result.add(relationship(shape, RelationshipType.CREATE, id)));
         shape.getRead().ifPresent(id -> result.add(relationship(shape, RelationshipType.READ, id)));
         shape.getUpdate().ifPresent(id -> result.add(relationship(shape, RelationshipType.UPDATE, id)));
         shape.getDelete().ifPresent(id -> result.add(relationship(shape, RelationshipType.DELETE, id)));
-        shape.getList().ifPresent(id -> result.add(relationship(shape, RelationshipType.LIST, id)));
+        shape.getPut().ifPresent(id -> result.add(relationship(shape, RelationshipType.PUT, id)));
+        // LIST and CREATE are, by definition, collection operations
+        shape.getCreate().ifPresent(id -> {
+            result.add(relationship(shape, RelationshipType.CREATE, id));
+            result.add(relationship(shape, RelationshipType.COLLECTION_OPERATION, id));
```
mtdowling: This behavior differs from how the OPERATION relationship is added. The OPERATION relationship is only formed for operations literally bound to the "operations" property, whereas the handling here for COLLECTION_OPERATION is to also emit this relationship for CREATE and LIST. We should probably make this work like the OPERATION binding.

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/RelationshipType.java
```diff
@@ -44,6 +44,13 @@
      */
     OPERATION("operation", RelationshipDirection.DIRECTED),
 
+    /**
+     * A collection operation relationship exists between a resource and the
+     * operations bound to the resource only in the "collectionOperations"
+     * property.
+     */
+    COLLECTION_OPERATION("collectionOperation", RelationshipDirection.DIRECTED),
```
mtdowling: I think this is right, but the description doesn't line up with how it's implemented in NeighborVisitor.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ResourceShape.java
```diff
@@ -103,6 +125,13 @@ public boolean hasIdentifiers() {
         return !identifiers.isEmpty();
     }
 
+    /**
+     * Gets the put lifecycle operation of the resource
+     */
```
mtdowling: Needs a `@return` or else javadocs will fail

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborVisitor.java
JordonPhillips: The OPERATION is bound to all operations, not just those in the `operations` key:

```
// Add all operation BINDING relationships (resource -> operation) and BOUND relations (operation -> resource).
shape.getAllOperations().forEach(id -> addBinding(result, shape, id, RelationshipType.OPERATION));
// Do the same, but for all of the lifecycle operations. Note: does not yet another BOUND relationships.
```

It would need to be `getOperations()` for it to only apply to what's in the `operations` key. Is that a bug? The comment immediately below seems to imply it's not, but then the docs in the relationship type imply it is.

mtdowling: Ah, good point. This needs some cleanup, and it looks like you also need to update the selectors docs: https://awslabs.github.io/smithy/spec/selectors.html#relationships

Based on the selector docs, I think it makes sense to have OPERATION continue to mean all operations (including collection operations), and we should add PUT.

This leaves COLLECTION_OPERATION as a bit of an outlier; we don't currently disambiguate between collection and instance operation bindings today since the OPERATION relationship includes both instance and collection operations. If we need to add this kind of relationship in the future, we could potentially add "OPERATIONS" to mean specifically operations bound to the "operations" property, and "COLLECTION_OPERATIONS" to mean operations bound to the "collectionOperations" property.

Thoughts?

JordonPhillips: We discussed this offline, and came to the decision to leave COLLECTION_OPERATION as is and also add in INSTANCE_OPERATION. You might want to, for instance, create traits that can only target one or the other.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ResourceShape.java
```diff
@@ -87,6 +96,19 @@ public Builder toBuilder() {
...
+    /**
+     * Gets the operations bound through the "collectionOperations" property.
+     *
+     * This will not include operations bound to resources using a lifecycle
+     * operation binding.
+     *
+     * @return Get the "collectionOperations" directly bound to this shape.
+     * @see #getAllOperations()
+     */
+    public Set<ShapeId> getCollectionOperations() {
```
kstich: Should this now also have a `getInstanceOperations` method?

JordonPhillips: This is specifically getting what's bound to `collectionOperations`, the equivalent would be `getOperations`

mtdowling:
One more change: ModelSerializer needs to be updated https://github.com/awslabs/smithy/blob/master/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ModelSerializer.java#L226

JordonPhillips:
Updated the serializer

PR #139, opened 2019-08-20
Allow plugins to be run serially

JordonPhillips:
*Issue #, if available:* N/A

*Description of changes:*

This change allows plugins to declare that they cannot be run in parallel
with other plugins. Any projection with such plugins will be run serially.
Parallelizeable projections will then be run in parallel afterwards.

Note that plugins already ran serially with respect to other plugins in the
same projection, and that behavior persists.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
@@ -153,14 +153,37 @@ final class SmithyBuildImpl {
     SmithyBuildResult applyAllProjections() {
         Model resolvedModel = createBaseModel();
         SmithyBuildResult.Builder builder = SmithyBuildResult.builder();
+        Map<String, ProjectionConfig> serialProjections = new TreeMap<>();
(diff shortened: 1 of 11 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
```diff
@@ -153,14 +153,37 @@ private static void validatePluginName(String projection, String plugin) {
     SmithyBuildResult applyAllProjections() {
         Model resolvedModel = createBaseModel();
         SmithyBuildResult.Builder builder = SmithyBuildResult.builder();
+        Map<String, ProjectionConfig> serialProjections = new TreeMap<>();
+        Map<String, ProjectionConfig> parallelProjections = new TreeMap<>();
         config.getProjections().entrySet().stream()
```
mtdowling: Can you leave a comment that describes the what and why of this branching? I'd also like to see a TODO item noted here that we could refactor this to be much more efficient by running all parallelizable plugins across all projections in parallel followed by all serial plugins. As each plugin completes, the SmithyBuildResult.Builder would be updated. This is a bigger refactor and all hidden as an implementation detail, which is why I'm fine delaying it.

PR #152, opened 2019-09-04
Allow using path expressions in paginator outputs

JordonPhillips:
*Description of changes:*

This allows the use of path expressions in paginator outputs. This enables
customers to have result wrapper shapes, which is a common pattern in XML
protocols.

### DIFF
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -4045,7 +4045,7 @@ settings from a service.
             }
         }
 
-The values for ``outputToken`` and ``items`` are paths. Paths are a series of
+The values for ``outputToken`` and ``items`` are paths. *Paths* are a series of
 identifiers separated by dots (``.``) where each identifier represents a
 member name in a structure. The first member name MUST correspond to a member
 of the output structure and each subsequent member name MUST correspond to a
@@ -4056,7 +4056,7 @@ following ABNF.
     path    :`identifier` *("." `identifier`)
 
 The following example defines a paginated operation which uses a result
-wrapper where the output token and items are indicated with paths.
+wrapper where the output token and items are referenced by paths.
 
 .. tabs::
 
@@ -4064,7 +4064,7 @@ wrapper where the output token and items are indicated with paths.
 
         namespace smithy.example
 
-        @collection @readonly
+        @readonly
         @paginated(inputToken: "nextToken", outputToken: "result.nextToken",
                    pageSize: "maxResults", items: "result.foos")
         operation GetFoos(GetFoosInput) -> GetFoosOutput
@@ -4101,7 +4101,6 @@ wrapper where the output token and items are indicated with paths.
                         "input" :"GetFoosInput",
                         "output": "GetFoosOutput",
                         "readonly": true,
-                        "collection": true,
                         "paginated": {
                             "inputToken": "nextToken",
                             "outputToken": "result.nextToken",
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -3883,7 +3883,7 @@ The ``paginated`` trait is an object that contains the following properties:
         the service that contains the operation.
     * - outputToken
       - ``string``
-      - The name of the operation output member that contains an optional
+      - The path to the operation output member that contains an optional
         continuation token. When this value is present in operation output,
         it indicates that there are more results to retrieve. To get the next
         page of results, the client passes the received output continuation
@@ -3896,7 +3896,7 @@ The ``paginated`` trait is an object that contains the following properties:
         the service that contains the operation.
     * - items
       - ``string``
-      - The name of a top-level output member of the operation that contains
+      - The path to an output member of the operation that contains
         the data that is being paginated across many responses. The named
         output member, if specified, MUST target a list or map.
     * - pageSize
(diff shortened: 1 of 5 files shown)


docs/source/spec/core.rst
```diff
@@ -4045,6 +4045,112 @@ settings from a service.
             }
         }
 
+The values for ``outputToken`` and ``items`` are paths. Paths are a series of
```
mtdowling: Why do we not support paths for inputToken and pageSize?

docs/source/spec/core.rst
```diff
@@ -4045,6 +4045,112 @@ settings from a service.
...
+member name in a structure. The first member name MUST correspond to a member
+of the output structure and each subsequent member name MUST correspond to a
+member in the previously referenced structure. Paths MUST adhere to the
+following ABNF.
+
+.. productionlist:: smithy
+    path    :`identifier` *("." `identifier`)
+
+The following example defines a paginated operation which uses a result
+wrapper where the output token and items are indicated with paths.
```
mtdowling: Minor: would this read better? "indicated with" -> "referenced by"

docs/source/spec/core.rst
mtdowling: Minor: can you put italics on "Path" to show it's defining a term?

docs/source/spec/core.rst
```diff
@@ -4045,6 +4045,112 @@ settings from a service.
...
+
+.. tabs::
+
+    .. code-tab:: smithy
+
+        namespace smithy.example
+
+        @collection @readonly
```
mtdowling: The `@collection` trait no longer exists

docs/source/spec/core.rst
```diff
@@ -4045,6 +4045,112 @@ settings from a service.
...
+        {
+            "smithy": "0.4.0",
+            "smithy.example": {
+                "shapes": {
+                    "GetFoos": {
+                        "type": "operation",
+                        "input" :"GetFoosInput",
+                        "output": "GetFoosOutput",
+                        "readonly": true,
+                        "collection": true,
```
mtdowling: Remove

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PaginatedTraitValidator.java
```diff
@@ -120,20 +122,26 @@
...
-        if (memberName == null) {
+        if (memberPath == null) {
             return service != null && validator.isRequiredToBePresent()
                    ? Collections.singletonList(error(operation, trait, String.format(
                     "%spaginated trait `%s` is not configured", prefix, validator.propertyName())))
                    : Collections.emptyList();
         }
 
-        MemberShape member = validator.getMember(opIndex, operation, trait).orElse(null);
+        if (!validator.pathsAllowed() && memberPath.split("\\.").length != 1) {
```
mtdowling: Minor: Storing the regex in a constant means we don't need to re-compile it each time. Could you just use `memberPath.contains(".")`?

docs/source/spec/core.rst
```diff
@@ -4045,6 +4045,112 @@ settings from a service.
...
 
+The values for ``outputToken`` and ``items`` are paths. Paths are a series of
+identifiers separated by dots (``.``) where each identifier represents a
+member name in a structure. The first member name MUST correspond to a member
+of the output structure and each subsequent member name MUST correspond to a
+member in the previously referenced structure. Paths MUST adhere to the
+following ABNF.
+
+.. productionlist:: smithy
+    path    :`identifier` *("." `identifier`)
```
mtdowling: Why limit this to two levels deep?

smithy-model/src/test/java/software/amazon/smithy/model/traits/PaginatedTraitTest.java
```diff
@@ -97,4 +97,35 @@ public void mergesWithOtherTrait() {
...
+
+    @Test
+    public void allowsNestedOutputToken() {
+        TraitFactory provider = TraitFactory.createServiceFactory();
+        Node node = Node.objectNode()
+                .withMember("inputToken", Node.from("inputToken"))
+                .withMember("outputToken", Node.from("result.outputToken"));
+
+        Optional<Trait> trait = provider.createTrait(
+                ShapeId.from("smithy.api#paginated"), ShapeId.from("ns.qux#foo"), node);
```
mtdowling: Minor: we typically use a newline to separate test setup from test assertions

docs/source/spec/core.rst
JordonPhillips: It isn't limited to two levels deep - you'll get a warning if it's more than that but your build will still pass. The warning exists because every level of nesting makes the api progressively more cumbersome to use. In an ideal world we would only have the result list and the output token in the response - anything else makes it hard to provide useful abstractions like iterators. A single level of nesting can be justified if it's a result wrapper shape, which is a fairly common pattern.

docs/source/spec/core.rst
JordonPhillips: We can always add support for this later, but it wasn't added because it makes the operations significantly more cumbersome to use. For reference, 0 of the several hundred existing AWS paginators use paths for input.

kstich: Should use the `:dfn:` directive.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated-trait-test.errors
```diff
@@ -18,3 +18,7 @@
 [ERROR] ns.foo#UnresolvedOutput$nextToken: member shape targets an unresolved shape `ns.foo#Missing` | Target
 [ERROR] ns.foo#Invalid9: When bound within the `ns.foo#Service` service, paginated trait `inputToken` is not configured | PaginatedTrait
 [ERROR] ns.foo#Invalid9: When bound within the `ns.foo#Service` service, paginated trait `outputToken` is not configured | PaginatedTrait
+[WARNING] ns.foo#DeeplyNestedOutputOperation: paginated trait `items` should not include a path with more than two parts | PaginatedTrait
```
mtdowling: Since this is the reference implementation of all things Smithy, this guidance here comes off almost as giving permission to implementers that more than 2 levels deep isn't supported. Can you try to rephrase this to be more about avoiding more than 2 levels to improve API quality?

docs/source/spec/core.rst
mtdowling: Ok, that sounds reasonable.

docs/source/spec/core.rst
mtdowling: 👌

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PaginatedTraitValidator.java
```diff
@@ -59,6 +62,7 @@
     private static final Set<ShapeType> ITEM_SHAPES = SetUtils.of(ShapeType.LIST, ShapeType.MAP);
     private static final Set<ShapeType> PAGE_SHAPES = SetUtils.of(ShapeType.INTEGER);
     private static final Set<ShapeType> STRING_SET = SetUtils.of(ShapeType.STRING);
+    private static final java.util.regex.Pattern PATH_PATTERN = Pattern.compile("\\.");
```
mtdowling: Whoops - can you remove this fully-qualified reference?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PaginatedTraitValidator.java
```diff
@@ -164,9 +182,53 @@
...
+        @Override
+        boolean pathsAllowed() {
+            return true;
+        }
+
+        Optional<MemberShape> getMember(
+                ShapeIndex index, OperationIndex opIndex, OperationShape operation, PaginatedTrait trait
+        ) {
+            // Split up the path expression into a list of member names
+            Optional<List<String>> memberNames = getMemberPath(opIndex, operation, trait)
```
mtdowling: Minor: this could be simplified to:

```java
Optional<List<String>> memberNames = getMemberPath(opIndex, operation, trait)
        .map(value -> Arrays.asList(value.split("\\.")))
        .orElse(Collections.emptyList());
```

PR #154, opened 2019-09-06
fix timestamp test cases

webern:
Trying to get a foothold in the project.

This may be a Java SDK version issue? Not sure. Using 12.

There were several timestamps that are valid per RFC 3339, but the tests were asserting that these were invalid. I double-checked RFC 3339. In any event, when I built master I received test failures and this PR fixed them. If the maintainers do not have test failures then it could be a Java version difference causing it.


webern:
Interesting! I'll have to try locally with Java 11 or 8.

webern:
Confirmed that there is a behavioral difference between Java 12 and everything before it that causes test failures and my 'fixes' are only relevant on Java 12. I will write this up in an issue.

PR #157, opened 2019-09-10
Add method to write class resource to manifest

mtdowling:
Writing a class resource to a file manifest by calling
getClass().getResourceAsStream("...") will trip up static analysis tools
like SpotBugs since the InputStream is never closed. This commit adds
methods to write resources from a class into a FileManifest that avoid
this issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/FileManifest.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/FileManifest.java
@@ -15,6 +15,7 @@
 
 package software.amazon.smithy.build;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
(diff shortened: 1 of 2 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/FileManifest.java
```diff
@@ -69,6 +70,34 @@ static FileManifest create(Path basePath) {
     @SuppressWarnings("unused")
     Path addFile(Path path);
 
+    /**
+     * Adds the files from another FileManifest into this FileManifest.
+     *
+     * @param manifest Other object to merge with.
+     */
+    default void addAllFiles(FileManifest manifest) {
```
mtdowling: Sorry about the churn, but these methods were wedged in between the other overloads of `writeFile`. Me adding two new methods made this divide even worse, so I went ahead and moved these.

PR #160, opened 2019-09-11
Allow authType to be specified

JordonPhillips:
This lets customers specify the `authType` themselves. Previously we
had always set it to `awsSigv4` or `custom` (depending on the
scheme), and that behavior is still the default.


aws/smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/apigateway/AuthorizerDefinition.java
```diff
@@ -187,6 +213,7 @@ public boolean equals(Object o) {
         AuthorizerDefinition that = (AuthorizerDefinition) o;
         return scheme.equals(that.scheme)
                && Objects.equals(type, that.type)
+               && authType.equals(that.authType)
```
mtdowling: I think this could cause a NullPointerException. You should use `Objects.equals(authType, that.authType)`.

JordonPhillips: Ah you're right, must have thought I was working with the `getAuthType` output

PR #163, opened 2019-09-14
Allow model assembling from symlink model files / directory

sawrav:
*Issue #163*

*Allow model assembling from symlink model files / directory. Model Assembler is currently ignoring symlink directories and files.*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URL;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -246,7 +247,7 @@ public final class ModelAssembler {
 
         if (Files.isDirectory(importPath)) {
             try {
-                Files.walk(importPath)
+                Files.walk(importPath, FileVisitOption.FOLLOW_LINKS)
                         .filter(p -> !p.equals(importPath))
                         .filter(p -> Files.isDirectory(p) || Files.isRegularFile(p))
                         .forEach(this::addImport);
@@ -254,7 +255,7 @@ public final class ModelAssembler {
                 throw new ModelImportException("Error loading the contents of " + importPath, e);
             }
         } else if (Files.isRegularFile(importPath)) {
-            stringModels.put(importPath.toString(), () -> IoUtils.readUtf8File(importPath.toString()));
+            stringModels.put(importPath.toString(), () -> IoUtils.readUtf8File(importPath));
         } else {
             throw new ModelImportException("Cannot find import file: " + importPath);
         }
--- a/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
@@ -30,6 +30,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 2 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -246,13 +248,21 @@ public ModelAssembler addImport(Path importPath) {
 
         if (Files.isDirectory(importPath)) {
             try {
-                Files.walk(importPath)
+                Files.walk(importPath, FileVisitOption.FOLLOW_LINKS)
```
mtdowling: After pulling this down and testing, I think this is actually the only line that needs to change. I was able to get the added test to pass by removing every other change but this one. Any reason for the other changes?

sawrav: The change comes into play when we pass a singular symlink targets instead of a directory. I will add appropriate test for this case as well by EOD

mtdowling: That would be great. The setting of NOFOLLOW_LINKS and the introduced or statement to check if a file is a symbolic link shouldn't be necessary since `Files.isRegularFile` should automatically follow symlinks: The setting of NOFOLLOW_LINKS and the introduced or statement to check if a file is a symbolic link shouldn't be necessary since `Files.isRegularFile` should automatically follow symlinks: https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isRegularFile-java.nio.file.Path-java.nio.file.LinkOption...-

sawrav: Yes you are right with your point of not having to isolate logic for detecting symlinks,  ```Files.isRegularFile``` should be sufficient to detect both symlinks and regular files. I did that in rev#1 since I was isolating the logic for importing regular files and symlinks. I have removed the extra check in rev#2.

So there are 2 use cases,

- **Symlinks under a directory / subdirectory**
In this case ```FileVisitOption.FOLLOW_LINKS``` change in files walk resolve the symlinks to actual target path and recursively trigger the logic for importing a regular file. Regular file import works fine since the import path = target path and does not require any additional change in the regular file import logic. For this I have added ```importsSymlinksDirectoryWithAllShapes``` test.

- **Single symlink file**
In this case the original import logic for regular file is not sufficient. This is because in this we try to import using the symlink (```importPath```) and not the actual target path. Therefore it is either required to have different import logic for symlink file (rev#1) or change the current regular file import logic to accommodate symlinks. In rev#2 I have gone with the latter and used ```importPath.toRealPath()``` instead of plain ```importPath```. This is because ```toRealPath()``` [works for both regular file and symlink](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#toRealPath(java.nio.file.LinkOption...)), for symlink this method yields the target path. To test this I have added ```importsSymlinkFileWithAllShapes``` test to confirm. If ```importPath.toRealPath()``` is removed, this test fails as importPath=symlink and nothing is imported.

In Rev#1 I covered both of the above cases by isolating the import logic for symlinks and regular files since I did not want to touch the regular file import logic. In Rev#2, it's the same but now I augmented the regular file import logic to accommodate symlinks by using ```importPath.toRealPath()``` instead of plain ```importPath```. Latter should be okay since we have good tests covering regular file imports as well.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -246,15 +247,20 @@ public ModelAssembler addImport(Path importPath) {
...
                         .filter(p -> Files.isDirectory(p) || Files.isRegularFile(p))
                         .forEach(this::addImport);
             } catch (IOException e) {
                 throw new ModelImportException("Error loading the contents of " + importPath, e);
             }
         } else if (Files.isRegularFile(importPath)) {
-            stringModels.put(importPath.toString(), () -> IoUtils.readUtf8File(importPath.toString()));
+            try {
+                Path path = importPath.toRealPath();
+                stringModels.put(path.toString(), () -> IoUtils.readUtf8File(path.toString()));
```
mtdowling: When storing the models, I thikn this should store the path name as `importPath.toString()` rather than `path.toString()` since we want to refer to the original path as the symlink rather than the target of the symlink.

Minor: I think `IoUtils.readUtf8File(path.toString())` can change to `IoUtils.readUtf8File(path)`.

sawrav: > When storing the models, I thikn this should store the path name as `importPath.toString()` rather than `path.toString()` since we want to refer to the original path as the symlink rather than the target of the symlink.
> 

The issue I see with this is the case when the symlink name is not the same as the target filename (specifically no file extensions exist in the symlink). Using `path.toString()` covers that case. This happens because the all `ModelLoader` validates filename before delegating to next loader. For example in the test `importsSymlinkFileWithAllShapes` I have used a different link name than the actual filename to prove that. If I have the same link name as the filename, there is no issue, otherwise the test fails. I would still prefer to store as `path.toString()` since symlink name could potentially be different than filename. Thoughts?

mtdowling: I see your point, but I think of the symlink as an implementation detail that shouldn't carry over into the API or error messages. I would guess that symlinks will usually be directories, and it would be very rare to symlink a single model file. If a single file is symlinked, then I'd be OK with requiring that these files end with ".json" or ".smithy".

Do you have a use case where you need to symlink individual files and can't use a valid extension for Smithy models? Given that you can access the contents of a file by pointing though a symlink, is it necessary to expose the files targeted by the symlink?

sawrav: Our logic for generating symlinks is pretty straight forward and it’s really upto the customer smithy model. Therefore we would not have a use case where we would need to have symlink names that is different to filenames. I just wanted to check this with you with an open source mindset. If there is no such concern then there is no need to expose the files targeted by the symlink, I will make the change.

sawrav: I have updated accordingly

PR #171, opened 2019-09-22
Refactor event streams to target members

mtdowling:
This commit removes the inputEventStream and outputEventStream traits
and replaces them with the eventStream trait that targets members. This
simplifies the model, validation, code generation, and I found that it
makes it easier to model operations correctly. In fact, several typos
were fixed in the documentation while making these changes due to the
fact that the previous design required that the member name of an
input/output shape is referenced from the operation.

The major drawbacks of this approach are:

1) It's a breaking change. While this is a breaking change, no models are
currently using the event stream feature in Smithy, and external tooling
hasn't been published that consumes them either.

2) Structures that contain an event stream cannot be targeted by anything
other than operation input or output. While it's awkward that event stream
structures are essentially "poisoned" and can't be targeted by other
members or used as errors, it does make event stream far easier to use
with code generation since the structure itself does not need to be
specialized when generating operation inputs and outputs. If we find
ourselves needing to use this for other abstractions, something more
reusable like a trait attribute could be introduced to automate this
validation.

*Issue #, if available:*

*Description of changes:*


smithy-model/src/test/java/software/amazon/smithy/model/knowledge/EventStreamIndexTest.java
```diff
@@ -22,15 +22,21 @@
...
-    private static Model model = Model.assembler()
-            .addImport(HttpBindingIndex.class.getResource("event-stream-index.json"))
-            .assemble()
-            .unwrap();
+    private static Model model;
+
+    @BeforeAll
+    public static void beforeAll() {
+        model = Model.assembler()
+                .addImport(HttpBindingIndex.class.getResource("event-stream-index.json"))
```
kstich: s/HttpBindingIndex/EventStreamIndexTest

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/event-stream-validator.json
```diff
@@ -43,35 +39,31 @@
...
       "ValidMultiEventBidirectionalOperation": {
         "type": "operation",
         "input": "ValidMultiEventInput",
-        "inputEventStream": "events",
-        "output": "ValidMultiEventOutput",
-        "outputEventStream": "events"
+        "output": "ValidMultiEventOutput"
       },
 
       "ValidMultiEventInput": {
```
kstich: The structures for `ValidMultiEvent[Input|Output]` should target the `ValidUnion` type, not `Event`

PR #173, opened 2019-09-24
Add a shape ID shader to help with codegen

mtdowling:
This abstraction can be used when generating code so that code is
generated into a consolidated namespace rather than spread across
all of the namespaces used in a Smithy model. `SymbolProvider`
implementations can use the `ShapeIdShader` to handle renaming
shape IDs before they convert the shape ID into a symbol appropriate
for that language. This abstraction essentially normalizes shape IDs
before the ID is converted into a language-specific symbol.

For example, if you want to generate code into the "EC2" namespace,
could set the `targetNamespace` to "EC2". If the service shape
that is being generated is in the "com.amazon.ec2" namespace, then you
could set the `rootNamespace` to "com.amazon.ec2" to completely
elide that namespace so that "com.amazon.ec2" is mapped to "EC2".

Any namespaces underneath the root namespace are then combined
with the target namespace. By default, they are just merged. For
example, given "com.amazon.ec2.nested.ns#Hello", the resulting
shape ID is "EC2.nested.ns#Hello". The merge behavior can be
customized by passing a `ShadeOption` to the `#shade(ShapeId, ShadeOption)`
method.

A shade option of `ShadeOption#SQUASH_INTO_NAME` combines
all trailing namespaces into the shaded shape ID's name. For example,
given "com.amazon.nested.ns#Hello", the shaded shape ID will become
"EC2#NestedNsHello". The first character of each namespace segment is
capitalized and prefixed on the original shape ID name.

A shade option of `ShadeOption#SQUASH_INTO_NAMESPACE` combines
all trailing namespaces into the shaded shape ID's namespace. For example,
given "com.amazon.nested.ns#Hello", the shaded shape ID will become
"EC2.NestedNs#Hello". The first character of each namespace segment is
capitalized and appended as additional segments after the target namespace.

Namespaces that are not inside of the given `rootNamespace`
are concatenated with the `targetNamespace` and then merged like
a sub-namespace. For example, given a "example.foo#Hi", the resulting
shape ID is "EC2.example.foo#Hi". When using `SQUASH_INTO_NAME`,
the resulting namespace is "EC2#ExampleFooHi". When using
`SQUASH_INTO_NAMESPACE`, the resulting namespace is
"EC2.ExampleFoo#Hi".

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapeIdShader.java
+++ b/codegen/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapeIdShader.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.codegen.core;
+
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
+ * Shades and renames shape IDs into a target namespace.
+ *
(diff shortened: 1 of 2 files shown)


codegen/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapeIdShader.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        return namespace;
+    }
+
+    /**
+     * Controls how shading is handled when there are namespace segments found
+     * outside of the {@code rootNamespace}.
+     */
+    public enum ShadeOption {
```
kstich: This should probably have an `APPEND` (or similar) that represents the default behavior.

codegen/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapeIdShader.java
```diff
@@ -0,0 +1,271 @@
...
+        /**
+         * Appends additional namespace segments outside of the target namespace
```
kstich: Should 'outside of the target namespace' be 'outside of the root namespace'?

PR #179, opened 2019-10-03
Disable authentication on generated CORS options

JordonPhillips:
CORS preflight options checks are sent without authentication, and so
the generated OPTIONS integrations should have an empty
"security" list.

To enable this, I made the list of security settings an `Optional` to distinguish between an empty list and a list that wasn't set. This is a bit odd in terms of the rest of the code base where we simply don't serialize empty lists. It's needed so that we can special case the OPTIONS methods from the general "security" setting. The alternative is to explicitly set the "security" setting on every method, which seems like it would be error prone.

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
@@ -56,7 +56,11 @@ public final class OperationObject extends Component implements ToSmithyBuilder<
         responses = Collections.unmodifiableMap(new TreeMap<>(builder.responses));
         deprecated = builder.deprecated;
         callbacks = Collections.unmodifiableMap(new TreeMap<>(builder.callbacks));
-        security = ListUtils.copyOf(builder.security);
+        if (builder.security != null) {
+            security = ListUtils.copyOf(builder.security);
+        } else {
+            security = null;
+        }
         servers = ListUtils.copyOf(builder.servers);
     }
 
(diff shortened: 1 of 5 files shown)


smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
```diff
@@ -104,8 +108,8 @@ public boolean isDeprecated() {
         return deprecated;
     }
 
-    public List<Map<String, List<String>>> getSecurity() {
-        return security;
+    public Optional<List<Map<String, List<String>>>> getSecurity() {
```
mtdowling: You could make this method always return a List by making the list of security schemes nullable. You'd return the value if present, otherwise an empty list.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
```diff
@@ -139,8 +143,8 @@ public boolean isDeprecated() {
                     .collect(ObjectNode.collectStringKeys(Map.Entry::getKey, Map.Entry::getValue)));
         }
 
-        if (!security.isEmpty()) {
-            builder.withMember("security", getSecurity().stream()
+        if (getSecurity().isPresent()) {
```
mtdowling: Then this would become a `if (security != null) {` check

smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
JordonPhillips: The problem then is from the outside you don't know whether the list was deliberately set to an empty list or just not set, unless you go out of you way to look at the generated node.

PR #180, opened 2019-10-04
Disable URLConnection cache in CLI

mtdowling:
When running Smithy in a build tool like Gradle, Smithy is run multiple
times within a thread. Java's URLConnection has a cache built-in for
loading JARs, and when a JAR is updated during a Smithy build, it can
cause checksum exceptions when attempting to load that same JAR from a
subsequent run through the cache. This commit disables the cache when
running Smithy through the CLI to prevent this issue (Gradle runs Smithy
via the CLI in a thread).

(I couldn't find a way to test this other than manually test that it fixed the issues I was running into when using it with the Gradle plugin)

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.file.FileVisitOption;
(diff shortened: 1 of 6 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -142,7 +157,6 @@ public ModelAssembler reset() {
         validators.clear();
         suppressions.clear();
         documentNodes.clear();
-        properties.clear();
```
mtdowling: These properties should have never of been cleared during a call to reset(). Reset should clear out transient state of the assembler, not configuration like properties.

PR #191, opened 2019-10-21
Allow the xmlFlattened trait to be on members

JordonPhillips:

### DIFF
--- a/docs/source/spec/xml.rst
+++ b/docs/source/spec/xml.rst
@@ -100,7 +100,7 @@ Summary
 Trait selector
     ``:test(collection, member:of(structure) > collection)``
 
-    *Any list or set*
+    *Any list or set or any structure member that targets a list or set*
 Value type
     Annotation trait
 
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
```diff
@@ -148,7 +148,7 @@ structure xmlAttribute {}
 
 /// Moves serialized collection members from their collection element to that of
 /// the collection's container.
-@trait(selector: "collection")
+@trait(selector: ":each(collection, structure > :test(member > collection))")
```
mtdowling: It would be simpler to do:

`:test(collection, member:of(structure) > collection)`

docs/source/spec/xml.rst
```diff
@@ -98,7 +98,7 @@ Summary
     Moves serialized collection members from their collection element to that
     of the collection's container.
 Trait selector
-    ``collection``
+    ``:test(collection, member:of(structure) > collection)``
 
     *Any list or set*
```
kstich: + "or a member that targets a list or set"

PR #204, opened 2019-11-14
Prevent bad list, set, map recursion

mtdowling:
We currently don't allow a directly recursive list or set shape.
However, we still allow for list, set, and map shapes to be
transitively recursive unconditionally (e.g., listA -> listA is invalid,
but listA -> set -> listA is valid). This is problematic for code
generators in various languages that want to use parameteric types and
standard libraries to codegen lists, sets, and maps. For example, in
Java, a list in Smithy should be generated as a `List<X>` where X is
the code generated member of the list. However, if the list is recursive
onto itself, then it's impossible to generate valid code (e.g.,
`List<List<List...`).

This change requires that recursive list, set, and map shapes must have
one or more structure or union members in their recursive path in order to
define a valid shape. This will result in shapes that are far easier to
generate and provides constraints that are very helpful when converting
to other formats (for example, this might be useful in JSON schema
conversions if we want to always inline list and set shape definitions).
This change should not have a practical impact on models since no
current AWS service defines an unconditionally recursive list, set,
or map shape.

*Issue #, if available:*

*Description of changes:*


docs/source/spec/core.rst
```diff
@@ -505,6 +504,13 @@ definition using an ``apply`` statement:
             }
         }
 
+The member of a list cannot directly or transitively target the list
```
kstich: Should similar paragraphs be added for set and map types?

docs/source/spec/core.rst
```diff
@@ -415,8 +415,7 @@ list
 The :dfn:`list` type represents a homogeneous collection of values. A list is
 defined using a :token:`list_statement`. A list statement consists of the
 shape named followed by an object with a single key-value pair of "member"
-that defines the :ref:`member <member>` of the list. The member of a list
-cannot target the containing list shape.
+that defines the :ref:`member <member>` of the list.
```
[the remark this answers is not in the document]
mtdowling: Whoops that was a leftover from a previous iteration. Removed.

docs/source/spec/core.rst
mtdowling: Also a removed leftover from a previous iteration

PR #206, opened 2019-11-15
Add and remove members to index automatically

mtdowling:
This commit updates the ShapeIndex to automatically add and remove
members from shapes that have members. This required some extra
functionality be introduced to shapes which then in turn resulted in
cleanup in other places in the codebase.

1. ShapeIndex now automatically adds and removes members when shapes are
   added.
2. There is now a members() method on all shapes to return the members
   contained in the shape (if any). Shapes that have no members just
   return an empty collection by default, and shapes that do contain
   members override the abstract method.
3. Because members are now less important to contain a reference to
   outside of their containing shapes (since you don't need to
   explicitly add them to a shape index), you can now automatically
   create member shapes for container shapes by just passing in the
   target of the member or passing both the target and a consumer that
   accepts the created MemberShape.Builder type. This removes the need
   to create the right shape ID for the shape (suffixing with "$" + the
   member name).
4. ShapeIndex doesn't really need a ConcurrentHashMap to cache different
   groups of shape types. The code that was used for that was overly
   complicated and likely made it slower rather than faster (the
   majority of models will use the prelude and will have more than 50
   shapes, so the optimizations that were in place don't really make
   sense any more).
5. Because there's so much shared code between StructureShape and
   UnionShape, I created an abstract class for their shape and builders
   to make them simpler to maintain.
6. A couple model transformations were updated to use the new members()
   method.
7. Members of structures and unions are now sorted rather than maintain
   insert order. Nothing should ever depend on a stable order based on
   inserts since Smithy models can be loaded and aggregated from
   multiple sources.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeIndex.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeIndex.java
@@ -19,12 +19,9 @@ import java.util.AbstractSet;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import software.amazon.smithy.utils.MapUtils;
 import software.amazon.smithy.utils.SmithyBuilder;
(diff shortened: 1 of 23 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeIndex.java
```diff
@@ -216,18 +186,21 @@ public Builder addShape(Shape shape) {
...
          *
          * @param shapes Collection of Shapes to add.
+         * @param <S> Type of shape being added.
          * @return Returns the builder.
          */
-        public Builder addShapes(Collection<Shape> shapes) {
-            shapes.forEach(this::addShape);
+        public <S extends Shape> Builder addShapes(Collection<S> shapes) {
+            for (Shape shape : shapes) {
+                addShapes(shape);
```
kstich: It's just redirected in the end, but should be `addShape` not `addShapes` for singular adds.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeIndex.java
```diff
@@ -251,7 +224,16 @@ public Builder addShapes(Shape... shapes) {
          * @return Returns the builder.
          */
         public Builder removeShape(ShapeId shapeId) {
-            shapeMap.remove(shapeId);
+            if (shapeMap.containsKey(shapeId)) {
+                Shape previous = shapeMap.get(shapeId);
+                shapeMap.remove(shapeId);
+
+                // Automatically remove any members contained in the shape.
```
kstich: Might be worth a note in the javadoc as well, like was added to `addShape`.

PR #211, opened 2019-11-21
Update smithy-build to be streaming

mtdowling:
smithy-build used to load every model from every projection into memory
at once. This was not ideal since some use cases utilize hundreds of
projections. For example, in AWS SDKs, we create a projection for every
AWS service, which is hundreds of them with tens-of-thousands of shapes
and operations.

This commit updates smithy-build to support two modes: in-memory and
streaming. The in-memory mode is useful for testing or one-off
integrations. The streaming mode is useful for actually executing
smithy-build from the CLI so that results are outputted to stdout as
they happen and not all models are loaded into memory.

The CLI was updated to significantly change its output to show more
information about things that go wrong and print more statistics. Every
projection and plugin is attempted to be executed -- the CLI previously
failed on the first exception.

Finally, executing tasks is now done with an ExecutorService rather than
parallel streams to have more control over task execution.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
@@ -17,20 +17,27 @@ package software.amazon.smithy.build;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Comparator;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
```diff
@@ -150,46 +157,99 @@ private static void validatePluginName(String projection, String plugin) {
...
+            projectionExceptionConsumer.accept(name, e);
+        }
 
-        return builder.build();
+        if (result != null) {
+            projectionResultConsumer.accept(result);
+        }
+    }
+
+    private void executeParallelProjections(
```
kstich: 🎉

PR #217, opened 2019-11-26
Add validation for length trait values

JordonPhillips:


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/LengthTraitValidator.java
```diff
@@ -0,0 +1,43 @@
+package software.amazon.smithy.model.validation.validators;
```
kstich: Missing license header.

PR #218, opened 2019-11-27
Allow page member to be required

JordonPhillips:
This allows the page size member to be required, but still defaults
it to failing the build when it is.


mtdowling:
Spec needs updating to say it SHOULD NOT be required: https://awslabs.github.io/smithy/spec/core.html#paginated-trait

PR #221, opened 2019-12-02
Limit streaming traits to top level members

JordonPhillips:


mtdowling:
Why do we need this? Ideally we could apply streaming traits to *only* shapes or *only* members.

The motivation for applying only to shapes is that the streaming trait should change the code generated type of the shape. We usually try to not have members alter the type that is generated in targeted programming languages.

However, an argument for the streaming trait to only be applied to members could be made based on the fact that only top-level input, output, and error shapes can every be streaming. These are actually the same locations as event streams. The rationale for making even streams apply to the member of a structure was that the structure that targets the event stream is essentially claimed as a special shape that can't be used like other shapes throughout the model.... Maybe we should apply this same treatment to the streaming trait.

PR #233, opened 2019-12-17
Relax validation for uri conflicts

JordonPhillips:
This relaxes uri conflict validation in the case where one of the
conflicting segments is a label and the other is static. Previously
this would always result in an ERROR. Now it will result in a DANGER
if and only if all such conflicting segments have constraints on the
label member that prevent it from evaluating to the value of the
static segment.


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -37,45 +48,145 @@
...
+                        return false;
+                    }
+
+                    // If the pattern on the label segment does not match the static segment, then this segment's
+                    // conflict is allowable.
+                    return !pattern.matcher(staticSegment).find();
+                });
+    }
+
+    private Map<String, Pattern> getLabelPatterns(Model model, OperationShape operation) {
```
mtdowling: This can be simplified since the `HttpBinding` type already contains the resolved `MemberShape` for each binding.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -37,45 +48,145 @@
...
+        return events;
+    }
+
+    private boolean isAllowableConflict(Model model, OperationShape operation, OperationShape otherOperation) {
+        UriPattern uriPattern = operation.getTrait(HttpTrait.class).get().getUri();
+        UriPattern otherUriPattern = otherOperation.getTrait(HttpTrait.class).get().getUri();
+
+        List<Pair<Segment, Segment>> conflictingSegments = uriPattern.getConflictingSegments(otherUriPattern);
+
+        // If there aren't any conflicting segments that means the uris are identical, which is not allowable.
```
mtdowling: When would this get triggered given the check was already made before this was called to see if the URIs conflict? Would it be when the conflicts are for nothing but literal segments? If that's the case, there might be a more descriptive name for `getConflictingSegments` that helps convey this...

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -37,45 +48,145 @@
...
-            return Optional.empty();
+        return conflictingSegments.stream()
+                // Only allow conflicts in cases where one of the segments is static and the other is a label.
+                .filter(conflict -> conflict.getLeft().isLabel() != conflict.getRight().isLabel())
+                // Only allow the uris to conflict if every conflicting segment is allowable.
+                .allMatch(conflict -> {
+                    Pattern pattern;
+                    String staticSegment;
+
+                    if (conflict.getLeft().isLabel()) {
```
mtdowling: Seems like this could be simplified if you first resolve which of the two are the label, place that result in a variable, and then extract the pattern and static segment from that.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -37,45 +48,145 @@
...
+                        staticSegment = conflict.getLeft().getContent();
+                    }
+
+                    if (pattern == null) {
+                        return false;
+                    }
+
+                    // If the pattern on the label segment does not match the static segment, then this segment's
+                    // conflict is allowable.
+                    return !pattern.matcher(staticSegment).find();
```
mtdowling: Please double check that this doesn't implicitly add regex anchors... I don't remember if it's find that does this or matches.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
JordonPhillips: I renamed it to `getConflictingLabelSegments`

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
JordonPhillips: I don't think that simplifies anything. I need to know which pattern map to search.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-conflict-validator.json
```diff
@@ -95,6 +122,167 @@
                     "method": "GET"
                 }
             }
+        },
+        "ns.foo#H": {
```
kstich: Can there be a test case that conflicts with a label that doesn't have a pattern at all to verify that's handled correctly?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -37,45 +44,124 @@
...
+        }
+
+        return events;
+    }
+
+    private boolean isAllowableConflict(Model model, OperationShape operation, OperationShape otherOperation) {
+        UriPattern uriPattern = operation.getTrait(HttpTrait.class).get().getUri();
+        UriPattern otherUriPattern = otherOperation.getTrait(HttpTrait.class).get().getUri();
+
+        List<Pair<Segment, Segment>> conflictingSegments = uriPattern.getConflictingLabelSegments(otherUriPattern);
```
kstich: Minor: Can this var be renamed `conflictingLabelSegments` for clarity, along with the comment on line 113?

PR #242, opened 2020-01-05
Clean up and rewrite spec for XML bindings

mtdowling:
This commit updates the XML traits spec to be made more generic and
framed in terms of binding to XML. XML traits have been restricted to
only apply to shapes in which they are actually supported. Many more
examples have been added to the XML spec (note that I selectively did
not include a JSON equivalent example when a JSON example adds no
expository value).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/spec/xml.rst
+++ b/docs/source/spec/xml.rst
@@ -1,15 +1,587 @@
-==========
-XML traits
-==========
+============
+XML bindings
+============
 
-Protocols that serialize messages using XML SHOULD honor the ``xmlAttribute``,
-``xmlFlattened``, ``xmlName``, and ``xmlNamespace`` traits when serializing
-payloads.
+This document defines how to bind Smithy shapes to XML documents. Smithy
+defines several traits that are used to influence the serialization of
+shapes with XML based protocols.
 
 .. contents:: Table of contents
     :depth: 2
     :local:
     :backlinks: none
+
+
+----------------------
+Serializing XML shapes
+----------------------
+
+This document provides recommendations on how Smithy structures and
+shapes within structures SHOULD be serialized with XML based protocols;
+however, protocols MAY choose to deviate from these recommendations
+if necessary.
+
+
+Structure and union serialization
+=================================
+
+All XML serialization starts with a structure. The shape name of a structure
+is used as the outermost XML element name. Members of a structure or union
+shape are serialized as nested XML elements where the name of the element is
+the same as the name of the member.
+
+For example, given the following:
+
+.. code-block:: smithy
+
+    structure MyStructure {
+        foo: String,
+    }
+
+The XML serialization is:
+
+.. code-block:: xml
+
+    <MyStructure>
+        <foo>example</foo>
+    </MyStructure>
+
+
+Custom XML element names
+------------------------
+
+Structure/union member element names can be changed using the
+:ref:`xmlname-trait`.
+
+
+XML attributes
+--------------
+
+The :ref:`xmlattribute-trait` is used to serialize a structure
+member as an XML attribute.
+
+
+``xmlName`` on structures
+-------------------------
+
+An ``xmlName`` trait applied to a structure changes the element name of the
+serialized structure; however, it does not influence the serialization of
+members that target it. Given the following:
+
+.. tabs::
+
+    .. code-tab:: smithy
+
+        @xmlName("AStruct")
+        structure A {
+            b: B,
+        }
+
+        @xmlName("BStruct")
+        structure B {
+            hello: String,
+        }
+
+    .. code-tab:: json
+
+        {
+            "smithy": "0.5.0",
+            "shapes": {
+                "smithy.example#A": {
+                    "type": "structure",
+                    "members": {
+                        "b": {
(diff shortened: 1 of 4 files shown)


docs/source/spec/xml.rst
```diff
@@ -1,26 +1,598 @@
...
+serialized in a nested element named ``entry`` that contains a nested
+``key`` and ``value`` element. For example, given the following:
+
+.. code-block:: smithy
+
+    structure Foo {
+        values: MyMap
+    }
+
+    list MyMap {
```
kstich: Multiple `list MyMap` instances that should be `map myMap`.

docs/source/spec/xml.rst
```diff
@@ -1,26 +1,598 @@
...
+to change the nested element names.  For example, given the following:
+
+.. code-block:: smithy
+
+    structure Foo {
+        values: MyMap
+    }
+
+    list MyMap {
+        @xmlName("Name")
```
kstich: Mismatched with example below using `"Attribute"`.

docs/source/spec/xml.rst
```diff
@@ -253,33 +897,32 @@ Given the following structure definition,
...
+        <bar>example</bar>
     </MyStructure>
 
-.. note::
+A namespace prefix can be inserted before the element name. Given the
+following
+
+.. code-block:: smithy
+
+    structure AnotherStructure {
```
kstich: Mismatched with below using `MyStructure`.

docs/source/spec/xml.rst
```diff
@@ -349,177 +994,37 @@ Given the following structure definition,
...
-                }
-            }
-        }
+Given the following:
 
-and the following values provided for ``MyStructure``,
+.. code-block:: smithy
 
-::
+    @xmlNamespace(uri: "http://foo.com", prefix: "bar")
```
kstich: Mismatched with example below using `baz` for with the `xmlns` attribute.

docs/source/spec/xml.rst
```diff
@@ -1,26 +1,598 @@
...
+                        }
+                    },
+                    "traits": {
+                        "smithy.api#xmlName": "BStruct"
+                    }
+                }
+            }
+        }
+
+The XML serialization of ``AStruct`` is:
```
kstich: The name of the structure shape itself is `A`, `AStruct` is the value of the `@xmlName` trait.

docs/source/spec/xml.rst
```diff
@@ -1,26 +1,598 @@
...
+
+.. code-block:: xml
+
+    <Choice>
+        <Hi>example1</Hi>
+        <Hi>example2</Hi>
+        <Hi>example3</Hi>
+    </Choice>
+
+The ``xmlName`` trait applied to the member of a list has no effect when
```
kstich: Nit: Should this paragraph use "list/set" like the opening paragraph, since "structure/union" is preserved as well? The proceeding section uses "list or set", which is different as well.

docs/source/spec/xml.rst
```diff
@@ -1,26 +1,598 @@
...
+
+The following table defines how simple types are serialized in XML documents.
+
+.. list-table::
+    :header-rows: 1
+    :widths: 10 90
+
+    * - Shape
+      - Serialization
+    * - blob
```
kstich: We usually code-wrap the types in tables like this.

docs/source/spec/xml.rst
mtdowling: Fixed in the prior rebase

docs/source/spec/xml.rst
mtdowling: Because `A` is the outermost structure being serialized, the `xmlName` trait takes effect.

docs/source/spec/xml.rst
mtdowling: I copied this table: https://awslabs.github.io/smithy/spec/core.html#simple-types

docs/source/spec/xml.rst
kstich: I'd read this sentence as referring to the XML serialization of the Smithy structure `smithy.example#A`.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
```diff
@@ -140,21 +140,19 @@ structure idempotencyToken {}
...
+/// Unwraps the values of a list, set, or map into the containing structure/union.
+@trait(selector: ":test(member:of(structure, union) > :each(collection, map))")
 @tags(["diff.error.const"])
 structure xmlFlattened {}
 
-/// Allows a serialized object property name to differ from a structure member name
-/// used in the model.
-@trait
+/// Changes the serialized element or attribute name of a structure or member.
+@trait(selector: ":test(structure, member)")
```
JordonPhillips: This only applies to top level structures right? So this selector should enforce that.

mtdowling: Also structures targeted by the `httpPayload` trait.

PR #251, opened 2020-01-10
Add ec2QueryNameTrait

JordonPhillips:
This adds a trait to specify the name of an EC2 structure member when
it is used as an input.

### DIFF
--- a/smithy-aws-traits/src/main/resources/META-INF/smithy/aws.api.json
+++ b/smithy-aws-traits/src/main/resources/META-INF/smithy/aws.api.json
@@ -238,6 +238,16 @@
                 },
                 "smithy.api#documentation": "Configures endpoint discovery for the service."
             }
+        },
+        "aws.api#ec2QueryName": {
+            "type": "string",
+            "traits": {
+                "smithy.api#trait": {
+                    "selector": "member:of(structure)"
+                },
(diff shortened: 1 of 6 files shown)


smithy-aws-traits/src/main/resources/META-INF/smithy/aws.api.json
```diff
@@ -238,6 +238,16 @@
...
             }
+        },
+        "aws.api#ec2QueryName": {
+            "type": "string",
+            "traits": {
+                "smithy.api#trait": {
+                    "selector": "member:of(structure)"
+                },
+                "smithy.api#documentation": "Indicates the serialized name of a structure member when that structure is serialized for the input of an EC2 operation.",
+                "smithy.api#pattern": "^[a-zA-Z_][a-zA-Z_0-9-]*(:[a-zA-Z_][a-zA-Z_0-9-]*)?$"
```
mtdowling: Where did this come from?

JordonPhillips: This is the same as the xmlname pattern

JordonPhillips:
Updates: added the license bit and updated the pattern to not include a `:` which only made sense in the context of xml

PR #259, opened 2020-01-15
Update Maven uri for HTTPS

kstich:


build.gradle.kts
```diff
@@ -141,7 +141,7 @@ subprojects {
         repositories {
             mavenLocal()
             maven {
```
mtdowling: This should actually just get changed to `mavenCentral()`

PR #263, opened 2020-01-19
Use Gradle plugin to build protocol tests

mtdowling:
Make various improvements geared towards Gradle/CLI

AWS protocol tests:
- Use the Gradle plugin to build the AWS protocol tests.

SmithyBuild:
- Only run parallel projections if there's more than one.

SmithyCli (breaking CLI changes):
- to allow the writing to stderr and stdout to be completely customized
  using a Consumer<String>. This allows the Gradle plugin to log writes
  to stdout rather than rely on the non-thread-safe default behavior of
  intercepting calls.
- Colors nows is used by calling out and err directly on an enum variant
  rather than a static method.
- Adding the --logging parameter to every command and removed the static
  `configureLogging` method.
- Adding `stdout` and `stderr` methods to the Cli. These are now called
  by Colors when writing.
- Enabling disabling ANSI colors is done on Cli and not on Colors now.
  All of the methods used to influence the CLI globally is now all on
  the Cli.
- Logging is only configured when a logging option is passed in. A
  custom logger is used that makes calls to the intercepted stderr
  method.
- Running `build` now shows validation results too.

SmithyModel:
- ValidationEvents are now sorted by filename, line number, column,
  severity, shape ID, message, then finally the event ID.
- Introducing the ValidationEventFormatter interface. There are
  now implementations of the current display (showing the event on
  a single line like CheckStyle), and a contextual formatter that
  shows the line of the source file that has the error.


mtdowling:
Note that this depends on the Gradle plugin being released first.

mtdowling:
This has been updated to use 0.4.3 of the Gradle plugin

PR #274, opened 2020-02-12
Rewrite JSON schema conversion

mtdowling:
This commit significantly cleans up the JSON schema conversion and
automatically inlines primitive references rather than inlining them
through a mapper.

The previous JSON schema implementation code had several issues:

1. It did not do a good job at handling shape ID conflicts when using
   namespace stripping. We had to add some pretty bad hacks to achieve
   this. For examplem, it had implicit state that was tricky to handle
   (like temporarily setting a ref strategy based on a converted shape
   index).
2. It didn't detect errors early in the process, resulting in strange
   errors when you try to use the schema.
3. It exposed too much public API (for example RefStrategy should not be
   public). Ideally with this trimmed down API surface area, we won't
   need another breaking change.
4. JSON schema names by default should not include a namespace.
5. Simple shapes by default should always be inlined. Things like list
   and set shapes aren't that important for generating good JSON
   schema or OpenAPI schemas. By inlining them, we also ensure that
   any member documentation attached to members that target list or
   set shapes isn't lost since that documentation comes from either
   the member or the targeted shape. This also reduces the possibility
   for naming conflicts when dropping the namespace from the Smithy
   shape ID and converting it to JSON Schema.
6. We were dropping member traits in some scenarios like
   documentation, pattern, range, length. This is now fixed.

Because converting shape IDs to JSON pointers can now result in a nested
JSON pointer, the ability to select schemas from a SchemaDocument using a
JSON pointer has been implemented.

Finally, the Smithy document shape is actually meant to be a simple type,
but it was correctly subclassing SimpleShape, resulting in JSON schema
conversions not working correctly (document types were creating distinct
named shapes, whereas they are intended to be inlined).

*Issue #, if available:*

*Description of changes:*

Note: this will fail due to OpenAPI not being updated yet. That'll follow soon.


smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConverter.java
```diff
@@ -109,216 +151,270 @@ public ObjectNode getConfig() {
...
+                .map(pair -> {
+                    LOGGER.fine(() -> "Converting " + pair.getRight() + " to JSON schema at " + pair.getLeft());
+                    return Pair.of(pair.getLeft(), pair.getRight().accept(visitor));
+                })
                 .forEach(pair -> builder.putDefinition(pair.getLeft(), pair.getRight()));
 
+        LOGGER.fine(() -> "Completed JSON schema document conversion (root shape: " + rootShape + ")");
+
+        // Add extensions if present.
+        config.getObjectMember(JsonSchemaConstants.SCHEMA_DOCUMENT_EXTENSIONS).ifPresent(builder::extensions);
```
kstich: This is doing the [same work as this line](https://github.com/awslabs/smithy/pull/274/files#diff-aaa37e69f32f0a6ca741784407a1fc55R220), just without going to the function.

PR #280, opened 2020-02-20
[0.10] Add protocol and auth trait requirements

mtdowling:
Protocol traits now list the traits that protocol implementations must
understand in order to successfully implement the protocol. This
provides the ability to better ensure correctness of protocol
implementations as protocols evolve.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -4675,7 +4675,21 @@ Summary
 Trait selector
     ``[trait|trait]``
 Value type
-    Annotation trait.
+    An object with the following properties:
+
+    .. list-table::
+       :header-rows: 1
+       :widths: 10 23 67
+
+       * - Property
+         - Type
+         - Description
+       * - traits
+         - [:ref:`shape-id`]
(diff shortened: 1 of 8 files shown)


kstich:
Should `authDefinition` receive the same treatment, allowing `aws.auth#sigv4` to specify it utilizing `aws.auth#unsignedPayload`?

mtdowling:
Sure. I updatred the auth trait to also have a list of trait requirements. We discussed this and don't see it as a general abstraction at this point (like a new property of trait definitions). These requirements are really a function of serialization and deserialization specific traits, which is what auth and protocol traits are.

docs/source/spec/core.rst
```diff
@@ -4928,7 +4942,21 @@ Summary
 Trait selector
     ``[trait|trait]``
 Value type
-    Annotation trait.
+    An object with the following properties:
```
kstich: JSON AST example should be updated to `{}` from `true`

PR #283, opened 2020-02-21
[0.10] Adds UnstableTrait and UnstableTraitValidator

srchase:
Adds `UnstableTrait` that can be used to indicated a shape is unstable and may change in the future.

Adds `UnstableTraitValidator` that emits warnings when a model contains shapes that apply unstable traits.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/traits/UnstableTrait.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/traits/UnstableTrait.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/traits/UnstableTrait.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JordonPhillips: 2020

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/UnstableTraitValidator.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JordonPhillips: 2020

smithy-model/src/test/java/software/amazon/smithy/model/traits/UnstableTraitTest.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JordonPhillips: 2020

JordonPhillips:
Are we only using this as a meta-trait, or can it be applied to any given shape? Your trait selector lets it be attached to anything, but you only emit meta-trait warnings

mtdowling:
I think it makes sense to only look at traits that are unstable in the IDL. Warnings should be emitted, when possible, in programming languages when unstable shapes are used, but in the model, it makes sense for shapes to refer to unstable shapes. For example, I could have an unstable operation bound to my service because I want to beta test it. Emitting a warning for that on the service would be just noise IMO.

mtdowling:
Can you also target 0.10 instead?

PR #290, opened 2020-02-24
[0.10] Adds UnstableTrait and UnstableTraitValidator

srchase:
Adds `UnstableTrait` that can be used to indicated a shape is unstable and may change in the future.

Adds `UnstableTraitValidator` that emits warnings when a model contains shapes that apply unstable traits.

### DIFF
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -5550,6 +5550,27 @@ Value type
         }
 
 
+.. _unstable-trait:
+
+``_unstable`` trait
+---------------
+
+Summary
+    Indicates a shape is unstable and may change in the future.
+Trait selector
+    ``*``
+
+Value type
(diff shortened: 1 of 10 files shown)


docs/source/spec/core.rst
```diff
@@ -5550,6 +5550,27 @@ Value type
         }
 
 
+.. _unstable-trait:
+
+``_unstable`` trait
+---------------
+
+Summary
+    Indicates a shape is unstable and may change in the future.
```
mtdowling: "may" -> "MAY".

I think this should give a bit more guidance after the first sentence:

"This trait can be applied to trait definitions to indicate that a trait is unstable or experimental. If possible, code generators SHOULD use this trait to warn when code generated from unstable features are used."

PR #293, opened 2020-02-26
[0.10] Add JSON pointer, patch, and OpenAPI mapper

mtdowling:
This commit adds a new mapper to OpenAPI that allows changing specific
parts of the generated schema using JSON pointers that inject values
using semantics similar to JSON Patch. This approach is preferred over
string substitutions because it is more explicit and does not encourage
modelers to include placeholders in their models.

The JSON substitutions mapper now allows for any string to be replaced.
This is still basically just as safe since it only replaces exact string
matches (it doesn't replace inside of larger strings). Most importantly,
the use of placeholder like strings encouraged modelers to include
placeholders in their models. By opening up the string to any value, you
can now replace whatever was generated from real model values.

To achieve this, I implemented a JSON Pointer parser and abstraction
over Node values that also include methods for getting a value from a
node by a pointer, and adding/replacing a value in a node using a
pointer (with immutable return values, of course). I was able to
integrate this abstraction into SchemaDocument and Schema too. The JSON
Patch "add" operation has been implemented in two ways: 1) we do it just
like JSON patch, where attempting to traverse through missing objects
and arrays causes the change to be ignored 2) we have a mode that
creates intermediate objects when needed, allowing this to be used in
the JSON Schema SchemaDocument code for generating JSON from an
in-memory document.

Implementing this change revealed an omission in ArrayNode where bounds
were not being checked. For example, attempting to access index -1 would
raise an exception. This is now fixed and covered in a test case.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/SchemaDocument.java
+++ b/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/SchemaDocument.java
@@ -15,16 +15,15 @@
 
 package software.amazon.smithy.jsonschema;
 
-import java.util.Arrays;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.logging.Logger;
 import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.NodePointer;
 import software.amazon.smithy.model.node.ObjectNode;
-import software.amazon.smithy.model.node.StringNode;
 import software.amazon.smithy.model.node.ToNode;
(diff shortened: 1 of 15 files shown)


smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/SchemaDocument.java
```diff
@@ -182,46 +146,37 @@ public Schema getRootSchema() {
...
-                prefix += "/" + refs[position];
-            } else {
-                prefix += refs[position];
-            }
+        // An empty pointer returns the root schema.
+        if (pointerParts.isEmpty()) {
+            return Optional.of(getRootSchema());
+        }
+
+        // Compute the part of the pointer that points add a literal entry in
```
kstich: s/add/at

PR #301, opened 2020-03-04
Add a Node object mapper

mtdowling:
Add Node object mapper

A Node object mapper is a useful abstraction for refactoring how
SmithyBuild transforms and plugins are implemented and simplifying
the definition of configurable validators. The NodeMapper is meant
to be very simple: there's no annotations, there's currently no
way to customize it beyond whether or not nulls are serialized and
if it fails/warns when missing keys are deserialized. It handles
the serialization/deserialization of primitives, strings, collections,
maps, POJOs, File, Path, URI, URL, Enum, and a few other things. The
way types are serialized and deserialized is meant to be compatible
with Jackson if we ever need to move over to it for whatever reason.
Compatibility means that we could swap out the NodeMapper implementation
in the future for Jackson or Jackson-JR if needed without breaking
the existing API. We can potentially open up the NodeMapper abstractions
for extension later, but that might lead us further away from Jackson
compatibility.

Why are we not just adding a dependency on Jackson? Mainly because it's
a ridiculously popular serialization library that spans lots of versions.
Adding a dependency on Jackson would cause a headache for the Smithy team,
service teams, and build teams within Amazon trying to deal with
depenency version conflicts. While implementing an object mapper adds a
lot of complexity and code to maintain, keeping dependencies minimal is
one of our primary project goals. Adding NodeMapper isn't a one-way door:
we can swap it out for Jackson databind in the future if needed.


smithy-model/src/main/java/software/amazon/smithy/model/node/DefaultNodeDeserializers.java
```diff
@@ -0,0 +1,594 @@
...
+            try {
+                return factory.create(value);
+            } catch (Exception e) {
+                throw NodeDeserializationException.fromContext(
+                        targetType, pointer, node, e, e.getMessage());
+            }
+        };
+    };
+
+    private static final List<ObjectCreatorFactory> DEFAULT_FACTORIES = ListUtils.of(
```
kstich: Minor: Would be good to have a comment here indicating that these are priority ordered, and what determined the priority.

smithy-model/src/main/java/software/amazon/smithy/model/node/DefaultNodeSerializers.java
```diff
@@ -0,0 +1,383 @@
...
+
+            // Remove the current value from the set to ensure that it can be serialized
+            // multiple times (like in a List<T>).
+            serializedObjects.remove(value);
+
+            return new ObjectNode(mappings, SourceLocation.NONE);
+        }
+    };
+
+    // The priority-ordered list of default serializers.
```
kstich: Minor: would be nice to know what determined this priority for any future updating.

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeMapper.java
```diff
@@ -0,0 +1,466 @@
...
+     *     method that matches the string value.</li>
+     *     <li>Built-in support for URI, URL, Pattern, Path, and File</li>
+     *     <li>When deserializing an object, any target object that provides a public static method named
+     *     {@code builder} that returns an instance of {@link SmithyBuilder} is invoked, and the builder is then
+     *     mutated using bean like setters (with an optional "set") prefix, until finally, the build method is
+     *     called and its result is returned.</li>
+     *     <li>When deserializing an object, the last thing tried is to find a public, zero-arg constructor,
+     *     and then the object is mutated using bean-style setter conventions for each key-value pair.</li>
+     *     <li>NodeMapper does not support non-static inner classes, classes with generic parameters, or
+     *     generic arrays. Support for these may be added int he future.</li>
```
kstich: s/int he/in the

PR #305, opened 2020-03-07
Update smithy-build to use NodeMapper

mtdowling:
* Update smithy-build to use NodeMapper
    
    ```
    The NodeMapper is now used to load smithy-build.json configuration
    files, reducing the boilerplate needed to deserialize it (and the
    error-prone need to manually update the deserialization code as things
    are added to the POJOs). The #toNode methods and methods were removed
    from the build/model/* classes for the same reasons.
    
    An explicit POJO that defines the format of smithy-build-info.json files
    was added, and the NodeMapper is now used to serialize it. This should
    give the format a well-defined structure and allow other tools to
    properly deserialize it too.
    
    Finally, a ConfigurableSmithyBuildPlugin abstract class was added that
    makes it easier to add NodeMapper deserialization into a plugin specific
    configuration POJO. Most plugins will ultimately extend from this class
    and get deserialization for free.
    ````

* Adds support for more build transform options
    ````
    This commit updates smithy-build so that model transformations accept
    an object of configuration values rather than just a list of strings.
    This makes transformers far more powerful and configurable. In order to
    maintain backward compatibility with existing smithy-build.json files,
    an array of strings provided for a transformer is automatically
    transformed into `{"__args": [X]}` where "X" is the original array of
    arguments.
    ```

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/TraitRemovalUtils.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/TraitRemovalUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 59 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/transforms/TraitRemovalUtils.java
```diff
@@ -16,27 +16,26 @@
...
+
+    static Pair<Set<ShapeId>, Set<String>> parseTraits(Set<String> ids) {
         Set<ShapeId> traitNames = new HashSet<>();
         Set<String> traitNamespaces = new HashSet<>();
 
-        for (String arg : arguments) {
+        for (String arg : ids) {
             if (arg.endsWith("#")) {
                 traitNamespaces.add(arg.substring(0, arg.length() - 1));
-            } else if (arg.equals(Prelude.NAMESPACE)) {
```
kstich: This seems like a pretty rare case to run in to, but calling out that we're removing this without having deprecated it in 0.9.x.

mtdowling: I'll add a PR to master to emit a warning.

PR #308, opened 2020-03-11
Fix idRef doc examples

srchase:
This PR fixes idRef doc examples to use the correct trait namespace, target type and fully qualified shapeId.

### DIFF
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -3558,35 +3558,35 @@ Given the following model,
                 "smithy.example#InvalidShape1": {
                     "type": "string",
                     "traits": {
-                        "smithy.api#integerRef": "NotFound"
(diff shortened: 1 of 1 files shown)


docs/source/spec/core.rst
```diff
@@ -3558,35 +3558,35 @@ Given the following model,
...
                 "smithy.example#ValidShape2": {
                     "type": "string",
                     "traits": {
-                        "smithy.api#integerRef": "MyShape"
+                        "smithy.example#integerRef": "smithy.example#MyShape"
                     }
                 },
                 "smithy.example#MyShape": {
-                    "type": "string"
+                    "type": "integer"
```
kstich: Looks like this also needs to be fixed in the IDL example!

srchase: Yes! fixed.

Issue #241, opened 2020-01-03
[0.10] Emit warnings for operations and resources not bound to any service


mtdowling:
When building models that span multiple files, it's easy to forget to bind an operation or resource to a service. We should add a built-in validator that emits warning events when operations or resources are found that are not bound within the closure of a service. This is similar to  the [UnreferencedShapes](https://awslabs.github.io/smithy/spec/validation.html#unreferencedshape) validator, but it is run automatically and doesn't check anything other than operations or resources.

PR #310, opened 2020-03-12
Add references to operations and resources

srchase:
This PR adds service references to all operations and resources in the test models for the following projects:
* smithy-aws-traits
* smithy-linters
* smithy-model
* smithy-mqtt-traits

### DIFF
--- a/smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/apigateway/errorfiles/valid-integration.json
+++ b/smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/apigateway/errorfiles/valid-integration.json
@@ -4,6 +4,11 @@
         "ns.foo#Service1": {
             "type": "service",
             "version": "2018-03-17",
+            "operations": [
+                {
+                    "target": "ns.foo#Operation"
+                }
+            ],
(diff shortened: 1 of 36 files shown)


mtdowling:
Why?

srchase:
> Why?

This clears the way for adding a validator as described in #241

kstich:
> > Why?
> 
> This clears the way for adding a validator as described in #241

Maybe the validator should only emit when a service shape exists in the model? If so, would we want something that emits when operations are found without a service at all?

mtdowling:
> Maybe the validator should only emit when a service shape exists in the model?

+1

> If so, would we want something that emits when operations are found without a service at all?

I think it would be fine to find orphans with a shallow search. Is this operation referenced by a resource or service? Then don't emit. Is this resource referenced by a resource or service? Then don't emit. This would acheive the goal of detemining if an operation or resource is completely disconnected and forgotten to be bound somewhere. While it wouldn't tell you if an operation that's bound to a resource is disconnected from a service, it would tell you that it transitively is disconnected because the resouce it's connect to isn't bound to a service.

srchase:
Updating models when introducing validator in #319.

Issue #278, opened 2020-02-19
[0.10] Add validator to check for traits applied to prelude shapes


mtdowling:
We should emit an ERROR for when traits are applied to prelude shapes outside of the prelude. Applying traits to prelude traits could have wide and unintended consequences when defining and merging models.

PR #317, opened 2020-03-15
Disallow traits applied to public prelude shapes

mtdowling:
Applying a trait to a prelude shape is a recipe for disaster. For
example, if someone marked the smithy.api#String shape as private,
virtually every model would break.

*Issue #, if available:*

*Description of changes:*


kstich:
This closes #278, correct?

PR #319, opened 2020-03-17
Migrate UnreferencedShapeValidator

srchase:
Implements https://github.com/awslabs/smithy/issues/241 by migrating the existing UnreferencedShapeValidator from smithy-linters to smithy-model so that the validation is always run against the model.

A check is added so that only models with services will emit `note` validation events when a shape is not referenced within the model.


JordonPhillips:
probably want to squash some of those commits

PR #324, opened 2020-03-26
Make error message more clear for duplicate shapes

mtdowling:
This commit updates model loaders to fail at the point that a duplicate
shape definition is detected in a file. This results in clearer error
messages when a shape is defined with the same name, particularly if one
conflicting shape supports members while the other does not. Previously,
in this case, a cryptic error message would be returned stating that a
member cannot be added to a builder:

    Member `example.foo#Bar$Baz` cannot be added to software.amazon.smithy.model.shapes.OperationShape$Builder

This now becomes:

    [ERROR] -: Duplicate shape definition for `example.foo#Bar$Baz` found at `/path/file.smithy [3, 1]` and `/path/file.smithy [5, 1]` | Model

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelLoader.java
```diff
@@ -54,28 +53,18 @@ private ModelLoader() {}
      * @throws SourceException if there is an error reading from the contents.
      */
     static boolean load(String filename, Supplier<InputStream> contentSupplier, LoaderVisitor visitor) {
-        try {
```
mtdowling: This try/catch wasn't necessary

PR #326, opened 2020-03-26
Make enum trait a list of structures

mtdowling:
The enum trait is currently a map of enum value to the enum definition.
This means that models often look like this:

```
@enum(foo: {}, bar: {})
string MyString
```

Ideally, a name and documentation are provided too:

```
@enum(
    foo: {name: "FOO", documentation: "foo docs..."},
    bar: {name: "BAR", documentation: "bar docs..."})
```

However, the difference between the key and "name" has resulted in
confusion as to which is which. This is caused by a couple things:

1. The key doesn't tell you what it defines. You have to know that the
the key defines the value of the enum to make sense of the trait.
2. The key has to be quoted if it deviates from the allows ABNF for node
object keys. This has also led to confusion.

Although this is a breaking change, using a list of structures where
everything has a name will result in models that are easier to
understand, and hopefully modelers will more often include properties
like `name` and `documentation`.

This now looks like:

```
@enum([
    {
        value: "foo",
        name: "FOO",
        documentation: "foo docs..."
    },
    {
        value: "bar",
        name: "BAR",
        documentation: "bar docs..."
    }
])
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
+++ b/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
@@ -127,7 +127,9 @@ structure httpApiKeyAuth {
 }
 
 @private
-@enum(header: {},  query: {})
+@enum([
+    {value: "header"},
+    {value: "query"}])
 string HttpApiKeyLocations
 
 /// Indicates that an operation can be called without authentication.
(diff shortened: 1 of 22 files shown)


smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
```diff
@@ -317,21 +320,25 @@ string title
...
+    member: EnumDefinition
 }
 
 /// An enum definition for the enum trait.
 @private
-structure EnumConstantBody {
+structure EnumDefinition {
+    /// Defines the enum value that is sent over the wire.
+    @required
+    value: String,
```
JordonPhillips: Since we're changing this anyway, we should add `@length(min: 1)`

JordonPhillips: To be clear, I meant that the value shouldn't be an empty string, though the list for sure should not be empty either

PR #328, opened 2020-03-27
Ensure HttpApiKeyAuth headers are added to CORS

mtdowling:
This change updates SecuritySchemeConverters to take in the trait being
converted so that things like HttpApiKeyAuth can add their designated
header to the CORS list of allowed headers.

*Issue #, if available:*

*Description of changes:*


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/security/HttpApiKeyAuthConverter.java
```diff
@@ -42,4 +44,9 @@ public SecurityScheme createSecurityScheme(Context<? extends Trait> context, Htt
                 .in(trait.getIn().toString())
                 .build();
     }
+
+    @Override
```
kstich: Since this is a new override, should a new test entry be expected?

PR #329, opened 2020-03-27
Adds security to individual operations

mtdowling:
We previously weren't adding security to individual operations if the
operation differend from the security of the operation differed from the
service because of an applied authorizer trait.

*Issue #, if available:*

*Description of changes:*


smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddAuthorizers.java
```diff
@@ -79,6 +83,25 @@
                 .orElse(requirement);
     }
 
+    @Override
+    public OperationObject updateOperation(Context context, OperationShape shape, OperationObject operation) {
+        ServiceShape service = context.getService();
+        AuthorizerIndex authorizerIndex = context.getModel().getKnowledge(AuthorizerIndex.class);
+
+        // Get the resolved security schemes of the service and operation, and
+        // only auth security if it's different than the service.
```
kstich: s/only auth security/only add security

PR #336, opened 2020-03-30
Make color scheme less clinical

mtdowling:
This commit updates the website color scheme to be a brownish red / iron
ore color scheme rather than the more clinical feeling blue. Several
other theme changes are made, including putting borders above headings
rather than below, automatically numbering headings in specification
pages using CSS, and adding contrast between the nav sidebar and the
content.

Homepage:

Screen Shot 2020-03-30 at 10 58 33 AM

Top of content pages:

Screen Shot 2020-03-30 at 11 01 36 AM

Inside of content pages:

Screen Shot 2020-03-30 at 10 59 17 AM

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/themes/smithy/landing.html
+++ b/docs/themes/smithy/landing.html
@@ -6,9 +6,9 @@
                 <div class="splash-column">
                     <h1>Smithy, an extensible API definition language</h1>
                     <div class="headline">
-                        Smithy is a protocol-agnostic interface definition language
-                        and set of tools for building clients, servers, and
-                        documentation for any programming language.
+                        Smithy <span class="splash-highlight">defines</span> and
+                        <span class="splash-highlight">generates</span> services,
+                        clients, and documentation for any protocol.
                     </div>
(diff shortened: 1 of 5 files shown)


kstich:
In general, I like the scheme a lot better. My only comment is that there appears to be no visual separation between the persistent navbar and the front page's "banner" area, color or otherwise.

mtdowling:
Updated with more visual distinction between the header and the splash thingy. I updated the tag line too to be less wordy... What do you think?

Screen Shot 2020-03-30 at 11 29 28 AM

jgalperin:
Looking good! To my eye, I think the difference in color is too contrasting. Maybe we could use something like [this tool](https://coolors.co/fa7921-fe9920-7f504b-b9a44c-0c4767) to find complimentary colors that are still distinct enough?

mtdowling:
How about this then?

Screen Shot 2020-03-30 at 11 50 23 AM

PR #337, opened 2020-03-31
Add service closure name conflict validation

mtdowling:
This commit validates and standardizes the behavior that has existed for
many years in AWS SDKs: requiring that shapes in the closure of a
service are case-insensitively unique regardless of their namespaces
(with some exceptions for simple types and lists/sets). This change
ensures that SDK code gen does not need to persist the namespaces in
their generated code, and that they can create their own
namespaced/siloed copies of shapes for each service (removing the need
to publish and pull in dependencies on shared types).

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ServiceValidator.java
```diff
@@ -16,55 +16,210 @@
...
+    private ValidationEvent conflictingNames(Severity severity, ServiceShape shape, Shape subject, Shape other) {
+        // Whether it's a should or a must based on the severity.
+        String declaration = severity == Severity.DANGER || severity == Severity.ERROR ? "must" : "should";
+
+        return ValidationEvent.builder()
+                .eventId(getName())
+                .severity(severity)
+                .shape(subject)
+                .message(String.format(
+                        "Shape name `%s` conflicts with `%s` in the `%s` service closure. Shapes in the closure "
```
kstich: Subtle tweak to the wording - "These shapes in the closure..." - to point that the sentence is referring to explicitly these shapes. I can foresee some reading the last part and being confused when encountering this taken from the test outputs, with one saying must and the other saying should worded to a general case:

```
[DANGER] another.ns#ListC: Shape name `ListC` conflicts with `ns.foo#ListC` in the `ns.foo#MyService` service closure. Shapes in the closure of a service must have case-insensitively unique names regardless of their namespaces. | Service
[NOTE] ns.foo#String: Shape name `String` conflicts with `smithy.api#String` in the `ns.foo#MyService` service closure. Shapes in the closure of a service should have case-insensitively unique names regardless of their namespaces. | Service
```

PR #340, opened 2020-03-31
[0.10] Apply streaming trait directly to shapes

JordonPhillips:
This updates the streaming trait to apply directly to shapes instead of
members. This was previously restricted to members because it made the
model easier to validate and grok at a glance. However, keeping it on
members introduces potential errors in code generation. A given blob
could have both the streaming trait and the media type trait, for
instance. Both of these traits could result in a new type being created,
but since the streaming trait is applied to a member that type is
implicit. The code generator would have to try to generate a name that
doesn't conflict with any existing names, which is very error prone and
likely would result in unfortunate names. Moving this trait to the
shape directly makes the naming concern an explicit part of modeling,
where it will only have to be handled once.

This also brings the trait in line with the general best practice of
having traits that impact the type generated apply directly to the
shapes they impact.

### DIFF
--- a/docs/source/spec/core.rst
+++ b/docs/source/spec/core.rst
@@ -3353,10 +3353,7 @@ Summary
     not be stored in memory, or that the size of the data stored in the shape
     is unknown at the start of a request.
 Trait selector::
-    ``operation -[input, output]-> structure > :test(member > blob)``
-
-    This trait may only be applied to top level members of operation inputs and
-    outputs.
+    ``blob``
 Value type
     ``structure``
 
@@ -3377,6 +3374,10 @@ optional members:
         In an HTTP-based protocol, for instance, this indicates that the
         ``content-length`` header must be set.
 
+Shapes targeted by this trait MAY NOT be used outside of top level operation
+inputs and operation outputs. Additionally, only one member of a structure may
+target a shape with this trait.
+
 .. tabs::
 
     .. code-tab:: smithy
@@ -3386,9 +3387,11 @@ optional members:
         }
 
         structure StreamingOutputWrapper {
-            @streaming
-            output: Blob,
+            output: StreamingBlob,
         }
+
+        @streaming
+        blob StreamingBlob
 
 Constraint traits
 =================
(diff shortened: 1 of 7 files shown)


docs/source/spec/core.rst
```diff
@@ -3377,6 +3374,10 @@ optional members:
         In an HTTP-based protocol, for instance, this indicates that the
         ``content-length`` header must be set.
 
+Shapes targeted by this trait MAY NOT be used outside of operation inputs and
```
mtdowling, line 3377: Maybe state it has to be top-level operation input and output members.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/StreamingTraitValidator.java
```diff
@@ -0,0 +1,67 @@
...
+        Set<ShapeId> topLevelIoShapes = model.shapes(OperationShape.class)
+                .flatMap(operation -> SetUtils.of(operation.getInput(), operation.getOutput()).stream())
+                .flatMap(OptionalUtils::stream)
+                .collect(Collectors.toSet());
+
+        Set<MemberShape> streamingMembers = model.shapes(MemberShape.class)
+                .filter(member -> member.getMemberTrait(model, StreamingTrait.class).isPresent())
+                .collect(Collectors.toSet());
+
+        List<ValidationEvent> events = streamingMembers.stream()
```
mtdowling, line 43: I think breaking this into two methods would make this validator easier to understand -- one method validates that the target is correct and the other that only a single member of a structure can target a streaming blob.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/StreamingTraitValidator.java
```diff
@@ -0,0 +1,67 @@
...
+                .filter(member -> !topLevelIoShapes.contains(member.getContainer()))
+                .map(member -> error(member, String.format(
+                        "The shape %s has the smithy.api#streaming trait, and so may only be targeted by "
+                                + "top-level operation inputs and outputs.",
+                        member.getTarget())))
+                .collect(Collectors.toList());
+
+        events.addAll(streamingMembers.stream()
```
mtdowling, line 51: This one-liner is pretty tricky and even plays with indentation in a non-standard way. It make make this more readable to break this across multiple lines.

PR #357, opened 2020-04-03
Update model version from 0.5.0 to 1.0

srchase:
Updates Model version from 0.5.0 to 1.0.


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelLoader.java
```diff
@@ -73,12 +73,13 @@ static boolean load(String filename, Supplier<InputStream> contentSupplier, Load
     // is then used to delegate loading to different versions of the
     // Smithy JSON AST format.
     //
-    // This loader supports version 0.5.0. 0.4.0 support was removed in 0.10.
+    // This loader supports version 1.0. 0.5.0 and 0.4.0 support was removed in
```
mtdowling, line 76: Looks like a typo or incomplete thought here?

srchase: Next line below completes the sentence.  Can this be on one line instead of wrapping?

kstich: Maybe this be phrased differently as to not have all the numbers right next to each other, likely the source of confusion.

>This loader supports version 1.0. Support for loading versions 0.5.0 and 0.4.0 was removed...

kstich:
I'd say `1.0.0` is the right call, we may have the need for minor vs patch level updates in the future.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelLoader.java
srchase: Swapped the wording on this to make it more clear.

PR #363, opened 2020-04-07
Update externalDocumentation to a named map

kstich:
This commit updates the @externalDocumentation trait to be a map of
named urls instead of a single url. This provides more context for
the given links and enables the ability to have multiple links per
trait application.

An update to the OpenApi conversion has been made to utilize this
when populating "externalDocs" fields. A configurable, priority
ordered list of url names is available through
`"openapi.externalDocs"`; it defaults to "Homepage", "Api
Reference", "User Guide", "Developer Guide", "Reference", and
"Guide". This list is compared case insensitively. This also
fixes an issue where "externalDocs" fields could be generated as
strings instead of the properly formatted external documentation
object in the resulting OpenApi document.

-----

### DIFF
--- a/docs/source/guides/converting-to-openapi.rst
+++ b/docs/source/guides/converting-to-openapi.rst
@@ -196,6 +196,28 @@ openapi.use.xml (boolean)
     Enables converting Smithy XML traits to OpenAPI XML properties. (this
     feature is not yet implemented).
 
+openapi.externalDocs ([string])
+    Limits the source of converted "externalDocs" fields to the specified
+    priority ordered list of names in an :ref:`externaldocumentation-trait`.
+    This list is case insensitive. By default, this is a list of the following
+    values: "Homepage", "API Reference", "User Guide", "Developer Guide",
+    "Reference", and "Guide".
+
+    .. code-block:: json
+
+        {
+            "version": "1.0",
+            "plugins": {
+                "openapi": {
+                    "service": "smithy.example#Weather",
+                    "openapi.externalDocs": [
+                        "Homepage",
+                        "Custom"
+                    ]
+                }
+            }
+        }
+
 openapi.keepUnusedComponents (boolean)
     Set to ``true`` to prevent unused components from being removed from the
     created specification.
--- a/smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.json
+++ b/smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.json
@@ -8,7 +8,9 @@
                     "selector": "service"
                 },
                 "smithy.api#documentation": "Specifies the source of the caller identifier that will be used to throttle API methods that require a key.",
-                "smithy.api#externalDocumentation": "https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-api-key-source.html",
+                "smithy.api#externalDocumentation": {
+                    "Developer Guide": "https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-api-key-source.html"
+                },
                 "smithy.api#tags": [
                     "internal"
                 ]
--- a/smithy-model/src/main/java/software/amazon/smithy/model/traits/ExternalDocumentationTrait.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/traits/ExternalDocumentationTrait.java
@@ -17,37 +17,112 @@ package software.amazon.smithy.model.traits;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Collections;
+import java.util.LinkedHashMap;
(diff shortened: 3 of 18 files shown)


smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.json
```diff
@@ -8,7 +8,9 @@
                     "selector": "service"
                 },
                 "smithy.api#documentation": "Specifies the source of the caller identifier that will be used to throttle API methods that require a key.",
-                "smithy.api#externalDocumentation": "https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-api-key-source.html",
+                "smithy.api#externalDocumentation": {
+                    "User Guide": "https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-api-key-source.html"
```
mtdowling, line 12: Should we call this "Developer Guide" so that it matches the nomenclature of the referenced docs?

docs/source/guides/converting-to-openapi.rst
```diff
@@ -196,6 +196,27 @@ openapi.use.xml (boolean)
     Enables converting Smithy XML traits to OpenAPI XML properties. (this
     feature is not yet implemented).
 
+openapi.use.convertedExternalDocs ([string])
```
mtdowling, line 199: I'm going to be renaming these anyways..., but I'd probably just call this `openapi.externalDocs`.

docs/source/guides/converting-to-openapi.rst
```diff
@@ -196,6 +196,27 @@ openapi.use.xml (boolean)
...
+    Limits the source of converted "externalDocs" fields to the specified set
+    of names in an :ref:`externaldocumentation-trait`. By default, this is a
+    set of the following values: "Homepage", "Api Reference", "User Guide",
```
mtdowling, line 202: 1. Can we add "Developer Guide" to this list?
2. Why Api and not API? Since these are arbitrary and meant to be displayed directly, I think we should probably use API.
3. Is this list case-insensitive? I think it *probably* should be. Thoughts?
4. Should probably not say "set of names", but "priority ordered list of names"

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExternalDocumentationTrait.java
```diff
@@ -17,37 +17,119 @@
...
+     *
+     * @return returns the external documentation mapping.
+     */
+    public Map<String, String> getUrls() {
+        return urls;
+    }
+
+    @Override
+    protected Node createNode() {
+        return urls.entrySet().stream()
```
mtdowling, line 74: Can this be simplified to: `ObjectNode.fromStringMap(urls)`

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude-traits.smithy
```diff
@@ -57,7 +57,10 @@ string documentation
 
 /// Provides a link to additional documentation.
 @trait
-string externalDocumentation
+map externalDocumentation {
```
mtdowling, line 60: I suggest applyinh the `@length` trait to force it to have > 1 entry

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExternalDocumentationTrait.java
```diff
@@ -17,37 +17,119 @@
...
-        this(value, SourceLocation.NONE);
-    }
+    public ExternalDocumentationTrait(Builder builder) {
+        super(ID, builder.sourceLocation);
+        this.urls = Collections.unmodifiableMap(builder.urls);
 
-    public static final class Provider extends StringTrait.Provider<ExternalDocumentationTrait> {
-        public Provider() {
-            super(ID, ExternalDocumentationTrait::new);
+        if (urls.isEmpty()) {
```
mtdowling, line 47: This validation could be removed and just handled by the model validation process if you add the `@length` trait to the map in the prelude model.

docs/source/guides/converting-to-openapi.rst
```diff
@@ -196,6 +196,28 @@ openapi.use.xml (boolean)
     Enables converting Smithy XML traits to OpenAPI XML properties. (this
     feature is not yet implemented).
 
+openapi.externalDocs ([string])
+    Limits the source of converted "externalDocs" fields to the specified
+    priority ordered list of names in an :ref:`externaldocumentation-trait`.
+    This list is case insensitive. By default, this is a list of the following
+    values: "Homepage", "API Reference", "User Guide", "Reference", and
```
mtdowling, line 203: Need to add "Developer Guide" here.

kstich: 🤦‍♂️

PR #364, opened 2020-04-08
Update open api json schema pojos

mtdowling:
*Issue #, if available:*

*Description of changes:*


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/Smithy2OpenApi.java
```diff
@@ -40,12 +46,29 @@ public String getName() {
...
-        context.getSettings().getStringMap().forEach(converter::putSetting);
         context.getPluginClassLoader().ifPresent(converter::classLoader);
 
-        ShapeId shapeId = ShapeId.from(context.getSettings()
-                .expectStringMember(OpenApiConstants.SERVICE)
-                .getValue());
+        // Remove deprecated "openapi." prefixes from configuration settings.
+        ObjectNode mapped = context.getSettings();
+        for (Map.Entry<String, Node> entry : mapped.getStringMap().entrySet()) {
+            if (entry.getKey().startsWith("openapi.")) {
```
kstich, line 54: The intent to help here is really nice, but if it's something we want to do it needs a little more work to not cause issues. This will give the wrong value for what used to be `openapi.aws.jsonContentType` - removing `.aws` is an option. It will give the wrong methodology completely for `openapi.use` entries - however only `openapi.use.xml` was documented and was not implemented.

PR #365, opened 2020-04-09
Collapse eventstream with the streaming trait

JordonPhillips:
The eventstream and streaming trait fundamentally work in similar
ways, and have identical validation rules. This merges the two traits
to simplify the model and reduce duplication.


smithy-model/src/main/java/software/amazon/smithy/model/knowledge/HttpBindingIndex.java
```diff
@@ -368,15 +368,13 @@ private String determineContentType(
...
-                if (binding.getMember().hasTrait(EventStreamTrait.class)) {
-                    return eventStreamContentType;
-                }
-
                 Shape target = model.getShape(binding.getMember().getTarget()).orElse(null);
                 if (target == null) {
                     // Can't determine the content-type because the model is broken :(
                     // Let other parts of the validation system point this out.
                     break;
+                } else if (target.isUnionShape() && target.hasTrait(StreamingTrait.class)) {
```
mtdowling, line 376: Should we add an `isEventStream` method somewhere to centralize this behavior? Like on the EventStreamTrait as a static method?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/streaming-trait.errors
```diff
@@ -1,3 +1,2 @@
-[ERROR] ns.foo#InvalidStreamingStructure$Body: The shape ns.foo#StreamingBlob has the smithy.api#streaming trait, and so may only be targeted by top-level operation inputs and outputs. | StreamingTrait
-[ERROR] ns.foo#InvalidStreamingStructure2$Body: The shape ns.foo#StreamingBlob2 has the smithy.api#streaming trait, and so may only be targeted by top-level operation inputs and outputs. | StreamingTrait
 [ERROR] ns.foo#InvalidStreamingOutput: Only a single member of a structure can target a shape marked with the `streaming` trait, but it was found on the following members: `StreamingBlob1`, `StreamingBlob2` | ExclusiveStructureMemberTrait
+[ERROR] ns.foo#InvalidNestedStream$NestedStream: This shape has an invalid `MEMBER_TARGET` relationship to a structure, `ns.foo#StreamingInput`, that contains a stream | StreamingTrait
```
mtdowling, line 2: Nice, did this get called out in the spec?

smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/validators/MqttPublishInputValidator.java
```diff
@@ -42,7 +42,10 @@
     private Stream<ValidationEvent> validateOperation(Model model, OperationShape operation) {
         return OptionalUtils.stream(operation.getInput().flatMap(model::getShape).flatMap(Shape::asStructureShape))
                 .flatMap(input -> input.getAllMembers().values().stream()
-                        .filter(member -> member.hasTrait(EventStreamTrait.class))
+                        .filter(member -> {
+                            Shape target = model.expectShape(member.getTarget());
+                            return target.isUnionShape() && target.hasTrait(StreamingTrait.class);
```
mtdowling, line 47: Could consolidate here if there's a helper method

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/streaming-trait.errors
JordonPhillips: This was already supposed to be a thing. I ended up cannibalizing the implementation from the event stream trait since it was easier to follow and less prone to bugs.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/HttpBindingIndex.java
JordonPhillips: sure

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/EventStreamIndex.java
```diff
@@ -26,7 +26,7 @@
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.ToShapeId;
-import software.amazon.smithy.model.traits.EventStreamTrait;
+import software.amazon.smithy.model.traits.StreamingTrait;
 
 /**
```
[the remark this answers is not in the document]
JordonPhillips: I think you would need to do a lot more massaging than just a rename. The `EventStreamInfo` has a bunch of stuff about things like initial input / initial output that don't make sense for blob streams, for instance.

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
-            room: String,
+    When applied to a blob, this simply means that the data could be very
+    large and thus should not be stored in memory or that the size is unknown
+    at the start of the request.
 
-            @eventStream
-            messages: Message,
-        }
+    When applied to a union, it indicates that shape represents an
+    `event stream <event-streams>`.
```
mtdowling, line 29: This is a broken link

docs/source/spec/core/stream-traits.rst
```diff
@@ -394,7 +303,7 @@ Initial messages
 ----------------
```
mtdowling, line 303: This should be under the "Event streams" section

docs/source/spec/core/stream-traits.rst
```diff
@@ -344,10 +253,10 @@ stream in its output:
             }
         }
 
-Multi-event client behavior
-===========================
+Client Behavior
```
mtdowling, line 256: Can you make this "Client behavior"? And it seems like this should go after "Initial messages" since it refers to that section before it's introduced.

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
-provide an abstraction to end-users that allows values to be produced or
-consumed asynchronously for the targeted event structure. Because a
-single-event event stream does not utilize named events like a multi-event
-event stream, functionality used to dispatch based on named events is
-unnecessary. Clients MUST provide access to the
-:ref:`initial-message <initial-messages>` of an event stream when necessary.
+.. _event-streams:
 
+-------------
+Event Streams
```
mtdowling, line 75: "Event streams"

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
-                    }
-                }
-            }
-        }
+    * - Property
+      - Type
+      - Description
+    * - requiresLength
+      - ``boolean``
+      - Indicates that the stream must have a known size.
```
mtdowling, line 47: This can only be set when attached to blobs, right? Should this maybe be a separate trait to make this discrepancy less weird?

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
 .. tabs::
 
     .. code-tab:: smithy
 
-        namespace smithy.example
-
-        operation SubscribeToMovements {
-            output: SubscribeToMovementsOutput
+        operation StreamingOperation {
+            output: StreamingOutputWrapper,
```
mtdowling, line 62: It's not really a "wrapper", it's the output. Can you name this "StreamingOperationOutput" or "StreamingOperationResponse"? Any maybe add another top-level member like... "streamId" or something contrived to show that other input can coexist?

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
--------------------------
+An operation can send an event stream as part of its input or output. An
+event stream is formed when an input or output member of an operation targets
+a union marked with the :ref:`streaming-trait`.
 
-A *multi-event event stream* is an event stream that streams any number of
-named event structure shapes defined by a union. It is formed when the
-``eventStream`` trait is applied to a member that targets a union. Each
+An event stream is capable of streaming any number of named event structure
+shapes defined by a union. It is formed when the ``streaming`` trait is
```
mtdowling, line 89: "It is formed when the ``streaming`` trait is applied to a union that is targeted by an input or output structure."

This is repetitive to the previous paragraph

docs/source/spec/core/stream-traits.rst
```diff
@@ -1,191 +1,100 @@
...
+An event stream is an abstraction that allows multiple messages to be sent
+asynchronously between a client and server. Event streams support both duplex
+and simplex streaming. The serialization format and framing of messages sent
+over event streams is defined by the :ref:`protocol <protocolDefinition-trait>`
+of a service.
 
--------------------------
-Multi-event event streams
--------------------------
+An operation can send an event stream as part of its input or output. An
```
mtdowling, line 84: Seems like the first sentence can be removed, this paragraph can be combined with the next one.

docs/source/spec/core/stream-traits.rst
```diff
@@ -201,10 +110,10 @@ stream in its input by referencing a member that targets a union:
         structure PublishMessagesInput {
             room: String,
```
mtdowling, line 112: Extra newline

docs/source/spec/core/stream-traits.rst
```diff
@@ -394,7 +303,7 @@ Initial messages
 ----------------
 
 An *initial message* is comprised of the top-level input or output members
-of an operation that are not targeted by the ``eventStream`` trait. Initial
+of an operation that do not target the event stream union. Initial
 messages provide an opportunity for a client or server to provide metadata
 about an event stream before transmitting events.
```
mtdowling, line 309: Let's just remove the "important" note below. All protocols should support initial messages, so it's strange to call it out here instead of just in the protocol documentation itself.

docs/source/spec/core/stream-traits.rst
```diff
@@ -440,8 +349,12 @@ service, followed by the events sent in the payload of the HTTP message.
             room: String,
```
mtdowling, line 349: This is my bad, but this needs to be marked as `@required` too :)

This also needs to change in the initial response example

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/EventStreamIndex.java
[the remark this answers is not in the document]
mtdowling: True

docs/source/spec/core/stream-traits.rst
JordonPhillips: Maybe. Is there not a use case for it for event streams?

mtdowling: I doubt it. Event streams usually don't have a predefined size of bounds.

PR #370, opened 2020-04-14
Convert options to text

mtdowling:
Use NodeMapper in API Gateway traits
    
    This commit updates API Gateway traits to use the NodeMapper and changes
    different getter/setter methods to use NodeMapper compatible API names.
    Further, support for omitting the toNode implementation of a class was
    added so that the NodeMapper can be used inside of a toNode method.
    Finally, support for omitting empty arrays and objects was added so the
    NodeMapper doesn't inject a bunch of empty values when it isn't
    desirable.

Add CONVERT_TO_TEXT to OPTIONS and cleanup
    
    CONVERT_TO_TEXT is necessary on OPTIONS request otherwise you may see
    500 errors from API Gateway. Further, this commit cleans up several
    getters/setters on the IntegrationTrait and MockIntegrationTrait to
    match what a code generator would produce and how the NodeMapper works.

### DIFF
--- a/smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/AuthorizerDefinition.java
+++ b/smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/AuthorizerDefinition.java
@@ -15,17 +15,13 @@
 
 package software.amazon.smithy.aws.apigateway.traits;
 
-import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import software.amazon.smithy.aws.traits.auth.SigV4Trait;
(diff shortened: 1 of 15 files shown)


smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/AuthorizerDefinition.java
```diff
@@ -182,7 +168,7 @@ public Builder toBuilder() {
                 .scheme(scheme)
                 .type(type)
                 .uri(uri)
-                .authType(authType)
```
mtdowling, line 185: Needed to change for the NodeMapper

PR #372, opened 2020-04-16
`smithy-model`: add support for `httpContentMd5` trait

skotambkar:
*Issue #, if available:*

*Description of changes:*
- add support for `httpContentMd5` trait

### DIFF
--- a/smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/HttpContentMd5Trait.java
+++ b/smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/HttpContentMd5Trait.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


mtdowling:
Thanks for the PR! Could you move this trait to the `smithy-aws-traits` package? I think that this trait isn't generic enough since I've never heard of any other service *requiring* a Content-MD5 on input like the way S3 does. The trait would go here: https://github.com/awslabs/smithy/blob/0.10/smithy-aws-traits/src/main/resources/META-INF/smithy/aws.protocols.json, in the `aws.protocols` namespace.

skotambkar:
I agree the trait isn't too generic, and can be moved to `smithy-aws-traits` package. But the trait isn't protocol specific, so curious why `aws.protocols` namespace? why not `aws.api` namespace instead as it seems to have traits like ClientDiscoveredEndpoint which target operations/services?

mtdowling:
Good question... I thought about aws.api too, but I kind of think this is protocol-specific in that it only applies to HTTP based application protocols.

skotambkar:
That seems more transport specific, but i guess it's better keeping it in `aws.protocol` rather than `aws.api`.

I have pushed a change for the trait to be in `aws.protocols` namespace.

A follow-up question wrt the pushed change is if we need to move the `HttpContentMd5Trait` class from `software.amazon.smithy.aws.traits` to `software.amazon.smithy.aws.traits.protocols` sub package?

Let me know once you get time to review this.

mtdowling:
Thanks. Yeah all the code needs to move to that package too.

skotambkar:
Moved `HttpContentMd5Trait` class  to `software.amazon.smithy.aws.traits.protocols` pkg 

Let me know if PR needs changes. Thanks!

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/HttpContentMd5Trait.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mtdowling, line 2: 2020

PR #379, opened 2020-04-16
Add CleanTraitDefinitions plugin

srchase:
Adds CleanTraitDefinitions transformer plugin that removes traits listed on AuthDefinitionTrait and ProtocolDefinitionTrait traits when a shape is removed from a model.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanTraitDefinitions.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanTraitDefinitions.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanTraitDefinitions.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private Stream<Shape> rebuildShapeWithAuthDef(
+            StructureShape shape,
+            SourceLocation location,
+            List<ShapeId> traits
+    ) {
+        AuthDefinitionTrait.Builder traitBuilder = AuthDefinitionTrait.builder();
+        traitBuilder.sourceLocation(location);
+        traits.forEach(traitBuilder::addTrait);
```
kstich, line 72: This could just be `traitBuilder.traits(traits);`

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanTraitDefinitions.java
```diff
@@ -0,0 +1,115 @@
...
+                    AuthDefinitionTrait authDefTrait = pair.getRight();
+                    List<ShapeId> traits = authDefTrait.getTraits();
+                    List<ShapeId> newTraits = excludeTraitsInSet(traits, removedShapeIds);
+
+                    // Return early if re-built list of traits is the same as existing list.
+                    if (traits.equals(newTraits)) {
+                        return Stream.empty();
+                    }
+
+                    return rebuildShapeWithAuthDef(pair.getLeft(), authDefTrait.getSourceLocation(), newTraits);
```
kstich, line 60: Since the only component of the trait here that needs to change is the `traits` field, this can be simpler. Instead of constructing a new trait, use the `toBuilder` and to modify it. Then the private `rebuild*` methods can be inlined since they would be reduced to only their current return statements.

PR #381, opened 2020-04-16
Remove allowing true for annotation traits

mtdowling:
This commit removes the support for setting annotation traits to true
because the previous rationale for supporting that no longer holds true.
The previous rationale was that supporting `true` makes writing the AST
JSON files easier. However, we've already crossed the point where we
don't want humans writing the JSON AST, and we want it to be as easy to
parse as possible.

The AST format now requires that `{}` is provided for annotation traits.
The IDL no longer supports setting a trait to `true` or `null`, as that
was only supported in order to provide isomorphism with the AST support.
We previously didn't persist which model format a pending trait
definition came from, so it was necessary to give them the same feature
set. However, now the Smithy loader maintains a boolean value that tells
the loader whether or not a value is an "annotation", meaning trait with
no value. This isn't a feature used in the AST loader, but is used in
the IDL loader.

Further, the BooleanTrait abstract class was renamed to AnnotationTrait.
This is something I wanted to do for a long time, but never did since it
would cause so much churn, and the name wasn't that far off from what it
represents. However, now that `true` isn't supported for annotation
traits, the name BooleanTrait makes no sense. This commit renames it and
takes the BC hit because it's now worth it.

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/traits/AnnotationTrait.java
```diff
@@ -15,31 +15,32 @@
...
     protected final Node createNode() {
-        return new BooleanNode(true, getSourceLocation());
+        return new ObjectNode(Collections.emptyMap(), getSourceLocation());
     }
 
     /**
      * Trait provider that expects a boolean value of true.
      */
-    public static class Provider<T extends BooleanTrait> extends AbstractTrait.Provider {
+    public static class Provider<T extends AnnotationTrait> extends AbstractTrait.Provider {
```
kstich, line 43: Looks like there's now dead code in this class's `createTrait` method that should be cleaned up.

PR #388, opened 2020-04-20
Major selector enhancements

mtdowling:
This commit adds several new features to selectors, updates its grammar, and rewrites the parser to be more adaptable to change.

* Add support for numeric attribute comparators. You can match values that are parsed as numbers using `>`, `>=`, `<`, `<=`. This can be used to find out things like `[trait|httpError >= 500]`.
* Add support for `!=` to check if an attribute exists but isn't equal to the given value.
   The "!=" attribute comparator is used to find shapes that have an attribute, but the attribute is not equal to a given value. For example, the following selector: `[trait|error!=client]`, is a simplification of the following: 
`[trait|error]:not([trait|error=client])`
* Add support for CSV values in attribute selectors. This enables checking if a resolved attribute value matches one or more comparison values, allowing for things like `[trait|httpError = 500, 501]`.
* Fix "~=" in docs to properly reference "^=". The docs previously documented the wrong comparator.
* Adds support for numbers in attribute selectors. You no longer need to quote numbers in selector attribute values.
* Shape IDs with members must be quoted. This is to remove a parsing ambiguity with the `$=` comparator.
* Add support for selecting nested trait properties. Nested trait properties can be selected using a "|" delimited syntax. Values or lists and objects can be "projected" using the special `(values)` property. Keys of objects can be projected using the special `(keys)` property.
* pseudo-properties are now formalized in the grammar and parser as`"(" identifier ")"`, allowing us to add more in the future as needed.
* This commit updates selectors to tolerate unknown functions and renames the ":each" selector to ":is" to match the CSS4 name of ":is", which is the same functionality. Support for ":each" is still present and functions as an alias.


docs/source/spec/core/selectors.rst
```diff
@@ -77,145 +139,261 @@ Attribute selectors support the following comparators:
...
+
+The following example matches all shapes that have an ``externalDocumentation``
+trait that has an entry named ``Homepage``:
 
 .. code-block:: none
 
-    resource[id$="$key"]
+    [trait|externalDocumentation|(keys)=Homepage]
 
+Like the ``(list)`` property, the ``(keys)`` property also treats empty
```
kstich, line 356: `(list)` should be `(values)`

docs/source/spec/core/selectors.rst
```diff
@@ -77,145 +139,261 @@ Attribute selectors support the following comparators:
...
 
 
 Neighbors
 =========
 
-The *current* shape evaluated by a selector is changed using a neighbor token,
-``>``. A neighbor token returns every shape that is connected to the current
-shape. For example, the following selector returns the key and value members of
-every map:
+The *current* shapes evaluated by a selector is changed using a
```
kstich, line 379: "shapes" should be singular or "is" should be "are"

docs/source/spec/core/selectors.rst
```diff
@@ -584,67 +773,33 @@ Selectors are defined by the following ABNF_ grammar.
...
+    selector_recursive_neighbor     :"~>"
+    selector_rel_type               :`identifier`
+    selector_attr                   :"[" `selector_key` *(`selector_comparator` `selector_values` ["i"]) "]"
+    selector_key                    :`identifier` *("|" `selector_key_path`)
+    selector_key_path               :`selector_pseudo_key` / `selector_value`
+    selector_values                 :`selector_value` *("," `selector_value`)
+    selector_value                  :`selector_text` / `number` / `root_shape_id`
+    selector_absolute_root_shape_id :`namespace` "#" `identifier`
+    selector_pseudo_key             :"(" `identifier` ")"
+    selector_comparator             :"^=" / "$=" / "*=" / "!=" / "="
```
kstich, line 796: Missing the mathematical comparators.

PR #392, opened 2020-04-22
Adjust auth scheme name for compatibility

srchase:
Adds getOpenApiAuthSchemeName default method which derives an auth scheme name from its shape ID by replacing the `#` with `.`.  Ensures compatibility with  API Gateway authorizer name requirements.

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
@@ -504,7 +504,7 @@ public final class OpenApiConverter {
                     context, authSchemeClasses);
             for (SecuritySchemeConverter<? extends Trait> converter : converters) {
                 List<String> result = createSecurityRequirements(context, converter, service);
-                String openApiAuthName = converter.getAuthSchemeId().toString();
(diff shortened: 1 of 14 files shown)


mtdowling:
Why not do it on all OpenAPI models?

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
```diff
@@ -637,7 +637,7 @@ private OperationObject addOperationTags(
         for (SecuritySchemeConverter<? extends Trait> converter : context.getSecuritySchemeConverters()) {
             if (defaultAuthTraits.contains(converter.getAuthSchemeType())) {
                 List<String> result = createSecurityRequirements(context, converter, context.getService());
-                String authSchemeName = converter.getAuthSchemeId().toString();
+                String authSchemeName = converter.getAuthSchemeId().toString().replace("#", ".");
```
mtdowling, line 640: This is repeated in lots of places -- let's move this to a single method to make it easier to not get wrong or forget about later.

srchase: added getOpenApiAuthSchemeName() method.

PR #397, opened 2020-04-23
Move validators into core spec, rethink suppressions

mtdowling:
Move validation into core
    
    This commit moves model validation and suppressions into the core.
    EmitEachSelector and EmitNoneSelector are now part of smithy-model and
    were removed from smithy-linters. The `@suppress` trait was added to
    suppress validaation events on specific shapes. The `validators`
    metadata property was updated so that it now only take `id`,
    `namespace`, and `reason`, where `namespace` can be set to `*` to
    suppress a validation event for all namespaces or validation events that
    aren't specific to a single shape.
    
    The ability to add custom suppressions to the ModeAssembler has been
    removed.

Move validation into core spec
    
    This commit moves validation concepts into the core Smithy spec while
    leaving more specific linters in a linters guide.


docs/source/spec/core/model-validation.rst
```diff
@@ -0,0 +1,442 @@
...
+:ref:`suppressions metadata <suppressions-metadata>`.
+
+
+.. _suppression-trait:
+
+``suppression`` trait
+=====================
+
+Summary
+    The suppression trait is used to suppress a validation event for a
```
kstich, line 192: s/a validation event/validation event(s)

docs/source/spec/core/model-validation.rst
```diff
@@ -0,0 +1,442 @@
...
+    specific shape. Each value in the suppression trait is the
```
kstich, line 193: s/is the/is a

PR #398, opened 2020-04-24
Allowing a shape to be selector scope

mtdowling:
A shape can now be used as a selector scope allowing for selecting and
comparing multiple top-level attributes of shapes.

This commit also updates AttributeValue to be ready to expose as a
public interface if needed for more advanced selector message formatting
functionality.

*Issue #, if available:*

*Description of changes:*


mtdowling:
> Realized this needs a changelog entry.

We just need to mention that they exist, right? Which we already did in the changelog, so what would be updated?

PR #399, opened 2020-04-28
requestBody and response content schemas as references

srchase:
This PR updates the OpenAPI Converter to write requestBody and response content schemas to as references instead of in-line schema definitions.


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -398,10 +403,17 @@ private void createResponsePayload(
             String mediaType,
             Context<T> context,
             HttpBinding binding,
-            ResponseObject.Builder responseBuilder
+            ResponseObject.Builder responseBuilder,
+            Shape operationOrError
     ) {
+        Schema schema = context.inlineOrReferenceSchema(binding.getMember());
```
mtdowling, line 409: Can you add a comment here (and in the other bit of updated code) to state why we're doing this?

srchase: Added documentation in both places.

PR #412, opened 2020-05-02
Improve validators through a trait index on model

mtdowling:
This commit is a fairly large update to validators because there's a new
super useful feature on Model where you can now get a Set of shapes that
have a specific trait (and the value is cached). This means that a large
majority of validators that utilized Trait.flatMapStream can now be much
more efficiently written by iterating over the shapes that actually have
the trait directly. Lots of validators were updated to take advantage of
this and be more efficient (particularly HTTP validators and trait
conflict validators).

I also noticed that the TraitValueValidator was validating the prelude
model, and that was pretty expensive. I updated this validator to only
validate the prelude when a specific metadata property is set. Then, in
unit tests, I added a test that ensures that the validator validates the
prelude so that we can be sure that changes made to the prelude are
valid. Because of this change and the trait indexing, JMH shows that the
time to load and validate a model was nearly cut in half
(from about 2ms to 1ms on my old laptop).

I noticed that the UnstableFeatureValidator was still being used on
document shapes even though the spec was updated to remove the
disclaimer that document shapes are unstable, so I removed this
validator.

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -141,46 +133,66 @@ public static ModelAssembler assembler(ClassLoader classLoader) {
...
+    }
+
+    /**
+     * Gets a set of shapes in the model marked with a specific trait.
      *
-     * @return Returns all trait definitions in the model.
+     * @param trait Trait shape ID to look for on shapes.
+     * @return Returns the immutable set of matching shapes.
      */
-    public Map<Shape, TraitDefinition> getTraitDefinitions() {
```
kstich, line 149: The public API migrations in this file need a changelog entry: `getTraitDefinitions` and `getTraitShapes`.

PR #413, opened 2020-05-03
Remove non-idiomatic ShapeVisitor builder

mtdowling:
We previously had a builder that could create a ShapeVisitor from a
bunch of closures. This is actually not very idiomatic Java when
compared to just implementing an anonymous class that implements
ShapeVisitor.Default.

*Issue #, if available:*

*Description of changes:*


smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
```diff
@@ -111,15 +108,27 @@ private ShouldHaveUsedTimestampValidator(Config config) {
...
-                .when(IntegerShape.class, s -> validateSimpleShape(s, patterns))
-                .when(LongShape.class, s -> validateSimpleShape(s, patterns))
-                .when(FloatShape.class, s -> validateSimpleShape(s, patterns))
-                .when(StructureShape.class, shape -> validateStructure(shape, model, patterns))
-                .when(UnionShape.class, shape -> validateUnion(shape, model, patterns))
-                .orElse(ListUtils.of());
+        ShapeVisitor<List<ValidationEvent>> visitor = new ShapeVisitor.Default<List<ValidationEvent>>() {
+            @Override
+            protected List<ValidationEvent> getDefault(Shape shape) {
+                if (shape.isStringShape() || shape instanceof NumberShape) {
```
kstich, line 114: The expansion on this to be `NumberShape` means a few more shape types will be validated: `byte`, `double`, and the `Big*` numbers. Seems like that would be intended but calling out in case.

PR #414, opened 2020-05-03
Improve JSON parsing

mtdowling:
This commit updates the forked version of our JSON parser to use
SourceLocation directly, not need a stack of source locations when
parsing node values because the parser now passes that in when calling
end* methods, and swaps our ParseException with ModelSyntaxException.
This improves the performance of the JSON parser a bit and also removes
code that was duplicative with things in the rest of the project (like
Location vs SourceLocation).

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/node/internal/JsonParser.java
```diff
@@ -199,12 +212,12 @@ private void readArray() throws IOException {
...
         if (readChar(']')) {
             nestingLevel--;
-            handler.endArray(array);
+            handler.endArray(array, location);
             return;
         }
         do {
             skipWhiteSpace();
-            handler.startArrayValue(array);
+            // handler.startArrayValue(array);
```
kstich, line 220: There are a decent amount of commented out lines here that, if uncommented, would route to methods on `JsonHandler` that no longer exist.

mtdowling: This is meant to help us figure out how to update this class if/when the upstream project gets updated

PR #416, opened 2020-05-04
Add prioritized least recently added ShapeId cache

mtdowling:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
@@ -86,6 +86,11 @@ public final class ShapeId implements ToShapeId, Comparable<ShapeId> {
             return false;
         }
 
+        // Shortcut for prelude namespaces.
+        if (namespace.equals(Prelude.NAMESPACE)) {
+            return true;
+        }
+
         boolean start = true;
         for (int position = 0; position < namespace.length(); position++) {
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
```diff
@@ -59,32 +67,12 @@ private ShapeId(String namespace, String name, String member) {
...
-        } else if (memberPosition < namespacePosition) {
-            throw new ShapeIdSyntaxException("Invalid shape ID: " + absoluteShapeId);
-        } else {
-            name = absoluteShapeId.substring(namespacePosition + 1, memberPosition);
-            memberName = absoluteShapeId.substring(memberPosition + 1);
-        }
-
-        validateParts(absoluteShapeId, namespace, name, memberName);
-        return new ShapeId(absoluteShapeId, namespace, name, memberName);
+    public static ShapeId from(String id) {
```
kstich, line 74: I think the `fromParts` functions should route through this cache as well. It's used in `LoaderVisitor`  for prelude loading.

mtdowling: Using the cache for that would probably be suboptimal since the check is a guess that a shape might be in the prelude, and many will not. I will add a method to shape ID called `withNamespace` to optimize this though.

PR #463, opened 2020-06-09
Trial doc update

kggilmer:
*Issue #, if available:*
N/A

*Description of changes:*
Add a link from the http trait spec to the endpoint trait spec upon first mention.  The http trait spec occurs first in the index, so a reader may not yet be familiar with endpoints at that point.

This is my first PR as part of the AWS org.  I'm ignorant of process or convention, please LMK what I have missed.

*Testing Performed*
* Built site locally using Sphinx tool as documented in README.  Made code change, re-generated the documentation, verified local copy link renders and navigates as expected.


docs/source/1.0/spec/core/http-traits.rst
```diff
@@ -120,7 +120,7 @@ request URI and labels which are used to insert named components into the
 request URI.
 
 The resolved absolute URI of an operation is formed by combining the URI of
-the operation with the endpoint of the service. For example, given a service
+the operation with the :ref:`endpoint <endpoint-trait>` of the service. For example, given a service
```
mtdowling, line 123: It isn't clear here, but in this context "endpoint" is referring to the host of a server + any base URL. The `endpoint` trait is subtly different.

The `endpoint` trait only allows the modification of an already resolved endpoint of a service. For example, this is used with S3 to dynamically prepend to the host based on the input of a request. We haven't yet modeled a trait for describing the actual set of endpoints for a service... It's something we need to decide on whether it even belongs in the model, in an external model, in some kind dynamic service registry, etc.

kggilmer: I see!  Then it would be confusing to link to the endpoint trait.  I can update the text rephrasing the status of endpoint, or just abandon the PR.

kggilmer: I'm not sure what "Resolve conversation" implies process wise, so leaving as-is.

kggilmer: Sure, but just to confirm, instead of my last edit?
```
...endpoint of the service. (Endpoints are currently not
modeled in Smithy, but rather supplied externally)...
```

kggilmer:
Requested change to documentation has been added.

mtdowling:
Thanks! Can you try rebasing your branch and pushing again? It looks like this pull request picked up a lot of other changes probably because it hasn't been rebased in a while.

PR #467, opened 2020-06-15
Add support message templates in EmitEachSelector

mtdowling:
This commit adds support for templated messages in the EmitEachSelector
that are able to reference variables captured while shapes are matched.
This allows for very rich and descriptive error messages that generally
do as good as error messages in code-driven Smithy validators.

In order to pull this, it required that attribute value selector context
values could be parsed from a message template. Since context values are
already parsed as part of parsing selectors, this motivated me to reuse
the existing selector's parsing functionality. To do this, I created a
SimpleParser abstraction that extracts out lots of the duplicate parsing
code in Smithy into a single class. This allowed parsing of attributes
to be shared, but it was also an opportunity to trim down duplicate code
and reuse parsing logic across the IDL and selectors.

*Issue #, if available:*

*Description of changes:*


smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleParser.java
```diff
@@ -0,0 +1,366 @@
...
+     *
+     * @param expression Expression to parser.
+     * @param maxNestingLevel The maximum allowed nesting level of the parser.
+     */
+    public SimpleParser(String expression, int maxNestingLevel) {
+        this.expression = Objects.requireNonNull(expression, "expression must not be null");
+        this.length = expression.length();
+
+        if (maxNestingLevel < 0) {
+            throw new IllegalArgumentException("maxNestingLevel be >= 0");
```
kstich, line 64: must be

smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleParser.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    /**
+     * Gets the expression being parsed.
+     *
+     * @return Returns the expression being parsed.
+     * @throws NullPointerException if the expression has not been set.
+     */
+    public final String expression() {
+        return Objects.requireNonNull(expression, "Expression has not been set on " + getClass());
```
kstich, line 77: This is a duplicate check with the constructor.

smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleParser.java
```diff
@@ -0,0 +1,365 @@
...
+     * <p>This method can be manually invoked when parsing in order to
+     * prevent parsing too deeply using recursive descent parsers.
+     *
+     * @throws RuntimeException if the nesting level is deeper than the max allowed nesting.
+     */
+    public final void increaseNestingLevel() {
+        nestingLevel++;
+
+        if (maxNestingLevel > 0 && nestingLevel > maxNestingLevel) {
+            throw syntax("Parser nesting level too deep");
```
JordonPhillips, line 342: Would it be useful to say what the max nesting level is?

PR #469, opened 2020-06-16
Clarify build tools

### LINKED ISSUE #437: Better explain the build process
So the readme of the package says

> The Smithy Gradle Plugin is the best way to get started with building a Smithy model.

which leads me to believe it's not the only way implemented here?

There's a page on awslabs.github.io/smithy called "SmithyBuild" which seem to reference something other than a Gradle plugin (it's a different page but I can't tell 100% if this is just configurations and prerequisites or is really an alternative to the gradle plugin??) it doesn't reference Gradle at all on the page, just "building" in an abstract way.

There are also pages about converting to openapi - is this something different from "SmithyBuild"? (Note, here it references "smithy-build" which sounds more like a command line tool?) but there's again nothing about Gradle or links to the Gradle stuff which makes it seem like there's some other tool out there somewhere?

Can you tighten up the expectation around what is actually supported for building Smithy? Does it require Gradle? Does it specifically require Kotlin or a JVM language? Is there a CLI or some standalone build tool that's just missing documentation?

Also, can you link to the "issues" area of this github repo from the github.io pages? (Maybe as a "contact us" kind of link? There's no clear way to give feedback from those pages, and requires either an understanding of github pages or some googling)

mtdowling:
closes #437

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/1.0/guides/building-models/index.rst
+++ b/docs/source/1.0/guides/building-models/index.rst
@@ -4,13 +4,30 @@
 Building Smithy Models
 ======================
 
-These guides describe how to build artifacts from Smithy models, including
-JARs, OpenAPI specifications, generated code, and more.
+These guides describe how build artifacts are generated from Smithy
+models, including JARs, OpenAPI specifications, generated code, etc.
+
+
+Build system overview
+=====================
+
+Smithy models are built using the :doc:`Smithy Gradle plugin <gradle-plugin>`.
+The Gradle plugin is a wrapper around a :doc:`smithy-build <build-config>` and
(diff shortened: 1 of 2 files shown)


docs/source/1.0/guides/building-models/index.rst
```diff
@@ -4,13 +4,30 @@
 Building Smithy Models
 ======================
 
-These guides describe how to build artifacts from Smithy models, including
-JARs, OpenAPI specifications, generated code, and more.
+These guides describe how to build artifacts are generated from Smithy
```
kstich, line 7: s/how to/how

PR #470, opened 2020-06-19
Add addTrait method to ModelAssembler

srchase:
This CR adds an `addTrait` method to the model assembler, allowing for a resolved trait to be added to a target shape within the model.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
@@ -517,9 +517,9 @@ public final class ModelAssembler {
 
         shapes.forEach(visitor::onShape);
         metadata.forEach(visitor::onMetadata);
-        pendingTraits.forEach(pendingTrait -> {
+        for (PendingTrait pendingTrait : pendingTraits) {
             visitor.onTrait(pendingTrait.id, pendingTrait.trait);
-        });
+        }
 
         for (Model model : mergeModels) {
             mergeModelIntoVisitor(model, visitor);
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
@@ -40,6 +40,8 @@ import software.amazon.smithy.model.SourceException;
 import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.shapes.AbstractShapeBuilder;
 import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.Trait;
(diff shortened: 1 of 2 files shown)


smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
```diff
@@ -88,6 +92,26 @@ public void addsExplicitShapes() {
         assertThat(result.unwrap().getShape(ShapeId.from("ns.foo#Bar")), is(Optional.of(shape)));
     }
 
+    @Test
+    public void addsExplicitTraits() {
+        StringShape shape = StringShape.builder().id("ns.foo#Bar").build();
+
+        Node node = Node.fromStrings("validator");
+        TraitFactory provider = TraitFactory.createServiceFactory();
```
mtdowling, line 100: A simpler test case could just directly create a SuppressTrait rather than needing to use a TraitFactory. Seems like the TraitFactory bit isn't under test here.

smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
```diff
@@ -88,6 +92,26 @@ public void addsExplicitShapes() {
...
+
+        SuppressTrait suppress = (SuppressTrait) trait.get();
+        ValidatedResult<Model> result = new ModelAssembler()
+                .addShape(shape)
+                .addTrait(shape.toShapeId(), suppress)
+                .assemble();
+
+        assertThat(result.getValidationEvents(), empty());
+        Shape resultShape = result.unwrap().getShape(ShapeId.from("ns.foo#Bar")).get();
+        assertTrue(resultShape.findTrait("smithy.api#suppress").isPresent());
```
mtdowling, line 112: Then here you could test both for the concrete trait class, no? (i.e., `shape.getTrait(SuppressTrait.class)`)

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -499,6 +517,9 @@ private LoaderVisitor createLoaderVisitor() {
 
         shapes.forEach(visitor::onShape);
         metadata.forEach(visitor::onMetadata);
+        pendingTraits.forEach(pendingTrait -> {
```
mtdowling, line 520: Nit, personal taste -- I know this comes right after some `forEach` calls, but I prefer a normal for loop when possible.

smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
srchase: Fixed.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
srchase: Fixed.

PR #473, opened 2020-06-23
Update json protocol tests

JordonPhillips:
This adds some protocol tests for `awsJson1_1` to assert behavior surrounding null and enums. It also fixes some timestamp tests that were using the wrong members.


jasdel:
Would you mind duplication some of these tests for rest-json as well? Underlying implementations by serder should be the same, but would help for completeness.

PR #481, opened 2020-07-02
Update opinionated protocol tests to not describe optional behavior

jasdel:
Removes the protocol tests for HTTP Prefix Header deserialization, deserializing to an empty map vs null value. Null vs empty map will be SDK implementation specific distinction. If there was a way to model `empty` as an expectation that would encompass both null and empty but initialized map this test would be more portable.

Updates the empty input output serialization protocol test to expect empty payload not empty JSON object as the serialized result. It doesn't look like the REST-JSON protocol should be serializing empty JSON object when no members are serialized.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
@@ -83,7 +83,7 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
         protocol: restJson1,
         method: "POST",
         uri: "/EmptyInputAndEmptyOutput",
-        body: "{}",
+        body: "",
         bodyMediaType: "application/json"
     },
 ])
@@ -91,7 +91,7 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
 apply EmptyInputAndEmptyOutput @httpResponseTests([
     {
         id: "RestJsonEmptyInputAndEmptyOutput",
-        documentation: "Empty output serializes no payload",
+        documentation: "Empty output deserializes no payload",
         protocol: restJson1,
         code: 200,
         body: "",
--- a/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
@@ -91,15 +91,15 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
 apply EmptyInputAndEmptyOutput @httpResponseTests([
     {
         id: "RestJsonEmptyInputAndEmptyOutput",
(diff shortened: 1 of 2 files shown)


smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
```diff
@@ -83,20 +83,28 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
...
         id: "RestJsonEmptyInputAndEmptyOutput",
-        documentation: "Empty output serializes no payload",
+        documentation: "Empty output deserializes no payload",
         protocol: restJson1,
         code: 200,
         body: "",
         bodyMediaType: "application/json"
     },
+    {
+        id: "RestJsonEmptyInputAndEmptyJSONObjectOutput",
```
mtdowling, line 101: Nit: we use "Json" everywhere else in these tests, so "RestJsonEmptyInputAndEmptyJsonObjectOutput"

smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
```diff
@@ -83,20 +83,28 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
...
+        body: "",
         bodyMediaType: "application/json"
     },
 ])
 
 apply EmptyInputAndEmptyOutput @httpResponseTests([
     {
         id: "RestJsonEmptyInputAndEmptyOutput",
-        documentation: "Empty output serializes no payload",
+        documentation: "Empty output deserializes no payload",
```
mtdowling, line 94: not an issue, just a clarification: these tests are about serializing requests and responses. They aren't meant to be oriented towards clients or servers, but rather just about how messages are serialized. So a service should be able to take these tests and make sure the service is implemented correctly. That's why I used the word serialized here instead of deseralize.

smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
```diff
@@ -83,20 +83,28 @@ apply EmptyInputAndEmptyOutput @httpRequestTests([
         protocol: restJson1,
         method: "POST",
         uri: "/EmptyInputAndEmptyOutput",
-        body: "{}",
+        body: "",
```
mtdowling, line 86: 👍 we found out that this test was incorrectly copied from the RPC protocols.

smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
jasdel: Thanks for clarifying that. Updating this one thinking it was a copy/paste bug. Can revert this back.

PR #482, opened 2020-07-03
Clarify the semantic model

mtdowling:
This commit intends to clarify the differences between the IDL and JSON AST by defining the "semantic model" (as originally defined by Martin Fowler in [Domain Specific Languages](https://martinfowler.com/dslCatalog/semanticModel.html). Basically features that are present in the IDL that aren't present in the JSON AST (like control statements) aren't actually part of the semantic model.

### DIFF
--- a/docs/source/1.0/spec/core/lexical-structure.rst
+++ b/docs/source/1.0/spec/core/lexical-structure.rst
@@ -4,18 +4,35 @@
 Smithy IDL lexical structure
 ============================
 
-This document defines the ABNF_ grammar and syntax for defining models using
-the *Smithy interface definition language* (IDL).
-
-Smithy models are defined using either the Smithy IDL or the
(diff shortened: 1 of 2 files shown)


docs/source/1.0/spec/core/lexical-structure.rst
```diff
@@ -333,12 +360,44 @@ Syntactic shape IDs in the IDL
...
+                "members": {
+                    "target": "smithy.other#MyString"
+                }
+            }
+        }
+    }
+
+.. rubric:: Use quotes for literal strings
+
+Values that are not meant to be shape IDs MUST be quoted. The following
```
kstich, line 396: Minor: "Values in Smithy IDL..." might help reset context after just talking about the AST. It looks like this is done below as well.

PR #487, opened 2020-07-07
Remove old style guide namespace guidance

mtdowling:
We used to allow multiple namespaces in a single IDL file, but removed
this feature prior to 1.0. This change just removes a left-over
recommendation that suggests using a single namespace per/file (which is
now a requirement).

*Issue #, if available:*

*Description of changes:*


mtdowling:
cc @johnstonskj

PR #491, opened 2020-07-08
Add Content-Length required header to request protocol tests

jasdel:
Adds Content-Length as an expected header to be serialized in AWS protocol requests for structured and blob payloads.


smithy-aws-protocol-tests/model/restXml/http-payload.smithy
```diff
@@ -31,6 +31,9 @@ apply HttpPayloadTraits @httpRequestTests([
         headers: {
             "X-Foo": "Foo"
         },
+        requireHeaders: [
```
mtdowling, line 34: Tight. Since this blob is not streaming, this makes sense

PR #499, opened 2020-07-16
Disallow problematic identifiers

mtdowling:
These newly disallowed identifiers were never actually supported in
practice and are problematic on multiple levels:

1. An identifier like `_` or `__` are bad names that don't help users understand an API.
2. Identifiers like `_` are problematic for languages like Rust that use this character for pattern matching.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/1.0/spec/core/model.rst
+++ b/docs/source/1.0/spec/core/model.rst
@@ -445,7 +445,9 @@ Shape IDs are formally defined by the following ABNF:
     root_shape_id          :`absolute_root_shape_id` / `identifier`
     absolute_root_shape_id :`namespace` "#" `identifier`
     namespace              :`identifier` *("." `identifier`)
-    identifier             :(ALPHA / "_") *(ALPHA / DIGIT / "_")
+    identifier             :identifier_start *identifier_chars
+    identifier_start       :*"_" ALPHA
+    identifier_chars       :ALPHA / DIGIT / "_"
     shape_id_member        :"$" `identifier`
 
 .. rubric:: Best practices for defining shape names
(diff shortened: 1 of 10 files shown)


docs/source/1.0/spec/core/model.rst
```diff
@@ -445,7 +445,9 @@ Shape IDs are formally defined by the following ABNF:
     root_shape_id          :`absolute_root_shape_id` / `identifier`
     absolute_root_shape_id :`namespace` "#" `identifier`
     namespace              :`identifier` *("." `identifier`)
-    identifier             :(ALPHA / "_") *(ALPHA / DIGIT / "_")
+    identifier             :identifier_start *identifier_chars
```
kstich, line 448: Went looking to see if this needs to be updated in the `idl.rst` file of the spec, but it looks like it isn't present there. Should probably add this section of [the ABNF over there as well, since its heading seems to indicate completeness](https://github.com/awslabs/smithy/blob/master/docs/source/1.0/spec/core/idl.rst#smithy-idl-abnf).

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
```diff
@@ -117,21 +117,40 @@ public static boolean isValidNamespace(CharSequence namespace) {
...
         }
 
-        for (int i = 1; i < identifier.length(); i++) {
-            if (!ParserUtils.isValidIdentifierCharacter(identifier.charAt(i))) {
-                return false;
+        // Parse the optional identifier_chars production.
+        while (offset < identifier.length()) {
+            if (!ParserUtils.isValidIdentifierCharacter(identifier.charAt(offset))) {
+                // Return the position of the character that stops the identifier.
+                // This is either an invalid case (e..g, isValidIdentifier), or
```
kstich, line 146: s/e..g/e.g.

docs/source/1.0/spec/core/model.rst
mtdowling: Sphinx seems broken and doesn't allow duplicate productionlists with different names to coexist. So instead, I just added a link from the IDL to the shape ID section of the model. I think this is more appropriate than the other way around because shape IDs exist in the semantic model, the IDL, and AST.

PR #500, opened 2020-07-16
Add clarifications on how to use mediaType trait

mtdowling:
This commit clarifies what the mediaType trait is used for, the fact
that it's a design-time contract, and adds a missing header
serialization test to ensure that mediaType trait strings are base64
encoded.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
@@ -324,7 +324,7 @@ structure NullAndEmptyHeadersIO {
     c: StringList,
 }
 
-/// The example tests how timestamp request and response headers are serialized.
+/// This example tests how timestamp request and response headers are serialized.
 @http(uri: "/TimestampFormatHeaders", method: "POST")
 operation TimestampFormatHeaders {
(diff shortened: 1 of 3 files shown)


smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
```diff
@@ -413,3 +413,57 @@ structure TimestampFormatHeadersIO {
...
+
+apply MediaTypeHeader @httpRequestTests([
+    {
+        id: "MediaTypeHeaderInputBase64",
+        documentation: "Headers that target strings with a mediaType are base64 encoded",
+        protocol: restJson1,
+        method: "GET",
+        uri: "/MediaTypeHeader",
+        headers: {
+            "X-Json": "dHJ1ZQ=="
```
kstich, line 433: Not sure we want to update this since HTTP headers are supposed to be case insensitive, but the headers in both test cases are capital-J Json while the `@httpHeader` strings are lower-j json.

mtdowling: Good catch. Fixed.

PR #512, opened 2020-07-28
Don't raise validation erros from renaming

JordonPhillips:
None of the other transformers raise validation errors when
they produce a model in an invalid state, so the rename transformer
shouldn't either. With this change a tool could attempt to resolve
any issue before the model is actually used.


smithy-model/src/main/java/software/amazon/smithy/model/transform/RenameShapes.java
```diff
@@ -69,9 +70,11 @@ Model transform(ModelTransformer transformer, Model model) {
         // Use visitor to traverse node and rebuild model.
         Node newModel = node.accept(new RenameShapeVisitor(toRename, renamed));
 
-        return assembler.addDocumentNode(newModel)
-                .assemble()
-                .unwrap();
+        ValidatedResult<Model> result = assembler.addDocumentNode(newModel).assemble();
```
mtdowling, line 73: If we're doing this, can you also disable validation in the model assembler?

PR #520, opened 2020-08-03
Refactor model loading to allow equal conflicts

mtdowling:
When using Smithy features like model discovery on the classpath, it's sometimes difficult to prevent the same Smithy models from being added to the model assembler multiple times. This is also true when using artifacts created for the "model" plugin for smithy-build along with model discovery. To address this, Smithy has been updated to allow conflicting shape definitions if the shapes, after they are fully built, are equivalent (meaning, they have the same exact members or references to other shapes and are the same type).


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -763,15 +691,27 @@ private void addTraits(ShapeId id, List<TraitEntry> traits) {
...
+            modelFile.onTrait(target, id, coerceTraitValue(id, traitValue, isAnnotation, typeProvider));
         });
     }
 
+    private Node coerceTraitValue(ShapeId traitId, Node value, boolean isAnnotation,
+            Function<ShapeId, ShapeType> typeProvider) {
+        if (isAnnotation && value.isNullNode()) {
+            ShapeType targetType = typeProvider.apply(traitId);
+            if (targetType != null) {
+                if (targetType == ShapeType.STRUCTURE || targetType == ShapeType.MAP) {
```
JordonPhillips, line 704: Couldn't this also be a union or document?

docs/source/1.0/spec/core/model.rst
```diff
@@ -149,15 +149,32 @@ together to form a valid semantic model.
...
+#. Shapes defined in a single model file are added to the semantic model as-is.
+#. Shapes with the same shape ID defined in multiple model files are merged
+   using the following rules:
+
+   #. All conflicting shapes MUST have the same shape type.
+   #. Conflicting :ref:`aggregate shapes <aggregate-types>` MUST contain the
+      same members that target the same shapes.
+   #. Conflicting :ref:`service shapes <service-types>` MUST contain the same
+      properties and target the same shapes.
+   #. The traits of conflicting shapes and the traits of their members are
```
JordonPhillips, line 166: This doesn't seem right, isn't it an equality check not a merge? Except if they're using apply statements?

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
mtdowling: No— a union requires a value as does a document.

docs/source/1.0/spec/core/model.rst
mtdowling: The traits are merged, but the shape itself isn’t. I’ll try to make this more clear.

smithy-model/src/test/java/software/amazon/smithy/model/loader/LoaderVisitorTest.java
```diff
@@ -1,209 +0,0 @@
...
-        visitor.onShape(StringShape.builder().id(id));
-        visitor.onTrait(id, DocumentationTrait.ID, Node.from("abc"));
-        visitor.onTrait(id, DocumentationTrait.ID, Node.from("def"));
-        List<ValidationEvent> events = visitor.onEnd().getValidationEvents();
-
-        assertThat(events, not(empty()));
-    }
-
-    @Test
-    public void createsDynamicTraitWhenTraitFactoryReturnsEmpty() {
```
kstich, line 127: This test isn't carried over even though the functionality is retained and would be a break if the behavior changed unexpectedly.

mtdowling: Good catch. Fixed.

PR #527, opened 2020-08-12
Add payloadFormatVersion to API Gateway Integration trait.

JuergenO:
Currently, when converting the Smithy model to API Gateway for HTTP APIs, it fails because it doesn't configure the `payloadFormatVersion`. This change adds the payload format version to the IntegrationTrait.

The payload format version specifies the format of the data that API Gateway sends to an integration, and how API Gateway interprets the response. If you create an integration for HTTP APIs programmatically, you **must** specify a payloadFormatVersion. The supported values are 1.0 and 2.0.

*Issue #, if available:* No issue #

*Description of changes:* See commit message above. Added the payloadFormatVersion that is required as document here: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

### DIFF
--- a/docs/source/1.0/spec/aws/amazon-apigateway.rst
+++ b/docs/source/1.0/spec/aws/amazon-apigateway.rst
@@ -397,6 +397,11 @@ following members:
     * - cacheNamespace
       - ``string``
       - An API-specific tag group of related cached parameters.
+    * - payloadFormatVersion
+      - ``string``
+      - Specifies the format of the payload sent to an integration. Required for HTTP APIs. For HTTP APIs,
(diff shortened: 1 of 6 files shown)


JuergenO:
> Can you add this new property to the trait definition too? https://github.com/awslabs/smithy/blob/master/smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.json#L124

Sorry, forgot about that one. Added it to the trait definition now + modified the documentation. 
Please let me know if there's anything else I missed.

Thanks

PR #529, opened 2020-08-16
Improve CodeWriter newlines and indentation

mtdowling:
This commit simplifies and improves how the CodeWriter handles the
automatic injection of newlines and indentation. Now, the *only* difference
between write() and writeInline() is that write() appends a newline
(there was previously a lot of special-cased code for writeInline).

* writeInline() now properly handles indentation when necessary (e.g., if
  a previous call injected a newline). This includes a few edge cases
  like if the first call to the CodeWriter is to writeInline() and uses
  indentation.
* The character used for newlines is now tracked by the current state rather
  than globally.
* Trailing newlines are now *always* inserted to the given text if enabled
  There was a bug previously where they were not appended if the last call
  on the CodeWriter was a writeInline(). This could potentially affect
  usage of CodeWriters that only wrote with inline calls.
* The automatic appending of a newline by methods like openBlock(),
  closeBlock(), and write() can now enabled or disabled using
  enableNewlines() and disableNewlines().

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


jasdel:
> The automatic appending of a newline by methods like openBlock(), closeBlock(), and write() can now enabled or disabled using enableNewlines() and disableNewlines().

Would this allow additional helpers like openBlockInline to be defined, so I could have an inline block, but not add the trailing newline?

mtdowling:
I'm not 100% sure what `openInlineBlock` would do, but it should be possible with this change. There's a test case that does something like that: https://github.com/awslabs/smithy/pull/529/files#diff-07be89dbdaaf4a400e8e226ec79bac02R606

kggilmer:
> This could potentially affect usage of CodeWriters that only wrote with inline calls.

Anything in particular we need to look out for w/ existing usage of CodeWriter?

PR #531, opened 2020-08-18
Make internal a trait and not just a tag

mtdowling:
We've been using the "internal" tag to indicate that something is
internal only and should not be exposed to external consumers of a model
(i.e., people that don't work at the same company as the service team).
The original motivation for using tags and not traits was that there are
more dimensions to filtering shapes out of models than just internal vs
external. However, the internal/external split is so common that it
makes sense to make it a trait, which is more configurable and less
based on convention. Tags can still be used for filtering things out of
models (for example, an operation could be tagged so that it is only
exposed to a particular customer).

There are other cases where we migrated away from tags and moved to
traits like the `aws.api#controlPlane` and `aws.api#dataPlane` traits.

To support this change, a SmithyBuild transform was added that removes
shapes marked as internal. Note that this does not perform tree shaking
since like other transforms, tree shaking is performed by a dedicated
transform.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/1.0/guides/building-models/build-config.rst
+++ b/docs/source/1.0/guides/building-models/build-config.rst
@@ -74,9 +74,9 @@ The following is an example ``smithy-build.json`` configuration:
                             }
                         },
                         {
-                            "name": "excludeTraitsByTag",
+                            "name": "excludeShapesByTrait",
                             "args": {
-                                "tags": ["internal"]
+                                "trait": "internal"
                             }
                         }
                     ],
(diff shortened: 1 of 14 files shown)


docs/source/1.0/guides/building-models/build-config.rst
```diff
@@ -287,6 +287,47 @@ the :ref:`tags trait <tags-trait>`.
...
+
+.. list-table::
+    :header-rows: 1
+    :widths: 10 20 70
+
+    * - Property
+      - Type
+      - Description
+    * - trait
+      - ``string``
```
kstich, line 305: Should this be a list of traits instead? All the `ByTag` transformers take lists, for reference.

mtdowling: Done

PR #538, opened 2020-08-22
Move from Kotlin to Groovy for Gradle

mtdowling:
Unfortunately, Kotlin isn't well-supported in VS Code, and we're looking
at adding VS plugins to Smithy in the future. We've also found that the
majority of Gradle plugins only provide Groovy examples, making it
harder to write out build scripts due to a lack of Kotlin examples. With
IDE plugins and a quicker compile/test feedback loop, the loss in type
safety realized by Kotlin isn't as bad as I thought it would be.

*Issue #, if available:*

*Description of changes:*


build.gradle
```diff
@@ -0,0 +1,248 @@
...
+/*
+ * Sonatype Staging Finalization
+ * ====================================================
+ *
+ * When publishing to Maven Central, we need to close the staging
+ * repository and release the artifacts after they have been
+ * validated. This configuration is for the root project because
+ * it operates at the "group" level.
+ */
+if (project.hasProperty("sonatypeUser") && hasProperty("sonatypePassword")) {
```
kstich, line 240: Should be `project.hasProperty("sonatypePassword")`

mtdowling: Fixed

PR #540, opened 2020-08-24
Tracefile naming

et22:
*Issue #, if available:*

*Description of changes:*


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/trace/Metadata.java
```diff
@@ -61,10 +61,10 @@ private ArtifactMetadata(Builder builder) {
      * @param value an ObjectNode that contains all children of the artifact tag in the trace file
      * @return ArtifactMetadata produced by deserializing the node.
      */
-    public static ArtifactMetadata fromNode(Node value) {
+    public static Metadata fromNode(Node value) {
```
mtdowling, line 64: Can we call this `TraceMetadata` or something more descriptive? I fear that there are lots of classes called Metadata, and it would make autocomplete harder to find, and cause import collisions.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/trace/TraceFile.java
```diff
@@ -226,8 +225,8 @@ public String getSmithyTrace() {
      *
      * @return an ArtifactMetadata object.
      */
-    public ArtifactMetadata getArtifactMetadata() {
-        return artifactMetadata;
+    public TraceMetadata getTraceMetadata() {
```
mtdowling, line 228: I think this should just be `getMetadata` to match the file format. The class name doesn't have to have a major impact on the getter name.

PR #545, opened 2020-08-28
Add reverse-topological index

mtdowling:
Some languages require types to be defined before they're referenced, so
they need a reverse topological ordering of shapes. They also need to
know which shapes are recursive so that they can use forward
declarations before referencing recursive types. This commit adds a
TopologicalIndex to smithy-codegen-core to provide a stable ordered
reverse-topological set of shapes, and a set of recursive shapes along
with their recursive closures. This recursive closure may also be of
interest to code generators when deciding whether or not a recursive
reference needs to be boxed or not to compute a type's size at compile
time.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TopologicalIndex.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TopologicalIndex.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.TreeSet;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.KnowledgeIndex;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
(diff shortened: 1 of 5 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TopologicalIndex.java
```diff
@@ -0,0 +1,192 @@
...
+
+        for (Map.Entry<Integer, Map<Shape, List<PathFinder.Path>>> entry : edgesToShapePaths.entrySet()) {
+            recursiveShapes.putAll(entry.getValue());
+        }
+    }
+
+    private void visitShape(NeighborProvider provider, Shape shape) {
+        // Visit members before visiting containers. Note that no 'visited'
+        // set is needed since only non-recursive shapes are traversed.
+        for (Relationship rel : provider.getNeighbors(shape)) {
```
JordonPhillips, line 111: Is the ordering of this stable?

mtdowling: It is stable, but not alphabetically ordered. I updated it to sort neighbors alphabetically to make it more predictable.

PR #547, opened 2020-08-31
adds protocol tests for self closing tag and empty xml tag

skotambkar:
*Description of changes:*
* Adds protocol tests for self closing tag and empty xml tag

### DIFF
--- a/smithy-aws-protocol-tests/model/awsQuery/xml-lists.smithy
+++ b/smithy-aws-protocol-tests/model/awsQuery/xml-lists.smithy
@@ -31,7 +31,7 @@ operation XmlLists {
 apply XmlLists @httpResponseTests([
     {
         id: "QueryXmlLists",
-        documentation: "Serializes XML lists",
+        documentation: "Deserializes XML lists",
         protocol: awsQuery,
         code: 200,
         body: """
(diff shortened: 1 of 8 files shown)


smithy-aws-protocol-tests/model/awsQuery/xml-lists.smithy
```diff
@@ -31,7 +31,7 @@ operation XmlLists {
 apply XmlLists @httpResponseTests([
     {
         id: "QueryXmlLists",
-        documentation: "Serializes XML lists",
+        documentation: "Deserializes XML lists",
```
mtdowling, line 34: Note that these tests are not client-specific, so there's no need to change this language. If this is a concern, then I suggest we start using sentences like "Tests for XML list serialization".

smithy-aws-protocol-tests/model/awsQuery/xml-lists.smithy
```diff
@@ -118,7 +118,29 @@ apply XmlLists @httpResponseTests([
                 }
             ]
         }
-    }
+    },
+    {
```
mtdowling, line 122: Lots of these tests seem to be testing for different possibilities that clients need to handle, but they don't seem like something a service could actually implement. Can you add a `client-only` tag to client-specific tests to indicate that they are client-specific tests? Like this one: https://github.com/awslabs/smithy/blob/master/smithy-aws-protocol-tests/model/restXml/http-query.smithy#L303

smithy-aws-protocol-tests/model/restXml/document-structs.smithy
```diff
@@ -32,6 +32,7 @@ apply SimpleScalarProperties @httpRequestTests([
         body: """
               <SimpleScalarPropertiesInputOutput>
                   <stringValue>string</stringValue>
+                  <emptyStringValue></emptyStringValue>
```
JordonPhillips, line 35: These also should probably be isolated with the client only tag

JordonPhillips (changes requested):

smithy-aws-protocol-tests/model/restXml/document-structs.smithy
skotambkar: updated

JordonPhillips (approved):

PR #558, opened 2020-09-10
Add transformer that removes trait definitions

srchase:
This CR adds a transformer that removes trait definitions.  This enables sharing models where the trait versions may vary.  When consuming models produced by this plugin, it is necessary to assure that the requisite traits sources are imported.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/RemoveTraitDefinitions.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/RemoveTraitDefinitions.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.build.transforms;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.logging.Logger;
+import software.amazon.smithy.build.TransformContext;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.TraitDefinition;
+import software.amazon.smithy.model.transform.ModelTransformer;
+
+/**
+ * {@code removeTraitShapes} removes all trait definitions from a model,
+ * but leaves if the trait definition contains any of the provided
+ * {@code tags}.
+ *
+ * <p>This transformer will not remove prelude trait definitions.
+ */
+public final class RemoveTraitDefinitions extends BackwardCompatHelper<RemoveTraitDefinitions.Config> {
+
+    private static final Logger LOGGER = Logger.getLogger(RemoveTraitDefinitions.class.getName());
+
+    /**
+     * {@code removeTraitShapes} configuration settings.
+     */
+    public static final class Config {
+
+        private Set<String> exportTagged = Collections.emptySet();
+
+        /**
+         * You can <em>export</em> trait definitions by applying specific tags
+         * to the trait definition and adding the list of export tags as an
+         * argument to the transformer.
+         *
+         * @param exportByTags Tags that cause trait definitions to be
+         *                     exported.
+         */
+        public void setExportTagged(Set<String> exportByTags) {
+            this.exportTagged = exportByTags;
+        }
+
+        /**
+         * Gets the set of tags that are used to export trait definitions.
+         *
+         * @return the tags that are used to export trait definitions.
+         */
+        public Set<String> getExportTagged() {
+            return exportTagged;
+        }
+    }
+
+    @Override
+    String getBackwardCompatibleNameMapping() {
+        return "removeTraitDefinitions";
+    }
+
+    @Override
+    public Class<Config> getConfigType() {
+        return Config.class;
+    }
+
+    @Override
+    protected Model transformWithConfig(TransformContext context, Config config) {
+        Model model = context.getModel();
+        ModelTransformer transformer = context.getTransformer();
+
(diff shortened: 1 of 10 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/plugins/RemoveSharedShapesPlugin.java
```diff
@@ -0,0 +1,45 @@
...
+package software.amazon.smithy.build.plugins;
+
+import software.amazon.smithy.build.PluginContext;
+import software.amazon.smithy.build.SmithyBuildPlugin;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.shapes.ModelSerializer;
+import software.amazon.smithy.model.transform.ModelTransformer;
+
+public final class RemoveSharedShapesPlugin implements SmithyBuildPlugin {
```
mtdowling, line 25: Why not call this RemoveTraitShapes / "remove-trait-shapes"?

srchase: Fixed.  This evolved a bit and I didn't update the naming.

mtdowling: This should probably be a `ProjectionTransformer` and not a `SmithyBuildPlugin` since it needs to modify a model and not necessarily create its own artifacts.

srchase: This has been reworked and is now a transformer.

smithy-build/src/main/java/software/amazon/smithy/build/transforms/RemoveTraitDefinitions.java
```diff
@@ -0,0 +1,106 @@
...
+import software.amazon.smithy.model.transform.ModelTransformer;
+
+/**
+ * {@code removeTraitShapes} removes all trait definitions from a model,
+ * but leaves if the trait definition contains any of the provided
+ * {@code tags}.
+ *
+ * <p>This transformer will not remove prelude trait definitions.
+ */
+public final class RemoveTraitDefinitions extends BackwardCompatHelper<RemoveTraitDefinitions.Config> {
```
mtdowling, line 35: This probably doesn't need the BackwardCompatHelper since it's new.

smithy-build/src/main/java/software/amazon/smithy/build/transforms/RemoveTraitDefinitions.java
```diff
@@ -0,0 +1,106 @@
...
+        // If no export tags have been configured, we can return the scrubbed model directly.
+        if (config.getExportTagged().isEmpty()) {
+            return modelWithoutTraitDefinitions;
+        }
+
+        Predicate<Shape> keepTraitDefsByTag = trait -> config.getExportTagged().stream().anyMatch(trait::hasTag);
+        Model.Builder builder = modelWithoutTraitDefinitions.toBuilder();
+
+        // Add trait definitions back to scrubbed model.
+        model.shapes()
```
mtdowling, line 94: I don't see how this would work if, say, a trait shape is removed that is a structure that references other shapes that were removed. It might be better to instead filter before removing trait shapes rather than re-add. (e.g., maybe you can add a `Predicate` to the `getModelWithoutTraitShapes` method.

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -387,6 +406,20 @@ public Model scrubTraitDefinitions(Model model) {
...
+    }
+
+    /**
+     * Gets all shapes from a model where shapes that define traits or shapes
+     * that are only used as part of a trait definition have been removed.
+     *
+     * @param model Model that contains shapes.
+     * @param keepFilter Predicate function that accepts a trait shape (that
+     *  has the {@link TraitDefinition} trait) and returns true to remove the
+     *  definition or false to keep the definition in the model.base.
```
mtdowling, line 419: model.base?

smithy-model/src/main/java/software/amazon/smithy/model/transform/ScrubTraitDefinitions.java
```diff
@@ -42,14 +45,21 @@
...
+    Model transform(ModelTransformer transformer, Model model, Predicate<Shape> keepFilter) {
         // Find all trait definition shapes and private shapes in the prelude.
-        Set<Shape> toMark = Stream.concat(
+        Set<Shape> traitDefinitions = Stream.concat(
                 model.shapes().filter(shape -> shape.hasTrait(TraitDefinition.class)),
                 model.shapes().filter(shape -> Prelude.isPreludeShape(shape) && shape.hasTrait(PrivateTrait.class))
         ).collect(Collectors.toSet());
 
+        // Filter any trait definitions from being marked for removal that we want to retain.
+        Set<Shape> toMark = traitDefinitions.stream().filter(keepFilter).collect(Collectors.toSet());
```
mtdowling, line 59: Can this filtering be done all on a single stream instead of needing to make two sets?

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -376,7 +376,26 @@ public Model removeUnreferencedTraitDefinitions(Model model, Predicate<Shape> ke
...
+    }
+
+    /**
+     * Removes trait definitions from a model and all shapes that are
+     * only connected to the graph either directly or transitively by a
+     * trait definition shape.
+     *
+     * <p>This can be useful when serializing a Smithy model to a format that
+     * does not include trait definitions and the shapes used by trait definitions
+     * would have no meaning (e.g., Swagger).
```
mtdowling, line 389: s/Swagger/OpenAPI

mtdowling (changes requested):

smithy-build/src/main/java/software/amazon/smithy/build/transforms/RemoveTraitDefinitions.java
srchase: Fixed.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
srchase: Fixed this and a couple other Swagger references.

mtdowling (approved):

PR #571, opened 2020-09-17
Add response code binding to http binding index

JordonPhillips:
This adds support for the httpResponseCode trait to the http binding
index, including a new RESPONSE_CODE location.


mtdowling:
Does this automatically ensure that members bound to the `httpResponseCode` trait aren't part of the payload of a response?

JordonPhillips:
It wouldn't have been possible to bind to the payload, because it's the wrong type, and it wouldn't get bound to the document because that implicit binding only happens when there's no explicit one. It could have, however, been bound to other parts of the request. I updated the conflicts lists so that can no longer happen.

mtdowling (approved):

mtdowling (approved):

PR #583, opened 2020-09-29
Add additional default CORS headers with config

srchase:
Adds additional CORS headers when converting to an OpenAPI model for Amazon API Gateway.

By default, ``amz-sdk-invocation-id`` and ``amz-sdk-request`` headers will be added to the CORS preflight request.

On the ``openapi`` configuration, the ``additionalAllowedCorsHeaders`` option can be used to set other additional headers in place of these default headers, or to omit these default headers entirely by setting an empty array.

Sample config:

```
{
    "version": "1.0",
    "plugins": {
        "openapi": {
            "service": "smithy.example#Weather",
            "additionalAllowedCorsHeaders": ["foo-header", "bar-header"]
        }
    }
}
```


docs/source/1.0/guides/converting-to-openapi.rst
```diff
@@ -691,6 +691,24 @@ disableCloudFormationSubstitution (``boolean``)
     .. seealso:: :ref:`openapi-cfn-substitutions`
 
 
+additionalAllowedCorsHeaders (``[string]``)
+    Sets additional allowed CORS headers on the preflight integration. If this
```
mtdowling, line 695: I'd call this preflight requests and not integrations. We'll use this same configuration for HTTP APIs, and they don't require a preflight integration the same was as the REST APIs do.

mtdowling (changes requested):

docs/source/1.0/guides/converting-to-openapi.rst
srchase: Updated the docs.

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddCorsPreflightIntegration.java
```diff
@@ -112,6 +112,14 @@ private static PathItem addPreflightIntegration(
         // add headers during the Smithy to OpenAPI conversion process will need to update this
         // list of headers accordingly.
         Set<String> headerNames = new TreeSet<>(corsTrait.getAdditionalAllowedHeaders());
+
+        // Sets additional allowed headers from the API Gateway config.
+        List<String> additionalAllowedHeaders = context.getConfig().getExtensions(ApiGatewayConfig.class)
+                .getAdditionalAllowedCorsHeaders();
+        for (String header : additionalAllowedHeaders) {
```
mtdowling, line 119: nit: `headerNames.addAll(additionalAllowedHeaders)`

mtdowling (approved):

PR #590, opened 2020-10-05
Optimize reverse neighbors for memory

mtdowling:
This method previously needed lots of intermediate representations
stored in memory to create a Map<ShapeId, List<RelationShip>> that
contains only unique relationships. It was done using Stream,
distinct, and groupingBy. However, when trying to load ridiculously
large models, that approach consume tons of heap. This approach
allocates as little as possible (I think), but does require creating
an ArrayList copy of a Set each time neighbors are returned.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborProvider.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborProvider.java
@@ -16,11 +16,13 @@
 package software.amazon.smithy.model.neighbor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.Shape;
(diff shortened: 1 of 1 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborProvider.java
```diff
@@ -122,13 +124,24 @@ static NeighborProvider reverse(Model model) {
...
     static NeighborProvider reverse(Model model, NeighborProvider forwardProvider) {
-        Map<ShapeId, List<Relationship>> targetedFrom = model.shapes()
-                .map(forwardProvider::getNeighbors)
-                .flatMap(List::stream)
-                .distinct()
-                .collect(Collectors.groupingBy(Relationship::getNeighborShapeId, ListUtils.toUnmodifiableList()));
+        // Note: this method previously needed lots of intermediate representations
+        // stored in memory to create a Map<ShapeId, List<RelationShip>> that contains
+        // only unique relationships. It was done using Stream, distinct, and groupingBy.
+        // however, when trying to load ridiculously large models, that approach consume
```
srchase, line 130: nit:
```suggestion
        // However, when trying to load ridiculously large models, that approach consumes
```

srchase (changes requested):

srchase (approved):

PR #594, opened 2020-10-06
Add example of dependency for openapi plugin

srchase:
Updates the ``openapi`` plugin configuration documentation to include an example of an implementation dependency that is required when using a protocol defined in another package.


docs/source/1.0/guides/converting-to-openapi.rst
```diff
@@ -124,6 +124,13 @@ specification from a Smithy model using a buildscript dependency:
         }
     }
 
+    dependencies {
```
mtdowling, line 127: This needs to appear in the buildscript too. The buildscript doesn't pull in normal dependencies -- it's isolated from normal dependencies.

mtdowling (changes requested):

docs/source/1.0/guides/converting-to-openapi.rst
srchase: Added to the buildScript.

mtdowling (approved):

PR #599, opened 2020-10-08
Add support for explicit sparse lists and maps

mtdowling:
The vast majority of all lists and maps in use today are dense, meaning
they cannot contain null values. However, we have historically had no
way to indicate that a list or map (value) supports nulls, so we had to
assume that all lists and maps are sparse. This change makes it so that
all lists and maps are considered dense by default, but services can
opt-in to sparse lists using the `sparse` trait. This matters because it
now allows languages that bake "null" into their type systems to provide
better generated types.

*Issue #, if available:*

*Description of changes:*


docs/source/1.0/spec/core/constraint-traits.rst
```diff
@@ -556,6 +556,48 @@ in a response.
         }
 
 
+.. _sparse-trait:
```
mtdowling, line 559: I wasn't sure if this should be a "constraint trait" or a "type refinement trait"...

kstich: Feels like this should be a "type refinement trait" as it's dealing in the same realm as `@box` which lives there as well.

docs/source/1.0/spec/core/model.rst
```diff
@@ -823,12 +863,55 @@ The following example defines a map of strings to integers:
...
 formats have a way to define ``null`` map keys, and map implementations
 across programming languages often do not allow ``null`` keys in maps.
 
+.. rubric:: Map value member nullability
+
+Maps values are considered *dense* by default, meaning they MAY NOT contain
+``null`` values. A map MAY be made *sparse* by applying the
+:ref:`sparse-trait`. The :ref:`box-trait` is not used to determine if a map
+is dense or sparse; a map with no ``@sparse`` trait is always considered
+dense.
```
kstich, line 878: Needs to close with "The following example defines a sparse map:"

kstich (changes requested):

kstich (approved):

kggilmer (approved):
LGTM

jasdel (approved):

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
```diff
@@ -0,0 +1,116 @@
...
+                // Only structure shapes look at the box trait.
+                return member.hasTrait(BoxTrait.class)
+                       || model.getShape(member.getTarget()).filter(NullableIndex::isShapeBoxed).isPresent();
+            case MAP:
+                // Map keys can never be null.
+                if (member.getMemberName().equals("key")) {
+                    return false;
+                } // fall-through
+            case LIST:
+                // Sparse lists and maps are considered nullable.
```
kstich, line 84: This should state that it's the members of the sparse lists and maps.

mtdowling: It does

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
```diff
@@ -0,0 +1,116 @@
...
+            ShapeType.SET,
+            ShapeType.MAP,
+            ShapeType.STRUCTURE,
+            ShapeType.UNION);
+
+    private final Set<ShapeId> nullableShapes = new HashSet<>();
+
+    public NullableIndex(Model model) {
+        for (Shape shape : model.toSet()) {
+            if (shape.asMemberShape().isPresent()) {
```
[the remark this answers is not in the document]
mtdowling: I like it the way it is since it doesn't warn in the IDE

kstich (approved):

PR #602, opened 2020-10-15
Maintain order of operations and resources

### LINKED ISSUE #591: EntityShape#operations should be ordered
Currently, operations (and resources) are exposed from an EntityShape as `Sets`, backed by `HasSets` This loses the ordering from the spec, which happens to be useful information to construct routing logic in REST-like protocols. 

https://github.com/awslabs/smithy/blob/acc7c95a2052b0597c1ea1eab2a3ed9273092bc5/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EntityShape.java#L28-L29

I'd personally like for the data structure to retain ordering, and for the ordering guarantee to be reflected in the exposed types.

mtdowling:
This commit updates service and operation shapes to maintain the order
in which operations and resources are defined in a Smithy model. Note
that resource lifecycle operation order is not maintained since those
are named operations that don't have any particular order in a model
like a list of operations, collectionOperations, or resources do.

To achieve this, I added a couple new helper methods to make copies of
Maps and Sets that maintains order.

Closes #591

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EntityShape.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EntityShape.java
@@ -16,7 +16,7 @@
 package software.amazon.smithy.model.shapes;
 
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 import software.amazon.smithy.utils.SetUtils;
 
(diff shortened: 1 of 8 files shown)


Baccata:
👍

kstich (commented):
Should there be associated spec updates with this, like with the recommendation to add members to the ends of their respective shapes?

mtdowling:
I don't see this as something for the spec right now for a few reasons:

1. I think there's an implicit expectation about ordering here because these operations and resources are defined in lists which are inherently ordered.
2. Adding operations or resources in a recommended order isn't the intention here; this just guarantees maintaining order in the in-memory semantic model, and order is completely arbitrary and up to the modeler.
3. I think this is more of an "implementer's note" which could be added later if we discover it's necessary to explicitly mandate this detail of the semantic model.

kstich (approved):

PR #610, opened 2020-10-20
Add warning for enums without names

JordonPhillips:
This adds a warning-level validation event for enums that do not use
the name property. It additionally updates all built in enums and
enums used for tests where the absense of the name isn't deliberate.


kstich (commented):
Should the spec docs for `@enum` be updated to indicate the "SHOULD have names"?

JordonPhillips:
There's already language saying "their use is strongly encouraged"

mtdowling (approved):

PR #615, opened 2020-10-26
Fix broken link in docs to protocol tests

rcoh:
*Description of changes:*
I updated the link to 1.3.0—let me know if you'd prefer master or somewhere else. Currently, the link just 404s.


kstich (changes requested):
These links should use the `__smithy_version__` version replacement instead of `1.3.0` directly so this stays up to date as we publish new versions.

rcoh:
Updated, generated docs locally, and manually verified the links are correctly generated.

kstich (approved):
Thanks for the fix!

PR #617, opened 2020-10-29
Add tests for writing in sections.  

kggilmer:
The inline test is failing due to  (what I believe to be) a bug.

*Issue #, if available:* N/A

*Description of changes:*

I have code that requires multiple integrations to all contribute a string to the same line in codegen output.  I use `CodeWriter.writeInLine()` but this produces no change in the codegen output.  This PR captures the bug.

I've added another test which passes which exercises `CodeWriter.write()` in addition to the breaking test.


mtdowling:
Thanks! I moved the tests to the smithy-utils package since the issue can be reproduced there. I also fixed the bug.

mtdowling (approved):

PR #626, opened 2020-11-06
Don't conflict URI if an endpoint prefix is present

JordonPhillips:
Two URIs can't conflict if one has an endpoint prefix, or if both have an endpoint prefix that can't resolve to the same prefix.


smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-conflict-validator.json
```diff
@@ -315,6 +333,120 @@
...
+                    "target": "smithy.api#String",
+                    "traits": {
+                        "smithy.api#hostLabel": {},
+                        "smithy.api#pattern": "foo",
+                        "smithy.api#required": {}
+                    }
+                }
+            }
+        },
+
```
mtdowling, line 402: ?

mtdowling (changes requested):
It seems like this also need a spec update. Maybe to the endpoint trait, but the URI trait conflict section could link to the longer description of deconflicting http traits in the endpoint trait?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-conflict-validator.json
JordonPhillips: whoops haha i put in that extra space to make it easier for me to compartmentalize my change but forgot to remove it

smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java
```diff
@@ -111,6 +113,43 @@ public boolean equals(Object other) {
...
 
+    /**
+     * Gets a list of explicitly conflicting label segments between this
+     * pattern and another.
+     *
+     * @param otherPattern SmithyPattern to check against.
+     * @return A list of Segment Pairs where each pair represents a conflict
+     *     and where the left side of the Pair is a segment from this pattern.
+     */
+    public List<Pair<Segment, Segment>> getConflictingLabelSegments(SmithyPattern otherPattern) {
```
mtdowling, line 124: Can we just return a map and state that the map maintains order?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpUriConflictValidator.java
```diff
@@ -159,6 +197,16 @@ private boolean isAllowableConflict(Model model, OperationShape operation, Opera
                 .collect(Collectors.toMap(Pair::getLeft, Pair::getRight));
     }
 
+    private Map<String, Pattern> getHostLabelPatterns(Model model, OperationShape operation) {
+        return OperationIndex.of(model).getInput(operation)
```
mtdowling, line 201: Personal nit: I'd replace this with good old-fashioned looking Java. Maybe it's because the indentation implies that the Optional composition doesn't wrap the streaming composition.

smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java
JordonPhillips: Yes, but this was an exiting method from UriPattern so it's gonna get ugly. Pushed the change in a separate commit, enjoy

mtdowling (approved):

Issue #616, opened 2020-10-28
Missing protocol test for xmlnamespace-trait applied on service shape


skotambkar:
[XmlNamespace trait](https://awslabs.github.io/smithy/1.0/spec/core/xml-traits.html#xmlnamespace-trait)  can be applied on service shapes. We should have protocol tests around this behavior.

Found this issue as `Route53` depend on this trait to correctly serialize xmlns into the request body.  
Related to https://github.com/aws/aws-sdk-go-v2/issues/850 .

PR #643, opened 2020-11-28
Clarify various aspects of xmlNamespace traits

mtdowling:
* The xmlNamespace trait can only be applied to simple types, list,
  map, set, structure, union, and service shapes. We previously
  allowed "*", but that includes resource and operation shapes where
  the trait isn't actually used.
* aws.protocols#awsQuery and aws.protocols#ec2Query protocol traits
  both now require that an xmlNamespace is set on the top-level
  service shape they are applied to so that the namespace is
  guaranteed to be known.
* Added various examples for how the xmlNamespace trait interacts
  with flattened lists and maps.
* Add a protocol test suite named "restXmlWithNamespace" to test that
  implementations add an xmlns to restXml services when necessary,
  closing #616

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-aws-protocol-tests/model/ec2Query/xml-lists.smithy
+++ b/smithy-aws-protocol-tests/model/ec2Query/xml-lists.smithy
@@ -23,7 +23,8 @@ use smithy.test#httpResponseTests
 /// 4. XML lists with @xmlName on its members
 /// 5. Flattened XML lists.
 /// 6. Flattened XML lists with @xmlName.
-/// 7. Lists of structures.
+/// 7. Flattened XML lists with @xmlNamespace.
+/// 8. Lists of structures.
 operation XmlLists {
     output: XmlListsOutput
 }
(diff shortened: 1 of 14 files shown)


smithy-aws-protocol-tests/model/ec2Query/xml-lists.smithy
```diff
@@ -23,7 +23,8 @@ use smithy.test#httpResponseTests
 /// 4. XML lists with @xmlName on its members
 /// 5. Flattened XML lists.
 /// 6. Flattened XML lists with @xmlName.
-/// 7. Lists of structures.
+/// 7. Flattened XML lists with @xmlNamespace.
```
kstich, line 26: `ec2Query` got the updates for lists, but not the new test for maps.

mtdowling: EC2 does not utilize any maps, so there aren't any tests for it in EC2.

kstich (approved):

PR #650, opened 2020-12-02
Add HTTP header trait validation

### LINKED ISSUE #614: Add httpHeader trait value validation
The `httpHeader` trait states that headers MUST adhere to RFC 7230#section-3.2, but we don't currently validate that in the model. https://github.com/awslabs/smithy/blob/master/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpHeaderTraitValidator.java#L40

mtdowling:
Closes #614 and also removes a non-inclusive term from the same
validator.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpHeaderTraitValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpHeaderTraitValidator.java
@@ -22,6 +22,7 @@ import static java.util.stream.Collectors.toList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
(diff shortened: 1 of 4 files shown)


kstich (approved):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpHeaderTraitValidator.java
```diff
@@ -68,15 +84,30 @@
...
+
+        if (BLOCKLIST.contains(header.toLowerCase(Locale.ENGLISH))) {
+            return Optional.of(danger(member, trait, String.format(
+                    "`%s` is not an allowed HTTP header binding", header)));
+        }
+
+        for (int i = 0; i < header.length(); i++) {
+            if (!TCHAR.contains(header.charAt(i))) {
+                return Optional.of(danger(member, trait, String.format(
+                        "`%s` is not a valid HTTP header field name according to section 3.2 of RFC 7230", header)));
```
JordonPhillips, line 104: Can you put the specific character that caused the failure in the error message?

JordonPhillips (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpHeaderTraitValidator.java
mtdowling: IMO, that isn't super important given we link to the RFC and show the header in the error message. I considered using a regular expression, but decided that this kind of check would be easier to understand. By keeping the validation message like this, we could theoretically switch to a regex and keep the same error message if we ever wanted to.

PR #661, opened 2020-12-08
Migrate CI to GitHub Actions

JordonPhillips:
*Description of changes:*

This migrates the CI testing from Travis to GitHub Actions.

See an example of what it looks like on my fork: https://github.com/JordonPhillips/smithy/pull/1


srchase (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+
+      - name: clean and build
```
kstich, line 25: Nit: name should probably be upper cased

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Java ${{ matrix.java }}
+    strategy:
+      matrix:
+        java: [8, 11]
```
kstich, line 15: Looks like these are OpenJDK versions, meaning we lose the OracleJDK from the previous matrix. This shouldn't be a problem, but noting in case someone else knows if we'd need it.

JordonPhillips: I don't think it's a problem.

kstich (approved):

PR #662, opened 2020-12-08
Fix truncated environment variable expansions

srchase:
Any environment variable expansions in smithy-build.json that did not start at the beginning of the field resulted in dropping the character that preceded the ${NAME} placeholder.

Consider the following `smithy-build.json` example, with the environment variable of `FOO` was set to `BAZ`:
```
{
    "version": "1.0",
    "projections": {
        "a": {
            "transforms": [
                {
                    "name": "includeShapesByTag",
                    "args": {
                        "tags": ["BAR${FOO}"]
                    }
                }
            ]
        }
    }
}
```
Currently, this results in the shapes with the tag `BABAZ` being included, instead of `BARBAZ` as expected.

This CR corrects the expansion behavior to assure preceding characters are included appropriately.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/model/ConfigLoader.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/model/ConfigLoader.java
@@ -115,6 +115,11 @@ final class ConfigLoader {
             while (matcher.find()) {
                 String variable = matcher.group(1);
                 String replacement = expand(node.getSourceLocation(), variable);
+                // INLINE over-matches to allow for escaping. If the over-matched first group does not start with
+                // '${', we need to prepend the first character from that group on the replacement.
+                if (!matcher.group(0).startsWith("${")) {
+                    replacement = matcher.group(0).charAt(0) + replacement;
+                }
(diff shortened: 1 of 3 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/model/ConfigLoader.java
```diff
@@ -115,6 +115,11 @@ public Node stringNode(StringNode node) {
             while (matcher.find()) {
                 String variable = matcher.group(1);
                 String replacement = expand(node.getSourceLocation(), variable);
+                // INLINE over-matches to allow for escaping. If the over-matched first group does not start with
+                // '${', we need to prepend the first character from that group on the replacement.
+                if (!matcher.group(0).startsWith("${")) {
```
mtdowling, line 120: This works, but it would probably be easier to understand if this used https://github.com/awslabs/smithy/blob/master/smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleParser.java. I think we can merge this for now to fix the bug, and then go back and improve this when we have time.

mtdowling (approved):

PR #692, opened 2021-01-22
Fix protocol test issues

### LINKED ISSUE #686: Clarifying questions on the restJson1 protocol spec
I have a couple clarifying questions on the restJson1 spec.

#### Question 1

[The spec states](https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html#content-type) that the default Content-Type of `application/json` should be used, and that if there is a @httpPayload member, then there must be a Content-Type header and provides a table of possible values. What's not clear to me is what the expected behavior is when there are no members, or when there is no input/output shape on the operation. From what I can see in the protocol tests, the content type is `application/json` in these cases, but is that actually correct? Does it make sense to have a Content-Type when there is no content?

Tests I'm referencing:
- [EmptyInputAndEmptyOutput](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy#L87), [IgnoreQueryParamsInResponse](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-query.smithy#L253), [HttpResponseCode](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-response-code.smithy#L29) - Each of these is setting Content-Type to `application/json` when there is no JSON payload.
- [GreetingWithErrors](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/errors.smithy#L38) - This one, at a glance, looks like it is JSON, but the greeting value is actually a HTTP header, so there is no payload? I think?
- [HttpPayloadTraits](https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-payload.smithy#L32) - From my understanding of the spec, I think this test should have a Content-Type of `application/octet-stream`?

I think I might be misunderstanding something.

#### Question 2

For HTTP headers in restJson1, I'm trying to figure out how the `@sparse` trait on lists/sets plays into header values. I couldn't find it mentioned in the spec, and didn't see any protocol tests that exercise it.

So if I have something like this:

```
@sparse
list SparseList {
    member: String
}

structure ExampleInput {
    @httpHeader("X-Example")
    example: SparseList,
}
```

Is this supported? If so, how are the null values expected to be serialized?

Thanks!

mtdowling:
Closes #686

*Issue #, if available:* 686

*Description of changes:*

This commit fixes various issues raised in #686. Some of this is related to just broken test cases (typically describing responses), and some of it is related to needing to indicate that a test case is meant to provide a client with an invalid response to ensure the client can gracefully handle it.

### DIFF
--- a/smithy-aws-protocol-tests/model/awsJson1_0/empty-input-output.smithy
+++ b/smithy-aws-protocol-tests/model/awsJson1_0/empty-input-output.smithy
@@ -16,8 +16,14 @@ operation NoInputAndNoOutput {}
 
 apply NoInputAndNoOutput @httpRequestTests([
     {
-        id: "AwsJson10NoInputAndNoOutput",
-        documentation: "No input serializes no payload",
+        id: "AwsJson10MustAlwaysSendEmptyJsonPayload",
+        documentation: """
+                Clients must always send an empty JSON object payload for
+                operations with no input (that is, `{}`). While AWS service
+                implementations support requests with no payload or requests
+                that send `{}`, always sending `{}` from the client is
+                preferred for forward compatibility in case input is ever
+                added to an operation.""",
         protocol: awsJson1_0,
         method: "POST",
         headers: {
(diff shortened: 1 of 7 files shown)


smithy-aws-protocol-tests/model/restJson1/errors.smithy
```diff
@@ -31,13 +31,8 @@ apply GreetingWithErrors @httpResponseTests([
         documentation: "Ensures that operations with errors successfully know how to deserialize the successful response",
         protocol: restJson1,
         code: 200,
-        body: """
```
mtdowling, line 34: TBH, I'm confused as to how any protocol test generators were able to pass this test case.

rcoh: I assume because none of the current protocol test implementors are _generating_ response bodies? So they probably just ignore the body since there are no fields in it for this shape?

jasdel: Looking at the v2 SDK's protocol test for this, it looks like it happily deserializes the response. I think I'm missing why generated tests should not be able to deserialize this response.

https://github.com/aws/aws-sdk-go-v2/blob/a55d57fffff519278eabd0e18ffc2bfeaac633c6/internal/protocoltest/awsrestjson/api_op_GreetingWithErrors_test.go#L30-L45

jasdel: Ah ! i see the greeting comes from a header, not the payload.

mtdowling: Yeah, it's likely because deserialization code in clients just looks at the things it needs to look at for this test case (e.g., the header that greeting is bound to), and ignores the payload. I could see this kind of test being useful for pathological testing to ensure clients tolerate unexpected response content, but that should be in addition to expected responses, and they should always be marked with the `client-only` tag.

smithy-aws-protocol-tests/model/awsJson1_0/empty-input-output.smithy
```diff
@@ -16,28 +16,97 @@ operation NoInputAndNoOutput {}
...
+                protocol.""",
+        protocol: awsJson1_0,
+        method: "POST",
+        headers: {
+            "Content-Type": "application/x-amz-json-1.0",
+            "X-Amz-Target": "JsonRpc10.NoInputAndNoOutput",
+        },
+        uri: "/",
+        body: "",
+        tags: ["server-only"]
```
rcoh, line 53: server only vs. client only seems fundamental enough that it may be worth modeling

mtdowling: It does feel that way. Let's do that in a follow up PR though since these tags are already in use.

smithy-aws-protocol-tests/model/awsJson1_0/empty-input-output.smithy
```diff
@@ -16,28 +16,97 @@ operation NoInputAndNoOutput {}
...
         headers: {
             "Content-Type": "application/x-amz-json-1.0",
         },
         code: 200,
-   }
+        // Service implementations must ignore this test.
+        tags: ["client-only"]
+    },
+    {
+        id: "AwsJson10Handles_unexpected_json_output",
```
JordonPhillips, line 76: nit: the mixing of cases is a bit awkward

smithy-aws-protocol-tests/model/restJson1/http-query.smithy
```diff
@@ -246,17 +246,32 @@ operation IgnoreQueryParamsInResponse {
 apply IgnoreQueryParamsInResponse @httpResponseTests([
     {
         id: "RestJsonIgnoreQueryParamsInResponse",
-        documentation: "Query parameters must be ignored when serializing the output of an operation",
+        documentation: """
+                Query parameters must be ignored when serializing the output
+                of an operation. As of January 2021, server implementations
+                are expected to response with a JSON object regardless of
```
JordonPhillips, line 252: expected to respond

JordonPhillips (approved):

JordonPhillips (approved):

PR #696, opened 2021-01-28
Add tags and appliesTo to protocol tests

mtdowling:
This commit adds support for applying tags to individual test cases,
allowing them to be grouped much more granularly. It also adds support
for appliesTo, which allows test cases to indicate if they should only
be implemented by clients or servers. Test cases that do not define an
appliesTo member are to be implemented by both client and server
implementations.

Existing AWS protocol tests are updated to now utilize appliesTo rather
than adding tags to operations.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
@@ -284,7 +284,8 @@ apply NullAndEmptyHeadersClient @httpRequestTests([
             a: null,
             b: "",
             c: [],
-        }
+        },
+        appliesTo: "client",
     },
 ])
 
@@ -309,7 +310,8 @@ apply NullAndEmptyHeadersServer @httpResponseTests([
             a: null,
             b: "",
             c: [],
-        }
+        },
+        appliesTo: "server",
     },
 ])
 
--- a/smithy-aws-protocol-tests/model/restXml/document-maps.smithy
+++ b/smithy-aws-protocol-tests/model/restXml/document-maps.smithy
@@ -125,7 +125,8 @@ apply XmlEmptyMaps @httpRequestTests([
         },
         params: {
             myMap: {}
-        }
+        },
+        appliesTo: "client"
     }
 ])
 
@@ -146,7 +147,8 @@ apply XmlEmptyMaps @httpResponseTests([
         },
         params: {
             myMap: {}
-        }
+        },
+        appliesTo: "client",
     },
     {
         id: "XmlEmptySelfClosedMaps",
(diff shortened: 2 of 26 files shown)


smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
```diff
@@ -265,7 +265,6 @@ structure InputAndOutputWithHeadersIO {
 /// Null and empty headers are not sent over the wire.
 @readonly
 @http(uri: "/NullAndEmptyHeadersClient", method: "GET")
-@tags(["client-only"])
 operation NullAndEmptyHeadersClient {
```
kstich, line 268: Can these operations be consolidated now that the distinction is on the tests themselves? This would require removing the operation level tags, though, and possibly cause a downstream issue. Maybe we just consolidate this as a followup.

smithy-aws-protocol-tests/model/restXml/document-maps.smithy
```diff
@@ -101,7 +101,6 @@ apply XmlMaps @httpResponseTests([
 ])
 
 @http(uri: "/XmlEmptyMaps", method: "POST")
-@tags(["client-only"])
 operation XmlEmptyMaps {
```
kstich, line 104: Looks like there's a set of `@httpResponseTests` for this operation. Should they have the `appliesTo` member added?

smithy-aws-protocol-tests/model/restXml/document-structs.smithy
```diff
@@ -227,7 +227,6 @@ structure SimpleScalarPropertiesInputOutput {
 // This example serializes empty string in the top level XML document.
 @idempotent
 @http(uri: "/XmlEmptyStrings", method: "PUT")
-@tags(["client-only"])
```
kstich, line 230: Looks like there's a set of `@httpResponseTests` for this operation. Should they have the `appliesTo` member added?

smithy-aws-protocol-tests/model/restXml/http-headers.smithy
```diff
@@ -265,7 +265,6 @@ structure InputAndOutputWithHeadersIO {
 /// Null and empty headers are not sent over the wire.
 @readonly
 @http(uri: "/NullAndEmptyHeadersClient", method: "GET")
-@tags(["client-only"])
 operation NullAndEmptyHeadersClient {
```
kstich, line 268: Can these operations be consolidated now that the distinction is on the tests themselves? This would require removing the operation level tags, though, and possibly cause a downstream issue. Maybe we just consolidate this as a followup.

kstich (changes requested):

smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
mtdowling: Since I'm adding tags back for now, this will need to wait. But it's a good idea.

kstich (approved):

PR #704, opened 2021-02-04
Add docs/tests for Glacier customizations

JordonPhillips:
This adds documentation and protocol tests for the customizations required to generate Glacier client.

For the protocol tests, I included a representative subset of the Glacier model with protocol tests added in. I'm unsure if this is the right strategy though. The other option would be to include the entire model. That would reduce the risk of excluding too much, but also potentially results in having some truly massive models buried in this repo.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy
@@ -44,7 +44,7 @@ service Glacier {
         params: {
             accountId: "foo",
             vaultName: "bar",
-        }
+        },
     },
     {
         id: "GlacierChecksums",
@@ -61,7 +61,8 @@ service Glacier {
         params: {
             accountId: "foo",
             vaultName: "bar",
-        }
+        },
+        appliesTo: "client",
     },
     {
         id: "GlacierAccountId",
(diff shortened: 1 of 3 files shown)


smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy
```diff
@@ -0,0 +1,168 @@
...
+)
+@restJson1
+@title("Amazon Glacier")
+@xmlNamespace(
+    uri: "http://glacier.amazonaws.com/doc/2012-06-01/",
+)
+service Glacier {
+    version: "2012-06-01",
+    operations: [
+        UploadArchive,
```
kstich, line 28: This should also include `UploadPart`, to validate that the checksum customizations are made to both operations.

smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy
```diff
@@ -0,0 +1,168 @@
...
+    ],
+}
+
+@httpRequestTests([
+    {
```
kstich, line 33: Should these tests have the `appliesTo: "client"` flag?

kstich (changes requested):

smithy-aws-protocol-tests/model/restJson1/services/glacier.smithy
JordonPhillips: The version header one no, the rest yes

kstich (approved):

PR #705, opened 2021-02-04
Add config definitions for aws protocol tests

JordonPhillips:
This  adds aws config definitions intended to be used alongside: https://github.com/awslabs/smithy/pull/702

It's not an exhaustive listing, but includes most of what should be needed for protocol tests.

### DIFF
--- a/smithy-aws-protocol-tests/model/aws-config.smithy
+++ b/smithy-aws-protocol-tests/model/aws-config.smithy
@@ -39,12 +39,14 @@ structure EnvironmentConfig {
     AWS_SECRET_ACCESS_KEY: String,
     AWS_DEFAULT_REGION: String,
     AWS_RETRY_MODE: RetryMode,
+    AWS_SESSION_TOKEN: String,
 }
 
 /// Config settings that can be set in the AWS config file.
 structure FileConfig {
-    access_key_id: String,
-    secret_access_key: String,
+    aws_access_key_id: String,
+    aws_secret_access_key: String,
+    aws_session_token: String,
     region: String,
     s3: S3Config,
     retry_mode: RetryMode,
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/aws-config.smithy
```diff
@@ -0,0 +1,95 @@
...
+/// Configuration specific to S3.
+structure S3Config {
+    use_accelerate_endpoint: Boolean,
+    use_dualstack_endpoint: Boolean,
+}
+
+/// Configuration that is set on the constructed client.
+structure ClientConfig {
+    access_key_id: String,
+    secret_access_key: String,
```
rcoh, line 63: missing session token.

smithy-aws-protocol-tests/model/aws-config.smithy
```diff
@@ -0,0 +1,95 @@
...
+structure EnvironmentConfig {
+    AWS_ACCESS_KEY_ID: String,
+    AWS_SECRET_ACCESS_KEY: String,
+    AWS_DEFAULT_REGION: String,
+    AWS_RETRY_MODE: RetryMode,
+}
+
+/// Config settings that can be set in the AWS config file.
+structure FileConfig {
+    access_key_id: String,
```
rcoh, line 46: I'm not positive I would model these—by modeling this explicitly, we'd need to upgrade Smithy for any new features that wanted to be tested (which is maybe OK?)

JordonPhillips: These are for the protocol tests here, you'd need to update this package anyway to add more tests to it. In any event, adding a line here is never gonna be the bottleneck.

kstich (approved):

kstich (approved):

PR #714, opened 2021-02-16
Throw syntax exceptions on duplicate bindings

kstich:
This commit updates the Smithy IDL parsing to throw when duplicate
bindings or members are found in the following declarations: service,
resource, and operation shapes, applied structure traits, and applied
traits with members that are structures. This behavior is unsafe for
model writers and shouldn't be valid.


kstich:
Looks like some windows-based newline issues, will take a look.

mtdowling (approved):

PR #726, opened 2021-02-28
Improve Selector usability

mtdowling:
This commit improves the usability of Selectors by adding new methods to
the Selector interface and by deprecating its runner functionality. The
new methods include methods to stream matching shapes, stream matches,
and consume matches. Streaming matches and shapes requires a minimal
amount of buffering, and they provide the ability to more easily perform
checks like determine if any shapes in a model match a selector. If all
matching shapes are necessary, calling select is preferred. If consuming
matches can be done in a way that is push-based (e.g., writing to CLI
output), then using consumeMatches is preferred to matches.

The previous runner() method of Selector is now deprecated and just
delegates calls back to the Selector. runner() was initially intended to
allow for the customization of the environment in which a selector is
executed, but this design turned out to be unnecessary, and as such, the
need to create a runner and use a kind of builder interface to use
Selectors made them awkward to work with for no benefit.

A final change is made to implement an IdentitySelector to optimize
matching all shapes in a model.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
@@ -15,14 +15,18 @@
 
 package software.amazon.smithy.model.selector;
 
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.shapes.Shape;
 
 /**
  * Provides a toString method that prints the expression.
- *
- * <p>This is the only concrete implementation of Selector that is
- * exported from the package.
  */
 final class WrappedSelector implements Selector {
     private final String expression;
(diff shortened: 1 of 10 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
```diff
@@ -47,23 +51,93 @@
...
+        });
+    }
+
+    @Override
+    public Stream<ShapeMatch> matches(Model model) {
+        Context context = createContext(model);
+        return streamStartingShape(model).flatMap(shape -> {
+            List<ShapeMatch> result = new ArrayList<>();
+            delegate.push(context.clearVars(), shape, (ctx, s) -> {
+                result.add(new ShapeMatch(s, ctx.copyVars()));
```
kstich, line 114: Is there a reason this uses `copyVars` with the above note on using `getVars` instead in `consumeMatches`? I don't see anything different enough to warrant it.

smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
```diff
@@ -47,23 +51,93 @@
...
 
     @Override
     public int hashCode() {
         return expression.hashCode();
     }
+
+    @Override
+    public Set<Shape> select(Model model) {
+        Set<Shape> result = new HashSet<>();
+        // This is more optimized that using shapes() and collecting to a Set
```
kstich, line 72: s/that/than

smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
```diff
@@ -47,23 +51,93 @@
...
+        pushShapes(model, (ctx, s) -> {
+            result.add(s);
+            return true;
+        });
+        return result;
+    }
+
+    @Override
+    public void consumeMatches(Model model, Consumer<ShapeMatch> shapeMatchConsumer) {
+        // This is more optimized that using matches() and collecting to a Set
```
kstich, line 84: s/that/than

smithy-model/src/main/java/software/amazon/smithy/model/selector/WrappedSelector.java
mtdowling: Oh good point. Will change this one.

kstich (approved):

PR #746, opened 2021-03-22
Failing Request Test Draft

rcoh:
This is a draft of how failing request test support may be added in a way that will support failing requests for both server and client implementations.

This has a few issues & required cleanups, just want to see if we're directionally on the same page.

*Issue #, if available:*

*Description of changes:*


srchase:
CI is failing on some checkstyle issues: Import ordering, an unused import, missing copyright header and wrong copyright year, and line lengths exceeding 120 chars.

mtdowling:
The idea of defining negative protocol tests are interesting, but do we have a use case in mind for them? I also think if we added them, it might be easier to add them to the existing traits.

rcoh:
yeah can close, I think adding a boolean to the existing traits is probably better.

PR #760, opened 2021-04-05
Support service renames when flattening namespaces

srchase:
Adds support for service renames (#734) when using the `flattenNamespaces` transformer.

Any shapes that have been configured on the service to be renamed will also be flattened into the target namespace.


JordonPhillips (changes requested):
This could use a doc update. I don't think it's necessarily obvious behavior.

JordonPhillips (approved):

smithy-build/src/main/java/software/amazon/smithy/build/transforms/FlattenNamespaces.java
```diff
@@ -164,9 +164,18 @@ private ShapeId updateNamespace(ShapeId shapeId, String namespace) {
         return shapeWalker.walkShapes(service).stream()
                 .filter(FunctionalUtils.not(Prelude::isPreludeShape))
                 .map(shape -> Pair.of(shape.getId(), updateNamespace(shape.getId(), config.getNamespace())))
+                .map(pair -> applyServiceRenames(pair, service))
```
mtdowling, line 167: Nit: ideally we can destructure pairs before calling methods with them. That would give the left and right shape ID a name so it's easier to understand the code.

PR #761, opened 2021-04-06
Add core contributors to PRs by default

JordonPhillips:
This adds a codeowners file that will automatically request reviews from the core contributors on each pr. We could make it more complicated but that doesn't seem necessary right now.


srchase (approved):
I think this might be a bit noisy, but I think it's worth trying.

PR #766, opened 2021-04-08
Parse numbers as BigDecimal or BigInteger when needed.

srchase:
Currently, numbers greater than 9223372036854775807 fail to be parsed from the IDL with `NumberFormatException`, even when the underlying implementation supports BigDecimal and BigInteger.

This PR updates the IdlModelParser and NodeHandler to use fall back to using `BigDecimal` or `BigInteger` where appropriate.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -724,7 +724,11 @@ final class IdlModelParser extends SimpleParser {
         String lexeme = ParserUtils.parseNumber(this);
 
         if (lexeme.contains("e") || lexeme.contains(".")) {
-            return new NumberNode(Double.valueOf(lexeme), location);
+            double value = Double.parseDouble(lexeme);
+            if (Double.isFinite(value)) {
+                return new NumberNode(value, location);
+            }
+            return new NumberNode(new BigDecimal(lexeme), location);
         } else {
             try {
                 return new NumberNode(Long.parseLong(lexeme), location);
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -17,6 +17,7 @@ package software.amazon.smithy.model.loader;
 
 import static java.lang.String.format;
 
+import java.math.BigDecimal;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -721,10 +722,15 @@ final class IdlModelParser extends SimpleParser {
     NumberNode parseNumberNode() {
         SourceLocation location = currentLocation();
         String lexeme = ParserUtils.parseNumber(this);
+
         if (lexeme.contains("e") || lexeme.contains(".")) {
             return new NumberNode(Double.valueOf(lexeme), location);
         } else {
-            return new NumberNode(Long.parseLong(lexeme), location);
+            try {
+                return new NumberNode(Long.parseLong(lexeme), location);
+            } catch (NumberFormatException e) {
+                return new NumberNode(new BigDecimal(lexeme), location);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/node/internal/NodeHandler.java
```diff
@@ -79,7 +79,12 @@ void endNumber(String string, SourceLocation location) {
...
             value = new NumberNode(new BigDecimal(string), location);
         } else {
-            value = new NumberNode(Long.parseLong(string), location);
+            try {
+                value = new NumberNode(Long.parseLong(string), location);
+            } catch (NumberFormatException e) {
+                value = new NumberNode(new BigDecimal(string), location);
+            }
+
         }
```
JordonPhillips, line 88: Shouldn't this have the same logic as loading from the IDL? Interesting that this always parsed as a big decimal. It looks like previously this would have failed on `1E-1` or something like that since it tries to parse a long. In any event complex numbers should be decimals like the idl parser tries to do.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -721,10 +722,19 @@ SourceLocation currentLocation() {
     NumberNode parseNumberNode() {
         SourceLocation location = currentLocation();
         String lexeme = ParserUtils.parseNumber(this);
+
         if (lexeme.contains("e") || lexeme.contains(".")) {
```
JordonPhillips, line 726: Wait this should be checking for both lower and upper case e - `ParserUtils.parseNumber` supports both and doesn't do any lowercasing.

JordonPhillips (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -721,10 +722,19 @@ SourceLocation currentLocation() {
...
-            return new NumberNode(Double.valueOf(lexeme), location);
+            double value = Double.parseDouble(lexeme);
```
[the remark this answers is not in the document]
srchase: Anything bigger than `1.7976931348623157E+309` returns just returns `Infinity`.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -721,10 +722,19 @@ SourceLocation currentLocation() {
...
+            if (Double.isFinite(value)) {
+                return new NumberNode(value, location);
+            }
+            return new NumberNode(new BigDecimal(lexeme), location);
         } else {
-            return new NumberNode(Long.parseLong(lexeme), location);
+            try {
+                return new NumberNode(Long.parseLong(lexeme), location);
+            } catch (NumberFormatException e) {
+                return new NumberNode(new BigDecimal(lexeme), location);
```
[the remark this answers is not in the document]
srchase: Fixed in a4dcceb.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
srchase: Fixed in a4dcceb.

smithy-model/src/main/java/software/amazon/smithy/model/node/internal/NodeHandler.java
srchase: Fixed in a4dcceb.

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/traits.smithy
```diff
@@ -134,3 +134,12 @@ apply E @deprecated
 
 @documentation("")
 string V
+
+@trait
+bigDecimal bDecimal
```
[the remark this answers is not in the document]
srchase: Added in a4dcceb.

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/traits.json
```diff
@@ -171,6 +171,48 @@
...
+        "com.example#Y": {
+            "type": "string",
+            "traits": {
+                "com.example#bDecimal": 2E+308
+            }
+        },
+        "com.example#Z": {
+            "type": "string",
+            "traits": {
+                "com.example#bDecimal": 2E+308
```
JordonPhillips, line 195: I think one of these was supposed to use a lowercase e

JordonPhillips (approved):

PR #779, opened 2021-04-23
Warn against potentially risky hostPrefixes

### LINKED ISSUE #778: hostPrefix property in endpoint trait should validate prefix ends in dot (.)
The endpoint trait is not required to end in a dot. Is this by design? In the Smithy docs for `hostLabel` it has this example:

```
@endpoint(hostPrefix: "{foo}.data")
```

But for an endpoint `example.com` and `foo = "foo"` it would become `foo.dataexample.com`, unless clients or code generators are validating against this. I think it's dangerous to allow this because it changes the configured host and not limited to prefixing as the trait suggests. If you agree, I think it also makes sense to call out this validation in the documentation, too.

JordonPhillips:
This updates the documentation and adds a warning for the case where a `hostPrefix` contains a host label but does not end in a period. This is potentially risky since it could change the domain to something that the modeler doesn't control. For example, a host label of `{foo}` combined with a configured endpoint `example.com` could result in a resolved host of `mistakeexample.com` when the modeler wanted it to resolve to `mistake.example.com`. If the modeler doesn't own EVERY domain suffixed by the configured endpoint then it is possible that customer data will be exposed.

Resolves #778 

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HostLabelTraitValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HostLabelTraitValidator.java
@@ -50,6 +50,8 @@ import software.amazon.smithy.model.validation.ValidationUtils;
  *     operation have corresponding hostLabel traits.</li>
  *     <li>Validates that an expanded version of each endpoint hostPrefix of
  *     each operation is a valid RFC 3896 host.</li>
+ *     <li>Validates that the host prefix SHOULD end in a period if it contains
+ *     host labels.</li>
  * </ul>
  */
 public final class HostLabelTraitValidator extends AbstractValidator {
@@ -81,6 +83,9 @@ public final class HostLabelTraitValidator extends AbstractValidator {
 
         // Validate the host can become a valid RFC 3986 Section 3.2.2 host.
         validateExpandedPattern(operation, endpoint).ifPresent(events::add);
+
+        // Validate the host prefix SHOULD end in a period if it has labels.
+        validateTrailingPeriod(operation, endpoint).ifPresent(events::add);
 
         // If the operation has labels then it must also have input.
         if (!operation.getInput().isPresent() && !endpoint.getHostPrefix().getLabels().isEmpty()) {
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HostLabelTraitValidator.java
```diff
@@ -154,4 +159,14 @@
         }
         return Optional.empty();
     }
+
+    private Optional<ValidationEvent> validateTrailingPeriod(OperationShape operation, EndpointTrait trait) {
+        SmithyPattern hostPrefix = trait.getHostPrefix();
+        if (!hostPrefix.toString().endsWith(".") && !hostPrefix.getLabels().isEmpty()) {
+            String message = "`endpoint` trait hostPrefix contains host labels and does not end in a period (`.`). "
+                    + "This can result in clients inadvertently sending data to domains you do not control.";
+            return Optional.of(warning(operation, trait, message));
```
JordonPhillips, line 168: I am also open to changing this to a DANGER. Right now 8 AWS services use a `hostPrefix` and none of them would run afoul of this validator.

mullermp:
Thanks for addressing the issue. The [hostLabel documentation](https://awslabs.github.io/smithy/1.0/spec/core/endpoint-traits.html#hostlabel-trait)
has an example `@endpoint(hostPrefix: "{foo}.data")`. Could this be changed too?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HostLabelTraitValidator.java
mullermp: DANGER (or an outright failure) I think is appropriate..

JordonPhillips:
Whoops, that was actually a bug - the other example on the page and the json version of that same example all end in periods

srchase (approved):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HostLabelTraitValidator.java
kstich: +1 on making this a DANGER, given it's specific to the label-having prefixes.

JordonPhillips: done

kstich (approved):

PR #781, opened 2021-04-23
add restXml tests for flattened list of structures

aajtodd:
*Description of changes:*
* Adds a test for ignoring the root element name since some services like S3 do not send a root element that matches the structure shape id from the model
* expanded the xml list test to include a flattened list of structures (we had an issue in Kotlin where the deserializer state got messed up by this case, maybe it helps someone else)

### DIFF
--- a/smithy-aws-protocol-tests/model/restXml/document-lists.smithy
+++ b/smithy-aws-protocol-tests/model/restXml/document-lists.smithy
@@ -28,6 +28,7 @@ use smithy.test#httpResponseTests
 /// 6. Flattened XML lists with @xmlName.
 /// 7. Flattened XML lists with @xmlNamespace.
 /// 8. Lists of structures.
+/// 9. Flattened XML list of structures
 @idempotent
 @http(uri: "/XmlLists", method: "PUT")
 operation XmlLists {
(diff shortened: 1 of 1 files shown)


aajtodd:
It's not clear if the `IgnoresRootElementName` belongs with the general `restXml` protocol tests suite or should be made specific to S3 customization test suite (or if the model should just be fixed but that's a longer resolution)

srchase:
https://github.com/awslabs/smithy/pull/789 was added which addresses the root element name, namely that it should not be ignored. Instead, models should properly set the root element with an `xmlName` trait if needed.

Could you remove the root element part of this in PR?

aajtodd:
> #789 was added which addresses the root element name, namely that it should not be ignored.

The test there is a nice addition but that doesn't address the root issue which was that the model is wrong. I'm assuming though that what you're saying is that the model(s) will be updated appropriately to include missing `xmlName` trait?

I've removed the test for ignoring the name, only the flattened list of structures remains.

JordonPhillips:
>  I'm assuming though that what you're saying is that the model(s) will be updated appropriately to include missing `xmlName` trait?

Yes

JordonPhillips (approved):

srchase (approved):

PR #794, opened 2021-05-10
Fix bug with cloudformation unions

JordonPhillips:
This fixes a bug where having unions in your model would cause the
cloudformation schema generation to fail.


mtdowling (approved):
LGTM, but we could also look into a future change that fixes this as the shapes are converted to JSON Schema rather than mapping over the already created JSON Schema. Maybe using this: https://github.com/awslabs/smithy/blob/main/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaMapper.java

PR #804, opened 2021-05-16
Update Protocol Test with more complex XML Unescaping

rcoh:
*Description of changes:*

Add protocol response test to cover XML escaping of literals as well as decimal & hexadecimal escape sequences.

### DIFF
--- a/smithy-aws-protocol-tests/model/restXml/document-structs.smithy
+++ b/smithy-aws-protocol-tests/model/restXml/document-structs.smithy
@@ -142,12 +142,14 @@ apply SimpleScalarProperties @httpResponseTests([
         }
     },
     {
-        id: "SimpleScalarPropertiesWithEscapedCharacter",
+        id: "SimpleScalarPropertiesComplexEscapes",
         documentation: """
         Serializes string with escaping.
 
         This validates the three escape types: literal, decimal and hexadecimal. It also validates that unescaping properly
         handles the case where unescaping an & produces a newly formed escape sequence (this should not be re-unescaped).
+
+        Servers may produce different output, this test is designed different unescapes clients must handle
         """,
         protocol: restXml,
         code: 200,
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restXml/document-structs.smithy
```diff
@@ -143,12 +143,17 @@ apply SimpleScalarProperties @httpResponseTests([
...
+
+        This validates the three escape types: literal, decimal and hexadecimal. It also validates that unescaping properly
+        handles the case where unescaping an & produces a newly formed escape sequence (this should not be re-unescaped).
+        """,
         protocol: restXml,
         code: 200,
         body: """
               <SimpleScalarPropertiesInputOutput>
-                  <stringValue>&lt;string&gt;</stringValue>
+                  <stringValue>escaped data: &amp;lt;&#xD;&#10;</stringValue>
```
JordonPhillips, line 156: I tested existing implementations and most will actually return the newline character un-escaped, though it's still reasonable to have this as a response test. Still, you should mark this as applying to clients only, and ideally add a server variant that reflects the existing behavior.

I also tested S3 by creating an object with this as a key, and it seems they prefer to url encode: `escaped+data%3A+%26lt%3B%0D%0A`

JordonPhillips (changes requested):

smithy-aws-protocol-tests/model/restXml/document-structs.smithy
rcoh: yeah good call, this should be clients only. Re: S3, are they returning that in an XML payload?

JordonPhillips: Yeah. Create an object with your test string as a key and then list objects

PR #805, opened 2021-05-17
Make various perf improvements

mtdowling:
This commit makes various performance improvements that particularly help
when validating models with 100k+ shapes.

* Fixed JMH benchmarks so they work with the updated Gradle version.
* Model now uses a synchronized IdentityHashMap now to back the
  blackboard cache rather than a ConcurrentHashMap. This will actually
  now prevent duplicate work in creating KnowledgeIndexes. An
  IdentityHashMap was used because it works well for the classes
  used to cache knowledge indexes.
* HttpBindingIndex now uses a WeakReference to Model. This was
  previously an unnecessary cyclical reference.
* HttpBindingIndex no longer throws when attempting to access the
  HTTP bindings of shapes that don't exist or aren't operations.
  This prevents it from having to store a List entry for every
  single operation.
* Model#getShapesWithTraits was used everywhere possible rather
  than streaming over all shapes and looking for traits.
* Made optimizations to NullableIndex to no longer need to
  traverse every shape.
* Removed unnecessary streaming from OperationIndex.
* Removed unnecessary streaming from PaginatedIndex.
* Removed unnecessary streaming from NeighborVisitor.
* Updated Node expect methods to also accept a Supplier to create
  error messages if the expectation fails. This prevents needing to
  evaluate String.format even for valid node assertions.
* AttributeSelector no longer uses a BiFunction key supplier, and instead
  the attribute path is just passed in. This allows for the selector
  to also perform optimizations when determining if a shape has a trait
  by leveraging Model#getShapesWithTraits.
* InternalSelectors used to implement Selectors now support more general
  optimizations. This was previously hardcoded to only support an optimization
  for selecting shapes by type, but now selecting shapes by trait is
  optimized too.
* Minor optimization to structure and union loading so that when validating
  that members have a shape ID compatible with the container, an intermediate
  shape ID no longer is constructed.
* The ShapeId cache was increased from 1024 to 8192. This helps significantly
  with large models. The ShapeId cache was also updated to implement the
  LRA cache inside of the computeIfAbsent method.
* NodeValidationVisitor now has a Context object that supports caching
  selectors evaluated against a model. This helps significantly with IdRef
  validation.  To make this caching reusable, the visitor is now mutable
  after it is constructed.
* NodeValidationVisitor idRef now special cases "*" and uses the context
  cache.
* TraitTargetValidator has been simplified, special cases "*", and now
  uses a cache to speed up evaluating traits that use the same selectors.
* TraitValueValidator now reuses the same NodeValidationVisitor in order
  to reuse the same selector cache.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/selector/InternalSelector.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/selector/InternalSelector.java
@@ -15,6 +15,9 @@
 
 package software.amazon.smithy.model.selector;
 
+import java.util.Collection;
+import java.util.function.Function;
+import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.Shape;
 
 /**
@@ -51,6 +54,21 @@ interface InternalSelector {
      * @return Returns true to continue sending shapes to the selector.
      */
     boolean push(Context ctx, Shape shape, Receiver next);
+
+    /**
+     * Returns a function that is used to optimize which shapes in a model
+     * need to be evaluated.
+     *
+     * <p>For example, when selecting "structure", it is far less work
+     * to leverage {@link Model#toSet(Class)} than it is to send every shape
+     * through every selector.
+     *
+     * @return Returns a function that returns null if no optimization can
+     *   be made, or a Collection of Shapes if an optimization was made.
+     */
+    default Function<Model, Collection<? extends Shape>> optimize() {
+        return null;
+    }
 
     /**
      * Receives shapes from an InternalSelector.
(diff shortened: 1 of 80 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/selector/InternalSelector.java
```diff
@@ -52,6 +55,21 @@
      */
     boolean push(Context ctx, Shape shape, Receiver next);
 
+    /**
+     * Returns a function that is used to stream shapes out of the given
+     * model in a more optimized way based on the selector.
+     *
+     * <p>For example, when selecting "structure", it is far less work
+     * to leverage {@code mode.shapes(StructureShape.class)} than it is
```
kstich, line 63: s/mode/model

kstich (approved):
Lots of bits that look like they add up, nice!

kstich (approved):

PR #806, opened 2021-05-18
Add methods to get shapes by type and trait

mtdowling:
Model now contains methods for every shape type to get a Set of shapes
or that type or a set of shapes of a type that have a specific trait.
These methods are simpler to use than `shapes(Class)` and
`toSet(Class)`, and they hopefully encourage the use of the caches that
Model uses. These methods also eliminated a lot of boilerplat that
getShapesWithTrait previously required when only shapes of a certain
type were needed.

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -192,6 +213,424 @@ private TraitCache getTraitCache() {
...
+
+    /**
+     * Gets an immutable set of all bigDecimals in the Model that have a specific trait.
+     *
+     * @param trait The exact trait class to look for on shapes.
+     * @return Returns the set of {@code bigdecimal}s that have a specific trait.
+     */
+    public Set<BigDecimalShape> getBigDecimalShapesWithTrait(Class<? extends Trait> trait) {
+        return new ShapeTypeFilteredSet<>(getShapesWithTrait(trait), BigDecimalShape.class);
+    }
```
JordonPhillips, line 233: I'm not really a huge fan of having methods for each shape type, but this would head-off fat-fingering and it's a bit easier to understand for people less experienced with java.

smithy-model/src/main/java/software/amazon/smithy/model/ShapeTypeFilteredSet.java
```diff
@@ -0,0 +1,111 @@
...
+
+        @SuppressWarnings("unchecked")
+        private T computeNext() {
+            // Filter out shapes of other types.
+            while (iterator.hasNext()) {
+                Shape nextCandidate = iterator.next();
+                if (shapeType == nextCandidate.getClass()) {
+                    return (T) nextCandidate;
+                }
+            }
```
JordonPhillips, line 90: What's the reason for skipping elements rather than creating a new set that only has those elements? Because it will likely only be run through once?

JordonPhillips (approved):

smithy-model/src/main/java/software/amazon/smithy/model/ShapeTypeFilteredSet.java
mtdowling: This filtered view of the Set means that we don't have to create a new Set and preemptively copy all the values from one set to the other. Almost all the code that uses these methods (at least in our libraries) will just be iterating over the Set and not need the entire container populated up front.

smithy-model/src/main/java/software/amazon/smithy/model/ShapeTypeFilteredSet.java
```diff
@@ -0,0 +1,111 @@
...
+    public Iterator<T> iterator() {
+        return new FilteredIterator<>(shapes.iterator(), shapeType);
+    }
+
+    @Override
+    public int size() {
+        // Compute and cache the number of elements in the set with the desired type.
+        int result = size;
+
+        // Note that a resulting size with no elements is inefficiently cached.
```
kstich, line 56: Is it worth using a boxed `Integer` for `size` to distinguish the `int` zero value and empty set cases?

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -192,6 +213,424 @@ private TraitCache getTraitCache() {
...
+     * Gets an immutable set of all bigIntegers in the Model.
+     *
+     * @return Returns the Set of {@code bigIntegers}s.
+     */
+    public Set<BigIntegerShape> getBigIntegerShapes() {
+        return toSet(BigIntegerShape.class);
+    }
+
+    /**
+     * Gets an immutable set of all bigintegers in the Model that have a specific trait.
```
kstich, line 245: s/bigintegers/bigIntegers

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
kstich: This is also much more usable than having two `Class` params generally.

smithy-model/src/main/java/software/amazon/smithy/model/ShapeTypeFilteredSet.java
mtdowling: Actually, we can just default the size to `-1`.

kstich (approved):

PR #822, opened 2021-06-08
Validate known media types in protocol tests

### LINKED ISSUE #821: Fix Ec2Query protocol codegen tests
**Observed behavior**

I believe that I have identified two tests in the Ec2Query protocol codegen tests that seem to contain invalid XML:

1. Test id `Ec2XmlNamespaces`, located [here](https://github.com/awslabs/smithy/blob/a3682362d104e4a7d1d680427f4a5703194763ee/smithy-aws-protocol-tests/model/ec2Query/xml-structs.smithy#L390) seems to include invalid xml because it has two different values for `xmlns=""` defined in the root element.  I believe we need a similar fix to in [this commit](https://github.com/awslabs/smithy/commit/32ef0f10193f612bd476066de1497cf7f344f217#diff-4403c51782de0df6ec1df844c308cf4d9898661d7ea10848b6f966fd3fd23e06) made for AWSQuery
2. Test id `Ec2IgnoresWrappingXmlName`, located [here](https://github.com/awslabs/smithy/blob/a3682362d104e4a7d1d680427f4a5703194763ee/smithy-aws-protocol-tests/model/ec2Query/xml-structs.smithy#L455) also include invalid xml due to `xmlns=` having two values

**Expected behavior**
Tests should not include invalid xml

mtdowling:
This commit updates protocol test validators to ensure that known media
types for XML and JSON contain valid content.

Closes #821

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-aws-protocol-tests/model/ec2Query/xml-structs.smithy
+++ b/smithy-aws-protocol-tests/model/ec2Query/xml-structs.smithy
@@ -394,7 +394,7 @@ apply XmlNamespaces @httpResponseTests([
         protocol: ec2Query,
         code: 200,
         body: """
-              <XmlNamespacesResponse xmlns="http://foo.com" xmlns="https://example.com/">
+              <XmlNamespacesResponse xmlns="https://example.com/">
                   <nested>
                       <foo xmlns:baz="http://baz.com">Foo</foo>
                       <values xmlns="http://qux.com">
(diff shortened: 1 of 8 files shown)


kstich (changes requested):
Looks like the `valid-xml.smithy` file is empty.

kstich (approved):

PR #832, opened 2021-06-14
Setup cross platform builds of the CLI

mtdowling:
This commit adds support for cross platform builds of the CLI using
jlink via each support version of JDK 16 of Amazon Corretto (we'll bump
to JDK 17 when it's available). Each supported build target is built by
downloading the appropriate JDK from Corretto and running jlink.

The Dockerfile has been updated to now use the prebuilt binary for
linux-x86_64, and will now use application class data sharing to improve
CLI startup time in the Docker container.

A couple dependencies were updated, including Gradle, in order to build
on JDK 16. JDK 16 also complained about creating an explicitly boxed
value, so that was addressed too.

Future work should include building signed tar.gz and zip files, creating a
Mac pkg file via jpackage, and creating an msi via jpackage.

### DIFF
--- a/smithy-cli/build.gradle
+++ b/smithy-cli/build.gradle
@@ -13,9 +13,11 @@
  * permissions and limitations under the License.
  */
 
+import org.apache.tools.ant.taskdefs.condition.Os
+
 plugins {
     id "application"
-    id "org.beryx.runtime" version "1.8.4"
+    id "org.beryx.runtime" version "1.12.5"
 }
 
 description = "This module implements the Smithy command line interface."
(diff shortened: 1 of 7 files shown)


kstich (approved):

smithy-cli/build.gradle
```diff
@@ -35,16 +59,51 @@ dependencies {
...
+    targetPlatform("linux-x86_64") {
+        jdkHome = jdkDownload("${correttoRoot}-x64-linux-jdk.tar.gz")
+    }
+
+    targetPlatform("linux-aarch_64") {
+        jdkHome = jdkDownload("${correttoRoot}-aarch64-linux-jdk.tar.gz")
+    }
+
+    targetPlatform("osx-x86_64") {
+        jdkHome = jdkDownload("${correttoRoot}-x64-macos-jdk.tar.gz")
```
JordonPhillips, line 84: what about arm macos

JordonPhillips (changes requested):

smithy-cli/build.gradle
JordonPhillips: ah coretto doesn't have a build target for it yet

JordonPhillips (approved):

PR #833, opened 2021-06-15
add new awsQuery tests for nested structs and fix endpoint test

aajtodd:
*Description of changes:*
We ran into an issue with our serializer implementation on a real service (SES) that had a nested structure with list input (specifically [SendEmailRequest.Destination$ToAddresses](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/sdk-codegen/aws-models/ses.2010-12-01.json#L1834). I've added tests for both lists and maps (double check for correctness). Also there appears to be a broken test unless I'm mistaken, I've fixed that as well. 

- Adds tests lists/maps in a nested structure
- Fix the expected version of the `AwsQueryEndpointTrait` test

### DIFF
--- a/smithy-aws-protocol-tests/model/ec2Query/input-lists.smithy
+++ b/smithy-aws-protocol-tests/model/ec2Query/input-lists.smithy
@@ -85,6 +85,23 @@ apply QueryLists @httpRequestTests([
             ListArgWithXmlName: ["A", "B"]
         }
     },
+    {
+        id: "Ec2ListNestedStructWithList",
+        documentation: "Nested structure with a list member",
+        protocol: ec2Query,
+        method: "POST",
+        uri: "/",
+        headers: {
+            "Content-Type": "application/x-www-form-urlencoded"
+        },
+        body: "Action=QueryLists&Version=2020-01-08&NestedWithList.ListArg.member.1=A&NestedWithList.ListArg.member.2=B",
+        bodyMediaType: "application/x-www-form-urlencoded",
+        params: {
+            NestedWithList: {
+                ListArg: ["A", "B"]
+            }
+        }
+    },
 ])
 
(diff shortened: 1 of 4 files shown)


JordonPhillips (approved):
Looks good! Would you mind replicating these into the ec2 query tests as well?

aajtodd:
> Looks good! Would you mind replicating these into the ec2 query tests as well?
@JordonPhillips Sure thing, do you want it in a separate PR or update this one?

JordonPhillips:
updating this one is good

smithy-aws-protocol-tests/model/ec2Query/input-lists.smithy
```diff
@@ -85,6 +85,23 @@ apply QueryLists @httpRequestTests([
...
+    {
+        id: "Ec2ListNestedStructWithList",
+        documentation: "Nested structure with a list member",
+        protocol: ec2Query,
+        method: "POST",
+        uri: "/",
+        headers: {
+            "Content-Type": "application/x-www-form-urlencoded"
+        },
+        body: "Action=QueryLists&Version=2020-01-08&NestedWithList.ListArg.member.1=A&NestedWithList.ListArg.member.2=B",
```
JordonPhillips, line 97: ec2 doesn't have the `member` segment - it looks like our docs are actually wrong on this

JordonPhillips (changes requested):

smithy-aws-protocol-tests/model/ec2Query/input-lists.smithy
aajtodd: Ahh good catch, haven't implemented EC2 protocol yet and missed that difference.

JordonPhillips (approved):

PR #836, opened 2021-06-16
Fix list and map ec2 query serialization docs

JordonPhillips:
This fixes the query serialization docs for ec2Query:

* Lists are always flat
* Maps are undefined

### DIFF
--- a/docs/source/1.0/spec/aws/aws-query-protocol.rst
+++ b/docs/source/1.0/spec/aws/aws-query-protocol.rst
@@ -95,6 +95,27 @@ that affect serialization:
 
 .. |quoted shape name| replace:: ``awsQuery``
 .. |name resolution text| replace:: The :ref:`xmlName-trait` can be used to serialize a property using a custom name
+.. |query collection text| replace::
+    Each value provided in the collection is serialized as a separate key with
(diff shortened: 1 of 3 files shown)


docs/source/1.0/spec/aws/aws-query-protocol.rst
```diff
@@ -95,6 +95,27 @@ that affect serialization:
 
 .. |quoted shape name| replace:: ``awsQuery``
 .. |name resolution text| replace:: The :ref:`xmlName-trait` can be used to serialize a property using a custom name
+.. |query collection text| replace::
```
mtdowling, line 98: Do you think the shared templates are still helping?

mtdowling: Or maybe we can make them a little more granular?

mtdowling (approved):

docs/source/1.0/spec/aws/aws-query-protocol.rst
JordonPhillips: These big blocks aren't just copied into the template, they're repeated later on in the doc where there's a long-form examples.

PR #838, opened 2021-06-19
Add proposal for mixins

mtdowling:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/designs/mixins.md
+++ b/designs/mixins.md
@@ -684,8 +684,8 @@ A `member` relationship exists from `ListSomethingInput` to the following
 shapes, in any order:
 
 - `smithy.example#ListSomethingInput$nameFilter`
-- `smithy.example#PaginatedInput$nextToken`
-- `smithy.example#PaginatedInput$pageSize`
+- `smithy.example#ListSomethingInput$nextToken`
+- `smithy.example#ListSomethingInput$pageSize`
 
 The following two selectors:
 
@@ -700,8 +700,8 @@ structure[id|name = ListSomethingInput] -[member]-> *
 Both yield the following shapes, in any order:
 
 - `smithy.example#ListSomethingInput$nameFilter`
-- `smithy.example#PaginatedInput$nextToken`
-- `smithy.example#PaginatedInput$pageSize`
+- `smithy.example#ListSomethingInput$nextToken`
+- `smithy.example#ListSomethingInput$pageSize`
 
 
 ### Mixins and member ordering
--- a/designs/mixins.md
+++ b/designs/mixins.md
@@ -1,0 +1,1139 @@
+# Smithy Mixins
+
+This proposal defines _mixins_, a modeling mechanism that allows for structure
+and union definition reuse.
+
+> Note: mixins will be considered for Smithy IDL 1.1, which treats commas as
+> optional whitespace.
+
+
+## Motivation
+
+When authoring Smithy models, the same members often need to be duplicated
+across multiple related structures. For example, the identifiers of a resource
+need to be present in each input structure of resource operations:
+
+```
+namespace smithy.example
+
+resource City {
+    identifiers: { cityId: CityId }
+    read: GetCity
+    delete: DeleteCity
+    update: UpdateCity
+}
+
+structure GetCityInput {
+    @required
+    @httpLabel
+    cityId: CityId
+}
+
+structure DeleteCityInput {
+    @required
+    @httpLabel
+    cityId: CityId
+}
+
+structure UpdateCityInput {
+    @required
+    @httpLabel
+    cityId: CityId
+}
+```
+
+Repetition is also required to define paginated operations in a service.
+
+```
+structure ListCitiesInput {
+    nextToken: String
+    pageSize: Integer
+}
+
+structure ListCitiesOutput {
+    nextToken: String
+
+    @required
+    items: CitySummaries
+}
+
+structure ListWeatherStationsInput {
+    nextToken: String
+    pageSize: Integer
+}
+
+structure ListWeatherStationsOutput {
+    nextToken: String
+
+    @required
+    items: WeatherStationSummaries,
+}
+```
+
+This repetition is tedious and error-prone as it can lead to unintentional
+inconsistencies due to copy/paste errors and drift between related shapes
+over time.
+
+
+## Proposal
+
+This proposal introduces *mixins* to reduce the amount of repetition in
+structures and unions, reduce copy/paste errors, and provide the ability to
+define reusable partial structures and unions.
+
+
+### Goals
+
+1. Improve maintainability of large models by reducing repetition and
+   copy/paste errors of related concepts.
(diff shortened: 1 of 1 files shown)


designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+
+```
+structure C {
+    a: String,
+    b: String,
+    c: String
+}
+```
+
+Unions can be mixins and use mixins too.
```
JordonPhillips, line 167: Can union mixins be used with structures and vice versa? e.g.

```
@mixin 
union UnionMixin {
    a: String
}

structure Struct with UnionMixin {
    b: String
}
```

Pretty sure the answer is no, but it should be said explicitly.

designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+
+The `@mixin` trait has a single optional member named `localTraits` that
+contains a list of shape IDs. Each shape ID MUST reference a valid trait that
+is applied directly to the mixin. The traits referenced in `localTraits` are
+not copied onto shapes that use the mixin. `localTraits` only affects traits
+applied to the mixin container shape and has no impact on the members
+contained within a mixin.
+
+> Note: `smithy.api#mixin` is considered implicitly present in the
+> `localTraits` property and does not  need to be defined in the list of
```
JordonPhillips, line 770:
```suggestion
> `localTraits` property and does not need to be defined in the list of
```

JordonPhillips (approved):

designs/mixins.md
mtdowling: No. There's a callout for this that states:
> Structure shapes can only use structure mixins, and union shapes can only use union mixins.

designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+    input: InputMixin // <- this is invalid
+}
+```
+
+
+### Mixins in the IDL
+
+To support mixins, the `structure_statement` and `union_statement` ABNF rules
+will be updated to contain an optional `with` clause that comes after the
+shape name and before `{`. If present, the `with` clause MUST contain one or
```
joguSD, line 585: How do `with` list members scale? Say I had a dozen or so mixins used on a particular shape, would they all be required to be defined in a single line? Or is there a way to breakup the definition like:

```
structure MyStruct with (
    MyMixinA,
    MyMixinB,
    MyMixinC,
...
) {
    ...
}
```

joguSD (approved):

designs/mixins.md
mtdowling: In the current proposal, whitespace and commas are insignificant, you can break them up onto different lines, and there's no need for a `(` or `)`:

```
structure MyStruct with
    MyMixinA
    MyMixinB
    MyMixinC
    MyMixinD
{
    @required
    foo: String
}
```

joguSD: Cool, makes sense.

designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+structure B with A {
+    /// B docs
+    @tags(["hello"])
+    @deprecated
+    b: String
+}
+
+structure ABC with A, B {
+    /// More specific B docs
+    @override
```
johnstonskj, line 369: So the override here applies to the usage of the trait?

johnstonskj, line 369: In this example we alter an existing trait, I can see how one would add one, but is it possible to remove a trait?

designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+
+### Mixins and member ordering
+
+The ordering of structure and union members is important for languages like C
+that require a stable ABI. Mixins provide a deterministic member ordering.
+Members are ordered by:
+
+- Members copied from mixins come before members defined directly in the shape.
+- Members are copied from mixins based on the order in which they applied to
+  a shape.
```
johnstonskj, line 706: Is it the order in which they were _first_, or _last_ applied to the shape?

designs/mixins.md
```diff
@@ -0,0 +1,1075 @@
...
+
+```
+@trait(selector: "is(structure, union) > member")
+structure override {}
+```
+
+
+### Reference implementation notes
+
+Mixins will be implemented in Smithy's Java reference implementation in a way
```
johnstonskj, line 786: Does this imply that other implementations could choose a different behavior, or is the reference implementation the preferred approach?

johnstonskj:
Will this imply a new version number? it seems like we introduced a compatibility issue when removing commas without a version bump. 

As for the BNF, is this a reasonable approximation?

```ebnf
structure_statement =
    "structure" ws identifier ws [mixins ws] shape_members

mixins =
    "with" ws root_shape_id *(ws comma root_shape_id) trailing_comma
```

mtdowling:
@johnstonskj the removal of commas was never released. It's only in the idl-1.1 branch. This proposal would be built on top of that branch and eventually released with a new IDL version of 1.1. The ABNF is defined in the proposal as:

```
structure_statement = "structure" ws `identifier` ws [with_clause] shape_members
union_statement = "union" ws `identifier` ws [with_clause] shape_members
with_clause = "with" ws 1*shape_id ws
```

designs/mixins.md
mtdowling: First. I'll clarify.

designs/mixins.md
mtdowling: This section is non-normative, and it's only specific to the reference implementation. Other implementations are free to implement however necessary. This document is not the specification, it's a design document. This is new to the repo, and it's the first one we've published. Once we have agreement here, we'll create a PR to add this to the specification (but also, we're not shipping this until we've finalized 1.1 of the IDL).

designs/mixins.md
mtdowling: There is not. We could allow traits to be filtered per member by extending the `@override` trait. However, I think that needing to _remove_ a trait from a specific member implies a case where the mixin is not appropriate, and the site of the mixin isn't actually opting into the mixin. Adding this kind of filtering per member, I think, would make the models harder to understand (plus harder to implement in Smithy implementations).

An alternative to removing traits but still leveraging mixins is to either not use a mixin for the one-off cases, or to make two layers of mixins. For example, say that some applications of a mixin want a member to be required, and others to be optional. In that case, make two mixins:

```
@mixin
structure FooMixinOptional {
    someMember: String
    otherMember: String
}

@mixin
structure FooMixinRequired with FooMixin {
    @override
    @required
    someMember: String
}
```

johnstonskj:
Awesome, thanks!

designs/mixins.md
mtdowling: Mulled this over with others, and I think we should start with a predefined ordering of members based on what we already had -- mixin members come before members defined locally, mixin members are ordered based on when they are first seen, and redefining members has no effect on ordering. If this proves insufficient for ordering, then we can later consider adding traits for explicit ordering.

kstich (approved):

PR #842, opened 2021-06-23
Tighten substitution pattern for Fn::Sub

gosar:
*Description of changes:*
CloudFormation variables can be parameter names, resource logical IDs
or resource attributes.

* Parameter names: can be alphanumeric or pseudo-parameters, which can
  be prefixed with "AWS::"
* Resource logical IDs: can be alphanumeric.
* Resource attributes: specified as resourceLogicalId.attributeName,
  where attributeName is also alphanumeric.

*Issue #, if available:*
This change was needed as part of https://github.com/adamthom-amzn/smithy-typescript-ssdk-demo/pull/28

### DIFF
--- a/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/CloudFormationSubstitution.java
+++ b/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/CloudFormationSubstitution.java
@@ -45,7 +45,7 @@ final class CloudFormationSubstitution implements ApiGatewayMapper {
 
     private static final Logger LOGGER = Logger.getLogger(CloudFormationSubstitution.class.getName());
     private static final String SUBSTITUTION_KEY = "Fn::Sub";
-    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{.+}");
+    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{[A-Za-z0-9.:]+}");
 
     /**
      * This is a hardcoded list of paths that are known to contain ARNs or identifiers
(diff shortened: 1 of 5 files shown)


smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/CloudFormationSubstitution.java
```diff
@@ -45,7 +45,7 @@
 
     private static final Logger LOGGER = Logger.getLogger(CloudFormationSubstitution.class.getName());
     private static final String SUBSTITUTION_KEY = "Fn::Sub";
-    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{.+}");
+    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{[A-Za-z0-9.:]+}");
```
JordonPhillips, line 48: I think this is too restrictive. `[\w:]+` would be a little better, though maybe we want hyphens too. Generally I'm just a little nervous about doing what is a potentially breaking change. Could you just specifically exclude that token pattern? For instance, [`\$\{(?!Token\[[^\[\]]*\])[^\{\}]+\}`](regexr.com/60l7a). (You'd need to add `Pattern.CASE_INSENSITIVE` to prevent it from being even more grim.)

gosar: I've based it on the rules in CFN.
Parameter names: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-requirements - _Each parameter must be given a logical name (also called logical ID), which must be alphanumeric_
Pseudo parameters: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html
Resource logical IDs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html - _The logical ID must be alphanumeric (A-Za-z0-9)_
Resource attributes: Couldn't find it in docs, but I believe they can be only alphanumeric too.

Agree the code change is going from looser to restrictive which is technically backwards incompatible, but given the above, anything else this code allowed earlier, would still have have worked end-to-end.

JordonPhillips: Cool, so other values never would have worked. Thanks for the context!

JordonPhillips (approved):

PR #843, opened 2021-06-23
Adds httpChecksum trait support

skotambkar:
Status: Needs review

Description of changes:
[x] Adds httpChecksum trait to smithy.
[x] Adds test cases and docs

Replaces https://github.com/awslabs/smithy/pull/741

### DIFF
--- a/docs/source/1.0/spec/core/behavior-traits.rst
+++ b/docs/source/1.0/spec/core/behavior-traits.rst
@@ -655,3 +655,223 @@ See
             input: PutSomethingInput,
             output: PutSomethingOutput
         }
+
+
+.. _httpChecksum-trait:
+
+----------------------
+``httpChecksum`` trait
+----------------------
+
+Summary
+    Indicates that an operation's HTTP request or response supports checksum
+    validation. At least one of request or response checksum properties must
+    be specified within the trait.
+
+Trait selector
+    ``operation``
+Value type
+    ``structure``
+
+The ``httpChecksum`` trait is a structure that contains the following members:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 10 10 80
+
+    * - Property
+      - Type
+      - Description
+    * - request
+      - ``[`` :ref:`HttpChecksumProperty structure <checksum-property>` ``]``
+      - The ``request`` property defines a list of checksum validation
+        behavior for HTTP requests.
+
+    * - response
+      - ``[`` :ref:`HttpChecksumProperty structure <checksum-property>` ``]``
+      - The ``response`` property defines a list of checksum validation
+        behavior for HTTP responses.
+
+.. tabs::
+
+    .. code-tab:: smithy
+
+        @httpChecksum(
+            request: [
+                { algorithm: "sha256", in: "header", name: "x-checksum-sha256"},
+                { algorithm: "crc32", in: "header", name: "x-checksum-crc32"}
+            ],
+            response: {
+                { algorithm: "sha256", in: "header", name: "x-checksum-sha256"},
+                { algorithm: "crc32", in: "header", name: "x-checksum-crc32"}
+            }
+        )
+        operation PutSomething {
+            input: PutSomethingInput,
+            output: PutSomethingOutput
+        }
+
+    .. code-tab:: json
+
+        {
+            "smithy": "1.0",
+            "shapes": {
+                "smithy.example#Example": {
+                    "type": "service",
+                    "version": "2019-06-27",
+                },
+                "smithy.example#PutSomething": {
+                    "type": "operation",
+                    "input": {
+                        "target": "smithy.example#PutSomethingInput"
+                    },
+                    "output": {
+                        "target": "smithy.example#PutSomethingOutput"
+                    },
+                    "traits": {
+                        "smithy.api#httpChecksum": {
+                            "request": [
+                                { "algorithm": "sha256", "in": "header", "name": "x-checksum-sha256"},
+                                { "algorithm": "crc32", "in": "header", "name": "x-checksum-crc32"}
+                            ],
+                            "response": [
+                                { "algorithm": "sha256", "in": "header", "name": "x-checksum-sha256"},
+                                { "algorithm": "crc32", "in": "header", "name": "x-checksum-crc32"}
+                            ]
+                        }
+                    }
+                }
+            }
+        }
+
+
+.. _checksum-property:
+
+HttpChecksumProperty structure
+================================
+
+``HttpChecksumProperty`` defines checksum validation behavior using the
+following members:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 10 10 80
+
+    * - Property
+      - Type
+      - Description
+
+    * - algorithm
+      - ``string``
+      - **Required**. The algorithm string represents a checksum algorithm
+        supported for the HTTP request or response. A valid algorithm
+        should follow ABNF:
+
+        .. code-block:: abnf
+
+            algorithms = 1*(lower-alpha / DIGIT)
+            lower-alpha = %x61-7A
+
+        That is, it should be comprised only of digits and lowercase letters.
+
+    * - in
+      - ``string``
+      - **Required**. The location where checksum can be supplied. Valid values
+        are:
+
+        * ``header`` - Indicates the checksum should be placed in an HTTP header.
+        * ``trailer`` - Indicates the checksum should be placed in the `chunked trailer part`_
+          of the body.
+
+        An algorithm may be supported at multiple locations. The first
+        supported ``in`` member value within the list of HttpChecksum property
+        takes precedence and SHOULD be used. See :ref:`here for more behavior
+        details <httpChecksum_trait_behavior>`.
+
+    * - name
+      - ``string``
+      - **Required**. The name used as a header or trailer name for a checksum
+        type. Recommended ABNF for name is as follows:
+
+        .. code-block:: abnf
+
+            name = lower-alpha *(["-"](lower-alpha / DIGIT))
+            lower-alpha = %x61-7A
+
+        That is, it should start with a lowercase letter followed by any
+        number of lowercase letters, digits, or non-sequential hyphens.
+
+        A member with the :ref:`httpHeader-trait` or :ref:`httpPrefixHeaders-trait`
+        MAY conflict with a ``httpChecksum`` header name, allowing a
+        checksum to be supplied directly. See :ref:`here for more behavior
+        details <httpChecksum_trait_header_conflict_behavior>`.
+
+.. _chunked trailer part: https://tools.ietf.org/html/rfc7230#section-4.1.2
+
+
+.. _httpChecksum_trait_behavior:
+
+HttpChecksum Behavior
+=====================
+
+.. rubric:: Client Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP
+request, the client MUST compute the request payload checksum, using an
+algorithm defined within the request section. The computed checksum SHOULD
+be supplied at the first supported checksum location as per the order of
+:ref:`checksum properties<checksum-property>` defined within the request
+section.
+
+If the resolved location is ``header``, client MUST put the checksum into the
+HTTP request headers. If the resolved location is ``trailer``, the client MUST
+put the checksum into the `chunked trailer part`_ of the body. The header or
+trailer name to use with an algorithm, location is defined by the ``name``
+member of the :ref:`checksum property<checksum-property>`.
+
+If the ``httpChecksum`` trait has a modeled ``response`` section, for HTTP
+response, the client MUST look for a supported checksum at all supported
+locations as per the defined checksum properties. If a checksum is found, the
+client MUST validate the received checksum value by computing the corresponding
+checksum of the received payload.
+
+.. _httpChecksum_trait_header_conflict_behavior:
+
+An operation's ``input``, ``output``, and ``error`` structures MAY contain a
+member bound to an HTTP header matching the header or trailer name defined
+within the ``httpChecksum`` trait. For an HTTP request, if customer provides value
+for an input shape member bound to an HTTP header matching the constructed
+header or trailer name, the client MUST use the customer provided value as is,
+and skip computing request payload checksum.
+
+.. _chunked trailer part: https://tools.ietf.org/html/rfc7230#section-4.1.2
+
+
+.. rubric:: Service Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP request,
+the service SHOULD validate the received checksum by computing corresponding
+checksum of the request payload to ensure data integrity.
+
+If the ``httpChecksum`` trait models a ``response`` section, for HTTP response,
+the service SHOULD send at least one supported payload checksum at the first
+supported location for the checksum algorithm. The ``name`` defined for the
+checksum algorithm and location within ``the httpChecksum`` trait MUST be used
+to place the checksum.
+
+
+.. _checksum_trait_with_checksum_required:
+
+Behavior with HttpChecksumRequired
+----------------------------------
+
+When both the ``httpChecksum`` trait with a modeled ``request`` section, and
+the :ref:`httpChecksumRequired <httpChecksumRequired-trait>` trait are applied
+on an operation, the client MUST prefer using a checksum algorithm modeled for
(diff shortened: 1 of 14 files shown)


docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+
+.. _httpChecksum-trait:
+
+----------------------
+``httpChecksum`` trait
+----------------------
+
+Summary
+    Indicates that an operation's HTTP request or response supports checksum
+    validation. At least one of request or response checksum properties must
```
mtdowling, line 668:
```suggestion
    validation. At least one of request or response checksum properties MUST
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+            lower-alpha = %x61-7A
+
+        That is, it should be comprised only of digits and lowercase letters.
+
+    * - in
+      - ``string``
+      - **Required**. The location where checksum can be supplied. Valid values
+        are:
+
+        * ``header`` - Indicates the checksum should be placed in an HTTP header.
```
mtdowling, line 782:
```suggestion
        * ``header`` - Indicates the checksum is placed in an HTTP header.
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+        * ``trailer`` - Indicates the checksum should be placed in the `chunked trailer part`_
```
mtdowling, line 783:
```suggestion
        * ``trailer`` - Indicates the checksum is be placed in the `chunked trailer part`_
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+          of the body.
+
+        An algorithm may be supported at multiple locations. The first
+        supported ``in`` member value within the list of HttpChecksum property
+        takes precedence and SHOULD be used. See :ref:`here for more behavior
```
mtdowling, line 788: For links like this, I usually give more descriptive details about what's being referenced. You could even just do:

```
.. seealso::

    See :ref:`httpChecksum_trait_behavior` for more details.
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+      - ``string``
+      - **Required**. The name used as a header or trailer name for a checksum
+        type. Recommended ABNF for name is as follows:
+
+        .. code-block:: abnf
+
+            name = lower-alpha *(["-"](lower-alpha / DIGIT))
+            lower-alpha = %x61-7A
+
+        That is, it should start with a lowercase letter followed by any
```
mtdowling, line 801: Can you give some context on why it's much more restrictive than the field-name production in [RFC 7230, section 3](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2)? I'm assuming it's because headers are case-insensitive, and we want to normalize it here?

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+        number of lowercase letters, digits, or non-sequential hyphens.
+
+        A member with the :ref:`httpHeader-trait` or :ref:`httpPrefixHeaders-trait`
+        MAY conflict with a ``httpChecksum`` header name, allowing a
+        checksum to be supplied directly. See :ref:`here for more behavior
```
mtdowling, line 806: Same comment above -- better to include the title and use `seealso` admonitions.

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+
+
+.. _httpChecksum_trait_behavior:
+
+HttpChecksum Behavior
+=====================
+
+.. rubric:: Client Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP
```
mtdowling, line 819:
```suggestion
If the ``httpChecksum`` trait has a modeled ``request`` section for the HTTP
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+    * - in
+      - ``string``
+      - **Required**. The location where checksum can be supplied. Valid values
+        are:
+
+        * ``header`` - Indicates the checksum should be placed in an HTTP header.
+        * ``trailer`` - Indicates the checksum should be placed in the `chunked trailer part`_
+          of the body.
+
+        An algorithm may be supported at multiple locations. The first
```
mtdowling, line 786:
```suggestion
        An algorithm MAY be supported at multiple locations. The first
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+
+.. _httpChecksum_trait_behavior:
+
+HttpChecksum Behavior
+=====================
+
+.. rubric:: Client Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP
+request, the client MUST compute the request payload checksum, using an
```
mtdowling, line 820: I thought it was only if there's an `httpChecksumRequired` trait?

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+algorithm defined within the request section. The computed checksum SHOULD
+be supplied at the first supported checksum location as per the order of
+:ref:`checksum properties<checksum-property>` defined within the request
+section.
+
+If the resolved location is ``header``, client MUST put the checksum into the
```
mtdowling, line 826:
```suggestion
If the location is ``header``, the client MUST put the checksum into the
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+HTTP request headers. If the resolved location is ``trailer``, the client MUST
+put the checksum into the `chunked trailer part`_ of the body. The header or
+trailer name to use with an algorithm, location is defined by the ``name``
```
mtdowling, line 829:
```suggestion
trailer name to use with an algorithm location is defined by the ``name``
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+member of the :ref:`checksum property<checksum-property>`.
+
+If the ``httpChecksum`` trait has a modeled ``response`` section, for HTTP
```
mtdowling, line 832:
```suggestion
If the ``httpChecksum`` trait has a modeled ``response`` section for the HTTP
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+response, the client MUST look for a supported checksum at all supported
+locations as per the defined checksum properties. If a checksum is found, the
+client MUST validate the received checksum value by computing the corresponding
+checksum of the received payload.
+
+.. _httpChecksum_trait_header_conflict_behavior:
+
+An operation's ``input``, ``output``, and ``error`` structures MAY contain a
+member bound to an HTTP header matching the header or trailer name defined
+within the ``httpChecksum`` trait. For an HTTP request, if customer provides value
```
mtdowling, line 842: Can you frame this so it's not about customers, clients, requests, or responses, but rather just matter of fact: If the HTTP header corresponding to a checksum is set explicitly, then the implementation MUST use the explicitly set header and skip computing a payload checksum.

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+header or trailer name, the client MUST use the customer provided value as is,
+and skip computing request payload checksum.
+
+.. _chunked trailer part: https://tools.ietf.org/html/rfc7230#section-4.1.2
+
+
+.. rubric:: Service Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP request,
+the service SHOULD validate the received checksum by computing corresponding
```
mtdowling, line 853: MUST?

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+for an input shape member bound to an HTTP header matching the constructed
+header or trailer name, the client MUST use the customer provided value as is,
+and skip computing request payload checksum.
+
+.. _chunked trailer part: https://tools.ietf.org/html/rfc7230#section-4.1.2
+
+
+.. rubric:: Service Behavior
+
+If the ``httpChecksum`` trait has a modeled ``request`` section, for HTTP request,
```
mtdowling, line 852:
```suggestion
If the ``httpChecksum`` trait has a modeled ``request`` section for the HTTP request,
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+the service SHOULD validate the received checksum by computing corresponding
+checksum of the request payload to ensure data integrity.
+
+If the ``httpChecksum`` trait models a ``response`` section, for HTTP response,
```
mtdowling, line 856:
```suggestion
If the ``httpChecksum`` trait models a ``response`` section for the HTTP response,
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+the service SHOULD send at least one supported payload checksum at the first
+supported location for the checksum algorithm. The ``name`` defined for the
```
mtdowling, line 858: Is the last sentence needed here? Seems like it was already defined earlier in the trait.

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+checksum algorithm and location within ``the httpChecksum`` trait MUST be used
+to place the checksum.
+
+
+.. _checksum_trait_with_checksum_required:
+
+Behavior with HttpChecksumRequired
+----------------------------------
+
+When both the ``httpChecksum`` trait with a modeled ``request`` section, and
```
mtdowling, line 868: Earlier in the doc, I think it's stated that the checksum is always required. I also don't think `httpChecksum` requires that `httpChecksum` to be present, which this seems to imply.

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+the :ref:`httpChecksumRequired <httpChecksumRequired-trait>` trait are applied
+on an operation, the client MUST prefer using a checksum algorithm modeled for
+request in the ``httpChecksum`` trait over an MD5 digest, and place checksum
+as per the client behavior defined in the
+:ref:`httpChecksum trait behavior<httpChecksum_trait_behavior>` section.
+
+The service MUST accept a checksum value received as per the ``httpChecksum``
```
mtdowling, line 875:
```suggestion
The service MUST require that a checksum value is received as defined in the ``httpChecksum``
```

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+trait's request property, to satisfy checksum validation requirements for the
```
mtdowling, line 876:
```suggestion
trait's request property to satisfy checksum validation requirements for the
```

mtdowling (changes requested):

docs/source/1.0/spec/core/behavior-traits.rst
skotambkar: Service may use md5 digest to satisfy data integrity requirements. Operations modeled with `httpChecksumRequired` default to sending md5 digest. 

The intent here is to have services support checksum values defined within `httpChecksum` trait to satisfy data integrity requirements, and not rely solely on md5.

docs/source/1.0/spec/core/behavior-traits.rst
skotambkar: Correct, this is to normalize the name and stay consistent.

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -655,3 +655,223 @@ See
...
+
+.. list-table::
+    :header-rows: 1
+    :widths: 10 10 80
+
+    * - Property
+      - Type
+      - Description
+
+    * - algorithm
```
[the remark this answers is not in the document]
skotambkar: I defined it to normalize the algorithm, and make it a little easier for clients/service to correctly map supported algorithm implementations.

docs/source/1.0/spec/core/behavior-traits.rst
skotambkar: Thanks for catching this. Updated to `SHOULD`.

docs/source/1.0/spec/core/behavior-traits.rst
skotambkar: Corrected this as stated in earlier comment.  Checksum is only required when `htttpChecksumRequired` trait is present.

kstich (changes requested):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/HttpChecksumTraitValidator.java
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * isTargetProtocol returns true if service uses a target protocol. By default,
+     * target protocol resolves to aws protocol.
+     *
+     * @param service is the service shape for which target protocol usage is checked.
+     * @return boolean indicating target protocol is used by the service.
+     */
+    protected boolean isTargetProtocol(ServiceShape service) {
+        // By default, target protocol is AWS protocol.
+        return service.hasTrait(AwsProtocolTrait.class);
```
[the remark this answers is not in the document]
skotambkar: Would need to explore `Supplier`. 

We don't have a use case for `awsQuery` or `ec2Query` yet, can we add this support when required?

smithy-model/src/main/java/software/amazon/smithy/model/traits/HttpChecksumProperty.java
```diff
@@ -0,0 +1,236 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        } else if (!(o instanceof HttpChecksumProperty)) {
+            return false;
+        }
+
+        HttpChecksumProperty that = (HttpChecksumProperty) o;
+        return algorithm.equals(that.algorithm)
+                && location.equals(that.location);
```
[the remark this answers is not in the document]
skotambkar: No, this is intended.  We consider checksum property equal when they are modeled for the same checksum algorithm and location. 

This is because modeling same checksum algorithm and location is not valid.  

I use equals to raise an error when we encounter multiple values with same checksum algorithm and location.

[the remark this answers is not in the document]
skotambkar: I raise an error if duplicate entries are found while adding in request list [1] or response list [2] 

[1] https://github.com/awslabs/smithy/pull/843/files#diff-c1d6f5ffdbf17d22712737361d41edb190675537fe90920e3844f1d7afeae8bfR181

[2] https://github.com/awslabs/smithy/pull/843/files#diff-c1d6f5ffdbf17d22712737361d41edb190675537fe90920e3844f1d7afeae8bfR203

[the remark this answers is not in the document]
kstich: `.equals` should be checking for true equality. You can always add a separate method that detects if there's a conflict, we do [this with `UriPattern`](https://github.com/awslabs/smithy/blob/076547fea82e04491c490b1e0675a6340900ae6f/smithy-model/src/main/java/software/amazon/smithy/model/pattern/UriPattern.java#L128).

kstich (approved):

JordonPhillips (approved):

PR #845, opened 2021-06-25
Add a protocol test for endpoints that contain path prefixes

adamthom-amzn:
*Description of changes:*
Sometimes, a service will have a path prefix that is not modeled. This adds a
test that ensures REST-JSON clients will properly take that path into account.

### DIFF
--- a/docs/source/1.0/spec/http-protocol-compliance-tests.rst
+++ b/docs/source/1.0/spec/http-protocol-compliance-tests.rst
@@ -105,8 +105,12 @@ that support the following members:
         the query string (for example, "/foo/bar").
     * - host
       - ``string``
-      - The host / endpoint provided to the client, not including the path or
-        scheme (for example, "example.com").
+      - The host / endpoint provided to the client (for example, "example.com").
+        ``host`` MAY contain a path to indicate a base path from which each
+        operation in the service is appended to. For example, given a ``host``
+        of ``example.com/foo/bar`` and an operation path of ``/MyOperation``,
+        the resolved host of the operation is ``example.com`` and the resolved
+        path is ``/foo/bar/MyOperation``.
     * - resolvedHost
       - ``string``
       - The host / endpoint that the client should send to, not including the
--- a/smithy-aws-protocol-tests/model/restJson1/endpoint-paths.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/endpoint-paths.smithy
@@ -1,0 +1,24 @@
+// This file defines tests to ensure that implementations support endpoints with paths
+
+$version: "1.0"
+
+namespace aws.protocoltests.restjson
+
+use aws.protocols#restJson1
+use smithy.test#httpRequestTests
+
+@httpRequestTests([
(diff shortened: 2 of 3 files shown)


smithy-aws-protocol-tests/model/restJson1/main.smithy
```diff
@@ -86,5 +86,8 @@ service RestJson {
         // @endpoint and @hostLabel trait tests
         EndpointOperation,
         EndpointWithHostLabelOperation,
+
+        // custom endpoints with paths
+        PathOperation,
```
gosar, line 91: `PathOperation` reads like an operation with path parameters. Maybe EndpointWithPathOperation? Or if that implies use of @endpoint trait like the couple of tests above, maybe HostWithPathOperation?

gosar (approved):

gosar (approved):

smithy-aws-protocol-tests/model/restJson1/endpoint-paths.smithy
```diff
@@ -0,0 +1,24 @@
...
+use smithy.test#httpRequestTests
+
+@httpRequestTests([
+    {
+        id: "RestJsonHostWithPath",
+        documentation: """
+                Custom endpoints supplied by users can have paths""",
+        protocol: restJson1,
+        method: "GET",
+        uri: "/custom/PathOperation",
```
mtdowling, line 17:
```suggestion
        uri: "/custom/HostWithPathOperation",
```

mtdowling (changes requested):

docs/source/1.0/spec/http-protocol-compliance-tests.rst
```diff
@@ -105,8 +105,9 @@ that support the following members:
         the query string (for example, "/foo/bar").
     * - host
       - ``string``
-      - The host / endpoint provided to the client, not including the path or
-        scheme (for example, "example.com").
+      - The host / endpoint provided to the client (for example, "example.com").
+        This will generally not include a path unless it is a non-modeled prefix,
```
mtdowling, line 109: Can you flip this sentence around and say when it would a path, an example of what it looks like, and how it works with paths in operations? It's important to state that here since we aren't using the same URI resolution logic defined in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-5).

For example:

> `host` MAY contain a path to indicate a base path from which each operation in the service is appended to. For example, given a `host` of `example.com/foo/bar` and an operation path of `/MyOperation`, the resolved host of the operation is `example.com` and the resolved path is `/foo/bar/MyOperation`.

mtdowling (changes requested):

smithy-aws-protocol-tests/model/restJson1/endpoint-paths.smithy
adamthom-amzn: Yeah, that was a bad miss.

mtdowling (approved):

PR #857, opened 2021-07-07
Adjust query string parameter tests for servers

adamthom-amzn:
*Description of changes:*
On the server side, `@httpQueryParams` holds all of the query parameter values
in the request, so if any of the query parameters has a list value, then
`@httpQueryParams` must target a map with a list value, otherwise a 400 will be
thrown during deserialization, instead of discarding excess query param values.

An alternative solution could be to adjust SSDK behavior where `@httpQueryParams` would only receive query parameter values that were not otherwise explicitly bound to a structure member.

Prior art in this area is for the service to truncate to the last specified query parameter value when binding all query parameters to a map of strings, and since that's not at all a desirable behavior, and this doesn't change what gets put on the wire, we have considerable leeway to choose the behavior we want.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-query.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-query.smithy
@@ -105,9 +105,9 @@ apply AllQueryStringTypes @httpRequestTests([
             "QueryParamsStringKeyB=Bar",
         ],
         params: {
-            queryParamsMapOfStrings: {
-                "QueryParamsStringKeyA": "Foo",
-                "QueryParamsStringKeyB": "Bar",
+            queryParamsMapOfStringList: {
+                "QueryParamsStringKeyA": ["Foo"],
+                "QueryParamsStringKeyB": ["Bar"],
             },
         }
     },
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restJson1/http-query.smithy
```diff
@@ -105,9 +105,9 @@ apply AllQueryStringTypes @httpRequestTests([
             "QueryParamsStringKeyB=Bar",
         ],
         params: {
-            queryParamsMapOfStrings: {
-                "QueryParamsStringKeyA": "Foo",
-                "QueryParamsStringKeyB": "Bar",
+            queryParamsMapOfStringList: {
+                "QueryParamsStringKeyA": ["Foo"],
+                "QueryParamsStringKeyB": ["Bar"],
```
JordonPhillips, line 110: I agree that we should have this test, but why do we need to get rid of the old one?

adamthom-amzn: If `@httpQueryParams` is on a member, it has to be on a map of lists if any other `@httpQuery` member can have multiple values. So this isn't really about removing or adding tests, it's about making this test compatible with servers.

JordonPhillips (approved):

PR #868, opened 2021-07-23
Update headers for sigv4 and restJson1 for apig

JordonPhillips:
This updates the lists of headers needed for sigv4 and for the restJson1 protocol in the openapi conversion. It also ensures that those headers are added to cors configurations for apig.

A few notes:

* I'm only 95% sure that `X-Amzn-Trace-Id` isn't present in responses.
* I didn't include `X-Amz-Retry-After` since only one SDK uses it, and it is mostly obsoleted by adaptive retries.
* I didn't include any service-specific headers, such as `X-Amz-Crc32`. These can be configured if the service actually makes use of them.


smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/ApiGatewayConfig.java
```diff
@@ -55,7 +55,7 @@
 
     private ApiType apiGatewayType = ApiType.REST;
     private boolean disableCloudFormationSubstitution;
-    private List<String> additionalAllowedCorsHeaders = ListUtils.of("amz-sdk-invocation-id", "amz-sdk-request");
```
kstich, line 58: Would these be deduplicated by a `set` or do they _need_ to be removed?

kstich (changes requested):

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/ApiGatewayConfig.java
JordonPhillips: That is the wrong place to have them, which is a good enough reason for me. They wouldn't be de-duped because they're lower case and the others are canonicalized.

kstich (approved):

PR #873, opened 2021-07-29
Flag enum insertions as errors

ianbotsf:
*Issue #, if available:*

(none)

*Description of changes:*

When calculating Smithy diffs, enum value insertions should be flagged as errors because they can cause backwards incompatibility concerns when enum ordinals are used for iterating, serialization, etc. Enum values appended to the end of the existing set of values should not be errors.

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
@@ -29,8 +29,9 @@ import software.amazon.smithy.utils.OptionalUtils;
 import software.amazon.smithy.utils.Pair;
 
 /**
- * Emits a NOTE when a new enum value is added, emits an ERROR when an
- * enum value is removed, and emits an ERROR when an enum name changes.
+ * Emits a NOTE when a new enum value is appended, emits an ERROR when an
+ * enum value is removed, emits an ERROR when an enum name changes, and
+ * emits an ERROR when a new enum value is inserted.
  */
 public final class ChangedEnumTrait extends AbstractDiffEvaluator {
     @Override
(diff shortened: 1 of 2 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
```diff
@@ -29,8 +29,9 @@
 import software.amazon.smithy.utils.Pair;
 
 /**
- * Emits a NOTE when a new enum value is added, emits an ERROR when an
- * enum value is removed, and emits an ERROR when an enum name changes.
+ * Emits a NOTE when a new enum value is appended, emits an ERROR when an
+ * enum value is removed, emits an ERROR when an enum name changes, and
+ * emits an ERROR when a new enum value is inserted.
```
mtdowling, line 34: Throughout this PR, can we be more specific about what inserted means? We mean that an enum value was added but not at the end of the list of enum values. I think clarifying with this little tweak will save countless hours of teams not understanding the error.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
```diff
@@ -67,11 +70,20 @@
...
         }
 
+        int newPosition = 0;
         for (EnumDefinition definition : newTrait.getValues()) {
             if (!oldTrait.getEnumDefinitionValues().contains(definition.getValue())) {
-                events.add(note(change.getNewShape(), String.format(
-                        "Enum value `%s` was added", definition.getValue())));
+                if (newPosition <= oldEndPosition) {
+                    events.add(error(change.getNewShape(), String.format(
+                            "Enum value `%s` was inserted. This can cause compatibility issues when ordinal values are "
```
mtdowling, line 78: I love the extra bit of context here about why this matters! Can you say something like "Enum value `%s` was inserted, but not at the end of the list of existing enum values. [...]" or something to that effect?

mtdowling (changes requested):

mtdowling (approved):

PR #875, opened 2021-08-03
Make sets ordered

JordonPhillips:
This deprecates the `uniqueItems` trait in favor of having ordered sets. Note: this PR is currently against the `idl-1.1` branch.

### DIFF
--- a/docs/source/1.0/spec/core/model.rst
+++ b/docs/source/1.0/spec/core/model.rst
@@ -458,7 +458,7 @@ reference other shapes using :ref:`members <member>`.
     * - :ref:`list`
       - Ordered collection of homogeneous values
     * - :ref:`set`
-      - Unordered collection of unique homogeneous values
+      - Collection of unique homogeneous values
     * - :ref:`map`
       - Map data structure that maps string keys to homogeneous values
     * - :ref:`structure`
@@ -565,7 +565,7 @@ example is ``smithy.example#MyList$member``.
 Set
 ===
 
-The :dfn:`set` type represents an unordered collection of unique homogeneous
+The :dfn:`set` type represents a collection of unique homogeneous
 values. A set shape requires a single member named ``member``.
 Sets are defined in the IDL using a :ref:`set_statement <idl-set>`.
 The following example defines a set of strings:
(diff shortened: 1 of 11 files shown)


smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -492,6 +491,7 @@ structure sparse {}
 
 /// Indicates that the items in a list MUST be unique.
 @trait(selector: "list")
+@deprecated(message: "The uniqueItems trait has been deprecated in favor of using sets.", since: "2.0")
```
kstich, line 494: Should deprecate the trait implementation class as well.

kstich (approved):

docs/source/1.0/spec/core/model.rst
```diff
@@ -618,6 +618,13 @@ SHOULD represent sets as a custom set data structure that can interpret value
 hash codes and equality, or alternatively, store the values of a set data
 structure in a list and rely on validation to ensure uniqueness.
 
+.. rubric:: Set member ordering
+
+Sets returned by generated tooling, such as the deserialized response to a
+client request, MUST be insertion ordered. Sets accepted as input by generated
+tooling, such as in a client request, SHOULD be ordered unless requiring an
```
mtdowling, line 625:
```suggestion
tooling, such as in a client request, SHOULD be insertion ordered unless requiring an
```

docs/source/1.0/spec/core/model.rst
```diff
@@ -618,6 +618,13 @@ SHOULD represent sets as a custom set data structure that can interpret value
...
+ordered set would be overly burdensome or conflict with language idioms.
```
mtdowling, line 626: Should these languages just use lists if they can't maintain order?

JordonPhillips: It depends on whether the unique property or the ordered property is more important. I mentioned this offline, but I don't think it's particularly ergonomic in any language to require an ordered set as an input.

mtdowling: I think the order would be very important if a service wants a set of unique items but order provides precedence. If in some languages it can't be relied on, then it's not guaranteed that the service will work interoperably across all of its clients, leaving the service to use a list instead, and now that uniqueItems is being removed, to have to implement validation and making the list unique manually rather than relying on a framework feature.

mtdowling (approved):

PR #881, opened 2021-08-05
Add malformed content-type / accept tests

JordonPhillips:
This adds malformed request tests for asserting content type and accept header behavior

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-accept.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-accept.smithy
@@ -1,0 +1,133 @@
+$version: "1.0"
+
+namespace aws.protocoltests.restjson
+
+use aws.protocols#restJson1
+use smithy.test#httpMalformedRequestTests
+use aws.protocoltests.shared#GreetingStruct
+use aws.protocoltests.shared#JpegBlob
+
+apply MalformedAcceptWithoutBody @httpMalformedRequestTests([
+    {
+        id: "RestJsonWithoutBodyExpectsEmptyAccept",
+        documentation: """
+        When there is no modeled output, accept must not be set.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedAcceptWithoutBody",
+            headers: {
+                // this should be omitted
+                "accept": "application/json"
+            }
+        },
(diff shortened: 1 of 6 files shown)


smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-accept.smithy
```diff
@@ -0,0 +1,133 @@
...
+use aws.protocols#restJson1
+use smithy.test#httpMalformedRequestTests
+use aws.protocoltests.shared#GreetingStruct
+use aws.protocoltests.shared#JpegBlob
+
+apply MalformedAcceptWithoutBody @httpMalformedRequestTests([
+    {
+        id: "RestJsonWithoutBodyExpectsEmptyAccept",
+        documentation: """
+        When there is no modeled output, accept must not be set.""",
```
adamthom-amzn, line 14: Do we want to be this strict? If a client that handles a JSON protocol set `Accept: application/json` on all requests without payload or event stream binding on the response I don't think I'd be too upset with that behavior.

smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-content-type.smithy
```diff
@@ -0,0 +1,137 @@
...
+        documentation: """
+        When there is a payload with a mediaType trait, the content type must match.""",
+        protocol: restJson1,
+        request: {
+            method: "POST",
+            uri: "/MalformedContentTypeWithPayload",
+            body: "{}",
+            headers: {
+                // this should be image/jpeg
+                "content-type": "application/json"
```
adamthom-amzn, line 72: `payload` below doesn't have an indicated mediaType, shouldn't this pass?

JordonPhillips: The payload is `JpegBlob` with has a media type `image/jpeg`

smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-accept.smithy
JordonPhillips: I guess I can go either way.

adamthom-amzn (approved):

PR #890, opened 2021-08-13
Update semantic validation of OPTIONS

mtdowling:
* OPTIONS do support payloads
* OPTIONS should rarely, if ever, be modeled. They should instead just
  be implementation details of how CORS works.


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpMethodSemanticsValidator.java
```diff
@@ -45,20 +45,25 @@
...
-            "DELETE", new HttpMethodSemantics(false, true, false),
-            "PUT", new HttpMethodSemantics(false, true, true),
-            "PATCH", new HttpMethodSemantics(false, null, true));
+            "GET", new HttpMethodSemantics(true, false, false, null),
+            "HEAD", new HttpMethodSemantics(true, false, false, null),
+            "OPTIONS", new HttpMethodSemantics(true, false, true,
+                    "OPTIONS requests are typically used as part of CORS and should not be modeled explicitly. They "
+                    + "are an implementation detail that should not appear in generated client or server code. "
+                    + "Instead, tooling should use the `cors` trait on a service to automatically configure CORS on "
+                    + "clients and servers."),
```
JordonPhillips, line 61: I think this should explicitly state that the apigateway integration will add this in. I know it's crossing the streams a bit, but I think it'll be helpful.

JordonPhillips (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpMethodSemanticsValidator.java
mtdowling: Good idea. Updated.

JordonPhillips (approved):

PR #900, opened 2021-08-31
Allow replacement of simple shape types in ModelTransformer

jdisanti:
*Issue #, if available:* No issue

*Description of changes:*

The AWS Rust SDK needs to customize S3's `Size` shape to be a `LongShape` instead of an `IntegerShape` so that larger files can be successfully represented in responses (see https://github.com/awslabs/aws-sdk-rust/issues/209). This change updates `ReplaceShapes` to allow simple shape changes like this, as well as `ListShape <-> SetShape` changes.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/transform/ReplaceShapes.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/transform/ReplaceShapes.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.CollectionShape;
 import software.amazon.smithy.model.shapes.ListShape;
 import software.amazon.smithy.model.shapes.MapShape;
 import software.amazon.smithy.model.shapes.MemberShape;
@@ -84,7 +85,7 @@ final class ReplaceShapes {
             return model;
         }
 
-        assertNoShapeChangedType(model, shouldReplace);
+        assertShapeTypeChangesSound(model, shouldReplace);
         Model.Builder builder = createReplacedModelBuilder(model, shouldReplace);
 
         // If a member shape changes, then ensure that the containing shape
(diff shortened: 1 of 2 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/transform/ReplaceShapes.java
```diff
@@ -121,18 +122,51 @@ Model transform(ModelTransformer transformer, Model model) {
...
+
+    private void validateListCollectionReplacement(Model model, CollectionShape left, CollectionShape right) {
+        Shape leftTarget = model.expectShape(left.getMember().getTarget());
+        Shape rightTarget = model.expectShape(right.getMember().getTarget());
+        if (!isReplacementValid(model, leftTarget, rightTarget)) {
+            throw new IllegalStateException(String.format(
+                    "Cannot change the type of %s from %s to %s",
+                    left.getMember().getId(), rightTarget.getType(), leftTarget.getType()));
+        }
+    }
```
JordonPhillips, line 168:
```suggestion
    private boolean isReplacementValid(Model model, Shape left, Shape right) {
        if (left instanceof CollectionShape && right instanceof CollectionShape) {
            validateListCollectionReplacement(model, (CollectionShape) left, (CollectionShape) right);
            return true;
        }
        return left instanceof SimpleShape && right instanceof SimpleShape;
    }

    private void validateListCollectionReplacement(Model model, CollectionShape left, CollectionShape right) {
        Shape leftTarget = model.expectShape(left.getMember().getTarget());
        Shape rightTarget = model.expectShape(right.getMember().getTarget());
        if (!isReplacementValid(model, leftTarget, rightTarget)) {
            throw new IllegalStateException(String.format(
                    "Cannot change the type of %s from %s to %s",
                    left.getMember().getId(), rightTarget.getType(), leftTarget.getType()));
        }
    }
```

JordonPhillips (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/transform/ReplaceShapes.java
jdisanti: That's much nicer, thanks!

JordonPhillips (approved):

PR #903, opened 2021-09-08
Remove base64 padding parser assertion tests

JordonPhillips:
This removes two malformed request tests, both of which enforce
validating that a base64 encoded blob is properly padded with the
`=` character to have a length that is a multiple of 4.

Padding isn't necessary when the size of the input data in
unambiguous. Between HTTPS and content-length validation, we can be
pretty certain we've got everything.

https://www.rfc-editor.org/rfc/rfc4648.html#section-3.2


adamthom-amzn:
I think I excluded strings with characters outside the alphabet to be more in-line with prior art (MIME specifies they should be ignored). If we want this to be the canonical behavior, because some decoders will reject unpadded data, we should take a note to have protocol tests for both.

adamthom-amzn (approved):

PR #905, opened 2021-09-13
Add more base64 malformed request tests

JordonPhillips:
This adds more base64 tests, including ones for strings that use
mediaType in places where they need to be encoded.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-string.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-string.smithy
@@ -1,0 +1,59 @@
+$version: "1.0"
+
+namespace aws.protocoltests.restjson
+
+use aws.protocols#restJson1
+use smithy.test#httpMalformedRequestTests
+
+@http(uri: "/MalformedString", method: "POST")
+operation MalformedString {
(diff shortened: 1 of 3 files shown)


smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-string.smithy
```diff
@@ -0,0 +1,59 @@
...
+            value: [
+                // Insufficient padding
+                "xyz",
+                // Extant, but also insufficient padding
+                "YmxvYg=",
+                // Invalid characters
+                "[][]",
+                // Invalid characters which are commonly used as filename-safe
+                // alternatives to + and /
+                "-_=="
```
JordonPhillips, line 47: Do we want to accept these or not? They're often used for uri-safe and filename-safe encoding. They can also be percent encoded in the uri case.

This also raises the question of whether media-typed strings should be encoded when used in a URI. Percent encoding should be sufficient there, I think.

adamthom-amzn: Yeah, percent encoding sounds right to me.

adamthom-amzn (approved):

PR #915, opened 2021-09-22
Add string split by character type utilities

JordonPhillips:
This adds in the apache string utilities to split by character type. This will be useful for languages that use snake casing.


mtdowling:
Do the utilities in https://github.com/awslabs/smithy/blob/main/smithy-utils/src/main/java/software/amazon/smithy/utils/CaseUtils.java not work the way you need?

PR #919, opened 2021-09-24
Allow common errors to be bound to a service

### LINKED ISSUE #894: Add `errors` to service shape
It's a common pattern in APIs to support a set of errors across all operations. An `errors` property could be added to the `service` shape to add a list of errors to every operation within the closure of the service.

mtdowling:
This commit allows a common list of errors to be bound to a service
shape so that every operation within the closure of the service
implicitly can return the common errors. This is a very common pattern,
and this change makes it easier to add common errors without needing
things like validation to enforce them being added to every operation.

Existing code generation tooling will either need to be updated to
perform a model transform that copies common errors onto each operation
bound within a service (see `copyServiceErrorsToOperations`), or they
need to use the introduced `getErrors` method on `OperationIndex` that
also takes in a service shape ID.

Closes #894

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedServiceError.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedServiceError.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.diff.evaluators;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import software.amazon.smithy.diff.ChangedShape;
+import software.amazon.smithy.diff.Differences;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.validation.ValidationEvent;
+
+/**
+ * Emits a warning when an error is added to a service.
+ */
+public final class AddedServiceError extends AbstractDiffEvaluator {
+    @Override
+    public List<ValidationEvent> evaluate(Differences differences) {
+        return differences.changedShapes(ServiceShape.class)
+                .flatMap(change -> createErrorViolations(change).stream())
+                .collect(Collectors.toList());
+    }
+
+    private List<ValidationEvent> createErrorViolations(ChangedShape<ServiceShape> change) {
+        if (change.getOldShape().getErrors().equals(change.getNewShape().getErrors())) {
+            return Collections.emptyList();
+        }
+
+        List<ValidationEvent> events = new ArrayList<>();
+        for (ShapeId id : change.getNewShape().getErrors()) {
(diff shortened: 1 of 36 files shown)


adamthom-amzn (approved):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedServiceError.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private List<ValidationEvent> createErrorViolations(ChangedShape<ServiceShape> change) {
+        if (change.getOldShape().getErrors().equals(change.getNewShape().getErrors())) {
+            return Collections.emptyList();
+        }
+
+        List<ValidationEvent> events = new ArrayList<>();
+        for (ShapeId id : change.getNewShape().getErrors()) {
+            if (!change.getOldShape().getErrors().contains(id)) {
+                events.add(warning(change.getNewShape(), String.format(
```
JordonPhillips, line 47: This should be omitted if the error was already bound to every operation, or at least relegated to an info.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
```diff
@@ -180,7 +180,7 @@ public Builder addError(String errorShapeId) {
          * @param errorShapeIds Error shape IDs to add.
          * @return Returns the builder.
          */
-        public Builder addErrors(List<ShapeId> errorShapeIds) {
+        public Builder addErrors(Collection<ShapeId> errorShapeIds) {
```
JordonPhillips, line 183: I know we have a thing in OperationIndex, but I feel like we should still have a helper method in OperationShape to get all errors and then deprecate `getErrors` or do something else to discourage its direct use.

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborVisitor.java
```diff
@@ -71,6 +71,10 @@
         for (ShapeId resource : shape.getResources()) {
             addBinding(result, shape, resource, RelationshipType.RESOURCE);
         }
+        // Add ERROR relationships from service -> errors.
+        for (ShapeId errorId : shape.getErrors()) {
```
JordonPhillips, line 75: Is there a way in the selector to get all errors for an operation bound to a service? Or do we just have to join them manually

JordonPhillips (changes requested):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedServiceError.java
mtdowling: I think it's worth calling out no matter what because every new operation added to the service will now get the added error. That's something you really need to pay attention to and make sure it makes sense.

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NeighborVisitor.java
mtdowling: You can do that using something like:

```
service $serviceErrors(-[error]->) ~> operation :is(${serviceErrors}, -[error]->)
```

1. Find each service
2. Store the errors of the service in a variable named `$serviceErrors`
3. Traverse every shape in the service
4. Only look at operations within the service
5. Emit every `$serviceError` plus every error attached to the operation

smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
mtdowling: I think the shape is fine though. It's part of the metamodel and it returns the errors directly bound to the shape itself. It would be like deprecating an accessor in an AST which isn't a great idea -- we need to know the errors bound directly to the operation and emit them when serializing the model, for example.

I can add a documentation pointer to the index.

mtdowling: I added another method like what you were suggesting -- `addErrors(ServiceShape service)`. It will return common errors of the given service plus operation errors.

JordonPhillips (approved):

PR #920, opened 2021-09-24
Add default trait design doc

mtdowling:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/designs/default-zero-value.md
+++ b/designs/default-zero-value.md
@@ -1,0 +1,816 @@
+# Default zero values
+
+* **Author**: Michael Dowling
+* **Created**: 2021-08-24
+* **Last updated**: 2021-12-16
+
+
+## Abstract
+
+This document describes a fundamental change to the way Smithy describes
+structure member nullability (also known as optionality). By adding a
+`@default` trait to structure members, code generators can generate non-null
+accessors (or getters) for members marked as `@default` or `@required`
+(with some caveats for backward compatibility). While this proposal is not
+AWS-specific, the practical impact of it is that when implemented, it will
+convert thousands of optional property accessors to non-optional in the
+Rust, Kotlin, and Swift AWS SDKs.
+
+In order to implement these changes, Smithy will need a 2.0 bump of the IDL.
+
+
+## Terms used in this proposal
+
+* null: This document uses the term null, nullable, and nullability to refer
+  to members that optionally have a value. Some programming languages don't
+  have a concept of `null` and need to map this to whatever abstraction is
+  appropriate for that programming language (for example, `Option` in Rust
+  or `Maybe` in Haskell).
+* accessor, getter: In this proposal, the terms "accessor" and "getter" should
+  be considered generic terms that translate to however a programming language
+  exposes structure member values. For example, some programming languages like
+  Java often use getter methods while others might expose struct style
+  properties directly.
+
+
+## Motivation
+
+Most structure member accessors generated from Smithy models return nullable
+values. As new languages like Rust and Kotlin that explicitly model nullability
+in their type systems adopt Smithy, excessive nullability in generated code
+becomes burdensome to end-users because they need to call methods like
+`.unwrap()` on everything. Generating every structure member as nullable makes
+it hard for customers to know when it is safe to dereference a value and when
+it will result in an error. Adding the ability to control when a value is
+nullable vs when a value is always present provides an ergonomic and safety
+benefit to end users of code generated types.
+
+For example, after this proposal is implemented, the following Rust code:
+
+```rust
+println!("Lexicons:");
+let lexicons = resp.lexicons.unwrap_or_default();
+for lexicon in &lexicons {
+    println!(
+        "  Name:     {}",
+        lexicon.name.as_deref().unwrap_or_default()
+    );
+    println!(
+        "  Language: {:?}\n",
+        lexicon
+            .attributes
+            .as_ref()
+            .map(|attrib| attrib
+                .language_code
+                .as_ref()
+                .expect("languages must have language codes"))
+            .expect("languages must have attributes")
+    );
+}
+println!("\nFound {} lexicons.\n", lexicons.len());
+```
+
+Can be simplified to:
+
+```rust
+println!("Lexicons:");
+for lexicon in &resp.lexicons {
+    println!(" Name:     {}",  lexicon.name);
+    println!(" Languages: {:?}\n",
+        lexicon.attributes.map(|attrib| attrib.language_code)
+    );
+}
+println!("\nFound {} lexicons.\n", resp.lexicons.len());
+```
+
+
+## Background
+
+Nullability in Smithy IDL 1.0 is controlled in the following ways:
+
+1. Map keys, values in sets, and unions cannot contain null values. This
+   proposal does not change this behavior.
+2. List and map values cannot contain null values by default. The `@sparse`
+   trait applied to list and map shapes allow them to contain null entries.
+   These kinds of collections rarely need to contain nullable members. This
+   proposal does not change this behavior.
+3. The nullability of structure members in Smithy today is resolved using the
+   following logic:
+   1. If the member is marked with `@box`, it's nullable.
+   2. If the shape targeted by the member is marked with `@box`, it's nullable.
+   3. If the shape targeted by the member is a byte, short, integer, long,
+      float, double, or boolean, it's non-null.
+   4. All other members are considered nullable.
+
+
+### Why the `@required` trait was unreliable for codegen
+
+Removing the `@required` trait from a structure member has historically been
+considered backwards compatible in Smithy because it is loosening a restriction.
+The `@required` trait in Smithy 1.0 is categorized as a constraint trait and
+only used for validation rather than to influence generated types. This gives
+service teams the flexibility to remove the required trait as needed without
+breaking generated client code. The `@required` trait might be removed if new
+use cases emerge where a member is only conditionally required, and more
+rarely, it might be added if the service team accidentally omitted the trait
+when the service initially launched.
+
+For context, as of May 2021, the `@required` trait has been removed from a
+structure member in 105 different AWS services across 618 different members.
+Encoding the `@required` trait into generated types would have made changing a
+member from required to optional a breaking change to generated code, which is
+something most services try to avoid given its frequency and the cost of
+shipping a major version of a service and client. While this data is
+AWS-specific, it's indicative of how often disparate teams that use Smithy
+(or its predecessor in of Amazon) sought to update their APIs so that members
+can become optional.
+
+
+## Goals and non-goals
+
+**Goals**
+
+1. Reduce the amount of nullability in code generated from Smithy models.
+2. Maintain similar backward compatibility guarantees that exist today like
+   being able to backward compatibly remove the `@required` trait from a
+   structure member without breaking previously generated clients.
+3. Maintain Smithy's protocol-agnostic design. Protocols should never influence
+   how types are generated from Smithy models; they're only intended to change
+   how types are serialized and deserialized. In practice, this means that
+   features of a protocol that extend beyond Smithy's metamodel and definitions
+   of structure member nullability should not be exposed in code generated
+   types that represent Smithy shapes.
+4. We should be able to use the same code-generated types for clients, servers,
+   and standalone type codegen.
+
+**Non-goals**
+
+1. Mandate how protocols are designed. This proposal intends to define
+   requirements and affordances about how nullability is defined the Smithy
+   metamodel, and in turn, in code generated from Smithy models.
+
+
+## High-level summary
+
+This proposal introduces a `@default` trait to structure members that
+initializes structure members with a default, zero value.
+
+```
+structure Message {
+    @required
+    title: String
+
+    @default
+    message: String
+}
+```
+
+In the above example:
+
+- `title` is marked as `@required`, so it is non-null in generated code.
+- `message` is also non-null in generated code and is assigned a default, zero
+   value of `""` when not explicitly provided. This makes it easier to use
+   `title` and `message` in code without having to first check if the value is
+   `null`. For example, you can call `message.message.size()` without first
+   checking if the value is non-null.
+
+If the service ever needed to make `title` optional, they can replace the
+`@required` trait with the `@default` trait:
+
+```
+structure Message {
+    @default
+    title: String
+
+    @default
+    message: String
+}
+```
+
+With the above change, codegen remains the same: both values are non-null.
+However, if `title` is not set in client code, server-side validation for the
+type will _not_ fail because `title` has a default, zero value of `""`. The
+trade-off to this change is that it is impossible to tell if `title` was
+omitted or explicitly provided.
+
+
+## Proposal overview
+
+This proposal will be implemented in two phases. The rules around nullability
+will be extended in Smithy IDL 1.0, and then changed in a backward incompatible
+way to simplify nullability rules in Smithy IDL 2.0. These breaking changes to
+the IDL will be made in such a way that 1.0 and 2.0 models can be loaded
+simultaneously without a major version bump in Smithy's Java libraries. This
+phased approach was chosen because we know that the number of models and
+tooling yet to be written for Smithy far exceeds the current number of models
+already written for Smithy. We'll provide a path for existing 1.0 models to more
+easily and confidently upgrade to IDL 2.0, and leave behind the more confusing
+nullability semantics of IDL 1.0 going forward as 2.0 becomes the default.
+
+
+### In Smithy IDL 1.0
+
+1. Add the `@default` trait to structure members.
+2. Warn when a structure member has non-nullable semantics but is not marked as
+   `@default`.
+3. Warn when `@box` is used on structure members.
+4. Warn when any byte, short, integer, long, float, double, or boolean shape is
+   _not_ marked with the `@box` trait. We want all shapes to be considered
+   nullable by default, and for structure members to opt-in to non-nullable
+   semantics through the `@default` trait or `@required` trait.
+5. Deprecate the `PrimitiveBoolean`, `PrimitiveShort`, `PrimitiveInteger`,
+   `PrimitiveLong`, `PrimitiveFloat`, and `PrimitiveDouble` shapes from the
+   Smithy prelude. Instead, target their corresponding prelude shapes like
+   `Boolean`, `Short`, `Integer`, etc.
+6. Create new guidance for code generators that allows them to generate
+   non-nullable accessors for structure members marked as `@required` or
+   `@default`.
+
+
+### In Smithy IDL 2.0
+
+1. Remove the `@box` trait from the Smithy 2.0 prelude. IDL 2.0 models will
+   fail if they use the `@box` trait.
+2. Remove the `PrimitiveBoolean`, `PrimitiveShort`, `PrimitiveInteger`,
+   `PrimitiveLong`, `PrimitiveFloat`, and `PrimitiveDouble` shapes from the
+   Smithy 2.0 prelude. IDL 2.0 models will fail if they target these shapes.
+3. Update the Smithy IDL 2.0 model loader implementation to be able to load
+   Smithy 1.0 models alongside Smithy 2.0 models.
+   1. Warn when a Smithy 1.0 model file is loaded.
+   2. Inject the `@default` trait on structure members when needed.
+   3. Remove the `@box` trait from the model.
+   4. Rewrite members that target one of the removed Primitive* shapes to
+      target the corresponding non-primitive shape in the prelude (for example,
+      `PrimitiveInteger` is rewritten to target `Integer`).
+
+
+## `@default` trait
+
+The `@default` trait can be applied to structure members to indicate that the
+targeted shape has a default, zero value. The `@default` trait *does not* allow
+for a custom default value.
+
+The following example defines a structure with a `@default` "title" member that
+has a default zero value:
+
+```
+structure Message {
+    @default
+    title: String // defaults to ""
+}
+```
+
+The `@default` trait is defined in Smithy as:
+
+```
+@trait(selector: """
+       "structure > member
+        :not(> :test(union, structure > :test([trait|required])))"""",
+       conflicts: [required, box])
+structure default {}
+```
+
+The `@default` trait conflicts with the `@box` trait and `@required` trait. The
+following model is invalid:
+
+```
+structure Message {
+    @default    // ERROR: this trait conflicts with @required.
+    @required
+    title: String
+}
+```
+
+
+### **Default zero values**
+
+The following list describes the default zero value of each kind of shape.
+Programming languages and code generators that cannot initialize structure
+members with the following default values SHOULD continue to represent those
+members as nullable as this is semantically equivalent to the default zero
+value.
+
+- Boolean: boolean false (`false`)
+- Numbers: numeric zero (`0`)
+- String: an empty string (`""`). Strings with the enum trait also have a
+  default value of "".
+- Blob: an empty blob. This includes blob shapes marked as `@streaming`.
+- Timestamp: zero seconds since the epoch (for example, `0`, or
+  `1970-01-01T00:00:00Z`).
+- Document: a null document value.
+- List: an empty list (`[]`).
+- Set: an empty set (`[]`).
+- Map: an empty map (`{}`).
+- Structure: an empty structure. Structures only have a default zero value if
+  none of the members of the structure are marked with the `@required` trait.
+- Union: no default zero value. A union MUST be set to one of its variants for
+  it to be valid, and unions have no default variant.
+
+
+### Impact on API design
+
+Members marked with the `@default` trait cannot differentiate between whether a
+property was set to its zero value or if a property was omitted. Default zero
+values are part of the type system of Smithy and not specific to any particular
+serialization of a shape. This makes the `@default` trait a poor choice for
+partial updates or patch style APIs where it is critical to differentiate
+between omitted values and explicitly set values.
+
+To help guide API design, built-in validation will be added to Smithy to detect
+and warn when `@default` members are detected in the top-level input of
+operations that start with `Update`, operation bound to the `update` lifecycle
+of a resource, or operations that use the `PATCH` HTTP method.
+
+For example, the following model:
+
+```
+$version: "2"
+namespace smithy.examnple
+
+operation UpdateUser {
+    input: UpdateUserInput
+}
+
+structure UpdateUserInput {
+    @default
+    username: String
+}
+```
+
+Would emit a warning similar to the following:
+
+```
+WARNING: smithy.example#UpdateUserInput$userName (DefaultOnUpdate)
+     @ /path/to/main.smithy
+     |
+ 4   | operation UpdateUser {
+     | ^
+     = This update style operation has top-level input members marked with the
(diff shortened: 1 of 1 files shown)


designs/default-zero-value.md
```diff
@@ -0,0 +1,778 @@
...
+  clients that think a structure member is required, but the service has since
+  transitioned the member to use the `@default` trait.
+- Serializing zero values can cause unintended information disclosure like
+  revealing internal-only members (this is the same reason that null values
+  SHOULD NOT be serialized).
+
+
+### Deserialization
+
+Structure members marked as `@required` are not in and of themselves a guarantee
```
Baccata, line 498: I have a lot of thoughts on this PR and writing them all will take a little bit of time that I'm hoping to find next week. 

However, the implications of this paragraph constitutes a significant risk for how we're designing the use of smithy in my organisation, and I find myself compelled to write something now. I completely understand that the project is primarily shaped by AWS usecases, but I think this statement (as well as a lot of details in your write-up) ought to be protocol-specific, at the very least. Otherwise you're essentially forcing all protocols to be akin to protobuf v3, which is somewhat controversial and unpractical in many cases. 

But, further than that, what constitutes a breaking change is extremely dependant on whether the types are used in covariant (output) or contravariant (input) position. It is important to recognise that, and I'll happily share a write-up I did on the subject, next monday. 

I 100% believe that having clients have to accept the absence of value on @required fields during deserialisation is wrong. When a service goes from @required to @default on an output type, my opinion is that the service should have to send the default value to not break existing clients, and for clients to protect against what is essentially breaking behaviour from services.

mtdowling: Thanks for taking a look and leaving your thoughts. Here are a few high level points that I can work back into the doc:

1. The goal of this doc isn't to dictate protocols, but rather how we think about nullabiltiy and backward compatibility around nullability in Smithy, and how that impacts codegen. If code generators are compliant with this proposal, then they won't get broken by services that think they're making backward compatible changes. If we are making claims that too broadly impact protocol design, we'll adjust the proposal.
2. In Smithy, protocols don't change how types are code-generated. The only thing that should change between protocols is how types are serialized and deserialized. Changing the protocol a client uses to call a service should not change code-generated types, which means what shapes do and don't have a zero value needs to be standardized in a protocol agnostic way. That's a kind of North Star for features in Smithy, which is what we'll need to arrive at with this proposal.
3. We should be able to use the same code-generated types for both client and server codegen use cases, and we also have use cases for Smithy where types are used outside of client/server interactions.
4. The default trait isn't only for transitioning something from required -> optional. It's used to provide a default value for members that don't need nullability, like when 0 and unset are equivalent, defaulting to 0 makes a service easier to use. You can actually do this today in Smithy with the `smithy.api#PrimitiveInteger` shape and other shapes in the prelude. This proposal is expanding that to other types.

> what constitutes a breaking change is extremely dependant on whether the types are used in covariant (output) or contravariant (input) position

We did consider this, and I can work this into the FAQ. For example, making an input member transition from required -> optional might not break a client because it's less restrictive, and neither would making an output member transition from optional -> required because output types are typically created only during deserialization (though things like mocking and injecting responses are cases where this isn't true).

But the thing about this proposal is that it's about the types we code generate from Smithy models. Using required traits and default traits to influence code generation means that the changes I just listed would be breaking changes to many Smithy code generators because the types would change from `T` -> `Option<T>` or `Option<T>` -> `T`. While this only will impact code generators that can bake required traits into their type systems, specifications around what is and isn't backward compatible in models gives these code generators the affordances they need in order to generate more idiomatic types.

I think one way to support different nullability rules for input and output is to bifurcate codegen so that shapes used as part of input are generated separately from shapes used as part of output. There are a lot of tradeoffs to this though, for example code size increase (2 generated types per shape) and awkwardness of roundtripping output shapes as input. 

More intrinsically, a design goal of Smithy is that we should be able to use the same code-generated types for both client and server codegen (we're building Smithy service codegen as we speak!). We also have use cases for Smithy where types are used outside of client/server interactions, which requires codegen guidance to be based on types and not if they're used as part of input or output shapes.

> I 100% believe that having clients have to accept the absence of value on `@required` fields during deserialisation is wrong. When a service goes from `@required` to `@default` on an output type, my opinion is that the service should have to send the default value to not break existing clients, and for clients to protect against what is essentially breaking behaviour from services.

I agree this could be up to a protocol to decide (for example, maybe you need to reuse a protocol that doesn't know about Smithy's default value semantics). I can update this section to more generally talk about how different protocols might choose to serialize or not serialize default values. Keep in mind though that the default trait isn't only for transitioning something from required -> optional; for these kinds of use cases, always serializing a default value could lead to information disclosure (e.g., a new member that hasn't yet been made available to all users yet), which is actually why we typically don't serialize null values in Smithy.

I'm curious why this matters for your use case though -- if a required member is missing during deserialization, you can fill in a deterministic and constant zero value (that's the part where clients are protecting themselves from breaking changes, right?).

Baccata: Thanks for the detailed answer. I'll start with the points I agree with : 

### Things with which I agree

> In Smithy, protocols don't change how types are code-generated 

> We should be able to use the same code-generated types for both client and server codegen use cases, and we also have use cases for Smithy where types are used outside of client/server interactions.

I very much agree with these points, and the code generator I'm developing abide by these rules. This is good design and greatly facilitates testability and conceptual understanding of what constitutes a service. It also greatly reduces bikeshedding discussions. That is one think I very much like with smithy : the fact that protocol semantics are separated from core semantics in the IDL guides the design of tooling to a state that unlocks a number of amazing possibilities, when protocol-level details more often pollute the generated code  

> More intrinsically, a design goal of Smithy is that we should be able to use the same code-generated types for both client and server codegen (we're building Smithy service codegen as we speak!)

Once again, I'm very much in agreement that this is good design. I have built both server/client side support using the same generated code. My generated code isn't even specialised based on whether the usecase is client or server, I just have a bunch of abstractions that let me interpret code in client-side or server-side context, it works beautifully and my users are very happy with it. 

>   what shapes do and don't have a zero value needs to be standardized in a protocol agnostic way. 

I can certainly agree with that. 

### What I do not agree with : Smithy dictating what code generators ought to do in the ecosystem. 

So one thing that I am not in agreement with is that the smithy specification (as found in https://awslabs.github.io/smithy/) dictates rules about the generated code. 

Let's take the example of timestamps, as seen in [this table] : 

> The serialization of a timestamp is an implementation detail that is determined by a protocol and MUST NOT have any effect on the types exposed by tooling to represent a timestamp value. 

The main thing that I disagree with in that statement is the use of the `MUST` word (as opposed to `SHOULD`). I think the documentation has to offer an educated advice on what works and what doesn't. I for one think that AWS has a lot more experience in this field than any other organisation in the world, and I very much agree with this particular advice regarding how timestamps are to be translated to generated code, but any third party building smithy tooling is completely empowered to make that decision for themselves, as long as the data that transits over the wire is compliant to the semantics of every protocol it integrates with. With that in mind, the use of `MUST` is not only impossible to test/enforce at the scale of the ecosystem (which the smithy developers have no control over), but also feels un-necessarily prescriptive and I'm willing to bet it will raise eyebrows (some of the spec has certainly surprised me). 

> But the thing about this proposal is that it's about the types we code generate from Smithy models

Going back to the concept of "null", some languages (like haskell for instance, but a great many other) do not even support it. The absence of value is always modelled as an algebraic `Option` or `Maybe` datatype. Having the smithy spec state how a `@default` trait `must` translate in languages that do not support nulls make it seem like the smithy spec "overextends" into topics the authors of the specification might not be fully versed in. It might be that haskell folks think would prefer for @default to translate to `Maybe`, as it may be more idiomatic to these languages. Whether this choice presents a risk around library compatibility when specs change should be (imho) the responsibility of the person maintaining code generators for this language, certainly not of the smithy-spec authors. 

Changing `MUST` by `SHOULD` would remove the impression that the smithy-spec authors assume to know best for every possible language. Rather, it means they offer a (very) educated opinion on what works well in the languages that they have experience with, which is a totally valuable thing to do.  

So, though I totally think it's the prerogative of the smithy-team (or AWS) to be dictating rules about the code generators they maintain themselves, I do however think it's unrealistic for these rules to be worded as "sacrosanct" in smithy's specification.  

#### Going back to the topic of protocols 

> I agree this could be up to a protocol to decide (for example, maybe you need to reuse a protocol that doesn't know about Smithy's default value semantics). I can update this section to more generally talk about how different protocols might choose to serialize or not serialize default values.

That would be greatly appreciated. 

> Keep in mind though that the default trait isn't only for transitioning something from required -> optional; 

I have very much kept this in mind throughout this discussion. However, the semantics of `optional` (as in the absence of value) and `default` are easy to conflate, at the protocol level.  In my organisation, services and clients are reasonably often on a footing, and the data that transits over the wire is the result of discussions between the two. Whatever is `@required` is always meant to be filled with a meaningful value, and for smithy to prescribe that a `@required` => `@default` modification is always non-breaking, whilst dictating that it's okay for @default fields to sometimes not appear over the wire, would mean that it's okay for services to start breaking the contractual promise to always fill a meaningful value there. 

Smithy dictating this (regardless of protocols) also remove organisations ability to define protocols where the absence of value and the presence of a default value (in the serialised messages) have the same meaning. This might not occur often, and you might think that the difference is meaningless in most usecases you think about, but I certainly think this should be discussed at the level of protocols. 

> for these kinds of use cases, always serializing a default value could lead to information disclosure (e.g., a new member that hasn't yet been made available to all users yet), which is actually why we typically don't serialize null values in Smithy.

I think if that actually something people worry about, then neither @default nor @optional are the right modelling. But I 100% think that it's AWS prerogative to use it as such in the AWS protocols 😄 . 

Also I kinda dislike the wording "we typically don't serialize null values in Smithy", when I think you mean "we typically don't serialise null values in the AWS-specific protocols we defined using Smithy". I think the distinction is important to be made if you want smithy to outlive the boundaries of AWS and become an industry standard ... which I very much think it should ! 

#### Summary : 

* For reasons I've already expressed, I very much agree with a plan to removal of `@box` and the shapes prefixed with  `Primitive`, because they are obviously tied to some language-level concerns (in particular JVM), which (imho) inadvertently signals some undesirable coupling between smithy and programming languages. 
* I think `@default` is a sensible path forward
* I think the smithy specification should not dictate what code generators do, but rather give educated opinion on the matter (ie replace MUST by SHOULD)
* I think whether `0` values transit on the wire when a shape is annotated with `@default` should be a protocol-level decision rather than a IDL-level decision (in a similar fashion to how protocols are able to state whether they support `Documents` or not). 

#### Post scriptum : 

I hope that this discussion is not perceived as a lack of appreciation or distrust. I think the `smithy` language and core libraries (`smithy-model`, `smithy-diff`, etc) are already showing immense value, and I'm really grateful for this product.

mtdowling: Regarding timestamps and codegen: I understand your broader point about codegen, but I want to address your specific example about timestamps. I think the current guidance is necessary because it sets expectations of what modelers can expect from timestamp shapes. Unambiguous codegen is a goal of Smithy, and these explicit statements about what affordances the metamodel and traits provide is important (even if some requirements can't be validated through compliance checks). If this timestamp requirement became a "SHOULD" instead of a "MUST", then there's ambiguity as to what implementations need to support when generating timestamp types. The "MUST" reinforces that the timestamp shape is part of Smithy's metamodel and has specific meaning as a type in the metamodel (that is, "an instant in time with no UTC offset or timezone").

If an implementation needs to deviate from the spec, then they can, for example, use a string shape with a custom trait to indicate that the string contains a timestamp in a specific format and with specific semantics (like a timezone or something). They can then add codegen extensions to the code generators they use to honor the trait.

> The absence of value is always modelled as an algebraic Option or Maybe datatype. Having the smithy spec state how a `@default` trait must translate in languages that do not support nulls make it seem like the smithy spec "overextends" into topics the authors of the specification might not be fully versed in.

This proposal was created in response to AWS building code generators in languages like Rust that do not have explicit nulls, but rather use `Option<T>` and `T` (like `Maybe` in Haskell). These kinds of languages were a key consideration when writing this proposal. We're confident that implementors for Haskell, Rust, and other programming languages are used to mapping null concepts from other systems like JSON Schema to their languages, and we didn't think the use of the word null would cause undue confusion. I mention in the motivation section that nullability is equivalent to optionality, but would a more tactical solution here be to move that guidance into the intro?

> It might be that haskell folks think would prefer for `@default` to translate to Maybe, as it may be more idiomatic to these languages.

Using `Maybe` for members marked with the `@default` trait would only make sense to me if an implementation just treats all members as optional (some of the generators we're building in AWS will probably function this way due to limitations in different PLs).

However, I think explicitly using `Maybe` for members marked with the `@default` trait would be a misunderstanding of the default trait, and it would make that implementation use features that are not afforded to it by Smithy's metamodel (which makes the implementation at risk of incompatibility with other Smithy implementations). The `@default` trait does not imply a tri-state of set, unset, or default. The usage of the `@default` trait forfeits the knowledge of whether a type is set explicitly to a zero value or defaulted to a zero value. If that information is required, then `@default` / `@required` shouldn't be used. For this use case you need to use nullable/optional/maybe/whatever you wanna call it members, which provides the member the ability to differentiate between set and unset.

>  Whether this choice presents a risk around library compatibility when specs change should be (imho) the responsibility of the person maintaining code generators for this language, certainly not of the smithy-spec authors.
> ...
> Changing MUST by SHOULD would remove the impression that the smithy-spec authors assume to know best for every possible language. Rather, it means they offer a (very) educated opinion on what works well in the languages that they have experience with, which is a totally valuable thing to do.

I think this is a major benefit of Smithy. By standardizing on backward compatibility guarantees and API lifecycle changes over time, we make implementations more consistent with each other and we easily diff models for backward compatibility changes. It becomes easier to reason about models generally in API reviews rather than needing to understand every possible and trait and protocol used by the model.

Ultimately, the decision does lie with the implementor, but the specification is setting expectations around what every other implementation expects. Deviating from the spec tells implementors that they're signing up for way more work because they need to customize every client/server implementation to adhere to their differences from the core specification.

> Whatever is `@required` is always meant to be filled with a meaningful value, and for smithy to prescribe that a `@required` => `@default` modification is always non-breaking, whilst dictating that it's okay for `@default` fields to sometimes not appear over the wire, would mean that it's okay for services to start breaking the contractual promise to always fill a meaningful value there.

I disagree that this would be a contractual break in Smithy, but I can see how a specific protocol might consider it a break in the serialization layer. The current wording says default values SHOULD NOT be serialized. I think that's a safe default stance to take for protocol implementations, and because it's a SHOULD, future protocols are free to deviate if necessary. As I said in my previous reply, I'll take a pass at the proposal to clear this up.

> Smithy dictating this (regardless of protocols) also remove organisations ability to define protocols where the absence of value and the presence of a default value (in the serialised messages) have the same meaning. This might not occur often, and you might think that the difference is meaningless in most usecases you think about, but I certainly think this should be discussed at the level of protocols.

Agreed that whether a default value is serialized or not isn't a hard requirement of this proposal.

Whether implementations can differentiate between explicit default values and default values *is* something this proposal and Smithy in general should standardize because it affects the compatibility of tools across implementations and programming languages.

> Also I kinda dislike the wording "we typically don't serialize null values in Smithy", when I think you mean "we typically don't serialise null values in the AWS-specific protocols we defined using Smithy". I think the distinction is important to be made if you want smithy to outlive the boundaries of AWS and become an industry standard ... which I very much think it should !

That's a fair point, and I'll keep that in mind. I often think about AWS because we have tens of thousands of models defined in Smithy or Smithy's internal predecessor, which gives a huge corpus of information for me to pull from, and lessons learned. I do think that the guidance around information disclosure is super useful to document, and teams should care about it, but you're right that Smithy != AWS.

Baccata: Hi Michael, and thanks for the detailed reply, this is much appreciated. 

> I mention in the motivation section that nullability is equivalent to optionality, but would a more tactical solution here be to move that guidance into the intro?

This has probably eluded me, and I think it should be the case. 

>  The usage of the @default trait forfeits the knowledge of whether a type is set explicitly to a zero value or defaulted to a zero value. If that information is required, then @default / @required shouldn't be used. For this use case you need to use nullable/optional/maybe/whatever you wanna call it members, which provides the member the ability to differentiate between set and unset.

I can personally agree with that statement. I can also say that I know a great many people that will have strong opinions about it (we Scala devs tend to be extremely opinionated about everything, as you may already have guessed 😅)

Anyhow, regardless of the MUST vs SHOULD debate, as long as it's made clear the expectations around how `@default` values are serialised (or absent from the serialisation) are protocol-specific, my main concern is mostly addressed ... except maybe for the case of structures / unions.  

#### Regarding structures and unions 

Somewhere else in the document, you state the following (quoting here to avoid splitting the convo) 

> The following list describes the default zero value of each kind of shape. Programming languages and code generators that cannot initialize structure members with the above default values MUST continue to represent those members as nullable.

and

> Structure: no zero value. Structures are always nullable and never have a
  default value. In order for a structure to have a default value, all members
  of the structure would need to be optional. Because it is a backward
  compatible change to remove the `@required` trait from a structure member,
  whether a structure has a zero value or not is too volatile to rely on for
  code generation.
> Union: no zero value. Unions are always nullable, MUST be set to one of its
  variants for it to be valid, and have no default variant. Adding some kind of
  special "default" or "unset" value to unions in code to account for default
  values would ultimately require _more_ special-case code in generated code to
  use than just dealing with nullable unions.

I do not see this as universally correct. This is making @required inherently useless on structures and unions, and implies that client code has to protect against nullity no matter what, as they cannot trust that the service will keep sending the data they rely on. It's making life a lot more convenient for the service (by relaxing its ability to evolve by a LOT) and a lot more inconvenient to the client. It drastically impacts developer experience on both sides, by forcing client users to over protect (which is like, THE thing that Scala devs find the most annoying about using the java AWS SDKs), and preventing server users from trusting the type systems to guide them through what it they should set and what they shouldn't set. 

I also find it paradoxal that the removal of `@required` on members targeting structures would be considered safe wrt compatibility, given the current [specs](https://awslabs.github.io/smithy/1.0/spec/core/constraint-traits.html?highlight=required#required-trait). The wording on the specs imply that clients should be able to trust that a required value will always appear in the output of operations. Making it okay for `@required` to be removed from output shapes would imply that clients cannot trust that information. 

#### Playing with an open hand 

I can tell you right now that as a third party, I will not abide by the directive to represent all structures/union fields as optional in the tooling that I'm building : in the generated Scala code, members targeting structures and unions and that have `@required` will be generated in an unwrapped fashion, whereas those who don't will be generated wrapped in `Option`. This is just more idiomatic, and better UX for Scala developers, in particular in the usecase that I'm setting up in my organisation. I might decide to bake some opt-in logic to abide by your directive in my code generator, so that in the context of a pure-Scala AWS SDK, clients would not start breaking when AWS performs what is, in my opinion, a breaking change (ie the removal of `@required` on field that appears in covariant/output position). 

I can tell you however that I am also building some tooling using `smithy-diff` to check for breaking changes, without relying on the default validation events, but rather implementing my own events, abiding by some rules of compatibility that I can prove formally (using a proof system). This tooling will be used internally in my organisation. 

I am **extremely appreciative** with the fact that he tooling empowers me to do so, and even though I disagree with some of aspects of the specification, as long as I'm able to apply discretion and that the smithy libraries do not try to overly protect against these usecases, I'll continue championing the use of smithy in my organisation. 

#### In conclusion 

I think we have fundamentally diverging opinions on the guarantees that users should be able to expect when using `@required`. The smithy libraries (`smithy-model`, `smithy-diff`) are, I think, very well designed and as long as they are flexible enough to cater to my organisation needs, I think the divergence of opinion isn't too problematic. In particular : as long as the use of `@required` does not become more constrained than it already is (ie, the smithy team might decide that it's not possible to put @required on members targeting structures/unions anymore), I'm happy to take responsibility for whatever pain points might come from me making the decision to deviate from the specs. 

Overall, I'm am in agreement with most of the specification as it exists, and I am partially in agreement with the `default` proposal. 

I do not have expectations for the smithy team to take steps towards something that would cater to my opinions/usecases, but if it was the case, a possible course of action would be for another trait to be introduced, with stricter semantics than `@required` (could call it `strictRequired` for instance) 

Kind regards,

designs/default-zero-value.md
```diff
@@ -0,0 +1,888 @@
...
+
+### How often has the `@required` trait been added to members in AWS?
+
+The required trait has been added to a member 9 different times across 5
+different services.
+
+
+### Why not base backward compatibility rules around inputs and outputs?
+
+This proposal talks about backward compatibility of structure members and does
```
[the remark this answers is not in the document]
mtdowling: I updated this section of the proposal to now special case input shapes as having more relaxed backward compatibility semantics. I'll do a followup design doc, but basically the `@input` trait will be used as a kind of contract that a structure is only ever used as the input of a single operation. Because of this knowledge, we can allow teams to remove the `@required` trait without replacing it with the `@default` trait, and code generators can account for that however necessary. There aren't any affordances made explicitly for output shapes, and it's not something the spec recommends, but you can implement more relaxed requirements there too if it's something you need.

designs/default-zero-value.md
```diff
@@ -0,0 +1,888 @@
...
+```
+
+If this structure is later updated so that `baz` is no longer required but has
+the `@default` trait, then the following constraint trait can never hold true
+when `baz` uses the zero value:
+
+```
+structure Foo {
+    @default
+    @length(min: 1, max: 100) // <-- ERROR: default value can never satisfy min > 1
```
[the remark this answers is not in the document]
mtdowling: I made a change here to say that constraint traits need to be aware of zero values. That is, if you say a number with a `@default` has to be between 10 and 15, then it can be `0` or any number between 10 and 15, inclusive. This will make it easier for teams to actually validate their constraints without needing to write custom code, and documentation will be more accurate about what the actual constraints are.

designs/default-zero-value.md
```diff
@@ -0,0 +1,872 @@
...
+   In this case, rather than failing, a client SHOULD set the member value to
+   its default zero value. Failing to deserialize the structure is a bad
+   outcome because what the service perceived as a backward compatible change
+   (i.e., replacing the `@required` trait with the `@default` trait) could
+   break previously generated clients.
+
+
+## AWS specific changes
+
+### AWS Protocols
```
Baccata, line 543: 👍

designs/default-zero-value.md
```diff
@@ -0,0 +1,872 @@
...
+attempts to dereference the value, and this kind of change is generally only
+made when a client opts-into some new kind of functionality or workflow. If an
+SDK fails during deserialization because a required output member is missing,
+then customer applications would be completely broken until they update their
+SDK.
+
+
+#### Special casing top-level input structures
+
+AWS SDKs SHOULD generate every required top-level input member as optional even
```
Baccata, line 597: I  greatly appreciate the indication that this expectation is AWS specific 👍

designs/default-zero-value.md
```diff
@@ -0,0 +1,872 @@
...
+Nearly 80% of all required traits used in AWS today appear in only top-level
+members of operation inputs (~24,500 total required members vs ~19,000
+required top-level input members). By special casing top-level operation input
+members as optional even if they are marked with the required trait, this
+proposal becomes significantly safer for AWS SDKs to adopt.
+
+
+#### `aws.api#serverRequired` trait
+
+The `aws.api#serverRequired` trait is used instead of `required` to indicate
```
Baccata, line 613: Does this mean that AWS specifications will see a transition of trait from `smithy.api#required` to `aws.api#serverRequired` ? That'd be really nice for third parties

designs/default-zero-value.md
```diff
@@ -0,0 +1,857 @@
...
+## Alternatives and trade-offs
+
+## A note on added complexity in Smithy 1.x
+
+Nullability rules in Smithy 1.x were already complex, and we're making them more
+complex in this proposal. Before this proposal, a member was considered nullable
+if it was marked with the `@box` trait or if it targeted a shape marked with the
+`@box` trait, or anything other than a boolean or number. With this proposal, we
+are also introducing the `@default` trait, further complicating nullability. We
+will address this added complexity if we release a Smithy 2.x that removes the
```
kstich, line 668: This sentence is out of date with the proposed IDL 2.0 that does remove the `@box` trait.

designs/default-zero-value.md
```diff
@@ -0,0 +1,857 @@
...
+   can be removed without breaking their end users.
+
+We will work to improve the accuracy and consistency of the required trait in
+AWS API models where possible, but it will take significant time and effort
+from hundreds of different teams within AWS. To accommodate this shift in
+modeling changes, we will introduce AWS-specific traits to influence code
+generation and provide AWS SDK specific code generation recommendations.
+
+
+#### `aws.api#clientOptional` trait
```
kstich, line 599: This is a downstream consideration, but generic shape generation will need to be set up to expose the right hooks for this AWS specific update.

designs/default-zero-value.md
```diff
@@ -0,0 +1,857 @@
...
+  default value of "".
+- Blob: an empty blob. This includes blob shapes marked as `@streaming`.
+- Timestamp: zero seconds since the epoch (for example, `0`, or
+  `1970-01-01T00:00:00Z`).
+- Document: a null document value.
+- List: an empty list (`[]`).
+- Set: an empty set (`[]`).
+- Map: an empty map (`{}`).
+- Structure: an empty structure. Structures only have a default zero value if
+  none of the members of the structure are marked with the `@required` trait.
```
JordonPhillips, line 301: This might get answered later, but doesn't this leave open a door for backwards-incompatible changes when child structures remove the required trait?

designs/default-zero-value.md
```diff
@@ -0,0 +1,857 @@
...
+
+Protocols MAY choose if and how the `@default` trait impacts serialization and
+deserialization. However, this proposal offers the following general
+best-practices for protocol designers:
+
+1. Serializing the default zero value of a member marked with the `@default`
+   trait can lead to unintended information disclosure. For example, consider
+   a newly introduced structure member marked with the `@default` trait that is
+   only exposed to customers of a service that are allowlisted into a private
+   beta. Serializing the zero value these members could expose the feature to
```
JordonPhillips, line 539:
```suggestion
   beta. Serializing the zero values of these members could expose the feature to
```

designs/default-zero-value.md
JordonPhillips: If nothing else, I know the current generators allow for pre-processing models. They could just strip out the required/default traits if the clientOptional trait is present.

JordonPhillips (changes requested):

designs/default-zero-value.md
mtdowling: How so?

I don't think so because you have to opt-in to `@default`, and you can only opt-in to `@default` on a member that targets a structure if the structure has all optional members.

designs/default-zero-value.md
```diff
@@ -0,0 +1,857 @@
...
+Is exactly equivalent to the following structure:
+
+```
+structure Message {
+    @default
+    title: String
+}
+```
+
+This pairs with the `@required` trait syntactic sugar that is also being added
```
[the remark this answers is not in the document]
mtdowling: That looks good to me. It's nice because it's on the same line as the member, and these important traits have a special place to look for them.

Alternatively, I'm also open to just removing the syntactic sugar of both `!` and `= default` -- I worry a little that adding too much sugar will end up making it harder to use.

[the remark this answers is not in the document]
mtdowling: Let's just yank this out. We can add sugar later if we feel we really need it.

kstich (approved):

PR #924, opened 2021-10-01
Add rest-json protocol test updates

nateprewitt:
These changes are to help codify current Smithy behavior and ensure all generated clients are conformant with expectations for the rest-json protocol. Each test should have an explanation of what it's trying to achieve and why that's relevant to clients generated to Smithy.

Added checks in `http-content-type.smithy`:
* Validating what an empty body looks like when handling structured and unstructured input.
* Validating correct Content-Type headers are sent with payloads and can be overwritten with members targeting an HTTP header.
* Validating methods that should not serialize a body are correctly omitting relevant headers.

From initial testing, Smithy clients appear to already follow these requirements but they aren't formally specified in any single test suite. Some of these checks are redundant with tests from `http-payload.smithy` and `http-headers.smithy`, but have been added to this file to reduce cognitive load when trying to understand how a client behaves in these cases.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
@@ -1,0 +1,293 @@
+// This file defines test cases that test both the HTTP body
+// and content-type handling.
+
+$version: "1.0"
+
+namespace aws.protocoltests.restjson
+
+use aws.protocols#restJson1
+use aws.protocoltests.shared#TextPlainBlob
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+
+/// This example operation serializes a structure in the HTTP body.
+///
+/// It should ensure Content-Type: application/json is
+/// used in all requests and that an "empty" body is
+/// an empty JSON document ({}).
+///
+@idempotent
+@http(uri: "/body", method: "POST")
+operation TestBodyStructure {
+    input: TestBodyStructureInputOutput,
+    output: TestBodyStructureInputOutput
+}
+
+apply TestBodyStructure @httpRequestTests([
+    {
+        id: "RestJsonTestBodyStructure",
+        documentation: "Serializes a structure",
+        protocol: restJson1,
+        method: "POST",
+        uri: "/body",
+        body: """
+              {"testConfig":
+                  {"timeout": 10}
+              }""",
+        bodyMediaType: "application/json",
+        headers: {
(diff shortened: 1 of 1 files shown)


jasdel:
I think the `main.smithy` file also needs to be updated to include the new `TestBodyStructure` operation in the service's set of operations.

https://github.com/awslabs/smithy/blob/f60c17e9b85181d868a55cba8536ea3ac240ac13/smithy-aws-protocol-tests/model/restJson1/main.smithy#L19

In addition, I think that some of these tests duplicate, or are very similar to, some of the existing `HttpPayload*`  operation tests. If these are all new maybe it would be better to be named similar to `HttpPayloadContent`, and grouped with the other `@httpPayload` tests.

smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
```diff
@@ -0,0 +1,293 @@
...
+        documentation: "Serializes a GET request with no modeled body",
+        protocol: restJson1,
+        method: "GET",
+        uri: "/no_payload",
+        body: "",
+        headers: {
+        },
+        requireHeaders: [
+            "Content-Length",
+            "Content-Type"
```
adamthom-amzn, line 263: Why is `Content-Type` required here?

nateprewitt: These TestNoPayload tests should both be using `forbidHeaders` not `requireHeaders`.

adamthom-amzn (approved):

smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
rcoh: wait, it `Content-Length` actually forbidden in this case? Why can't `Content-Length: 0` be sent here?

nateprewitt: We're being slightly overly prescriptive here but this is the behavior we're intending for the SDKs. 

>A user agent SHOULD NOT send a Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a body.
   
([RFC 7230 § 3.3.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2))
   
   This was left as SHOULD NOT because there are a few semi-common implementations with this behavior. It's broadly considered incorrect though and the goal is for new clients we're creating to do the right thing by default.

PR #931, opened 2021-10-08
Add inclusive words filter

DavidOgunsAWS:
*Issue #, if available:*
* Adding validator to catch usage of non-inclusive words in the text of a Smithy model

*Description of changes:*
* Since the list of non-inclusive words is statically small, those are hard coded
* The full text scanner internal to the validator can be abstracted further so a single pass over the model runs multiple validators more efficiently.

*TODO:*
Testing so far has been local. Follow up commit will add checked-in unit tests

### DIFF
--- a/docs/source/1.0/guides/model-linters.rst
+++ b/docs/source/1.0/guides/model-linters.rst
@@ -128,6 +128,73 @@ Example:
     ]
 
 
+.. _NoninclusiveTerms:
+
+NoninclusiveTerms
+=================
+
+Validates that all text content in a model (i.e. shape names, member names,
+documentation, trait values, etc.) does not contain words that perpetuate cultural
+biases. This validator has a built-in set of bias terms that are commonly found
+in APIs along with suggested alternatives.
+
+Noninclusive terms are case-insensitively substring matched and can have any
+number of leading or trailing whitespace or non-whitespace characters.
+
+This validator has built-in mappings from noninclusive terms to match model
+text to suggested alternatives. The configuration allows for additional terms
+to suggestions mappings to either override or append the built-in mappings. If
+a match occurs and the suggested alternatives is empty, no suggestion is made
+in the generated warning message.
+
+Rationale
+    Intent doesn't always match impact. The use of noninclusive language like
+    "whitelist" and "blacklist" perpetuates bias through past association of
+    acceptance and denial based on skin color. Other words should be used that
+    are not only inclusive, but more clearly communicate meaning. Words like
+    allowList and denyList much more clearly indicate that something is
+    allowed or denied.
+
+Default severity
+    ``WARNING``
+
+Configuration
+    .. list-table::
+       :header-rows: 1
+       :widths: 20 20 60
+
+       * - Property
+         - Type
+         - Description
+       * - terms
+         - { ``keyword`` -> [ ``alternatives`` ] }
+         - A set of noninclusive terms to suggestions to either override or replace
+           the built-in mappings. This property is not required unless
+           ``excludeDefaults`` is true. The default value is the empty set.
+       * - excludeDefaults
+         - ``boolean``
+         - A flag indicating whether or not the mappings set specified by ``terms``
+           configuration replaces the built-in set or appends additional mappings.
+           This property is not required and defaults to false.
+
+Example:
+
+.. code-block:: smithy
+
+    $version: "1.0"
+
+    metadata validators = [{
+        name: "NoninclusiveTerms"
+        configuration: {
+            excludeDefaults: false,
+            terms: {
+                mankind: ["humankind"],
+                mailman: ["mail carrier", "postal worker"]
+            }
+        }
+    }]
+
+
 .. _ReservedWords:
 
 ReservedWords
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.linters;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.knowledge.TextIndex;
+import software.amazon.smithy.model.knowledge.TextInstance;
+import software.amazon.smithy.model.node.NodeMapper;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationUtils;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.MapUtils;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
+ * <p>Validates that all shape names and values do not contain non-inclusive terms.
+ */
+public final class NoninclusiveTermsValidator extends AbstractValidator {
+    static final Map<String, List<String>> BUILT_IN_NONINCLUSIVE_TERMS = MapUtils.of(
+            "master", ListUtils.of("primary", "parent", "main"),
+            "slave", ListUtils.of("secondary", "replica", "clone", "child"),
+            "blacklist", ListUtils.of("denyList"),
+            "whitelist", ListUtils.of("allowList")
+        );
+
+    public static final class Provider extends ValidatorService.Provider {
+        public Provider() {
+            super(NoninclusiveTermsValidator.class, node -> {
+                NodeMapper mapper = new NodeMapper();
+                return new NoninclusiveTermsValidator(
+                        mapper.deserialize(node, NoninclusiveTermsValidator.Config.class));
+            });
+        }
+    }
+
+    /**
+     * NoninclusiveTermsValidator configuration.
+     */
+    public static final class Config {
+        private Map<String, List<String>> terms = MapUtils.of();
+        private boolean excludeDefaults;
+
+        public Map<String, List<String>> getTerms() {
+            return terms;
+        }
+
+        public void setTerms(Map<String, List<String>> terms) {
+            this.terms = terms;
+        }
+
+        public boolean getExcludeDefaults() {
+            return excludeDefaults;
+        }
+
+        public void setExcludeDefaults(boolean excludeDefaults) {
+            this.excludeDefaults = excludeDefaults;
+        }
+    }
+
+    private final Map<String, List<String>> termsMap;
+
+    private NoninclusiveTermsValidator(Config config) {
+        Map<String, List<String>> termsMapInit = new HashMap<>(BUILT_IN_NONINCLUSIVE_TERMS);
+        if (!config.getExcludeDefaults()) {
+            termsMapInit.putAll(config.getTerms());
+            termsMap = Collections.unmodifiableMap(termsMapInit);
+        } else {
+            if (config.getTerms().isEmpty()) {
+                //This configuration combination makes the validator a no-op.
+                throw new IllegalArgumentException("Cannot set 'excludeDefaults' to true and leave "
+                                                 + "'terms' empty or unspecified.");
+            }
+            termsMap = Collections.unmodifiableMap(config.getTerms());
+        }
+    }
+
+    /**
+     * Runs a full text scan on a given model and stores the resulting TextOccurrences objects.
+     *
+     * Namespaces are checked against a global set per model.
+     *
+     * @param model Model to validate.
+     * @return a list of ValidationEvents found by the implementer of getValidationEvents per the
+     *          TextOccurrences provided by this traversal.
+     */
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        TextIndex textIndex = TextIndex.of(model);
+        List<ValidationEvent> validationEvents = new ArrayList<>();
+        for (TextInstance text : textIndex.getTextInstances()) {
+            validationEvents.addAll(getValidationEvents(text));
+        }
+        return validationEvents;
+    }
+
+    /**
+     * Generates zero or more @see ValidationEvents and returns them in a collection.
+     *
+     * @param occurrence text occurrence found in the body of the model
+     */
+    private Collection<ValidationEvent> getValidationEvents(TextInstance instance) {
+        final Collection<ValidationEvent> events = new ArrayList<>();
+        for (Map.Entry<String, List<String>> termEntry : termsMap.entrySet()) {
+            final String termLower = termEntry.getKey().toLowerCase();
+            final int startIndex = instance.getText().toLowerCase().indexOf(termLower);
+            if (startIndex != -1) {
+                final String matchedText = instance.getText().substring(startIndex, startIndex + termLower.length());
+                switch (instance.getLocationType()) {
+                    case NAMESPACE:
+                        //Cannot use any warning() overloads because there is no shape associated with the event.
+                        events.add(ValidationEvent.builder()
+                                .sourceLocation(SourceLocation.none())
+                                .id(this.getClass().getSimpleName().replaceFirst("Validator$", ""))
+                                .severity(Severity.WARNING)
+                                .message(formatNonInclusiveTermsValidationMessage(termEntry, matchedText, instance))
+                                .build());
+                        break;
+                    case APPLIED_TRAIT:
+                        events.add(warning(instance.getShape(),
+                                instance.getTrait().getSourceLocation(),
+                                formatNonInclusiveTermsValidationMessage(termEntry, matchedText, instance)));
+                        break;
+                    case SHAPE:
+                    default:
+                        events.add(warning(instance.getShape(),
+                                instance.getShape().getSourceLocation(),
+                                formatNonInclusiveTermsValidationMessage(termEntry, matchedText, instance)));
+                }
+            }
+        }
+        return events;
+    }
+
+    private static String formatNonInclusiveTermsValidationMessage(
+            Map.Entry<String, List<String>> termEntry,
+            String matchedText,
+            TextInstance instance
+    ) {
+        final List<String> caseCorrectedEntryValue = termEntry.getValue().stream()
+            .map(replacement -> Character.isUpperCase(matchedText.charAt(0))
+                  ? StringUtils.capitalize(replacement)
+                  : StringUtils.uncapitalize(replacement))
+            .collect(Collectors.toList());
+        String replacementAddendum = !termEntry.getValue().isEmpty()
+                ? String.format(" Consider using one of the following terms instead: %s",
+                    ValidationUtils.tickedList(caseCorrectedEntryValue))
+                : "";
+        switch (instance.getLocationType()) {
+            case SHAPE:
+                return String.format("%s shape uses a non-inclusive term `%s`.%s",
+                        StringUtils.capitalize(instance.getShape().getType().toString()),
+                        matchedText, replacementAddendum);
+            case NAMESPACE:
+                return String.format("%s namespace uses a non-inclusive term `%s`.%s",
+                        instance.getText(), matchedText, replacementAddendum);
+            case APPLIED_TRAIT:
+                if (instance.getTraitPropertyPath().isEmpty()) {
+                    return String.format("'%s' trait has a value that contains a non-inclusive term `%s`.%s",
+                            Trait.getIdiomaticTraitName(instance.getTrait()), matchedText,
+                            replacementAddendum);
+                } else {
+                    String valuePropertyPathFormatted = formatPropertyPath(instance.getTraitPropertyPath());
+                    return String.format("'%s' trait value at path {%s} contains a non-inclusive term `%s`.%s",
+                            Trait.getIdiomaticTraitName(instance.getTrait()), valuePropertyPathFormatted,
+                            matchedText, replacementAddendum);
+                }
+            default:
+                throw new IllegalStateException();
+        }
+    }
+
+    private static String formatPropertyPath(List<String> traitPropertyPath) {
+        return String.join("/", traitPropertyPath);
+    }
+}
--- a/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.model.knowledge;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.loader.Prelude;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.ReferencesTrait;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.validators.TraitValueValidator;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+/**
+ * Index containing the full set of {@link TextInstance}s associated with a model.
+ */
+@SmithyUnstableApi
+public final class TextIndex implements KnowledgeIndex {
+    private final List<TextInstance> textInstanceList = new ArrayList<>();
+
+    public TextIndex(Model model) {
+        Set<String> visitedNamespaces = new HashSet<>();
+        // Validating the prelude is a feature for internal-only Smithy development
+        Node validatePreludeNode = model.getMetadata().get(TraitValueValidator.VALIDATE_PRELUDE);
+        boolean validatePrelude = validatePreludeNode != null
+                ? validatePreludeNode.expectBooleanNode().getValue()
+                : false;
+
+        for (final Shape shape : model.toSet()) {
+            if (validatePrelude || !Prelude.isPreludeShape(shape)) {
+                if (visitedNamespaces.add(shape.getId().getNamespace())) {
+                    textInstanceList.add(TextInstance.createNamespaceText(shape.getId().getNamespace()));
+                }
+                computeShapeTextInstances(shape, textInstanceList, model);
+            }
+        }
+    }
+
+    public static TextIndex of(Model model) {
+        return model.getKnowledge(TextIndex.class, TextIndex::new);
+    }
+
+    public Collection<TextInstance> getTextInstances() {
+        return Collections.unmodifiableList(textInstanceList);
+    }
+
+    private static void computeShapeTextInstances(
+            Shape shape,
+            Collection<TextInstance> textInstances,
+            Model model
+    ) {
+        textInstances.add(TextInstance.createShapeInstance(shape));
+
+        for (Trait trait : shape.getAllTraits().values()) {
+            Shape traitShape = model.expectShape(trait.toShapeId());
+            computeTextInstancesForAppliedTrait(trait.toNode(), trait, shape, textInstances,
+                    new ArrayDeque<>(), model, traitShape);
+        }
+    }
+
+    private static void computeTextInstancesForAppliedTrait(
+            Node node,
+            Trait trait,
+            Shape parentShape,
+            Collection<TextInstance> textInstances,
+            Deque<String> propertyPath,
+            Model model,
+            Shape currentTraitPropertyShape
+    ) {
+        if (trait.toShapeId().equals(ReferencesTrait.ID)) {
+            //Skip ReferenceTrait because it is referring to other shape names already being checked
+        } else if (node.isStringNode()) {
+            textInstances.add(TextInstance.createTraitInstance(
+                node.expectStringNode().getValue(), parentShape, trait, propertyPath));
+        } else if (node.isObjectNode()) {
+            ObjectNode objectNode = node.expectObjectNode();
+            objectNode.getStringMap().entrySet().forEach(memberEntry -> {
+                propertyPath.offerLast(memberEntry.getKey());
+                Shape memberTypeShape = getChildMemberShapeType(memberEntry.getKey(),
+                        model, currentTraitPropertyShape);
+                if (memberTypeShape == null) {
+                    //This means the "property" key value isn't modeled in the trait's structure/shape definition
+                    //and this text instance is unique
+                    textInstances.add(TextInstance.createTraitInstance(
+                        memberEntry.getKey(), parentShape, trait, propertyPath));
+                }
+                computeTextInstancesForAppliedTrait(memberEntry.getValue(), trait, parentShape, textInstances,
+                        propertyPath, model, memberTypeShape);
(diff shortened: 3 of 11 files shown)


mtdowling (changes requested):

smithy-linters/src/main/java/software/amazon/smithy/linters/AbstractModelTextValidator.java
```diff
@@ -0,0 +1,72 @@
...
+ * <p>Base class for any validator that wants to perform a full text search on a Model.
+ *
+ * <p>Does a full scan of the model and gathers all text along with location data
+ * and uses a generic function to decide what to do with each text occurrence. The
+ * full text search traversal has no knowledge of what it is looking for and descends
+ * fully into the structure of all traits.
+ *
+ * <p>Prelude shape definitions are not examined, however all values
+ */
+abstract class AbstractModelTextValidator extends AbstractValidator {
```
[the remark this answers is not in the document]
DavidOgunsAWS: I don't feel too strongly about keeping it. Can always introduce the abstraction later if useful.

smithy-linters/src/main/java/software/amazon/smithy/linters/AbstractModelTextValidator.java
```diff
@@ -0,0 +1,292 @@
...
+                        propertyPath, namespaces, model, memberTypeShape);
+                propertyPath.pop();
+                ++index[0];
+            });
+        }
+    }
+
+    private static Shape getChildMemberShapeType(String memberKey, Model model, Shape fromShape) {
+        if (fromShape != null) {
+            Shape childShape = null;
```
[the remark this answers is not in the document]
DavidOgunsAWS: Looking at the latest state of the code, I've made this change.

smithy-linters/src/main/java/software/amazon/smithy/linters/AbstractModelTextValidator.java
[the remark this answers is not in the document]
DavidOgunsAWS: Marked

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,153 @@
...
+import software.amazon.smithy.model.loader.Prelude;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.ReferencesTrait;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.validators.TraitValueValidator;
+
+public final class TextIndex implements KnowledgeIndex {
```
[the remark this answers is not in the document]
DavidOgunsAWS: Marked

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,153 @@
...
+        return textInstanceList;
+    }
+
+    private static void getTextInstances(Shape shape,
+                                           Collection<TextInstance> textInstances,
+                                           Model model) {
+        TextInstance.Builder builder = TextInstance.builder()
+                .locationType(TextInstance.TextLocation.SHAPE)
+                .shape(shape);
+        if (shape.isMemberShape()) {
```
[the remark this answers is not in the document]
DavidOgunsAWS: Adjusted

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+/**
+ * <p>Validates that all shape names, and values does not contain non-inclusive terms.
+ * *
+ * <p>See AbstractModelTextValidator for scan implementation details.
+ */
+public final class NoninclusiveTermsValidator extends AbstractValidator {
+    static final Map<String, List<String>> BUILT_IN_NONINCLUSIVE_TERMS = MapUtils.of(
+            "master", ListUtils.of("primary", "parent", "main"),
+            "slave", ListUtils.of("secondary", "replica", "clone", "child"),
+            "blacklist", ListUtils.of("denylist"),
```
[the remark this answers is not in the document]
DavidOgunsAWS: When used as a made up compound verb, the lowercase always fits better. But allowList and denyList are not likely to be used as a verb and more as a noun. In that case camel case on the word barrier is probably better for most API shape names.

I'll look into carrying through the same initial match letter case through to the suggestion.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationUtils;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.MapUtils;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
+ * <p>Validates that all shape names, and values does not contain non-inclusive terms.
```
kstich, line 42: a/names, and values does not/names and values do not

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+ * *
+ * <p>See AbstractModelTextValidator for scan implementation details.
```
kstich, line 44: This class doesn't exist anymore.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    private static boolean containsTerm(String text, String term) {
+        return text.toLowerCase().contains(term.toLowerCase());
+    }
+
+    private static String formatNonInclusiveTermsValidationMessage(Map.Entry<String, List<String>> termEntry,
+                                                                   TextInstance instance) {
+        String replacementAddendum = termEntry.getValue().size() > 0
```
kstich, line 164: Can use `!termEntry.getValue().isEmpty()` instead.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+     * @param model Model to validate.
+     * @return a list of ValidationEvents found by the implementer of getValidationEvents per the
+     *          TextOccurrences provided by this traversal.
+     */
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        TextIndex textIndex = TextIndex.of(model);
+        List<ValidationEvent> validationEvents = new ArrayList<>();
+        for (TextInstance text : textIndex.getTextInstances()) {
+            getValidationEvents(text, validationEvent -> {
```
kstich, line 115: Instead of using an anonymous consumer here, this should be refactored towards `validationEvents.addAll(getValidationEvents(text))`.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * This is a somewhat abstracted interface for validating text instances in a model
+     * by pushing ValidationEvents into a provided consumer.
+*
+     * @param occurrence text occurrence found in the body of the model
+     * @param validationEventConsumer consumer to push ValidationEvents into
+     */
+    private void getValidationEvents(TextInstance instance,
```
kstich, line 129: re: Above refactoring comment, this would return a list of events like other validators instead of utilizing a `Consumer`.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+                                       Consumer<ValidationEvent> validationEventConsumer) {
+        for (Map.Entry<String, List<String>> termEntry : termsMap.entrySet()) {
+            //lower casing the term will be more necessary when the terms are from config
```
kstich, line 132: Comments should be full sentences, including capitalization and punctuation.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+            if (containsTerm(instance.getText(), termEntry.getKey())) {
+                switch (instance.getLocationType()) {
+                    case NAMESPACE:
+                        validationEventConsumer.accept(ValidationEvent.builder()
```
kstich, line 136: If this can't just use `warning` (since a namespace doesn't have a singular Shape to pass to it?) it's probably worth documenting.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+        public boolean isAppendDefaults() {
+            return appendDefaults;
+        }
+
+        public void setAppendDefaults(boolean appendDefaults) {
+            this.appendDefaults = appendDefaults;
+        }
+    }
+
+    final Map<String, List<String>> termsMap;
```
kstich, line 88: Should probably be `private` and initialized here, as that allows cleanup in the constructor as well.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private NoninclusiveTermsValidator(Config config) {
+        if (config.isAppendDefaults()) {
```
kstich, line 91: This gets simpler if the map is initialized for the validator:

```
termsMap.putAll(config.getNonInclusiveTerms());
if (config.isAppendDefaults()) {
    termsMap.putAll(BUILT_IN_NONINCLUSIVE_TERMS);
}

File: smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java [hunk 15]
```diff
@@ -0,0 +1,151 @@
...
+            textInstances.add(TextInstance.builder()
+                    .locationType(TextInstance.TextLocation.APPLIED_TRAIT)
+                    .shape(parentShape)
+                    .trait(trait)
+                    .traitPropertyPath(propertyPath)
+                    .text(node.expectStringNode().getValue())
+                    .build());
+        } else if (node.isObjectNode()) {
+            ObjectNode objectNode = node.expectObjectNode();
+            objectNode.getMembers().entrySet().forEach(memberEntry -> {
```
kstich, line 108: This can use `getStringMap()` instead of `getMembers()` to remove the repeated `.getKey().getValue()` calls.

kstich (changes requested):

docs/source/1.0/guides/model-linters.rst
```diff
@@ -128,6 +128,68 @@ Example:
...
+
+Default severity
+    ``WARNING``
+
+Configuration
+    .. list-table::
+       :header-rows: 1
+       :widths: 20 20 60
+
+       * - Property
```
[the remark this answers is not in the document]
DavidOgunsAWS: Will add documentation. It is worth noting there is a non-trivial, maybe counter intuitive defaults behavior:

Though `appendDefaults` defaults to false, if `noninclusiveTerms` mappings is entirely unset or empty, `appendDefaults` behaves as if it were true -- the built in mappings are present. `noninclusiveTerms` has to be non-empty before `appendDefaults` behavior applies. If this behavior is acceptable, then I'll focus on clear and concise documentation for it. If not, then I should change the implementation.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static TextIndex of(Model model) {
+        return model.getKnowledge(TextIndex.class, TextIndex::new);
+    }
+
+    public Collection<TextInstance> getTextInstances() {
+        return textInstanceList;
+    }
+
+    private static void getTextInstances(Shape shape,
```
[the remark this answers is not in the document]
DavidOgunsAWS: I agree it doesn't need to be static, but given the computation only operates on passed in parameters it seems like a proper descriptor even if unnecessary. I think it makes it easier to see what data access the method uses and easier to potentially refactor.

I like the rename though.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,151 @@
...
+        });
+
+        for (String namespace : visitedNamespaces) {
+            textInstanceList.add(TextInstance.builder()
+                    .locationType(TextInstance.TextLocation.NAMESPACE)
+                    .text(namespace)
+                    .build());
+        }
+
+        //no point in allowing the list to change
```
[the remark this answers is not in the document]
DavidOgunsAWS: It's highly preferred given the primary interface to get the Collection&lt;TextInstance&gt; does not create a defensive copy.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,151 @@
...
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.validators.TraitValueValidator;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+@SmithyUnstableApi
+public final class TextIndex implements KnowledgeIndex {
+    private List<TextInstance> textInstanceList;
+
+    public TextIndex(Model model) {
+        textInstanceList = new ArrayList<>();
```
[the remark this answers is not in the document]
DavidOgunsAWS: At the end of this method `textInstanceList` is re-assigned to make the list un-modifiable after it is computed. This initial initialization is for local method usage to build the computation. I could have used a separate local variable to make it more obvious and assigned to `textInstanceList` once.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,170 @@
...
+        APPLIED_TRAIT,
+        NAMESPACE
+    }
+
+    public enum PathElementType {
+        KEY,
+        ARRAY_INDEX
+    }
+
+    public static final class PathElement {
```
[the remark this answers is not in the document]
DavidOgunsAWS: I agree that it's the granularity of the location tracking inside trait values that adds substantial complexity to this work. I think it has value though.

If a filtered term is in an `HttpTrait`'s `uri` property value, it would direct the user to the `uri` property. That doesn't seem too valuable since the `code` or `method` properties of `HttpTrait` would not likely have a noninclusive term. However, in making no assumptions on other traits defined outside of the prelude, there could be more than one property with more free form string data, and it could be nested further.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,151 @@
...
+        Set<String> visitedNamespaces = new HashSet<>();
+        Node validatePreludeNode = model.getMetadata().get(TraitValueValidator.VALIDATE_PRELUDE);
+        boolean validatePrelude = validatePreludeNode != null
+                ? validatePreludeNode.expectBooleanNode().getValue()
+                : false;
+
+        model.shapes().filter(shape -> validatePrelude || !Prelude.isPreludeShape(shape)).forEach(shape -> {
+            visitedNamespaces.add(shape.getId().getNamespace());
```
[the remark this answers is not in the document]
DavidOgunsAWS: `visitedNamespaces` is a set so it's necessary to avoid generating validation events multiple times for the same namespace. I can perform the de-dup of namespaces while adding unique ones to be validated into the same loop.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private static void getTextInstances(Shape shape,
+                                           Collection<TextInstance> textInstances,
+                                           Model model) {
+        TextInstance.Builder builder = TextInstance.builder()
+                .locationType(TextInstance.TextLocation.SHAPE)
+                .shape(shape);
+        builder.text(shape.getId().getMember().orElseGet(() -> shape.getId().getName()));
+        textInstances.add(builder.build());
```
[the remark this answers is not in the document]
DavidOgunsAWS: Strange. I already have it as one statement.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
DavidOgunsAWS: Great catch!

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
DavidOgunsAWS: Done

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
DavidOgunsAWS: Fixed

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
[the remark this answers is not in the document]
DavidOgunsAWS: Went back and adjusted `getTextInstances()` to return an unmodifiable list so I could made this adjustment.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
DavidOgunsAWS: Removed comment

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
DavidOgunsAWS: Refactored a bit

docs/source/1.0/guides/model-linters.rst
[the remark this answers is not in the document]
DavidOgunsAWS: Property structure have changed a bit. But current properties are documented for required or not, along with what the default values are

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,127 @@
...
+            this.shape = shape;
+            return this;
+        }
+
+        public Builder trait(Trait trait) {
+            this.trait = trait;
+            return this;
+        }
+
+        public Builder traitPropertyPath(Deque<String> traitPropertyPath) {
```
mtdowling, line 100: Why a Deque?

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,127 @@
...
+import software.amazon.smithy.utils.ListUtils;
+
+public final class TextInstance {
+    private final TextLocation locationType;
+    private final String text;
+    private final Shape shape;
+    private final Trait trait;
+    private final List<String> traitPropertyPath;
+
+    private TextInstance(Builder builder) {
```
mtdowling, line 34: I wonder if we even need a builder here -- for example, could we use named factory methods? This would capture each invariant more explicitly.

```
TextInstance createNamespaceText(String namespace);
TextInstance createShapeInstance(Shape shape, anythingElse?);
TextInstance createTraitInstance(Shape shape, Trait trait, Collection<String> traitPath);
```

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,145 @@
...
+    private final List<TextInstance> textInstanceList = new ArrayList<>();
+
+    public TextIndex(Model model) {
+        Set<String> visitedNamespaces = new HashSet<>();
+        Node validatePreludeNode = model.getMetadata().get(TraitValueValidator.VALIDATE_PRELUDE);
+        boolean validatePrelude = validatePreludeNode != null
+                ? validatePreludeNode.expectBooleanNode().getValue()
+                : false;
+
+        model.shapes().filter(shape -> validatePrelude || !Prelude.isPreludeShape(shape)).forEach(shape -> {
```
mtdowling, line 48: nit: I think this would probably be more readable as a normal for loop.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,145 @@
...
+import software.amazon.smithy.model.validation.validators.TraitValueValidator;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+@SmithyUnstableApi
+public final class TextIndex implements KnowledgeIndex {
+    private final List<TextInstance> textInstanceList = new ArrayList<>();
+
+    public TextIndex(Model model) {
+        Set<String> visitedNamespaces = new HashSet<>();
+        Node validatePreludeNode = model.getMetadata().get(TraitValueValidator.VALIDATE_PRELUDE);
```
mtdowling, line 43: It might be nice to add a comment that this is an internal-only feature used to validate the prelude when building Smithy and isn't a general-purpose feature we're exposing.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
DavidOgunsAWS: Deque is the structure used while building up the path. I didn't want to have the caller reverse that structure into a List, only to have that List again defensively copied by this builder.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
DavidOgunsAWS: Builders looked be the more common pattern in the rest of the code. After a second look, I see some factory creation methods as well, but most of those are using Builders right underneath. How about having both? Builder exists, along with the factory methods on top?

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
DavidOgunsAWS: `model.shapes()` returns a stream and appears to be the only method that returns all of the shapes. I don't think it would be worth extracting from the stream into a collection just to change the syntax here.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
DavidOgunsAWS: Done

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
mtdowling: You can use `model.toSet`. It's a very cheap method to call.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,89 @@
...
+            final String text,
+            final Shape shape,
+            final Trait trait,
+            final Deque<String> traitPropertyPath) {
+        this.locationType = locationType;
+        this.text = text;
+        this.shape = shape;
+        this.trait = trait;
+        this.traitPropertyPath = traitPropertyPath != null
+            ? traitPropertyPath.stream().collect(ListUtils.toUnmodifiableList())
```
mtdowling, line 42: nit - this is going to create hundreds of thousands of streams for a super simple creation. Can't you do something like: `Collections.unmodifiableList(new ArrayList<>(traitPropertyPath))`

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,131 @@
...
+        textInstances.add(TextInstance.createShapeInstance(shape));
+
+        for (Trait trait : shape.getAllTraits().values()) {
+            Shape traitShape = model.expectShape(trait.toShapeId());
+            computeTextInstancesForAppliedTrait(trait.toNode(), trait, shape, textInstances,
+                    new ArrayDeque<>(), model, traitShape);
+        }
+    }
+
+    private static void computeTextInstancesForAppliedTrait(Node node,
```
mtdowling, line 79: nit - the formatting for arguments is inconsistent in this file -- some things are on the same line others aren't

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,131 @@
...
+        // Validating the prelude is a feature for internal-only Smithy development
+        Node validatePreludeNode = model.getMetadata().get(TraitValueValidator.VALIDATE_PRELUDE);
+        boolean validatePrelude = validatePreludeNode != null
+                ? validatePreludeNode.expectBooleanNode().getValue()
+                : false;
+
+        for (final Shape shape : model.toSet()) {
+            if (validatePrelude || !Prelude.isPreludeShape(shape)) {
+                if (visitedNamespaces.add(shape.getId().getNamespace())) {
+                textInstanceList.add(TextInstance.createNamespaceText(shape.getId().getNamespace()));
```
mtdowling, line 52: nit - not indented

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
DavidOgunsAWS: Changed to `ListUtils.copyOf(traitPropertyPath)`

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
DavidOgunsAWS: Fixed

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
DavidOgunsAWS: Fixed

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,133 @@
...
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.ReferencesTrait;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.validators.TraitValueValidator;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+@SmithyUnstableApi
+public final class TextIndex implements KnowledgeIndex {
```
kstich, line 38: The class and public methods here should have javadoc added.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,90 @@
...
+package software.amazon.smithy.model.knowledge;
+
+import java.util.Deque;
+import java.util.List;
+import java.util.Objects;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.utils.ListUtils;
+
+public final class TextInstance {
```
kstich, line 25: The class and public methods here should have javadoc added.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,90 @@
...
+    private final Shape shape;
+    private final Trait trait;
+    private final List<String> traitPropertyPath;
+
+    private TextInstance(
+            final TextLocation locationType,
+            final String text,
+            final Shape shape,
+            final Trait trait,
+            final Deque<String> traitPropertyPath) {
```
kstich, line 37: nit: We usually put the `) {` for multi-line declarations on the next line, inline with the start of the declaration.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public Collection<TextInstance> getTextInstances() {
+        return Collections.unmodifiableList(textInstanceList);
+    }
+
+    private static void computeShapeTextInstances(
+            Shape shape,
+            Collection<TextInstance> textInstances,
+            Model model) {
```
kstich, line 70: nit: We usually put the `) {` for multi-line declarations on the next line, inline with the start of the declaration. (Others in the file as well.)

docs/source/1.0/guides/model-linters.rst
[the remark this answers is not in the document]
kstich: Is the reasoning for this structure change because you weren't happy with the behavior? I feel like a terms map and a defaults boolean that defaults to true would be pretty clear and usable for customers.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -0,0 +1,203 @@
...
+        final List<String> caseCorrectedEntryValue = termEntry.getValue().stream()
+            .map(replacement -> Character.isUpperCase(matchedText.charAt(0))
+                  ? StringUtils.capitalize(replacement)
+                  : StringUtils.uncapitalize(replacement))
+            .collect(Collectors.toList());
+        String replacementAddendum = !termEntry.getValue().isEmpty()
+                ? String.format(" Consider using one of the following terms instead: %s",
+                    ValidationUtils.tickedList(caseCorrectedEntryValue))
+                : "";
+        switch (instance.getLocationType()) {
```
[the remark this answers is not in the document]
DavidOgunsAWS: I see what you're saying. This whole thing breaks down to:

```
   builder.append("{textLocationSymbol}");
   builder.append("{locationType} {TypeDependentMessage}");
   builder.append("contains/uses a non-inclusive term {nonInclusiveTerm}.");
   builder.append("{replacementAddendum}");
```

It would create a top level section of a code that reads cleaner, but require considerably more code as I would need 3 switches to evaluate those components versus one which calls format() differently. I wouldn't want to put message formatting logic in TextInstance.TextLocationType methods to avoid the switch-case statements because I consider TextInstance.TextLocationType agnostic to this particular validator's event message formatting.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextIndex.java
DavidOgunsAWS: I added a class level comment. The public methods are the constructive and `of(Model model)` methods which other knowledge index classes do not have javadoc for.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,120 @@
...
+     * Retrieves the text content of the TextInstance.
+     */
+    public String getText() {
+        return text;
+    }
+
+    /**
+     * Gets the shape the text is associated with.
+     *
+     * Returns null if {@link TextLocationType} is NAMESPACE
```
kstich, line 88: Instead of returning `null`, we almost always use `Optional<Shape>` for the return type and use `Optional.ofNullable()` when returning. Applies to `getTrait` as well.

Also, the javadoc should use `@return` to describe return values so they're properly integrated with the generated documentation.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/TextInstance.java
```diff
@@ -0,0 +1,120 @@
...
+     * Returns null if {@link TextLocationType} is NAMESPACE or SHAPE
+     */
+    public Trait getTrait() {
+        return trait;
+    }
+
+    /**
+     * Gets the ordered path components within a trait's value the text is associated with.
+     *
+     * Returns null if {@link TextLocationType} is NAMESPACE or SHAPE
```
kstich, line 106: For collection and map return values, we usually return an empty list via `ListUtils.of()` instead of null. It's probably the right call along with a note that this method will only provide useful information for `APPLIED_TRAIT`.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -62,44 +62,40 @@ public Provider() {
      * NoninclusiveTermsValidator configuration.
      */
     public static final class Config {
-        private Map<String, List<String>> appendNoninclusiveTerms = MapUtils.of();
-        private Map<String, List<String>> replaceNoninclusiveTerms = MapUtils.of();
+        private Map<String, List<String>> terms = MapUtils.of();
+        private Boolean excludeDefaults = Boolean.FALSE;
```
kstich, line 66: This can just be a primitive `boolean`, since the null state isn't relevant.

DavidOgunsAWS: Fixed

kstich (approved):
Let's squash this PR on merge if it's viable.

PR #937, opened 2021-10-12
Add 'required' CFN schema property

### LINKED ISSUE #751: smithy-aws-cloudformation-traits does not mark top-level properties as required
The CloudFormation Resource type schema https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html allows to define properties required. Smithy currently does not mark top-level non-optional properties in the `required` section:

```
{
   ...
    "properties": {
         "propertyName": {
            "description": "string",
            "type": "string",
             . . . 
        },
 
        },
    },
    "required": [   // not added in Smithy
        "propertyName"
    ],
   ...
```

kstich:
Required members of create and put inputs are marked as required in the output schemas.

*Issue #, if available:*
Resolves #751 

### DIFF
--- a/smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/mappers/RequiredMapperTest.java
+++ b/smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/mappers/RequiredMapperTest.java
@@ -17,18 +17,22 @@ package software.amazon.smithy.aws.cloudformation.schema.fromsmithy.mappers;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
 
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import software.amazon.smithy.aws.cloudformation.schema.CfnConfig;
 import software.amazon.smithy.aws.cloudformation.schema.fromsmithy.CfnConverter;
 import software.amazon.smithy.model.Model;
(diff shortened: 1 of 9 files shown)


smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/mappers/RequiredMapperTest.java
```diff
@@ -0,0 +1,74 @@
...
+        CfnConfig config = new CfnConfig();
+        config.setOrganizationName("Smithy");
+        config.setService(ShapeId.from("smithy.example#TestService"));
+
+        ObjectNode resourceNode = CfnConverter.create()
+                .config(config)
+                .convertToNodes(model)
+                .get("Smithy::TestService::FooResource");
+
+        Assertions.assertEquals(ListUtils.of("/properties/FooRequiredProperty"),
```
prestomation, line 49: This is not the expected format for `required`. 

See the spec: https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html

According to the spec and the below examples, this list should contain the property names and not their path(no `/properties/`).

See this Nimble example: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-nimblestudio/blob/main/aws-nimblestudio-studio/aws-nimblestudio-studio.json#L68

or KMS key: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-kms/blob/master/key/aws-kms-key.json#L104

PR #943, opened 2021-10-18
Allow to define maven dependencies in smithy-build

Baccata:
Draft because no unit-tests yet. I'd ideally like for the smithy-maintainers to validate the approach before committing the time it'll take to add unit-tests. 

#### Changes 

Allows smithy-build to refer to (and fetch) maven dependencies. Custom repositories can be added, but by default fetches from maven-central. 

What might be controversial in this pull-request is the reliance on [coursier-interface](https://github.com/coursier/interface) to do so. [Coursier](https://get-coursier.io/) is a very efficient dependency resolver that the entirety of the scala ecosystem depends on. Coursier-interface is a pure-java interface for coursier, that shades coursier. 

The coursier-interface jar weighs about 4MB 

#### Rationale for this change

My interest mostly has to do with editor experience. Right now, the LSP client needs to instantiate the LSP server. If the opened smithy file depends on some library smithy model, it the LSP client takes the responsibility of pulling the dependencies itself, and calls `java` with the relevant classpath. The smithy model-assembler then [discover models](https://github.com/awslabs/smithy-language-server/blob/5804392edd8fc274201ee21887bdb9e6cde33223/src/main/java/software/amazon/smithy/lsp/SmithyInterface.java) from the classpath. 

Though this works, it's not great, because the dependencies referenced by the LSP client (say, vscode) need also to be referenced by whatever build tool is used to load the models and do something with them. Therefore there is a repetition of configuration between editor and build tool is not desirable, and create a discrepancy in terms of "what is the source of truth ?". 

From what I gather from [this PR](https://github.com/awslabs/smithy-language-server/pull/21), the longer-term vision would be for the LSP server to rely on smithy-build to allow for some granularity in how models rely to each other within a workspace. I think it makes sense, however my organisation's usecases are heavily reliant on sharing smithy models via jars stored in some artifact repository. 

So I think smithy-build configuration should be considered source of truth when it comes to declaring dependencies. The declarative configuration makes it easy for users to define dependencies and edit their files without having to interact with a higher-level build-tool, and those build-tools could leverage smithy-build directly without having to fiddle about passing jars dependencies to it. 

It'd open the door to making the editing and building experience very smooth. 

### DIFF
--- a/smithy-build/build.gradle
+++ b/smithy-build/build.gradle
@@ -24,4 +24,5 @@ ext {
 dependencies {
     api project(":smithy-utils")
     api project(":smithy-model")
+    implementation 'io.get-coursier:interface:1.0.4'
 }
--- a/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
@@ -15,6 +15,7 @@
 
 package software.amazon.smithy.build;
 
+import java.net.MalformedURLException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -255,6 +256,14 @@ final class SmithyBuildImpl {
         }
     }
 
+    private java.net.URL asUrl(String s) {
+      try {
+        return new java.net.URL(s);
+      } catch (MalformedURLException e) {
+        throw new SmithyBuildException(e);
+      }
+    }
+
     private Model createBaseModel() {
         Model resolvedModel = model;
 
@@ -262,6 +271,7 @@ final class SmithyBuildImpl {
             LOGGER.fine(() -> "Merging the following imports into the loaded model: " + config.getImports());
             ModelAssembler assembler = modelAssemblerSupplier.get().addModel(model);
             config.getImports().forEach(assembler::addImport);
+            config.getMavenImports().stream().map(this::asUrl).forEach(assembler::addImport);
             resolvedModel = assembler.assemble().unwrap();
         }
 
--- a/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
@@ -16,6 +16,8 @@
 package software.amazon.smithy.build;
 
 import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -39,6 +41,7 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import software.amazon.smithy.build.model.ProjectionConfig;
 import software.amazon.smithy.build.model.SmithyBuildConfig;
 import software.amazon.smithy.build.model.TransformConfig;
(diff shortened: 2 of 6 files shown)


mtdowling:
This is great, thanks for putting this up! It could help with both the LSP/editor experience, but also making the Smithy CLI more of a standalone tool that doesn't require Gradle for common tasks of just generating code from models (you're still probaby use Gradle or a build tool if you're developing custom traits, build plugins, and JARs though).

I like that the Coursier dependencies are shaded. I'm not too worried about the 4MB JAR.

The other thing about smithy-build.json files is that they also refer to build plugins that need to be present on the classpath. If we could wire those up through this mechanism too, then we could get a self-contained CLI and even make configuring Smithy's Gradle plugin easier since you'd no longer need to use buildScript dependencies. In order to pull this off, smithy-build would need to create a new thread with a custom ClassLoader based on the given classpath. It could send that classpath through Smithy model discovery as well to pick up models found in the classpath.

Can you show an example of what this would look like in smithy-build.json?

mtdowling:
I think we need to think through how this will be used... for example, I'd expect that when this is all used with the Smithy Gradle plugin, smithy-build and the Smithy CLI would rely solely on Gradle to fetch dependencies and manage running the Smithy CLI with the right classpath (either in a thread or when invoking the Smithy CLI in a process from Gradle). But when the CLI is used as a standalone tool, I'd expect it to use the data stored in smithy-build.json to download dependencies and run things in a Thread correctly using a custom ClassLoader.

How do you envision the IDE using this? Directly through smithy-build or through smithy-cli? If it's also through smithy-cli, then I think that convinces me to move the coursier stuff to the CLI, but leave the new getters/setters so they're just part of the smithy-build.json data model. Then smithy-cli could have some kind of flag where you can opt-out of fetching Maven dependencies. Opt-out seems like the right call, since tools like Gradle can easily set the opt-out flag with no inconvenience, whereas when actually using the CLI from the command line, you'd probably get annoyed if you had to constantly opt-in to it.

Baccata:
> I think we need to think through how this will be used... for example, I'd expect that when this is all used with the Smithy Gradle plugin, smithy-build and the Smithy CLI would rely solely on Gradle to fetch dependencies and manage running the Smithy CLI with the right classpath (either in a thread or when invoking the Smithy CLI in a process from Gradle). But when the CLI is used as a standalone tool, I'd expect it to use the data stored in smithy-build.json to download dependencies and run things in a Thread correctly using a custom ClassLoader.

> How do you envision the IDE using this? 

TBH I was hoping that smithy-build would be used as source of truth. That would unify the editor experience and the smithy-build use in the context of a build tool plugin. The IDE would call smithy-build directly (because why use a CLI when you can call directly java functions from java, it'd make implementation harder). 

That being said I can understand why you wouldn't want to commit to a design that forces artifact resolution on all downstream tooling. I'm happy to trim the PR down to the changes in SmithyConfig (getters/setters), and you guys could use the current state of this branch as a reference when comes the time to implement resolution in smithy-cli (or wherever you deem relevant). Having the configuration elements present in smithy-build would be enough to unlock some LSP work. 

Additionally, maybe an abstraction could be defined somewhere (possibly in smithy-model), and used where relevant : 

``` 
interface ArtifactResolver {
    List<URL> fetch(List<String> repositories, List<String> mavenDependencies); 
}
```

A coursier-based implementation in an opt-in module would be provided, to prevent a hard dependency. 

Obviously this wouldn't happen in this PR.

mtdowling:
There's a ton to think about here, so it's sadly going to take a little time to get back to you. I noticed that both Gradle and Maven can set various other properties on repos. One thing I think we probably need here is to make `mavenRepos` use a list of objects instead of strings so that we can add other repo configuration settings in the future if needed:

```
{
    ...
    "mavenRepos": [{"url": "xyz"}]
}
```

Baccata:
I'm closing this for now, we've implemented artifact resolution in our [fork of the LSP](https://github.com/disneystreaming/smithy-language-server), though we haven't documented it because we're terrible people 😅. 

Happy to revisit whenever, if you the smithy maintainers are interested.

PR #962, opened 2021-10-27
Add design for inline operation input/output shapes

JordonPhillips:

### DIFF
--- a/designs/inline-io.md
+++ b/designs/inline-io.md
@@ -115,25 +115,19 @@ operation GetUser {
 
 ```
 operation_statement =
-    "operation" ws identifier ws "{" *operation_body ws "}"
+    "operation" ws identifier ws inlineable_properties
 
-operation_body =
-    operation_input / operation_output / operation_errors
+inlineable_properties =
+    "{" *(inlineable_property ws) ws "}"
 
-operation_input =
-    "input" ws (io_reference / inline_structure)
-
-operation_output =
-    "output" ws (io_reference / inline_structure)
-
-io_reference =
-    ":" ws shape_id ws
+inlineable_property =
+    node_object_kvp / inline_structure
 
 inline_structure =
-    ":=" ws trait_statements [identifier] [mixins ws] "{" shape_members "}"
+    node_object_key ws ":=" ws inline_structure_value
 
-operation_errors =
-    "errors" ws ":" ws "[" *(ws shape_id) ws "]"
+inline_structure_value =
+    trait_statements [mixins ws] shape_members
 ```
 
 The following demonstrate customizing the suffixes.
--- a/designs/inline-io.md
+++ b/designs/inline-io.md
@@ -1,0 +1,207 @@
+# Inline Operation Inputs / Outputs
+
+This document describes a way to write input and output shapes as an inline
+part of an operation’s definition.
+
+## Motivation
+
+Operation input and output shapes are always structures, almost always have
+boilerplate names, and critically are almost never re-used. In some cases, they
+may not even have types generated for them. Because of those properties, the
+need to fully define them separately from an operation can feel like needless
+boilerplate. Additionally, separating them makes reading an operation at a high
+level more difficult since you have to jump around to get the information you
+need.
+
+## Proposal
+
+Operations will allow inlining input and output definitions, indicated by a
+walrus operator (`:=`). Inlined structures will deviate from normal
+structure definitions in two respects. Firstly, the structure keyword will be
+omitted. Additionally, providing a name for the inlined shape will be optional.
+
+### Walrus Operator
+
+Rather than using the same standalone colon (`:`) that is used in other
+cases, a walrus operator will be used to indicate an inline definition. The
+reason for this is to visually distinguish it at the outset, as well as to make
+tooling a bit cleaner to write.
+
+This usage of the operator is analogous to how some programming languages use
+it. For instance, Python uses it to assign a name to the result of an
+expression in places where initializing a variable was previously forbidden.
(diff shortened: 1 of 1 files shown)


designs/inline-io.md
```diff
@@ -0,0 +1,207 @@
...
+their IO shapes are almost never refrenced elsewhere and even if they were the
+default name makes it pretty clear where to look.
+
+There is at least one other place where this may make sense: resource
+identifiers. When defining a resource, you could use this syntax to define a
+structure that contains only the resource identifiers. This could be mixed in
+to other shapes in the model. This usage, while interesting, is out of scope
+for this document.
+
+### Can the suffix be customized?
```
mtdowling, line 198: Being able to customize the name of the input and output type would be something most _new_ models would never need to do or care about. However, teams migrating older models to Smithy _will_ want to maintain their existing conventions. For example, EC2 uses Request/Result I think. They wouldn't be able to use this feature, which would be a huge inconvenience for them.

I came up with a reasonable option: using a control statement. Control statements are per/file parser directives. Currently the only control statement we support is `$version`, but we can backward compatibly add other statements. We could, for example, add the following control statements:

```
$version: "2"
$operationInputSuffix: "Request"
$operationOutputSuffix: "Result"

namespace smithy.example

operation GetFoo {
    input:= {}
    output:= {}
}
```

The above would use the parser directives and create GetFooRequest and GetFooResult.

The drawback of this is that it's per/file, so it would be a minor inconvenience for service teams to remember to add these control statements. However, using validators, teams can easily catch when this is accidentally omitted (and most teams that care about maintain their existing naming schemes use tons of custom validators).

designs/inline-io.md
```diff
@@ -0,0 +1,207 @@
...
+than what is trivially generated. Therefore, requiring users to write it out
+is effectively pointless boilerplate. In AWS models, for instance, over 98% of
+operation input/output structures are named by suffixing the operation name.
+
+### Examples
+
+```
+operation GetUser {
+    // This uses the derived name: GetUserInput
+    input := {
```
mtdowling, line 58: Given we're adding `@input` and `@output` traits to Smithy, this new syntax should automatically apply the `@input` and `@output` traits to inline defined shapes. Given that, I'm kind of liking this syntax too:

```
operation GetUser {
    @input: {
        userId: String
    }
    @output: {
        userId: String
    }
    errors: []
}
```

If we wanted to extend this inline syntax out to other places, we could just say that `@` as a prefix is the convention for defining inline properites on shapes. It's just as generic as `:=` IMO.

JordonPhillips: `@` has a cognitive conflict with trait application though. While it's true that we may automatically apply those traits, it isn't what is most important to call out visually. Or, it's not so important that it makes it worth imposing the extra cognitive load.

designs/inline-io.md
```diff
@@ -0,0 +1,228 @@
...
+boilerplate. Additionally, separating them makes reading an operation at a high
+level more difficult since you have to jump around to get the information you
+need.
+
+## Proposal
+
+Operations will allow inlining input and output definitions, indicated by a
+walrus operator (`:=`). Inlined structures will deviate from normal
+structure definitions in two respects. Firstly, the structure keyword will be
+omitted. Additionally, providing a name for the inlined shape will be optional.
```
mtdowling, line 21: >Additionally, providing a name for the inlined shape will be optional.

Is this still true? You can only influence the name, not provide one.

designs/inline-io.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Walrus Operator
+
+Rather than using the same standalone colon (`:`) that is used in other
+cases, a walrus operator will be used to indicate an inline definition. The
+reason for this is to visually distinguish it at the outset, as well as to make
+tooling a bit cleaner to write.
```
mtdowling, line 28: nit: tooling is vague here. Maybe "parsers"? Also, "a bit cleaner to write" is vague too-- maybe "as well as to make parsers simpler because the walrus operator removes the need for arbitrary look ahead"?

designs/inline-io.md
```diff
@@ -0,0 +1,228 @@
...
+expression in places where initializing a variable was previously forbidden.
+In Go, it’s used to initialize and assign a variable in one step.
+
+### Omitting the structure Keyword
+
+When used at the top level of a Smithy IDL file, the type keywords are
+necessary to indicate what type of shape you’re making. Since operation
+inputs and outputs may only be structures, this isn’t necessary.
+
+### Default Names
```
mtdowling, line 41: I think we should remove the ability to provide an explicit name. Allowing an optional name makes the parser harder to write because we now need to know if "with" is a name or a mixin, or maybe we need to make "with" a keyword. We already added the ability to influence the generated suffix, so I think that's going far enough for people to migrate. Anyone else that uses completely bespoke naming patterns simply can't use inline input and output definitions.

designs/inline-io.md
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    // This uses an explicit name.
+    output := GetUserOutput {
+        username: String
+        userId: String
+    }
+}
+
+// Inlined inputs/outputs with explicit names and traits. Only 20% of current
```
mtdowling, line 83: 20% of what? AWS services? Be specific.

mtdowling (changes requested):

designs/inline-io.md
JordonPhillips: Yes, this is still true. You can still fully override the name:

```
operation Operation {
    input := Foo {}
}
```

mtdowling (approved):

PR #963, opened 2021-10-27
implementation of inlined operation input / output shapes

JordonPhillips:
This is an implementation of: #962

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -112,6 +112,9 @@ final class IdlModelParser extends SimpleParser {
     private final String filename;
     private TraitEntry pendingDocumentationComment;
 
+    private String operationInputSuffix = "Input";
+    private String operationOutputSuffix = "Output";
+
     // A pending trait that also doesn't yet have a resolved trait shape ID.
     static final class TraitEntry {
         final String traitName;
@@ -181,6 +184,7 @@ final class IdlModelParser extends SimpleParser {
     }
 
     private void parseControlSection() {
+        Set<String> definedKeys = new HashSet<>();
         while (peek() == '$') {
             expect('$');
             ws();
@@ -189,15 +193,19 @@ final class IdlModelParser extends SimpleParser {
             expect(':');
             ws();
 
-            // Validation here for better error location.
-            if (key.equals("version") && modelFile.getVersion() != Version.UNKNOWN) {
-                throw syntax("Cannot define multiple versions in the same file");
+            if (definedKeys.contains(key)) {
+                throw syntax(format("Duplicate control statement `%s`", key));
             }
+            definedKeys.add(key);
 
             Node value = IdlNodeParser.parseNode(this);
 
             if (key.equals("version")) {
                 onVersion(value);
+            } else if (key.equals("operationInputSuffix")) {
+                operationInputSuffix = value.expectStringNode().getValue();
+            } else if (key.equals("operationOutputSuffix")) {
+                operationOutputSuffix = value.expectStringNode().getValue();
             } else {
                 modelFile.events().add(ValidationEvent.builder()
                         .id(Validator.MODEL_ERROR)
@@ -519,6 +527,12 @@ final class IdlModelParser extends SimpleParser {
 
         ws();
         expect(':');
+
+        if (peek() == '=') {
+            throw syntax("Defining structures inline with the `:=` syntax may only be used when "
+                    + "defining operation input and output shapes.");
+        }
+
         ws();
         ShapeId memberId = parent.withMember(memberName);
         MemberShape.Builder memberBuilder = MemberShape.builder().id(memberId).source(memberLocation);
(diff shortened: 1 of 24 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -198,6 +202,10 @@ private void parseControlSection() {
 
             if (key.equals("version")) {
                 onVersion(value);
+            } else if (key.equals("operationInputSuffix")) {
+                operationInputSuffix = value.expectStringNode().getValue();
```
mtdowling, line 206: How much do we care about dupes in a single file? Seems like it would be silly for someone to duplicate these keys right, since they have to be at the top of the file... so guarding against it is probably going overboard.

Guarding against multiple versions per/file is still reasonable IMO because it's going to be used with low-level tools like grep that don't know Smithy model semantics. But these inline things probably won't be.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlNodeParser.java
```diff
@@ -135,6 +135,10 @@ static ObjectNode parseObjectNode(IdlModelParser parser, String parent) {
                 String key = parseNodeObjectKey(parser);
                 parser.ws();
                 parser.expect(':');
+                if (parser.peek() == '=') {
+                    throw parser.syntax("Defining structures inline with the `:=` syntax may only be used when "
```
mtdowling, line 139: This doesn't seem accurate. We're not defining a structure here, it's an object node.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -570,37 +584,120 @@ private void parseStructuredShape(
...
+                case "input":
+                    parseInlineableOperationMember(id, operationInputSuffix, builder::input);
+                    break;
+                case "output":
+                    parseInlineableOperationMember(id, operationOutputSuffix, builder::output);
+                    break;
+                case "errors":
+                    parseIdList(builder::addError);
+                    break;
+                default:
```
mtdowling, line 627: Could we ditch the OPERATION_PROPERTY_NAMES stuff and the validation in parseProperties, and instead handle that here in the switch? It seems redundant.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -570,37 +584,120 @@ private void parseStructuredShape(
...
+        });
+        modelFile.onShape(builder);
+    }
+
+    private void parseProperties(ShapeId id, Collection<String> allowedProperties, Consumer<String> valueParser) {
+        ws();
+        expect('{');
+        ws();
+
+        Set<String> defined = new TreeSet<>();
```
mtdowling, line 639: Was there a reason for making this a TreeSet instead of a HashMap?

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -570,37 +584,120 @@ private void parseStructuredShape(
...
+                    parseIdList(builder::addError);
+                    break;
+                default:
+                    break;
+            }
+        });
+        modelFile.onShape(builder);
+    }
+
+    private void parseProperties(ShapeId id, Collection<String> allowedProperties, Consumer<String> valueParser) {
```
mtdowling, line 634: I like this abstraction!

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -570,37 +584,120 @@ private void parseStructuredShape(
...
+        expect('}');
+    }
+
+    private void parseInlineableOperationMember(ShapeId id, String suffix, Consumer<ShapeId> consumer) {
+        if (peek() == '=') {
+            if (!modelFile.getVersion().supportsInlineOperationIO()) {
+                throw syntax(id, "Inlined operation inputs and outputs can only be used with Smithy version 2 or "
+                        + "later. Attempted to use inlined IO with version `" + modelFile.getVersion() + "`.");
+            }
+            expect('=');
```
mtdowling, line 664: I think you need to clear pending docs here. For example, what happens right now with this?

```
operation Foo {
    input 
    /// These should not be part of the shape!
    := {}
}
```

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: I don't think it matters that much to be honest, but I can add some generic check to make sure no control statement is duplicated.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlNodeParser.java
JordonPhillips: I updated the wording to be more clear

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: At one point I had all the defined keys as part of the error message, but there's no point anymore

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlNodeParser.java
```diff
@@ -159,7 +163,7 @@ static String parseNodeObjectKey(IdlModelParser parser) {
         }
     }
 
-    private static ArrayNode parseArrayNode(IdlModelParser parser) {
+    static ArrayNode parseArrayNode(IdlModelParser parser) {
```
mtdowling, line 166: Is this change needed? Or did I miss the new class where this is being used?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -145,15 +145,39 @@ private String serialize(Model fullModel, Collection<Shape> shapes) {
...
         shapes.stream()
                 .filter(FunctionalUtils.not(Shape::isMemberShape))
+                .filter(shape -> !inlineableShapes.contains(shape.getId()))
                 .sorted(new ShapeComparator())
                 .forEach(shape -> shape.accept(shapeSerializer));
 
         return serializeHeader(fullModel, namespace) + codeWriter.toString();
     }
 
+    private Set<ShapeId> getInlineableShapes(Model fullModel, Collection<Shape> shapes) {
```
mtdowling, line 160: We'll need to modify this once #980 lands

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlNodeParser.java
JordonPhillips: I was using it at one point for the operation errors, but I ended up removing it. It's changed back now.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
JordonPhillips: Yeah, we'll also need to add the input and output shapes by default, so we can change it as part of that.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -575,34 +583,89 @@ private void parseStructuredShape(
...
+            ws();
+        } while (peek() != '{');
     }
 
     private void parseOperationStatement(ShapeId id, SourceLocation location) {
         ws();
         OperationShape.Builder builder = OperationShape.builder().id(id).source(location);
-        ObjectNode node = IdlNodeParser.parseObjectNode(this, id.toString());
+
+        ObjectNode node = IdlNodeParser.parseObjectNode(this, id.toString(), (parser, key) -> {
```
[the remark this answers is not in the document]
mtdowling: Let's do service and resource in a seperate PR.

mtdowling (approved):

kstich (approved):

PR #969, opened 2021-11-02
Refactor queued projections

adamthom-amzn:
*Description of changes:*
TransformContext's queued projections were a side effect of transformer
execution, and TransformContext instances were not maintained throughout
projection execution. In particular, BackwardCompatHelper would create a new
instance of TransformContext for its subclasses, but relying on a side effect
of transformation was dangerous in any case.

This adds a new method to ProjectionTransformer that explicitly models
composition of transforms by returning the names of projections to be run
subsequently. It is defaulted to an empty list.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/ProjectionTransformer.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/ProjectionTransformer.java
@@ -16,11 +16,13 @@
 package software.amazon.smithy.build;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/ProjectionTransformer.java
```diff
@@ -42,6 +44,18 @@
...
 
+    /**
+     * Allows the composition of projections by returning additional
+     * projections to run after the current one.
+     *
+     * @param context Transformation context.
+     * @return a collection of named projections to run.
+     * @throws IllegalArgumentException if the arguments are invalid.
+     */
+    default Collection<String> getAdditionalProjections(TransformContext context) {
```
mtdowling, line 55: This should probably be a list since the order is very important.

mtdowling (approved):

PR #977, opened 2021-11-11
Clarify doc about sources plugin

gosar:
* All model _names_ referenced by the manifest, may be confused with shape names. All model _files_ seems clearer.
* _building_ a projection seems clearer than _applying_ a projection. And matches the earlier sentence - _When building the source projection_.

### DIFF
--- a/docs/source/1.0/guides/building-models/build-config.rst
+++ b/docs/source/1.0/guides/building-models/build-config.rst
@@ -1446,14 +1446,15 @@ Smithy models contained within the JAR are copied as a source model while the
 JAR itself is not copied. If there are no source models, an empty manifest is
 created.
 
-When applying a projection, a new model file is created that contains only
-the shapes, trait definitions, and metadata that were defined in a source
-model *and* all of the newly added shapes, traits, and metadata.
+When building projections other than ``source``, a new model file is created
+that contains only the shapes, trait definitions, and metadata that were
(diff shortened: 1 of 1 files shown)


docs/source/1.0/guides/building-models/build-config.rst
```diff
@@ -1446,14 +1446,14 @@ Smithy models contained within the JAR are copied as a source model while the
 JAR itself is not copied. If there are no source models, an empty manifest is
 created.
 
-When applying a projection, a new model file is created that contains only
-the shapes, trait definitions, and metadata that were defined in a source
+When building a different projection, a new model file is created that contains
```
mtdowling, line 1449: I don't see the reason for the word "different" here. Not sure what it means.

docs/source/1.0/guides/building-models/build-config.rst
```diff
@@ -1446,14 +1446,14 @@ Smithy models contained within the JAR are copied as a source model while the
...
+only the shapes, trait definitions, and metadata that were defined in a source
 model *and* all of the newly added shapes, traits, and metadata.
```
mtdowling, line 1451: Would be nice to clarify what "newly added" means here. I thought that was the main reason for the PR.

mtdowling (changes requested):

mtdowling (approved):

PR #986, opened 2021-11-16
Add tests for quoted strings in headers

adamthom-amzn:
*Description of changes:*
RFC 7230 sec 3.2.6 defines header values as potentially quoted strings, which
makes commas in header values unambiguous. It's also the only way to escape a
reserved character. This adds a test for list values with a comma and for list
values with escaped double quotes.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
@@ -47,6 +47,21 @@ apply InputAndOutputWithHeaders @httpRequestTests([
         }
     },
     {
+        id: "RestJsonInputAndOutputWithQuotedStringHeaders",
+        documentation: "Tests requests with string list header bindings that require quoting",
+        protocol: restJson1,
+        method: "POST",
+        uri: "/InputAndOutputWithHeaders",
+        headers: {
+            "X-StringList": """
+            "b,c","\"def\"",a"""
+        },
+        body: "",
+        params: {
+            headerStringList: ["b,c", "\"def\"", "a"]
+        }
+    },
+    {
         id: "RestJsonInputAndOutputWithNumericHeaders",
         documentation: "Tests requests with numeric header bindings",
         protocol: restJson1,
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
```diff
@@ -46,6 +46,22 @@ apply InputAndOutputWithHeaders @httpRequestTests([
...
+    {
+        id: "RestJsonInputAndOutputWithQuotedStringHeaders",
+        documentation: "Tests requests with string list header bindings that require quoting",
+        protocol: restJson1,
+        method: "POST",
+        uri: "/InputAndOutputWithHeaders",
+        headers: {
+            "X-StringList": """
+            a, "b,c", "\"def\""
+            """
```
mtdowling, line 58: This should go on the previous line. By putting it on the next line, it's inserting a "\n".

smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
```diff
@@ -187,6 +203,20 @@ apply InputAndOutputWithHeaders @httpResponseTests([
...
     },
+    {
+        id: "RestJsonInputAndOutputWithQuotedStringHeaders",
+        documentation: "Tests responses with string list header bindings that require quoting",
+        protocol: restJson1,
+        code: 200,
+        headers: {
+            "X-StringList": """
+            a, "b,c", "\"def\""
+            """
```
mtdowling, line 214: Same issue - needs to be on the previous line

mtdowling (changes requested):

smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
adamthom-amzn: Yes, but if I turn that into:

```
        headers: {
            "X-StringList": """
            a, "b,c", "\"def\"""""
        },
```

Then the entire file doesn't parse. I can work around this by flipping the order of the parameters, or by adding a whitespace (which is ignored according to spec)

mtdowling (approved):

PR #989, opened 2021-11-19
Fix filterSuppressions handling of members

gosar:
*Issue:*

FilterSuppressions uses builder.addShape [here](https://github.com/awslabs/smithy/blob/6ace7180592717eced4c01b5c317d68b9602c968/smithy-build/src/main/java/software/amazon/smithy/build/transforms/FilterSuppressions.java#L296) ) to update the traits on all shapes, including member shapes. However,  [`builder.addShape`](https://github.com/awslabs/smithy/blob/6ace7180592717eced4c01b5c317d68b9602c968/smithy-model/src/main/java/software/amazon/smithy/model/Model.java#L888-L889) does not support updating member shapes
directly.

*Description of changes:*
Instead use ReplaceShapes which handles replacing member shapes too.

*Testing:*
Updated the traits.smithy test cases to cover the member case. They failed before this change.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/FilterSuppressions.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/FilterSuppressions.java
@@ -32,6 +32,7 @@ import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.node.ObjectNode;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.traits.SuppressTrait;
+import software.amazon.smithy.model.transform.ModelTransformer;
 import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
 import software.amazon.smithy.model.validation.suppressions.Suppression;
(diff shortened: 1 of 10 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/transforms/FilterSuppressions.java
```diff
@@ -293,25 +295,24 @@ private void filterSuppressionTraits(
...
                 trait = trait.toBuilder().values(allowed).build();
-                builder.addShape(Shape.shapeToBuilder(shape).addTrait(trait).build());
+                replacementShapes.add(Shape.shapeToBuilder(shape).addTrait(trait).build());
             }
         }
+        return transformer.replaceShapes(model, replacementShapes);
     }
 
-    private void filterMetadata(
+    private Model filterMetadata(
```
mtdowling, line 307: minor: Can you add a check to make sure the model has suppressions defined before recreating it?

mtdowling (approved):

smithy-build/src/main/java/software/amazon/smithy/build/transforms/FilterSuppressions.java
gosar: Good suggestion. Added a check to make sure suppressions changed before recreating. Also added couple tests cases for 1. no suppressions 2. no change to suppressions.

PR #994, opened 2021-11-26
Change OpenAPI security schema names to lowercase

jzorn:
According to the [OpenAPI specification](https://spec.openapis.org/oas/latest.html), security schema names should
be used in lowercase. This causes issues e.g. in the [OpenAPI generator](https://github.com/OpenAPITools/openapi-generator/blob/06faa289bd93c5fe1f2caa376e32ec6cc07627a2/modules/openapi-generator/src/main/java/org/openapitools/codegen/DefaultCodegen.java#L4778)
which checks for case sensitive values.


JordonPhillips:
Where in the specification does it say they should be lower cased? This is all I see:

> REQUIRED. The name of the HTTP Authorization scheme to be used in the Authorization header as defined in [[RFC7235](https://spec.openapis.org/oas/latest.html#bib-RFC7235)]. The values used SHOULD be registered in the [IANA Authentication Scheme registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml).

If you look at that iana registry, all the names are capitalized.

jzorn:
Hm, that's fair. Maybe I barked up the wrong tree. Thanks for your reply! I will try to fix this in the Kotlin code generator.

PR #995, opened 2021-11-29
Reduce builder copies using a managed reference

mtdowling:
This commit reduces the number of copies builders need to make when
building up immutable objects.

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/DefaultBuilderRef.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/DefaultBuilderRef.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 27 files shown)


smithy-utils/src/main/java/software/amazon/smithy/utils/DefaultBuilderRef.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Default implementation of {@link BuilderRef}.
+ *
+ * @param <T> The type of value being built, borrowed, and copied.
+ */
+final class DefaultBuilderRef<T> implements BuilderRef<T> {
```
adamthom-amzn, line 27: should this be documented as explicitly not threadsafe?

mtdowling:
Relevant benchmarks:

Before:

```
Benchmark                                  Mode  Cnt   Score   Error  Units
Selectors.loadsIdlModelWithoutValidation   avgt    3  57.376 ± 5.336  us/op
Selectors.loadsJsonModelWithoutValidation  avgt    3  73.979 ± 9.156  us/op
```

After:

```
Benchmark                                  Mode  Cnt   Score    Error  Units
Selectors.loadsIdlModelWithoutValidation   avgt    3  50.852 ± 11.004  us/op
Selectors.loadsJsonModelWithoutValidation  avgt    3  61.740 ±  6.504  us/op
```

smithy-utils/src/main/java/software/amazon/smithy/utils/DefaultBuilderRef.java
mtdowling: Done

adamthom-amzn (approved):

PR #1000, opened 2021-12-04
Adding Github Action for CodeGuru Reviewer

martinschaef:
*Description of changes:*
Runs CodeGuru Reviewer on push and pull_request events and posts recommendations in the [Security tab](https://github.com/awslabs/smithy/security). 
Example run can be seen here: https://github.com/martinschaef/smithy/runs/4414789132?check_suite_focus=true

This uses the new OIDC-way to assume an IAM role. That is, no credentials are required. Only authorized repos can assume the role. 


martinschaef:
Somehow the action fails to assume the role in the Pull Request; probably because I am not a member of the Smithy repo. Created an issue with [aws-actions/configure-aws-credentials](https://github.com/aws-actions/configure-aws-credentials/issues/188#issuecomment-986085773). It works if the person starting the action has permissions on the repo. See [here](https://github.com/martinschaef/smithy/runs/4418820965?check_suite_focus=true)

.github/workflows/codeguru-reviewer.yml
```diff
@@ -0,0 +1,48 @@
...
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        role-to-assume: arn:aws:iam::048169001733:role/GuruGitHubCICDRole
+        aws-region: us-west-2
+
+    - name: Run CodeGuru Reviewer
+      uses: aws-actions/codeguru-reviewer@v1.1
+      continue-on-error: false
+      with:          
+        s3_bucket: codeguru-reviewer-github-profiler-demo-048169001733-uw2
```
adamthom-amzn, line 42: Where does this bucket come from, and who owns it?

martinschaef: It's owned by the CodeGuru team and was created using this Stack:
https://github.com/aws-samples/aws-codeguru-reviewer-cicd-cdk-sample

adamthom-amzn (approved):

adamthom-amzn:
I'm reverting this in https://github.com/awslabs/smithy/pull/1008, as it did not work even when the PR was opened by a smithy developer: https://github.com/awslabs/smithy/pull/1007

martinschaef:
Ugh ... that's a problem. The permissions are per account, so it makes sense that it doesn't work. You don't want people that do a PR against your repo be allowed to assume a role in your account. 

Would it be ok if the action just succeeds if it cannot assume the role?

adamthom-amzn:
Yes, I think that's fine, but who would this role assumption work for?

martinschaef:
On second thought, to only way to do this is to trigger the action only on `push` events. So the action would only run *after* you merge the pull request. This is a little bit counter intuitive but makes sense from an IAM point of view. Anyone could open a PR against your repo and they should not be allowed to just assume that role because they could create charges in that account.

martinschaef:
Oh wait ... this is a different issue: https://github.com/awslabs/smithy/runs/4458443115?check_suite_focus=true#step:8:38 here the analysis actually timed out

adamthom-amzn:
https://github.com/awslabs/smithy/runs/4459885012?check_suite_focus=true is the run that caused me to do the revert

martinschaef:
Yeah, this makes sense. This is because JordonPhillips is not allowed to assume the IAM Role. I'll add proper error handling to the action.

adamthom-amzn:
Who has permission to assume it? Jordon's a core smithy developer

martinschaef:
This works via OICD, so it works by repo. The container creates this JWT token to authenticate which is matched against an allow-list of orgs. So only if the container is run under `awslabs/smithy` it'll work, but this container tried to authenticate as `JordonPhillips/smithy`. That's why you basically can only use this in push events.

Fwiw, this is an IAM thing, not a CodeGuru thing.

martinschaef:
I changed it so it only does any of the steps if assuming the role succeeded:
https://github.com/martinschaef/smithy/runs/4462569719?check_suite_focus=true

adamthom-amzn:
Will that require a new PR, or were the changes on your side such that a revert of the revert should be sufficient?

martinschaef:
No, it requires code changes, I'll do a new PR

PR #1004, opened 2021-12-06
Reduce allocations and garbage

mtdowling:
This commit attempts to cut down on the amount of allocations used when
validating Smithy models, including limiting the use of Streams in core
APIs, no longer copying validation events passed to ValidatedResult, and
no longer caching the result of every selector used in trait defintions
when evaluating trait targets.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationUtils.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationUtils.java
@@ -26,7 +26,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
+import java.util.StringJoiner;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -60,9 +60,15 @@ public final class ValidationUtils {
      * @return Returns the string.
      */
     public static String orderedTickedList(Collection<?> values) {
-        return values.size() == 0
-               ? ""
-               : "`" + values.stream().map(Objects::toString).collect(Collectors.joining("`, `")) + "`";
+        if (values.size() == 0) {
+            return "";
+        }
+
+        StringJoiner joiner = new StringJoiner("`, `", "`", "`");
+        for (Object value : values) {
+            joiner.add(value.toString());
+        }
+        return joiner.toString();
     }
 
     /**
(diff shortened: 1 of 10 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationUtils.java
```diff
@@ -80,21 +86,10 @@ public static String tickedList(Collection<?> values) {
...
-     *
-     * <p>The returned map is sorted by the conflicting name, and the list of
-     * conflicting shape values is also sorted to ensure that validation
-     * events can just print the map directly with consistent error messages.
-     *
-     * @param shapes Shapes collection to check.
-     * @param <T> The type of shape being checked.
-     * @return Returns a map that only contains the conflicting shapes.
-     */
+    @Deprecated
```
kstich, line 92: Has this been unreferenced the whole time? It doesn't look like this commit removes usage of it, but it's being deprecated.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PrivateAccessValidator.java
```diff
@@ -38,26 +35,26 @@
 
     @Override
     public List<ValidationEvent> validate(Model model) {
-        Set<Shape> privateShapes = model.getShapesWithTrait(PrivateTrait.class);
-        NeighborProvider provider = NeighborProviderIndex.of(model).getProvider();
-        return model.shapes()
-                .filter(shape -> !(shape instanceof SimpleShape))
```
kstich, line 44: I don't see this filter carried forward, but assume that's intentional given the `*` selector on `@private` - should this come with a test?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/TraitTargetValidator.java
```diff
@@ -45,50 +45,53 @@
...
-                String sanitized = SANITIZE.matcher(selector.toString()).replaceAll(" ");
-                events.add(error(shape, shape.findTrait(trait).get(), String.format(
-                        "Trait `%s` cannot be applied to `%s`. This trait may only be applied "
-                        + "to shapes that match the following selector: %s",
-                        Trait.getIdiomaticTraitName(trait.toShapeId()),
-                        shape.getId(),
-                        sanitized)));
+        for (ShapeId traitId : traits) {
+            // Find all shapes that have the used trait applied to it.
+            for (Shape shape :  model.getShapesWithTrait(traitId)) {
```
kstich, line 83: nit: extra space after `:`

kstich (approved):

PR #1007, opened 2021-12-08
Add input and output traits with inlined IO

JordonPhillips:
This updates inlined IO shapes to automatically include the input and output traits.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -61,6 +61,8 @@ import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.TimestampShape;
 import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.DocumentationTrait;
+import software.amazon.smithy.model.traits.InputTrait;
+import software.amazon.smithy.model.traits.OutputTrait;
 import software.amazon.smithy.model.traits.RequiredTrait;
 import software.amazon.smithy.model.traits.TraitFactory;
 import software.amazon.smithy.model.validation.Severity;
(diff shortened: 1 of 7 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -662,15 +671,18 @@ private void parseInlineableOperationMember(ShapeId id, String suffix, Consumer<
...
             ws();
             modelFile.addForwardReference(ParserUtils.parseShapeId(this), consumer);
         }
     }
 
-    private ShapeId parseInlineStructure(String name) {
+    private ShapeId parseInlineStructure(String name, TraitEntry defaultTrait) {
         List<TraitEntry> traits = parseDocsAndTraits();
+        if (defaultTrait != null) {
+            traits.add(defaultTrait);
```
kstich, line 684: What happens if an `@input`/`@output` is applied manually? I don't see a test for that either.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -165,16 +169,20 @@ private String serialize(Model fullModel, Collection<Shape> shapes) {
                 continue;
             }
             OperationShape operation = shape.asOperationShape().get();
-            operation.getInput().ifPresent(shapeId -> {
-                if (shapes.contains(fullModel.expectShape(shapeId))) {
-                    inlineableShapes.add(shapeId);
+            if (!operation.getInputShape().equals(UnitTypeTrait.UNIT)) {
+                Shape inputShape = fullModel.expectShape(operation.getInputShape());
+                if (shapes.contains(inputShape) && inputShape.hasTrait(InputTrait.ID)
+                        && operation.getInputShape().getName().equals(operation.getId().getName() + "Input")) {
```
kstich, line 175: Looks like this is only going to inline shapes with the default suffix control statements. It'd bring a lot of complexity to try and set these statements, but calling it for confirmation.

JordonPhillips: Yes. There's already plenty of complexity without introducing that too.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: The dupe is ignored. Added a test case for it.

kstich (approved):

PR #1011, opened 2021-12-09
Fix escaping sets in http traits

rcoh:
*Issue #, if available:* https://github.com/awslabs/smithy-rs/pull/953/files, https://github.com/awslabs/aws-sdk-rust/issues/331

*Description of changes:* Update the set of characters that need to be escaped to include characters that are never valid in URIs.

### DIFF
--- a/docs/source/1.0/spec/core/http-traits.rst
+++ b/docs/source/1.0/spec/core/http-traits.rst
@@ -673,7 +673,7 @@ structure with the ``httpLabel`` trait MUST have a corresponding
   ``1985-04-12T23%3A20%3A50.52Z``). The :ref:`timestampFormat-trait`
   MAY be used to use a custom serialization format.
 - Reserved characters defined in :rfc:`section 2.2 of RFC3986 <3986#section-2.2>`
-  and the `%` itself MUST be percent-encoded_ (that is, ``:/?#[]@!$&'()*+,;=%``).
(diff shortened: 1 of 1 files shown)


docs/source/1.0/spec/core/http-traits.rst
```diff
@@ -673,7 +673,7 @@ structure with the ``httpLabel`` trait MUST have a corresponding
   ``1985-04-12T23%3A20%3A50.52Z``). The :ref:`timestampFormat-trait`
   MAY be used to use a custom serialization format.
 - Reserved characters defined in :rfc:`section 2.2 of RFC3986 <3986#section-2.2>`
-  and the `%` itself MUST be percent-encoded_ (that is, ``:/?#[]@!$&'()*+,;=%``).
+  and the `%` itself MUST be percent-encoded_ (that is, ``:/?#[]@!$&'()*+,;=%><"`? ``). Note the inclusion of a literal space, ` `.
```
adamthom-amzn, line 676: At this point it'd probably be shorter to list the things that aren't percent encoded according to RFC-3986. That'd be `ALPHA / DIGIT / "-" / "." / "_" / "~" / ":" / "@" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="` where ALPHA is valid alphabetic characters in US-ASCII.

Interesting that we've called out `@` and `:` where it's generally allowed in a path segment as-is.

PR #1025, opened 2021-12-22
Allow mixins on all shape types

JordonPhillips:
This expands mixins to allow them on any shape type.

One other thing that has been done is allowing shapes to re-define members. This *could* allow us to continue to require that lists / sets / maps always define their members. So instead of

```
list A {
    @required
    member: String
}

list B with A {}
```

You'd have

```
list A {
    @required
    member: String
}

list B with A {
    // This inherits the @required trait from A$member
    member: String
}
```

Trait merging works precisely like it does for the shapes themselves, which removes the need to have a bunch of awkward apply statements everywhere. The one restriction is that the target has to be the same shape.


smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/mixins/operations.flattened.smithy
```diff
@@ -0,0 +1,14 @@
...
+@error("client")
+structure ConcreteError {}
+
+@error("server")
+structure MixinError {}
+
+@internal
+operation ConcreteOperation {
+    errors: [MixinError, ConcreteError]
+}
```
mtdowling, line 14: A lot of tests are missing newlines

smithy-model/src/test/resources/software/amazon/smithy/model/shapes/idl-serialization/cases/service-mixins.smithy
```diff
@@ -0,0 +1,39 @@
...
+    operations: [
+        MixinOperation
+    ]
+    resources: [
+        MixinResource
+    ]
+    errors: [
+        MixinError
+    ]
+    rename: {
```
mtdowling, line 21: Could the shape mixing this in extend this rename map? Can you add a test if so?

smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleParser.java
```diff
@@ -103,6 +103,12 @@ public final int column() {
         return column;
     }
```
mtdowling, line 105: One day we may want to try to make this support some streaming use cases. If we do, it would be better to use a marker that is set and then rewound to so the parser knows when it needs to start buffering. So instead of rewinding to an absolute position, you call mark(int numberofbytes) when you might need to backtrack, followed by reset to go back to the last set marker. This is just based on streams in Java: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#mark(int)

designs/mixins.md
```diff
@@ -692,12 +698,88 @@ The members are ordered as follows:
...
+operation OperationA {}
+
+service A {
+    version: "A"
+    operations: [OperationA]
+}
+
+operation OperationB {}
+
+service B with A {
```
mtdowling, line 722: Doesn’t this need a mixin trait? Any shape used as a mixin previously needed a mixin trait.

designs/mixins.md
```diff
@@ -692,12 +698,88 @@ The members are ordered as follows:
...
+        "smithy.example#OperationB": "OperB"
+        "smithy.example#OperationC": "OpC"
+    }
+    operations: [OperationA, OperationB, OperationC]
+}
+```
+
+### Resource mixins
+
+Resource shapes with the `@mixin` trait MAY NOT define any properties. This is
```
mtdowling, line 766: Wouldn’t defining identifiers be ok? You could then extend the identifiers map in shapes that use the mixin. This might make it easier to define resource hierarchies.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -607,13 +607,39 @@ private void parseMixins(ShapeId id) {
...
+        } while (!peekEndWith());
+    }
+
+    private boolean peekEndWith() {
+        if (peek() == '{' || peek() == '@' || eof()) {
+            return true;
+        }
+
+        // We could make this more efficient with a prefix trie
+        for (String shapeType : SHAPE_TYPES) {
```
mtdowling, line 623: Does this mean that shape types basically are keywords in the grammar? I couldn’t name a shape integer for example and try to mix it in…. But only after the first mixin?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EntityShape.java
```diff
@@ -25,12 +27,39 @@
...
+            resources = builder.resources.copy();
+            introducedResources = resources;
+            operations = builder.operations.copy();
+            introducedOperations = operations;
+        } else {
+            Set<ShapeId> computedResources = new TreeSet<>();
+            Set<ShapeId> computedOperations = new TreeSet<>();
+
+            for (Shape shape : builder.getMixins().values()) {
+                if (shape instanceof EntityShape) {
```
mtdowling, line 47: I think you’re doing checks during the loading process to make sure mixins are coherent right? If so, it would be fine to throw here and do some up front validation.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ServiceShape.java
```diff
@@ -46,11 +89,11 @@ public static Builder builder() {
...
         return updateBuilder(builder())
-                .version(version)
-                .errors(errors)
-                .rename(rename)
-                .operations(getOperations())
-                .resources(getResources());
+                .version(introducedVersion)
+                .errors(introducedErrors)
+                .rename(introducedRename)
+                .operations(getIntroducedOperations())
```
mtdowling, line 95: I’m on a phone… sorry… how critical is it that we compute the “introduced” stuff up front? Is t it mostly only used when serializing a model? I’m wondering if we can yank all of this stuff out and then compute it when serializing a model by traversing the model. It could be wrapped in a knowledge index too if we think people might need it for other uses.

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -88,10 +89,46 @@
         }
     }
 
-    protected void validateMixins(Collection<ShapeId> mixins) {
+    protected void validateMixins(Map<ShapeId, Shape> mixins, Map<ShapeId, Trait> introducedTraits) {
         // This method is overridden in subclasses that allow mixins (structure, union, member).
         // This situation can only happen by manipulating the semantic model in code.
-        throw new SourceException("Mixins are not allowed on " + getId() + ", a " + getType(), this);
+        //throw new SourceException("Mixins are not allowed on " + getId() + ", a " + getType(), this);
```
kstich, line 95: Remove commented code.

kstich (changes requested):

designs/mixins.md
JordonPhillips: I'm not sure what this makes easier. Resource hierarchies are modeled using the `resources` property.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: > Does this mean that shape types basically are keywords in the grammar?

Yes. I think we'd need an explicit continuation marker to make this not the case, that or an explicit end marker.

> I think it also precludes shapes that start with the name of a shape type

This is a bug that can be fixed

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EntityShape.java
JordonPhillips: Yeah it ensures the shape types match before this is run.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ServiceShape.java
JordonPhillips: It's not super critical. It's just a lot harder to do it after the fact.

mtdowling: Or maybe these introduced things could be computed lazily. Possibly traits too, but we’d want to benchmark it — to get a trait via getTrait(), you check the current shape then iterate backwards over mixins for example. Each “introduced” method could be computed on demand (e.g., getting version means first check current shape, then check mixins, every time). My idea for moving this to a knowledge index wouldn’t work since all this information is local to shapes. Plus we do want to serialize traits and introduced things even if they repeat one of their mixins as that is an explicitly set value on a shape (something you can’t tell from outside of a shape).

I’m mostly concerned with time and space costs of introducing the additional sets, maps, up front validation, etc. I’m thinking of the use case of loading every AWS model at once and running a selector. The way it’s done now could be the best way and I could be totally wrong though.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: With the move to bracketing, this isn't relevant anymore

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ServiceShape.java
JordonPhillips: You actually can do it with a knowledge index, i have updated the pr to do it. It's just super tedious

designs/mixins.md
```diff
@@ -518,27 +570,52 @@ structure A1 {
...
-To support mixins, `structure_statement` and `union_statement` ABNF rules
+To support mixins, shape ABNF rules
 will be updated to contain an optional `mixins` production
-that comes after the shape name and before `{`. Each shape ID referenced in
+that comes after the shape name and before any `{`. Each shape ID referenced in
 the `mixins` production MUST target a shape of the same type as the
 shape being defined and MUST be marked with the `@mixin` trait.
 
 ```
+simple_shape_statement = simple_type_name ws identifier [ws mixins]
```
mtdowling, line 609: This would be a pretty big simplification for the parser I think, and it would get rid of the inability for us to easily add new keywords that start with "w" in the future.

```
simple_shape_statement = simple_type_name ws identifier [mixins]
mixins = sp "with" ws "[" 1*(ws shape_id) ws "]"
```

So that means these are ok:

```
string Foo with [Bar]
string Foo with [
    Bar
    Baz
]
```

But these are not:
```
string Foo
with [Bar]

string Foo
    with [Bar]
```

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -584,14 +589,20 @@ private void parseMixins(ShapeId id) {
         }
 
         ws();
+        expect('[');
+        ws();
+
         do {
+            clearPendingDocs();
```
mtdowling, line 596: Do you need to clear pending docs in each iteration? Couldn't you do that at the end, once?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
```diff
@@ -34,12 +36,37 @@
...
+            for (Shape shape : builder.getMixins().values()) {
+                shape.asOperationShape().ifPresent(mixin -> computedErrors.addAll(mixin.getErrors()));
+            }
+            introducedErrors = builder.errors.copy();
+            computedErrors.addAll(introducedErrors);
+            errors = Collections.unmodifiableList(new ArrayList<>(computedErrors));
+        }
+
+        if (hasTrait(MixinTrait.ID) && (!input.equals(UnitTypeTrait.UNIT) || !output.equals(UnitTypeTrait.UNIT))) {
+            throw new IllegalStateException(String.format(
```
mtdowling, line 64: Can we through a SourceException here instead?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ServiceShape.java
```diff
@@ -29,14 +34,52 @@
...
+        if (getMixins().isEmpty()) {
+            version = builder.version;
+            introducedVersion = version;
+            rename = builder.rename.copy();
+            introducedRename = rename;
+            errors = builder.errors.copy();
+            introducedErrors = errors;
+        } else {
+            String computedVersion = "";
+            Map<ShapeId, String> computedRename = new HashMap<>();
```
mtdowling, line 55: Rename should be an ordered map just like the builder uses

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -88,10 +89,43 @@
...
+    protected void validateMixins(Map<ShapeId, Shape> mixins, Map<ShapeId, Trait> introducedTraits) {
+        Set<String> invalid = new TreeSet<>();
+        for (Shape mixin : mixins.values()) {
+            if (mixin.getType() != getType()) {
+                invalid.add(mixin.getId().toString());
+            }
+        }
+        if (!invalid.isEmpty()) {
+            String invalidList = String.join("`, `", invalid);
+            throw new IllegalStateException(String.format(
```
mtdowling, line 101: SourceException

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -88,10 +89,43 @@
...
+                            .target(member.getTarget())
+                            .source(getSourceLocation())
+                            .addMixin(member)
+                            .build();
+                    break;
+                }
+            }
+        }
+        if (mixedMember == null) {
+            throw new IllegalStateException(String.format("Missing required member of shape `%s`: %s", getId(), name));
```
mtdowling, line 126: SourceException

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -88,10 +89,43 @@
...
+                    "Mixins may only be mixed into shapes of the same type. The following mixins were applied to the "
+                            + "%s shape `%s` which are not %1$s shapes: [`%s`]", getType(), getId(), invalidList));
+        }
+    }
+
+    protected MemberShape getRequiredMixinMember(
+            AbstractShapeBuilder<?, ?> builder,
+            String name
+    ) {
+        MemberShape mixedMember = null;
```
mtdowling, line 111: Is this method called in a way that it should return any members set locally on the builder vs members of mixins?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -88,10 +89,43 @@
...
+        for (Shape shape : builder.getMixins().values()) {
+            for (MemberShape member : shape.members()) {
+                if (member.getMemberName().equals(name)) {
+                    mixedMember = MemberShape.builder()
+                            .id(getId().withMember(name))
+                            .target(member.getTarget())
+                            .source(getSourceLocation())
+                            .addMixin(member)
+                            .build();
+                    break;
```
mtdowling, line 121: just return here?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -389,7 +389,9 @@ public SmithyIdlModelSerializer build() {
         @Override
         protected Void getDefault(Shape shape) {
             serializeTraits(shape);
-            codeWriter.write("$L $L", shape.getType(), shape.getId().getName()).write("");
+            codeWriter.writeInline("$L $L ", shape.getType(), shape.getId().getName());
```
mtdowling, line 392: I'm guessing CodeWriter is handling trimming spaces at the end of lines, right?

mtdowling (changes requested):
Looking good so far!

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/mixins/mixin-names.smithy
```diff
@@ -0,0 +1,21 @@
...
+
+namespace smithy.example
+
+@mixin
+integer integerPrefixed
+
+@mixin
+integer NotPrefixed
+
+// This is related to a bug that was present where only
```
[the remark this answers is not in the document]
JordonPhillips: No, but there's no harm in leaving the test case in

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: This was an artifact of when I needed to do more backtracking.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
JordonPhillips: No, we need the most recently introduced version. I added an explanatory comment.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
JordonPhillips: `AbstractShapeBuilder` doesn't have members, so the callers have to supply the local version if present.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
JordonPhillips: yes

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/mixins/mixin-names.smithy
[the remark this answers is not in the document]
mtdowling: The comment would be misleading then if the implementation and grammar has completely changed.

designs/mixins.md
```diff
@@ -518,27 +569,52 @@ structure A1 {
...
+that comes after the shape name. Each shape ID referenced in
 the `mixins` production MUST target a shape of the same type as the
 shape being defined and MUST be marked with the `@mixin` trait.
 
 ```
-structure_statement = "structure" ws identifier ws [mixins ws] structure_members
-union_statement = "union" ws identifier ws [mixins ws] union_members
-mixins = "with" 1*(ws shape_id)
+simple_shape_statement = simple_type_name ws identifier [mixins]
+list_statement = "list" ws identifier [mixins ws] shape_members
```
mtdowling, line 609: `shape_members` doesn't have a leading `ws` token in the current spec:

```
shape_members =
    empty_shape_members / populated_shape_members

empty_shape_members =
    "{" ws "}"

populated_shape_members =
    "{" ws shape_member_kvp
    *(comma shape_member_kvp ws) trailing_comma "}"
```

So this, and others, needs ws outside of `[mixins]` right?

```
list_statement = "list" ws identifier [mixins] ws shape_members
```

smithy-model/src/test/resources/software/amazon/smithy/model/loader/invalid/mixins/with-on-next-line.smithy
```diff
@@ -0,0 +1,10 @@
+// Parse error at line 10, column 9 near ` `: Unexpected shape type: with
```
mtdowling, line 1: We probably should have a test for each shape type :(... Sorry

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -566,9 +563,13 @@ private void parseStructuredShape(
         // Parse optional "with" statements to add mixins, but only if it's supported by the version.
         parseMixins(id);
         parseMembers(id, Collections.emptySet());
+        clearPendingDocs();
     }
 
     private void parseMixins(ShapeId id) {
+        // This is fine for now, but if we ever add any other keywords that start with
```
mtdowling, line 570: This comment can be removed now

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -690,6 +698,8 @@ private void parseIdList(Consumer<ShapeId> consumer) {
     }
 
     private void parseServiceStatement(ShapeId id, SourceLocation location) {
+        ws();
```
mtdowling, line 701: `ws` here would allow mixins on the next line

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -718,6 +729,8 @@ private void optionalIdList(ObjectNode node, String name, Consumer<ShapeId> cons
     }
 
     private void parseResourceStatement(ShapeId id, SourceLocation location) {
+        ws();
```
mtdowling, line 732: `ws` here would allow mixins on the next line

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -454,22 +452,25 @@ private ShapeId parseShapeName() {
 
     private void parseSimpleShape(ShapeId id, SourceLocation location, AbstractShapeBuilder builder) {
         modelFile.onShape(builder.source(location).id(id));
+        parseMixins(id);
+        // Pending docs get cleared by parseMixins. It isn't done here because the
```
mtdowling, line 456: Is this still true?

mtdowling (changes requested):

mtdowling (approved):

PR #1034, opened 2022-01-05
Make adoption of Unit easier for existing models

adamthom-amzn:
*Description of changes:*
Prevents Unit from being added to the service closure if it is only used by operation inputs and/or outputs, and
special-cases Unit for both operations and unions in JSONSchema conversion.

### DIFF
--- a/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/DeconflictingStrategy.java
+++ b/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/DeconflictingStrategy.java
@@ -17,6 +17,7 @@ package software.amazon.smithy.jsonschema;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import software.amazon.smithy.model.Model;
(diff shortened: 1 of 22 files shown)


smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/DefaultRefStrategy.java
```diff
@@ -130,6 +135,11 @@ public boolean isInlined(Shape shape) {
             return isInlined(target);
         }
 
+        // smithy.api#Unit is always inlined
+        if (UnitTypeTrait.UNIT.equals(shape.getId())) {
```
mtdowling, line 139: I think things like API Gateway will generate names for anonymous objects (which this code would cause). Is there a way to generate a Unit shape in the JSON Schema for tagged unions members, but not for operation input?

adamthom-amzn: I have hacked this PR to do this, but the implementation via a boolean flag screams "this is not the right way to accomplish this."

This would be so much easier if we had service attached shapes; then I could meaningfully use relationship data to examine what's actually attached to the scope of the conversion, and a lot of this code would be a lot simpler.

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/DeconflictingStrategy.java
```diff
@@ -49,12 +49,12 @@
...
     private final Map<String, ShapeId> reversePointers = new HashMap<>();
 
-    DeconflictingStrategy(Model model, RefStrategy delegate) {
+    DeconflictingStrategy(Model model, RefStrategy delegate, boolean preventUnitConflicts) {
         this.delegate = delegate;
 
         // Pre-compute a map of all converted shape refs. Sort the shapes
         // to make the result deterministic.
-        model.shapes().filter(FunctionalUtils.not(this::isIgnoredShape)).sorted().forEach(shape -> {
+        model.shapes().filter(s -> !isIgnoredShape(s, preventUnitConflicts)).sorted().forEach(shape -> {
```
mtdowling, line 57: Instead of the boolean flag, could you iterate over every tagged union shape, then check if any of its members target smithy.api#Union? If that's the case, then include Union, otherwise, don't include it. Should be pretty fast since it would only need to check union shape members and that kind of filtering is cached in Model.

mtdowling: Ah I see that's how you're determining the boolean flag. Nevermind.

mtdowling (approved):

mtdowling (approved):

PR #1037, opened 2022-01-07
Fix timestamp list headers restXml protocol test

jdisanti:
It looks like this test case was missed in https://github.com/awslabs/smithy/pull/986. I think this is the last of them. After applying this fix, all the protocol tests passed against smithy-rs with the code adjustments to support quoted values in the header lists.


jdisanti:
I missed that the timestamp lists being quoted was reverted in 1.16.

PR #1038, opened 2022-01-07
Add unwrite() to CodeWriter to help with CSVs

mtdowling:
It's sometimes easier to deal with comma separated values by just always
throwing a ", " at the end of each element in the list, even for the
last element. The unwrite() method allows you to remove the last text
written to the CodeWriter if it matches a given substring. This
allows, for example, ", " to be "unwritten" to removing the trailing
comma without having to first check if the last text written was a
trailing comma. This is similar to Ruby's chomp() method.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
@@ -1264,6 +1264,49 @@ public class CodeWriter {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kstich (approved):

mullermp:
LGTM. This is also useful for new lines. If we're iterating to code generate methods, we want new lines after each method unless it's the last. Ruby classes might be like

```
class Foo
  def meth1
  end
              <- new line
  def meth2
  end <-- no new line
end
```

PR #1040, opened 2022-01-07
Fix some RestJson query string protocol request tests for servers

david-perez:
From a server perspective, `params` is the expected output after request
deserialization. Since `queryParamsMapOfStringList` is bound with
`@httpQueryParams` to the query string, its value must be present and
derived from the input `queryParams` field of the request test.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-query.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-query.smithy
@@ -139,6 +139,10 @@ apply AllQueryStringTypes @httpRequestTests([
         params: {
             queryFloat: "NaN",
             queryDouble: "NaN",
+            queryParamsMapOfStringList: {
+                "Float": ["NaN"],
+                "Double": ["NaN"],
+            }
         }
     },
     {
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restJson1/http-query.smithy
```diff
@@ -139,6 +139,10 @@ apply AllQueryStringTypes @httpRequestTests([
         params: {
             queryFloat: "NaN",
```
david-perez, line 140: This test doesn't make sense from the server perspective anyway, since in IEEE 754 floating point numbers, `NaN != NaN`. As such, what should we assert we received after request deserialization, if we can't compare the deserialized value to anything? Shall I tag this test so that it only applies to the client, or should I assert something special in this case?

JordonPhillips (approved):
omg I wrote this comment ages ago but forgot to post it

smithy-aws-protocol-tests/model/restJson1/http-query.smithy
david-perez: @JordonPhillips I think what you're saying is that a protocol test involving `NaN` only cares that the deserialized value is _a_ `NaN` and that we shouldn't compare the bits, right? That's the approach we took https://docs.rs/aws-smithy-protocol-test/latest/aws_smithy_protocol_test/trait.FloatEquals.html, let me know if something else should be asserted.

In any case, this is subtle and should be called out in the documentation, perhaps in this page? https://awslabs.github.io/smithy/1.0/spec/http-protocol-compliance-tests.html

PR #1041, opened 2022-01-07
Add exact match configuration to repeated shape name validator

JordonPhillips:
This adds a configuration option to the repeating shape name validator to specify that it should be an exact match rather than a prefix check.

This also deprecates the old validator in favor of the new one due to its less than ideal name.

### DIFF
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/RepeatedShapeNameValidator.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/RepeatedShapeNameValidator.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.node.NodeMapper;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.UnionShape;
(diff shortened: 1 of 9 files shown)


mtdowling (changes requested):

smithy-linters/src/main/java/software/amazon/smithy/linters/RepeatedShapeNameValidator.java
```diff
@@ -0,0 +1,99 @@
...
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidatorService;
+
+/**
+ * Validates that structure members and union member names do not
+ * repeat their shape name as prefixes of their member or tag names.
+ */
+public final class RepeatedShapeNameValidator extends AbstractValidator {
+
+    public static final class Config {
```
kstich, line 38: This configuration isn't mentioned in the spec.

smithy-linters/src/main/java/software/amazon/smithy/linters/RepeatedShapeNameValidator.java
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            return lowerCaseMemberName.startsWith(lowerCaseShapeName);
+        }
+    }
+
+    private ValidationEvent repeatedMemberName(Model model, Shape shape, String shapeName, String memberName) {
+        Shape member = model.expectShape(shape.getId().withMember(memberName));
+        return warning(member, String.format(
+                "The `%s` %s shape repeats its name in the member `%s`; %2$s member names should not be "
+                        + "prefixed with the %2$s name.", shapeName, shape.getType(), memberName));
```
kstich, line 97: Should `prefixed` toggle with the `getExactMatch` status?

kstich (changes requested):

kstich (approved):

mtdowling (approved):

PR #1044, opened 2022-01-08
Reduce HttpBindingsMissing severity for non-REST services

adamthom-amzn:
*Description of changes:*
HttpBindingsMissing emits an ERROR event for all services that have
inconsistently @http traited operations, but not all protocols support @http.
This change inspects the list of supported traits for every protocol trait
applied to the service, and if none of them support the http trait, then
HttpBindingsMissing is emitted at WARNING, instead of ERROR.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpBindingsMissingValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpBindingsMissingValidator.java
@@ -20,11 +20,14 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.ServiceIndex;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
 import software.amazon.smithy.model.shapes.OperationShape;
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpBindingsMissingValidator.java
```diff
@@ -41,29 +43,44 @@
...
         return operations.stream()
                 .filter(shape -> !shape.getTrait(HttpTrait.class).isPresent())
-                .map(shape -> createEvent(service, shape))
+                .map(shape -> createEvent(severity, service, shape))
                 .collect(Collectors.toList());
     }
 
+    // Only emit ERROR events if any of the protocols for the service support HTTP traits
+    private Severity determineSeverity(ServiceShape service, Model model) {
+        if (service.getAllTraits().values()
```
mtdowling, line 67: You could use https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java#L78 instead

mtdowling (changes requested):

mtdowling (approved):

PR #1056, opened 2022-01-14
Add outputToken value clarification for clients

### LINKED ISSUE #1045: Clarify Pagination behavior when tokens are empty vs. unset
EC2 replies with `<nextToken></nextToken>` which is `""` and not `None`. On the surface, it _seems_ like if the token type is a string, we should consider it as unset if it's None or empty. However, It seems like it would also be possible for this to behave similarly with structures that have been filled with defaults because no output was sent.

1. The pagination docs should clearly call out this edge case
2. Clarify/investigate behavior of pagination for unset structured shapes.

https://github.com/awslabs/aws-sdk-rust/issues/391

It may also be worth suggesting that paginators should include a check of an unchanging outputToken and terminate to prevent runaway queries.

kstich:
Resolves #1045 

### DIFF
--- a/docs/source/1.0/spec/core/behavior-traits.rst
+++ b/docs/source/1.0/spec/core/behavior-traits.rst
@@ -201,12 +201,12 @@ The ``paginated`` trait is a structure that contains the following members:
     * - outputToken
       - ``string``
       - The path to the operation output member that contains an optional
-        continuation token. When this value is present in operation output,
-        it indicates that there are more results to retrieve. To get the next
-        page of results, the client passes the received output continuation
-        token to the input continuation token of the next request. This
-        output member MUST NOT be marked as ``required`` and SHOULD target a
-        string shape. It can, but SHOULD NOT target a map shape.
+        continuation token. When this value is present and not empty in
+        operation output, it indicates that there are more results to retrieve.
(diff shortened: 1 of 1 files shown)


rcoh:
LGTM! It might be nice to include some example APIs to test against that expose uncommon behavior, eg. DynamoDB scan items (uses a hashmap), the EC2 APIs with empty tokens. Ideally, of course, we'd have some form of protocol tests for pagination but that's a bigger lift. I understand if we don't want to call out AWS specific stuff in the docs though

docs/source/1.0/spec/core/behavior-traits.rst
```diff
@@ -406,8 +406,9 @@ client MUST follow when iterating over paginated API calls:
    at which pagination occurs.
 
 #. If the received response does not contain a continuation token in the
-   referenced ``outputToken`` member, then there are no more results to
-   retrieve and the process is complete.
+   referenced ``outputToken`` member (either the member is not set or is set to
+   an empty value,) then there are no more results to retrieve and the process
```
mtdowling, line 410:
```suggestion
   an empty value), then there are no more results to retrieve and the process
```

mtdowling (approved):

PR #1059, opened 2022-01-14
Add ability to serialize the JMESPath AST

mtdowling:
This allows the JMESPath AST to be parsed, then modified, then reserialized. Use cases include things like being able to modify
JMESPath expressions found in waiters so that field extraction matches any normalization changes code generators make to object keys (e.g., using snake_case in Ruby as opposed to the member names provided in the model).

To test, every valid expression from the jmespath test suite is parsed, serialized, and reparsed. The originally parsed AST is compared against the reparsed AST to ensure they are equal. Every JMESPath expression used in AWS waiters is included as a test case too.

*Issue #, if available:*

*Description of changes:*


JordonPhillips (approved):

mullermp:
Nice!

PR #1061, opened 2022-01-18
Allow shapes to be bound directly to a service

kstich:
This feature allows shapes to be directly bound to a service, meaning
that the service closure can now contain shapes not bound through its
resources or operations. This is useful for vending shapes that match
other integrations, like CloudWatch events or SNS messages. It can
also provide shapes that match known schemas for document contents.

It also updates the `@cfnResource` trait to require that any shape
listed as an `additionalSchema` is within the closure of any service
that contains the resource.

Existing code generation tooling will need to be updated to handle
generating these shapes and [de]serialization code for them.

### DIFF
--- a/docs/source/1.0/spec/core/json-ast.rst
+++ b/docs/source/1.0/spec/core/json-ast.rst
@@ -405,6 +405,12 @@ shapes defined in JSON support the same properties as the Smithy IDL.
         closure of the service can return. Each provided shape ID MUST target
         a :ref:`structure <structure>` shape that is marked with the
         :ref:`error-trait`.
+    * - shapes
+      - [:ref:`AST shape reference <ast-shape-reference>`]
+      - Defines a list of shapes bound directly to the service. Each provided
+        shape ID MUST NOT target a :ref:`service <service>`, :ref:`resource
+        <resource>`, :ref:`operation <operation>`, :ref:`member <member>`, or
+        :ref:`trait shape <trait-shapes>`.
     * - traits
       - map of :ref:`shape ID <shape-id>` to trait values
       - Traits to apply to the service
@@ -434,6 +440,11 @@ shapes defined in JSON support the same properties as the Smithy IDL.
                 "errors": [
                     {
                         "target": "smithy.example#SomeError"
+                    }
+                ],
+                "shapes": [
+                    {
+                        "target": "smithy.example#SomeShape"
                     }
                 ],
                 "traits": {
--- a/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceTraitValidator.java
+++ b/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceTraitValidator.java
@@ -20,8 +20,8 @@ import java.util.List;
 import java.util.Set;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
-import software.amazon.smithy.model.selector.PathFinder;
-import software.amazon.smithy.model.selector.Selector;
+import software.amazon.smithy.model.neighbor.RelationshipDirection;
+import software.amazon.smithy.model.neighbor.Walker;
 import software.amazon.smithy.model.shapes.ResourceShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.ShapeId;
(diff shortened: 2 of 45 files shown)


smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceTraitValidator.java
```diff
@@ -0,0 +1,91 @@
...
+            CfnResourceTrait trait
+    ) {
+        List<ValidationEvent> events = new ArrayList<>();
+
+        for (ShapeId additionalSchemaId : trait.getAdditionalSchemas()) {
+            Selector selector = Selector.parse("[id = " + additionalSchemaId + "]");
+
+            // Emit an event if we don't have a path to the additionalSchema from the
+            // service shape this resource is bound to.
+            if (PathFinder.create(model).search(service.getId(), selector).isEmpty()) {
```
JordonPhillips, line 82: nit: This is very clever, but wouldn't it be more performant to compute the shapes connected to the service once and then do a contains check for each additional schema?

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedServiceBoundShape.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private List<ValidationEvent> createShapeBindingEvents(ChangedShape<ServiceShape> change) {
+        if (change.getOldShape().getShapes().equals(change.getNewShape().getShapes())) {
+            return Collections.emptyList();
+        }
+
+        List<ValidationEvent> events = new ArrayList<>();
+        for (ShapeId id : change.getOldShape().getShapes()) {
+            if (!change.getNewShape().getShapes().contains(id)) {
+                events.add(warning(change.getNewShape(), String.format(
```
JordonPhillips, line 47: This should at least be a danger right? Since we're expecting these to be code generated

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConfig.java
```diff
@@ -368,7 +371,12 @@ public void setSupportNonNumericFloats(boolean supportNonNumericFloats) {
         this.supportNonNumericFloats = supportNonNumericFloats;
     }
 
+    @Deprecated
     public boolean isEnableOutOfServiceReferences() {
+        if (enableOutOfServiceReferences) {
+            LOGGER.warning("Deprecated JSON Schema setting `enableOutOfServiceReferences` enabled. "
+                    + "Bind shapes directly to the service instead.");
+        }
```
JordonPhillips, line 379: Were there any docs for this that need updated?

JordonPhillips (changes requested):
I'm sure it's on your to-do list, but these need to be accounted for in mixins on the idl-2 branch.

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConfig.java
kstich: No, this was a code-only configuration option that was provided.

docs/source/1.0/spec/core/json-ast.rst
```diff
@@ -405,6 +405,12 @@ shapes defined in JSON support the same properties as the Smithy IDL.
         closure of the service can return. Each provided shape ID MUST target
         a :ref:`structure <structure>` shape that is marked with the
         :ref:`error-trait`.
+    * - shapes
+      - [:ref:`AST shape reference <ast-shape-reference>`]
+      - Defines a list of shapes bound directly to the service. Each provided
+        shape ID MUST NOT target a :ref:`service <service>`, :ref:`resource
+        <resource>`, :ref:`operation <operation>`, :ref:`member <member>`, or
+        :ref:`trait shape <trait-shapes>`.
```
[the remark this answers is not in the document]
kstich: The alt text is used to prevent strange capitalizations in the documentation, as is done elsewhere in the spec.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedServiceBoundShape.java
kstich: I'll up this to `danger`, as it's possible that the shape is bound through another instead of directly now.

JordonPhillips (approved):

kstich:
Closing as this is outdated with recent changes to model parsing.

kaspar-p:
This issue is linked in the codegen documentation (https://smithy.io/2.0/guides/building-codegen/overview-and-concepts.html#type-generation) but is marked here as cancelled. Is it being replaced by another task to track "type generation on a service"?

PR #1064, opened 2022-01-21
Default zero value docs

JordonPhillips:


docs/source/1.0/spec/core/model.rst
```diff
@@ -790,22 +786,34 @@ by ``$``, followed by the member name. For example, the shape ID of the ``foo``
...
+Code generated types for structures SHOULD use the
+:ref:`default <default-trait>` and :ref:`required <required-trait>` traits to
+provide member accessors that always return non-null values.
+
+- When the :ref:`default-trait` is present on a member, the corresponding
+  accessor SHOULD always return a non-null value by defaulting missing members
+  with their :ref:`zero values <default-values>`.
+- When the :ref:`required-trait` is present on a member, the corresponding
+  accessor SHOULD always return a non-null value.
+- Smithy implementations in languages like TypeScript that do not provide a kind
```
mtdowling, line 807: I think this might be too much in the weeds.

Another callout that could be here is something like: "An explicitly provided default zero value and a member that defaults to the zero value are indistinguishable." (at least everywhere outside the (de)serialization layer of protocols). This in effect makes members set to the zero value equivalent to nullable members in clients that do not support default zero values.

docs/source/1.0/spec/core/type-refinement-traits.rst
```diff
@@ -11,60 +11,126 @@ the type of a shape.
...
+    Marks a structure member as having a default, zero value.
 Trait selector
-    .. code-block:: none
-
-        :test(boolean, byte, short, integer, long, float, double,
-              member > :test(boolean, byte, short, integer, long, float, double))
+    ``structure > member :not(> :test(union, structure > :test([trait|required])))``
 
-    *A boolean, byte, short, integer, long, float, double shape or a member that targets one of these shapes*
+    A member of a structure that does not target a union or structure.
```
mtdowling, line 27: The default trait works on structures so long as the structure has all optional members (no required members)

docs/source/1.0/spec/core/type-refinement-traits.rst
```diff
@@ -11,60 +11,126 @@ the type of a shape.
...
-                    }
-                }
-            }
-        }
+Default zero values
+===================
+
+The following table describes the default zero value of each kind of shape.
+Programming languages and code generators that cannot initialize structure
+members with the following default values SHOULD continue to represent those
```
mtdowling, line 51:
```suggestion
members with the following default values SHOULD represent those
```

docs/source/1.0/spec/core/type-refinement-traits.rst
```diff
@@ -11,60 +11,126 @@ the type of a shape.
...
+      - Description
+    * - boolean
+      - ``false``
+      - Boolean false.
+    * - numbers
+      - ``0``
+      - Numeric zero.
+    * - string
+      - empty string
+      - Strings with the enum trait also have the same default value.
```
mtdowling, line 70:
```suggestion
      - Strings with the :ref:`enum-trait` also have an empty string zero value.
```

mtdowling (changes requested):

mtdowling (approved):

PR #1073, opened 2022-02-01
Move traits to a single page

JordonPhillips:
This moves the non-aws trait definitions to a single page. To keep things somewhat simple in this pr, each previous page is now a template file, and all the template files are `include`'d. The only material change to the contents was to change the heading levels where necessary.

Additionally, the definition of the trait trait was moved onto this page.


mtdowling:
Something is funny with the `trait` trait. It's showing up under the "Traits" page, but also as a separate top-level entry in the Table of Contents.

image

mtdowling:
image

That link links within the same document. Can be removed.

JordonPhillips:
That seelaso link links to the description in the smithy model docs, not the trait trait on the same page

mtdowling (approved):

JordonPhillips:
This seems to have been undone in a recent rebase, so I re-merged it.

PR #1077, opened 2022-02-03
Add presignable trait to Smithy

kggilmer:
*Issue #, if available:* N/A

*Description of changes:*
* Adds an annotation trait that provides context about which specific service operations may have client-side presigners.  This state is currently modeled in customizations in various SDKs.  Some samples: Java v2 (in code, eg [S3](https://github.com/aws/aws-sdk-java-v2/blob/8586a54bf926316e3236a6f11d8142287c97fcc7/services/s3/src/main/java/software/amazon/awssdk/services/s3/presigner/S3Presigner.java)), [Go v2](https://github.com/aws/aws-sdk-go-v2/blob/4f7aaaf16d6d6c59223bbbf069ddc9c862c52095/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java#L83), and [Kotlin SDKs](https://github.com/awslabs/aws-sdk-kotlin/blob/72828c24da95eab7093739922618cd172c529487/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/PresignableOperations.kt#L12).

NOTE: this trait only models a specific use case for presigners: generators.  It does *not* model "presign url autofill parameters" of some services due to current migration away from this to other approaches.

NOTE: This PR was initially created [here](https://github.com/awslabs/smithy/pull/897), but moved to personal fork as repo rights were removed during PR work.

### DIFF
--- a/docs/source/1.0/spec/aws/aws-auth.rst
+++ b/docs/source/1.0/spec/aws/aws-auth.rst
@@ -142,6 +142,41 @@ literal string ``UNSIGNED-PAYLOAD`` is used when constructing a
 `canonical request`_, and the same value is sent in the
 `x-amz-content-sha256`_ header when sending an HTTP request.
 
+.. smithy-trait:: aws.auth#presignable
+.. _aws.auth#presignable-trait:
+
+------------------------------
+``aws.auth#presignable`` trait
+------------------------------
+
+Trait Summary
+    Indicates that an operation can be presigned. A presigner is a client-side utility 
+    that generates a presigned request for a given operation invocation that may be 
+    used in contexts where direct calls may be difficult or impossible. This trait 
+    indicates that a client presigner can be generated for the given operation.
+Trait selector
+    ``operation``
+Trait value
+    Annotation trait.
+
+The following example defines a service operation for which presigned requests can be generated.
+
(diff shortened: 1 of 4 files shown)


mullermp:
I like the idea of code generated presigners. It provides a good customer experience similar to waiters, paginators, resource models, etc. I do think however that presigning is a function of Sigv4/a and the signer, not generic service modeling. I think if there was a presignable trait, it certainly needs to only apply to sigv4 services or possibly a property baked into sigv4/sigv4a traits. Since pre-signed urls work for any sigv4 operation in theory, I would prefer to see SDKs have a "presigned_url: true" option passed to operation methods as options, like `create_bucket(bucket: 'foo', presigned_url: true)`, that generated a pre-signed url (as opposed to sending the request) for any sigv4 operation for any service.

kstich:
We haven't validated that an operation-level trait is the correct path forward for modeling if an operation can have a presigned url generated _or_ how this would need to interact with other traits, like `sigv4`. Given this and that this PR is now significantly out of date, I'm closing it. We can reopen this PR if it's the basis for the future design

PR #1088, opened 2022-02-14
Enum shapes

JordonPhillips:
This PR adds enum shapes. Documentation will be part of a separate PR to keep this a bit smaller, but also because I'm still waiting on another review for #1073 which makes it hard to add new trait docs.


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -367,6 +368,18 @@ public ModelAssembler addShapes(Shape... shapes) {
         return this;
     }
 
+    /**
+     * Sets the Smithy version to use for parsed shapes added directly to the
+     * assembler.
```
mtdowling, line 373: What happens when this isn't set? Can that be documented?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -511,6 +541,18 @@ private boolean isEmptyStructure(Node node, Shape shape) {
             return !shape.isDocumentShape() && node.asObjectNode().map(ObjectNode::isEmpty).orElse(false);
         }
 
+        @Override
+        public Void enumShape(EnumShape shape) {
+            shapeWithMembers(shape, new ArrayList<>(shape.members()), true);
```
mtdowling, line 546: Why the need to create a separate ArrayList of the members?

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumTrait.java
```diff
@@ -28,19 +28,23 @@
 /**
  * Constrains string values to one of the predefined enum constants.
  */
-public final class EnumTrait extends AbstractTrait implements ToSmithyBuilder<EnumTrait> {
+public class EnumTrait extends AbstractTrait implements ToSmithyBuilder<EnumTrait> {
```
mtdowling, line 31: Can you add some docs about this trait and reference the synthetic trait variant? Maybe document that this is deprecated and to use the enum shape.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumTrait.java
```diff
@@ -28,19 +28,23 @@
...
     public static final ShapeId ID = ShapeId.from("smithy.api#enum");
 
-    private final List<EnumDefinition> definitions;
+    protected final List<EnumDefinition> definitions;
```
mtdowling, line 34: Why protected? I think protected properties should be extremely rare and hidden behind, at most, a protected method to access the property. Bonus points would be to make sure this class can't be extended outside this package if possible.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/IntEnumShape.java
```diff
@@ -0,0 +1,259 @@
...
+    private final Map<String, MemberShape> members;
+    private volatile List<String> memberNames;
+
+    private IntEnumShape(Builder builder) {
+        super(builder);
+        members = NamedMemberUtils.computeMixinMembers(
+                builder.getMixins(), builder.members, getId(), getSourceLocation());
+        validateMemberShapeIds();
+        if (members.size() < 1) {
+            throw new SourceException("intEnum shapes must have at least one member", getSourceLocation());
```
mtdowling, line 40: Oh right. This will be good to also document in the spec.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/NamedMembers.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    Map<String, MemberShape> getAllMembers();
+
+    /**
+     * Returns an ordered list of member names based on the order they are
+     * defined in the model, including mixin members.
+     *
+     * @return Returns an immutable list of member names.
+     */
+    List<String> getMemberNames();
```
mtdowling, line 37: Can this not have a default implementation?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -0,0 +1,334 @@
...
+        return members;
+    }
+
+    /**
+     * Returns an ordered list of member names based on the order they are
+     * defined in the model, including mixin members.
+     *
+     * @return Returns an immutable list of member names.
+     */
+    public List<String> getMemberNames() {
```
mtdowling, line 63: Can you add the `@Override` annotations?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/IntEnumShape.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+import software.amazon.smithy.model.SourceException;
+import software.amazon.smithy.model.traits.EnumValueTrait;
+import software.amazon.smithy.model.traits.UnitTypeTrait;
+import software.amazon.smithy.utils.BuilderRef;
+import software.amazon.smithy.utils.ListUtils;
+
+public final class IntEnumShape extends IntegerShape {
```
mtdowling, line 29: Would it be useful to have a method like `getEnumValues` that returns a map of enum member names to their values?

mtdowling (changes requested):
Will spec updates and default value updates be done in a different PR?

designs/enum-shapes.md
```diff
@@ -0,0 +1,259 @@
...
+    suit: Suit
+}
+```
+
+To account for this, enums MAY define a default member by making the
+`@enumValue` of the member "":
+
+```
+enum Suit {
+    @enumValue(string: "")
```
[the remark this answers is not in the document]
JordonPhillips: Are you suggesting we disallow setting those values explicitly?

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,18 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
         return new ChangeShapeType(shapeToType).transform(this, model);
     }
 
+    /**
+     * Changes each compatible string shape with the enum trait to an enum shape.
+     *
+     * <p>Strings with enum traits that don't define names are not converted.
```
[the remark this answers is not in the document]
JordonPhillips: Not safely, at least not in all cases. Those values could be literally anything that fits in a string. Automatically converting them would be risky because we could introduce collisions. We could convert those that are safely convertible though.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ChangeShapeType.java
```diff
@@ -96,6 +113,17 @@ public Shape integerShape(IntegerShape shape) {
             return copyToSimpleShape(to, shape);
         }
 
+        @Override
+        public Shape intEnumShape(IntEnumShape shape) {
+            if (to.getCategory() != ShapeType.Category.SIMPLE) {
+                throw invalidType(shape, to, "Enum types can only be converted to simple types.");
```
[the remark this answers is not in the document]
JordonPhillips: No, this can be converted to other simple types too. The string enum checks for it explicitly because it needs to do some extra leg work when the target is a string, but that isn't the case for intEnum since it doesn't need to mess with the enum trait.

smithy-build/src/main/java/software/amazon/smithy/build/transforms/ChangeStringEnumsToEnumShapes.java
```diff
@@ -0,0 +1,32 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.build.transforms;
+
+import software.amazon.smithy.build.ProjectionTransformer;
+import software.amazon.smithy.build.TransformContext;
+import software.amazon.smithy.model.Model;
+
+public final class ChangeStringEnumsToEnumShapes implements ProjectionTransformer {
```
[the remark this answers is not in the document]
JordonPhillips: Yeah the proper docs I'm leaving to another pr since this is already huge. Mentioned in another comment, but we can synthesize in most cases.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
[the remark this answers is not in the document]
mtdowling: I think we should have something that does a best effort through at least an opt-in. Otherwise no AWS SDK team could use this I think.

designs/enum-shapes.md
```diff
@@ -0,0 +1,259 @@
...
+members will now be present in simple shapes, we will add a new shape type
+called “member” to go alongside simple, aggregate, and service types.
+
+## Smithy 1.0 → 2.0
+
+Since the enum trait does not require names, it is not possible to
+automatically convert string shapes using the `@enum` trait into enum traits.
+Instead, a transformer will be added that upgrades those shapes that can
+be upgraded. To make this change backward compatible for existing code
+generators, enum shapes in IDL 2.0 will always contain an enum trait with
```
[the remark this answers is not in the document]
mtdowling: Wouldn't old generators just see a string since enum is a subtype of string? That seems forward compatible to me. It would work the same way as intEnum and integer.

designs/enum-shapes.md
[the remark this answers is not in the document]
mtdowling: Yes, mostly because of how `0` has a big impact on `intEnum` when used with the `@default` trait. People might not realize when setting the value to `0` that they are also defining the default value of the enum. By adding a new property or trait, they have to realize it because it's on the tin.

designs/enum-shapes.md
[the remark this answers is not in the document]
JordonPhillips: They would see a string *without the enum trait*. If somebody were to switch their string shapes to enum shapes, that would be a backwards incompatible change for generators that haven't updated.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,18 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
...
+     *
+     * @param model Model to transform.
+     * @return Returns the transformed model.
+     */
+    public Model changeStringEnumsToEnumShapes(Model model) {
```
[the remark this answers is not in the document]
JordonPhillips: This mostly shouldn't be necessary, except if they're looking at shape type. But maybe they are.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -455,6 +464,13 @@ private void parseSimpleShape(ShapeId id, SourceLocation location, AbstractShape
         parseMixins(id);
     }
 
+    private void parseEnumShape(ShapeId id, SourceLocation location, AbstractShapeBuilder builder) {
```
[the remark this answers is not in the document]
JordonPhillips: No, that's handled in `onShape` in `AbstractMutableModelFile`. I did it that way so as to not have to duplicate the check for AST and IDL.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -511,21 +531,27 @@ private void parseMember(ShapeId parent, Set<String> allowed, Set<String> define
...
-        String target = ParserUtils.parseShapeId(this);
         modelFile.onShape(memberBuilder);
+        String target;
+
+        if (!targetsUnion) {
+            ws();
+            expect(':');
+
+            if (peek() == '=') {
+                throw syntax("Defining structures inline with the `:=` syntax may only be used when "
```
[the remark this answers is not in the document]
JordonPhillips: No, you can only expect a particular character. I'm really not sure what the issue you have with the message is, it seems extremely clear to me and doesn't to my mind imply you can use it elsewhere. In fact, it says exactly the opposite.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -493,7 +513,7 @@ private void parseMembers(ShapeId id, Set<String> requiredMembers) {
         expect('}');
     }
 
-    private void parseMember(ShapeId parent, Set<String> allowed, Set<String> defined) {
+    private void parseMember(ShapeId parent, Set<String> allowed, Set<String> defined, boolean targetsUnion) {
```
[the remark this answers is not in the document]
JordonPhillips: haha whoops, that was supposed to be `targetsUnit`

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumValueTrait.java
```diff
@@ -0,0 +1,118 @@
...
+        return Optional.ofNullable(integer);
+    }
+
+    public static final class Provider extends AbstractTrait.Provider {
+        public Provider() {
+            super(ID);
+        }
+
+        @Override
+        public Trait createTrait(ShapeId target, Node value) {
```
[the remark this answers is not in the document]
JordonPhillips: We haven't rebased that into idl-2.0 yet

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
JordonPhillips: Ah it was just because the method expects a list. There's no real need for it to be a list, just for it to be ordered. But since those are ordered anyway I changed it to collection.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumTrait.java
JordonPhillips: That was a holdover from a point in time where the synthetic trait was doing more complicated things. It has been fixed.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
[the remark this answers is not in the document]
mtdowling: What I meant that these checks aren't needed and are getting put in multiple places across the parser. Without this check, parsing would still fail if someone uses ":=".

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -0,0 +1,413 @@
...
+                            "Unable to convert enum trait entry with name: `%s` and value `%s` to an enum member.",
+                            definition.getName().orElse(""), definition.getValue()
+                    ));
+                }
+            }
+
+            return this;
+        }
+
+        public Builder members(EnumTrait trait) {
```
mtdowling, line 255: I don't get what this is doing. Is this actually something like "setMembersFromEnumTrait" or something?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -0,0 +1,413 @@
...
+        }
+
+        @Override
+        public Builder addMember(MemberShape member) {
+            if (!member.getTarget().equals(UnitTypeTrait.UNIT)) {
+                throw new SourceException(String.format(
+                        "Enum members may only target `smithy.api#Unit`, but found `%s`", member.getTarget()
+                ), getSourceLocation());
+            }
+            if (!member.hasTrait(EnumValueTrait.ID) && !member.hasTrait(EnumDefaultTrait.ID)) {
```
mtdowling, line 290: It's interesting that this is adding a default EnumValueTrait. That would be good to call out somewhere in javadocs.

A consequence of this is that you couldn't add a selector that warns when `@enumValue` isn't set. Not saying that's necessary, but it's not possible when we add default traits like this.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -0,0 +1,413 @@
...
+    public static final class Builder extends StringShape.Builder {
+        private final BuilderRef<Map<String, MemberShape>> members = BuilderRef.forOrderedMap();
+
+        @Override
+        public EnumShape build() {
+            addSyntheticEnumTrait();
+            return new EnumShape(this);
+        }
+
+        private void addSyntheticEnumTrait() {
```
mtdowling, line 204: This really needs some docs as to why we're doing this

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -0,0 +1,413 @@
...
+        if (enumValues == null) {
+            Map<String, String> values = new LinkedHashMap<>(members.size());
+            for (MemberShape member : members()) {
+                if (member.hasTrait(EnumDefaultTrait.ID)) {
+                    values.put(member.getMemberName(), "");
+                    continue;
+                }
+                values.put(member.getMemberName(), member.expectTrait(EnumValueTrait.class).expectStringValue());
+            }
+            enumValues = MapUtils.orderedCopyOf(values);
```
mtdowling, line 67: nit you don't need to copy here since the map doesn't escape this method. Just a `Collections.unmodifiableMap` works.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/IntEnumShape.java
```diff
@@ -0,0 +1,320 @@
...
+        if (enumValues == null) {
+            Map<String, Integer> values = new LinkedHashMap<>(members.size());
+            for (MemberShape member : members()) {
+                if (member.hasTrait(EnumDefaultTrait.ID)) {
+                    values.put(member.getMemberName(), 0);
+                    continue;
+                }
+                values.put(member.getMemberName(), member.expectTrait(EnumValueTrait.class).expectIntValue());
+            }
+            enumValues = MapUtils.orderedCopyOf(values);
```
mtdowling, line 63: nit no need to copy here

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeType.java
```diff
@@ -34,27 +35,42 @@
     BIG_DECIMAL("bigDecimal", BigDecimalShape.class, Category.SIMPLE),
     BIG_INTEGER("bigInteger", BigIntegerShape.class, Category.SIMPLE),
 
+    ENUM("enum", EnumShape.class, Category.SIMPLE, (a, b) -> b.equals(a) || b.equals(STRING)),
```
mtdowling, line 38: nit: I think since these are enums, you can use `==`

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumDefaultTrait.java
```diff
@@ -0,0 +1,52 @@
...
+import software.amazon.smithy.model.shapes.IntEnumShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+/**
+ * Indicates that an enum member is the default value member.
+ *
+ * On an {@link IntEnumShape} this implies the enum value is 0. On an
+ * {@link  EnumShape} this implies the enum value is an empty string.
+ */
+public class EnumDefaultTrait extends AnnotationTrait {
```
mtdowling, line 32: final

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ExclusiveStructureMemberTraitValidator.java
```diff
@@ -51,30 +51,30 @@
         }
 
         List<ValidationEvent> events = new ArrayList<>();
-        for (StructureShape shape : model.getStructureShapes()) {
+        model.shapes().filter(shape -> shape instanceof NamedMembers).forEach(shape -> {
```
mtdowling, line 54: nit: Can this just use a for loop with an if statement? It removes all the stream pipelining abstractions

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/EnumShapeValidator.java
```diff
@@ -0,0 +1,120 @@
...
+import software.amazon.smithy.model.shapes.EnumShape;
+import software.amazon.smithy.model.shapes.IntEnumShape;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.EnumDefaultTrait;
+import software.amazon.smithy.model.traits.EnumValueTrait;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.ValidationEvent;
+
+public final class EnumShapeValidator extends AbstractValidator {
```
mtdowling, line 35: Can you list out what this validates?

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,62 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
...
+     *     <li>Union to structure</li>
+     * </ul>
+     *
+     * @param model Model to transform.
+     * @param shapeToType Map of shape IDs to the new type to use for the shape.
+     * @param synthesizeEnumNames Whether enums without names should have names synthesized if possible.
+     * @return Returns the transformed model.
+     * @throws ModelTransformException if an incompatible type transform is attempted.
+     */
+    public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType, boolean synthesizeEnumNames) {
```
mtdowling, line 503: Rather than pass in a boolean, for future evolution of this method, it would be nice to pass in a variadic list of ChangeShapeTypeOption values, one of which is `SYNTHESIZE_ENUM_NAMES`.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,62 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
...
+    }
+
+    /**
+     * Changes each compatible string shape with the enum trait to an enum shape.
+     *
+     * @param model Model to transform.
+     * @param synthesizeEnumNames Whether enums without names should have names synthesized if possible.
+     * @return Returns the transformed model.
+     */
+    public Model changeStringEnumsToEnumShapes(Model model, boolean synthesizeEnumNames) {
```
mtdowling, line 514: This boolean is probably fine here since the method is so special purpose

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,62 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
...
+        return ChangeShapeType.upgradeEnums(model, false).transform(this, model);
+    }
+
+    /**
+     * Changes each enum shape to a string shape and each intEnum to an integer.
+     *
+     * @param model Model to transform.
+     * @return Returns the transformed model.
+     */
+    public Model changeEnumsToBaseShapeTypes(Model model) {
```
mtdowling, line 536: I kind of like just calling this "downgradeEnums"

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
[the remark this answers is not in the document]
mtdowling: Can we just require them to pass in `true` / `false` rather than add the overload?

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -481,6 +481,62 @@ public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType) {
...
+     * @param synthesizeEnumNames Whether enums without names should have names synthesized if possible.
+     * @return Returns the transformed model.
+     * @throws ModelTransformException if an incompatible type transform is attempted.
+     */
+    public Model changeShapeType(Model model, Map<ShapeId, ShapeType> shapeToType, boolean synthesizeEnumNames) {
+        return new ChangeShapeType(shapeToType, synthesizeEnumNames).transform(this, model);
+    }
+
+    /**
+     * Changes each compatible string shape with the enum trait to an enum shape.
```
mtdowling, line 508: Can you document what this actually means?

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumTrait.java
```diff
@@ -21,26 +21,39 @@
...
+import software.amazon.smithy.model.shapes.EnumShape;
 import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.synthetic.SyntheticEnumTrait;
 import software.amazon.smithy.utils.BuilderRef;
 import software.amazon.smithy.utils.ToSmithyBuilder;
 
 /**
  * Constrains string values to one of the predefined enum constants.
+ *
+ * This trait is deprecated, use an {@link EnumShape} instead.
```
mtdowling, line 33: nit: paragraphs should use `<p>` tags

designs/enum-shapes.md
```diff
@@ -0,0 +1,272 @@
...
+    CLUB
+    HEART
+    SPADE
+}
+```
+
+Each value listed in the enum is a member that implicitly targets
+`smithy.api#Unit`. The string representation of an enum member defaults to the
+member name. The string representation can be customized by applying the
+`@enumValue` trait.
```
mtdowling, line 52: In the implementation, you add a default `@enumValue` trait if one isn't present nor is the `@enumDefault` present. Should that baked into the spec? Something about how members with neither of these traits are indistinguishable from members that define an explicit `@enumValue` trait? That way, it won't matter that we default the trait, and people can always rely on using things like selectors to get enum values without needing to worry about what the implementation does (if it defaults or not).

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumDefinition.java
```diff
@@ -108,6 +114,93 @@ public static EnumDefinition fromNode(Node node) {
...
+        return getName().isPresent() || (withSynthesizedNames && CONVERTABLE_VALUE.matcher(getValue()).find());
+    }
+
+    /**
+     * Converts an enum member into an equivalent enum definition object.
+     *
+     * @param member The enum member to convert.
+     * @return An {@link EnumDefinition} representing the given member.
+     */
+    public static EnumDefinition fromMember(MemberShape member) {
```
mtdowling, line 186: These newly introduced methods are only ever used in the model transform right? I kinda wish these were hidden away in that transform and not publicly callable. Is that possible, or are these used elsewhere?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
JordonPhillips: The alternative to making this default is that you always have to check two things. I'm not against it necessarily, especially since we've already got the method to just get a map of name-value pairs.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumDefinition.java
JordonPhillips: It's used in the actual shape, which has a `fromStringShape` static method, but we could also just move that.

mtdowling: The more hidden this is the better IMO as that allows us to change it more over time.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumValueTrait.java
```diff
@@ -97,13 +96,12 @@ public Provider() {
...
-                    .map(v -> v.expectStringNode().getValue())
-                    .ifPresent(builder::stringValue);
-            objectNode.getMember("int")
-                    .map(v -> v.expectNumberNode().getValue().intValue())
-                    .ifPresent(builder::intValue);
+            value.asStringNode().ifPresent(node -> builder.stringValue(node.getValue()));
+            value.asNumberNode().ifPresent(node -> {
+                if (node.isNaturalNumber()) {
+                    builder.intValue(node.getValue().intValue());
+                }
```
mtdowling, line 103: Shouldn't this have an `else` that throws?

JordonPhillips: The constructor will throw if neither is set, but this could probably use a more specific error.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumDefinition.java
JordonPhillips: I did end up still needing to expose a method to ask whether a shape can be converted, but everything else was able to be hidden away on the enum shape itself.

smithy-model/src/main/java/software/amazon/smithy/model/traits/EnumValueTrait.java
mtdowling: If the number has a fractional part, then the constructor will throw saying no value was provided? Ideally it would say we don't support fractional precision right? Only integers.

smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java
```diff
@@ -617,4 +622,24 @@ public Model createDedicatedInputAndOutput(Model model, String inputSuffix, Stri
     public Model flattenAndRemoveMixins(Model model) {
         return new FlattenAndRemoveMixins().transform(this, model);
     }
+
+    /**
+     * Options that can be enabled when changing shape types.
+     */
+    public enum ChangeShapeTypeOption {
```
mtdowling, line 629: nit: this class is alrady big, can we just make this top-level?

designs/enum-shapes.md
```diff
@@ -125,6 +125,41 @@ structure GetFooInput {
 
 Every enum shape MUST define at least one member.
 
+#### enum members always have a value
+
+Enum members that have neither the `@enumValue` nor the `@enumDefault` trait
```
mtdowling, line 130: Can we require that a default enumValue is added then? Otherwise, you can't reliably use a selector that gets an enumValue from an enum member without knowing the implementation details of the Smithy model implementation.

PR #1089, opened 2022-02-14
docs: `length` trait counts number of Unicode scalar values for strings

david-perez:
Since Smithy defines string shapes to be UTF-8 encoded, it is more
correct for the spec to explicitly say the `length` trait counts the
number of Unicode scalar values when applied to strings.

### DIFF
--- a/docs/source/1.0/spec/core/constraint-traits.rst
+++ b/docs/source/1.0/spec/core/constraint-traits.rst
@@ -360,7 +360,7 @@ Shape        Length constrains
 list         The number of members
 set          The number of members
 map          The number of key-value pairs
-string       The number of Unicode code points
+string       The number of `Unicode scalar values <https://www.unicode.org/glossary/#unicode_scalar_value>`__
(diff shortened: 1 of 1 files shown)


david-perez:
The previous wording could be confusing in that someone could misinterpret the `length` trait as counting the Unicode code points in the surrogate pairs range from D800 to DFFF too, which is UTF-16 specific. Since Smithy only contemplates "UTF-8 encoded `string` shapes[^1]", it's impossible to count Unicode code points from that range. The set of Unicode scalar values is precisely the set of all Unicode code points minus that range.

So both wordings are _technically_ correct given the context, but this new one is context-independent and in my opinion clearer.

> Is it saying the same thing as what JSON schema's length validator in Java already does (i.e., value.codePointCount(0, value.length())), which matches how API Gateway interprets length constraints too?

Yes.

[^1]: whatever that may mean, since in my opinion there's not a clear explanation of what that statement actually entails in the spec

mtdowling (approved):

PR #1097, opened 2022-02-18
Add a request test for sets of documents

adamthom-amzn:
*Description of changes:*
Documents are deserialized as presented, unlike structures, which are
typically parsed and then specific values extracted, which can erase the
difference between a key with a null value and an undefined key. This is
particularly interesting for sets, where a null value may make a value unique
where it would not in a set of structures.

Tested these by running them in the JS v3 SDK repository after fixing smithy-typescript.


mtdowling (approved):
Pending https://github.com/awslabs/smithy/issues/1072

gosar (approved):

mtdowling:
Closing given the changes made in #1106

PR #1104, opened 2022-02-23
Rewrite code writer formatting to allow for Runnable evaluation

mtdowling:
CodeWriter's formatter was rewritten to make it easier to understand, maintain, and evolve. As part of this evolution, this change adds the ability to pass a Runnable into a CodeWriter 'C' formatter (meaning "call"). When encountered, CodeWriter will invoke the Runnable and expect it to make calls to mutate the writer. Any text written is captured and used as the parameter value. This allows larger templates to be defined but still delegate functionality to separate methods. When coupled with text blocks in newer versions of Java, this largely makes things like openBlock/closeBlock irrelevant.

This change also is technically backward incompatible because it removes the ability to register a CodeWriter formatter using '[',']', '{', '}', '|', '<', '>'. Registering a formatter with these characters would either not work at all (e.g., '{'), or would be highly unreadable. By reserving these characters, we can also evolve CodeWriter's functionality as needed with additional sigils.

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
@@ -15,369 +15,323 @@
 
 package software.amazon.smithy.utils;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.BiFunction;
 import java.util.regex.Pattern;
 
-/**
- * TODO: Rewrite the formatter parser to use a custom {@link SimpleParser}.
- */
+@SmithyInternalApi
 final class CodeFormatter {
-    private static final Pattern NAME_PATTERN = Pattern.compile("^[a-z]+[a-zA-Z0-9_.#$]*$");
-    private static final Set<Character> VALID_FORMATTER_CHARS = SetUtils.of(
-            '!', '#', '%', '&', '\'', '(', ')', '*', '+', ',', '-', '.', '/', ':', ';', '<', '=', '>', '?', '@',
-            'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
-            'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '[', ']', '^', '_', '`', '{', '|', '}', '~');
 
-    private final Map<Character, BiFunction<Object, String, String>> formatters = new HashMap<>();
-    private final CodeFormatter parentFormatter;
+    private CodeFormatter() {}
 
-    CodeFormatter() {
-        this(null);
-    }
-
-    /**
-     * Create a CodeFormatter that also uses formatters of a parent CodeFormatter.
-     *
-     * @param parentFormatter Optional parent CodeFormatter to query when expanding formatters.
-     */
-    CodeFormatter(CodeFormatter parentFormatter) {
-        this.parentFormatter = parentFormatter;
-    }
-
-    void putFormatter(Character identifier, BiFunction<Object, String, String> formatFunction) {
-        if (!VALID_FORMATTER_CHARS.contains(identifier)) {
-            throw new IllegalArgumentException("Invalid formatter identifier: " + identifier);
-        }
-
-        formatters.put(identifier, formatFunction);
-    }
-
-    /**
-     * Gets a formatter function for a specific character.
-     *
-     * @param identifier Formatter identifier.
-     * @return Returns the found formatter, or null.
-     */
-    BiFunction<Object, String, String> getFormatter(char identifier) {
-        BiFunction<Object, String, String> result = formatters.get(identifier);
-
-        if (result == null && parentFormatter != null) {
-            result = parentFormatter.getFormatter(identifier);
-        }
-
-        return result;
-    }
-
-    String format(CodeWriter writer, Object content, Object... args) {
-        String expression = String.valueOf(content);
-        char expressionStart = writer.getExpressionStart();
-        String indent = writer.getIndentText();
-
-        // Simple case of no arguments and no expressions.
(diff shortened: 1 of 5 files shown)


smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
```diff
@@ -15,369 +15,331 @@
...
+        static Operation block(Operation delegate) {
+            return (sink, writer, column) -> delegate.apply(new BlockAppender(sink, column), writer, column);
         }
     }
 
-    private void parseNamedArgument(CodeWriter writer, State state, int startingBraceColumn) {
-        // Expand a named context value: "$" key ":" identifier
-        String name = parseNameUntil(writer, state, ':');
-        state.next();
+    private static final class BlockAppender extends DecoratedAppendable {
```
JordonPhillips, line 105: This doesn't seem to account for the indentation settings on the codewriter. Go, for instance, uses tabs instead of spaces. So for Go it would produce:

```go
func main() {
 fmt.Println("Hello world")
}
```

But also, and maybe I'm missing something here, it looks like the column is just column and not whitespace so it would end up being more like:

```go
func main() {
             fmt.Println("Hello world")
}
```

JordonPhillips (dismissed):

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeFormatter.java
mtdowling: That's exactly how it works today too, before this refactor. If you use `|` for block formatting, it takes into account the number of spaces before that interpolation, dynamically accounting for other things that might have been interpolated before it. In order to properly indent with tabs and potentially a mix of tabs and spaces, we'd need to introduce static and dynamic formatting for blocks.

* Static would be used if the characters on a line before interpolation are all spaces and tabs, then that's the text used at each new line.
* Dynamic would be used when the characters before the text are not all spaces and tabs. This can only use spaces since we can't base the exact alignment on anything else. Relying on the configured indent text of CodeWriter doesn't work since block formatting is supposed to align perfectly with where the interpolation is defined.

mtdowling: I updated the PR to use this approach.

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
```diff
@@ -73,13 +74,25 @@
...
- *     <li>{@code S}: Adds double quotes around the result of formatting a
+ *
+ *     <li>{@code C} (call): Runs a {@link Runnable} argument that is expected to
+ *     write to the same writer. Any text written to the CodeWriter inside of the
+ *     Runnable is used as the value of the argument. Note that a single trailing
+ *     newline is removed from the captured text.
+ *
+ *     <pre>{@code
+ *     CodeWriter writer = new CodeWriter();
+ *     writer.write("Hello, $C.", () -> writer.write("there"));
```
rcoh, line 91: interesting—I wonder if this is a potential for bugs + limiting. It seems like it might be possible to end up accidentally calling a different writer since you're relying on capturing the writer in a closure. 

It also makes it a bit tricker because you _can't_ create a Runnable without also having access to the base code writer at the same time.

The alternative being `(CodeWriter) -> Unit` where `CodeWriter` handles passing itself through

mtdowling: Yeah I thought about that too. We could support both `Runnable` and `Consumer<CodeWriter>`. The issue there is that I don't think it can be `Consumer<T extends CodeWriter>`since we only know it's a CodeWriter and not some special subtype.

rcoh: ah I see, yeah you'd need to add a self-type to code writer, but I don't think that could be done in a backwards compatible way.

rcoh: actually I think you could do it with a refactor like this [note: I don't really know Java]
```java
abstract class BaseCodeWriter<Self extends BaseCodeWriter<Self>> {
  // all of the codewriter functionality actually goes here, but now you can name `Self`
}

class CodeWriter extends BaseCodeWriter<CodeWriter> {
}

class RustWriter extends BaseCodeWriter<RustWriter> { 
  // rust specific functionality
}
```

mtdowling: That would work, but turns out isn't necessary for the problem I was describing. We can just do a cast in Java and call the Consumer anyways. Your solution would be useful for cleaning up a different problem of ugly parameters, but would need more helper methods on CodeWriter to essentially be an identity function that makes a safe cast of a Runnable or Consumer.

For example, this:
```
CodeWriterSubtype w = new CodeWriterSubtype();
w.write("$C", (Consumer<CodeWriterSubtype>) writer -> writer.write2("Hello!"));
```

Could become:
```
CodeWriterSubtype w = new CodeWriterSubtype();
w.write("$C", w.call(writer -> writer.write2("Hello!")));
```

Slightly less verbose, but it requires the use of a helper function, and requires a big refactor of CodeWriter. I'm thinking it would be prudent to leave this out for now.

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
```diff
@@ -385,6 +401,33 @@
...
  * // Outputs: "Names: Bob\n       Karen\n       Luis\n"
  * }</pre>
+ *
+ * <p>Alignment occurs either statically or dynamically based on the characters
+ * that come before interpolation. If all of the characters in the literal
+ * template that come before interpolation are spaces and tabs, then those
+ * characters are used when indenting newlines. Otherwise, the number of
+ * characters written as the template result that come before interpolation
+ * are used when indenting (this takes into account any interpolation that
+ * may precede block interpolation).
```
adamthom-amzn, line 411: I struggled to understand this for a while. I don't have any suggestions to make it clearer, so it may just be a problem with the reader; maybe more examples would help.

adamthom-amzn (approved):

PR #1106, opened 2022-02-24
Breaking Change to limit sets to string, blob, byte, short, integer, long, bigInteger, bigDecimal

### LINKED ISSUE #1093: Exclude some types (e.g. floating point types) from `@uniqueItems` lists
In a similar vein to https://github.com/awslabs/smithy/issues/1072, it doesn't make sense to have a list of floating point numbers with `@uniqueItems` applied.

mtdowling:
Sets can now only contain string, blob, byte, short, integer, long, bigInteger,
and bigDecimal shapes. Sets with other types of values are either difficult
to implement in various programming languages (for example, sets of floats in
Rust), or highly problematic for client/server use cases. Clients that are out
of sync with a service model could receive structures or unions from a
service, not recognize new members and drop them, causing the hash codes of
members of the set to collide, and this would result in the client discarding
set entries. For example, a service might return a set of 3 structures, but
when clients deserialize them, they drop unknown members, and the set
contains fewer than 3 entries.

Existing models that already use a set of other types will need to migrate to
use a list rather than a set, and they will need to implement any necessary
uniqueness checks server-side as needed.

While this is a breaking change, allowing sets to contain other types of values has proven to be an unsafe misfeature and something most clients have been unable to support (in fact, most Smithy implementations actually treat sets as lists in generated code).

Closes #1106
Closes #1093


adamthom-amzn (commented):
We deprecated @uniqueItems but it isn't removed, should we restrict the types it can apply to as well?

docs/source/1.0/spec/core/model.rst
```diff
@@ -613,10 +614,11 @@ example is ``smithy.example#StringSet$member``.
 
 .. rubric:: Language support for sets
 
-Not all programming languages support set data structures. Such languages
-SHOULD represent sets as a custom set data structure that can interpret value
-hash codes and equality, or alternatively, store the values of a set data
-structure in a list and rely on validation to ensure uniqueness.
+Not all programming languages support an ordered set data structures. Such
+languages SHOULD represent sets as a custom set data structure that can
+interpret value hash codes and equality, or alternatively, store the values
```
david-perez, line 619: This is hinting at the fact that the backing data structure should be a `HashSet`. If that's the case I would call it out more explicitly; if it's not I would mention that an ordered set whose keys are not hashable is fine too.

mtdowling: I don't think so. We should try to avoid that level of implementation detail showing up in the spec. This was the previous text, and I'll just remove it.

mtdowling:
Sure, I can adjust uniqueItems too given it's deprecated _and_ we will not allow it's use in IDL v2.

adamthom-amzn (approved):

PR #1109, opened 2022-02-26
Add interface for codegen integration and settings

mtdowling:
Codegen integrations typically all follow a similar format -- they
provide hooks for modifying the code generator and are discovered using
SPI. This commit adds a base interface for codegen integrations and
settings objects. The integration object is as minimal as possible
whill still providing some utility and conformity across
implementations, and the settings object just provides a minimal
`getService` method. Each Java implementation of Smithy codegen has such
a method in their settings object, and so does the Kotlin based Rust
generator.

The integration interface is very simimlar to both TypeScript and Go
integrations, though not exactly the same as either. The TypeScript
generator passes the entire PluginContext to preprocessModel, but that
is never actually needed and makes it harder to mock (Go does not
require the full context, only the Model so it can take this change).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
@@ -27,7 +27,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.logging.Logger;
 
-final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extends SmithyIntegration<S>> {
+final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extends SmithyIntegration<S, ?>> {
 
     private static final Logger LOGGER = Logger.getLogger(IntegrationTopologicalSort.class.getName());
 
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
@@ -27,7 +27,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.logging.Logger;
 
-final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extends SmithyIntegration<S, ?>> {
+final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extends SmithyIntegration<S, ?, ?>> {
 
     private static final Logger LOGGER = Logger.getLogger(IntegrationTopologicalSort.class.getName());
 
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
@@ -76,10 +76,11 @@ final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extend
         I previous = this.integrationLookup.put(integration.name(), integration);
         insertionOrder.put(integration.name(), insertionOrder.size());
         if (previous != null) {
-            throw new IllegalArgumentException("Conflicting SmithyIntegration names detected for "
-                                               + integration.name() + ": "
-                                               + integration.getClass().getCanonicalName() + " and "
-                                               + previous.getClass().getCanonicalName());
+            throw new IllegalArgumentException(String.format(
+                    "Conflicting SmithyIntegration names detected for '%s': %s and %s",
+                    integration.name(),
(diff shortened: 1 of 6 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/SmithyIntegration.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * <p>By default, this method will return the given {@code symbolProvider}
+     * as-is.
+     *
+     * @param model Model being generated.
+     * @param settings Setting used to generate.
+     * @param symbolProvider The original {@code SymbolProvider}.
+     * @return The decorated {@code SymbolProvider}.
+     */
+    default SymbolProvider changeSymbolProvider(Model model, S settings, SymbolProvider symbolProvider) {
```
JordonPhillips, line 68: nit: `decorateSymbolProvider`?

JordonPhillips (approved):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/SmithyIntegration.java
```diff
@@ -0,0 +1,71 @@
...
+ * (e.g., register protocol generators, register auth scheme integrations,
+ * attach middleware, intercept and augment CodeWriter sections, etc).
+ *
+ * <p>This interface is currently unstable as more requirements
+ * may be added in the future.
+ *
+ * @param <S> The settings object used to configure the generator.
+ */
+@SmithyUnstableApi
+public interface SmithyIntegration<S extends SmithyCodegenSettings> {
```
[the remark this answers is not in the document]
mtdowling: Yeah this is an interesting problem. It would be nice to have a centralized solution around it that's flexible enough for any generator. Let me noodle on this.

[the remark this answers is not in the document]
mtdowling: I took your idea and built on it a bit. You can now order integrations using a priority or a relative ordering based on before/after other integrations by name. I also included a topological sort code generators can use to order integrations.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.PriorityQueue;
+import java.util.Queue;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.logging.Logger;
+
+final class IntegrationTopologicalSort<S extends SmithyCodegenSettings, I extends SmithyIntegration<S>> {
```
JordonPhillips, line 31: Nice!

smithy-codegen-core/src/test/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSortTest.java
```diff
@@ -0,0 +1,218 @@
...
+        assertThat(result, contains("d", "a", "c", "b"));
+    }
+
+    @Test
+    public void sortsIntegrationsBasedOnBefore() {
+        List<MyIntegration> integrations = new ArrayList<>();
+        integrations.add(new MyIntegration("a", createOrder((byte) 0, "b")));
+        integrations.add(new MyIntegration("b", createOrder((byte) 0, "c")));
+        integrations.add(new MyIntegration("c", createOrder((byte) 0, "d")));
+        integrations.add(new MyIntegration("d", createOrder((byte) 0)));
```
JordonPhillips, line 148: You should shuffle this around so the expected result isn't also insertion order

JordonPhillips (approved):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+    }
+
+    private void addIntegration(I integration) {
+        I previous = this.integrationLookup.put(integration.name(), integration);
+        insertionOrder.put(integration.name(), insertionOrder.size());
+        if (previous != null) {
+            throw new IllegalArgumentException("Conflicting SmithyIntegration names detected for "
+                                               + integration.name() + ": "
```
kstich, line 80: nit: can this name be wrapped in single quotes?

kstich (approved):

smithy-codegen-core/src/test/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSortTest.java
mtdowling: There are other tests that do out of order checks too

PR #1110, opened 2022-02-27
CodeWriter rehaul to allow typed sections

mtdowling:
This commit is a major refactor to CodeWriter to introduce a recursively typed `AbstractCodeWriter` that allows fluent interfaces to work with subtypes and allows the introduction of strongly typed section interceptors to CodeWriter.

This adds a major new feature to CodeWriter: typed sections and interceptors. This provides a much safer plugin system that can be used in code generators. Because the binding of an "interceptor" to a "section" is an interface in and of itself, code generation plugin systems don't need to actually hand off CodeWriters to plugins but rather just ask plugins for interceptors.

```java
// Say you have your own AbstractCodeWriter subclass:
MyWriter writer = new MyWriter();

// Say you have an event named NameEvent that can be intercepted.
// The following lines register section interceptors for that event that just
// append to any previously written text in the section. The first interceptor
// uses one of the accessor methods of the provided NameEvent object.
writer.onSection(CodeInterceptor.appender(NameEvent.class, (w, section) -> w.write("$L", section.name())));
writer.onSection(CodeInterceptor.appender(NameEvent.class, (w, section) -> w.write("Foo")));
writer.onSection(CodeInterceptor.appender(NameEvent.class, (w, section) -> w.write("Bar")));

// injectSection is a shortcut for pushState(section).popState().
// popState() is when code interceptors are applied.
writer.injectSection(new NameEvent("Thomas"));

assert(writer.toString().equals("Name?\nThomas\nFoo\nBar\n"));
```

You can create a strongly typed section by passing a CodeSection to `injectSection` or `pushState`. A CodeSection is just an interface with a `sectionName()` method that returns the name of the class by default.

```java
record NameEvent(String name) implements CodeSection;

NameEvent event = new NameEvent("Thomas");
```

You can create `CodeInterceptors` by implementing `CodeInterceptor`, or even better, `CodeInterceptor.Appender` / `CodeInterceptor.Prepender` that perform writing the previously emitted content correctly (only if non-empty and using writeInlineWithNoFormatting).

```java
class SpecificNameInterceptor extends CodeInterceptor.Appender<NameEvent, MyWriter>() {
    private final String name;

    public SpecificNameInterceptor(String name) {
        this.name = name;
    }

    @Override
    public Class< NameEvent > sectionType() {
        return NameEvent.class;
    }

    @Override
    public void append(MyWriter writer, NameEvent section) {
        if (section.name().equals(name)) {
            writer.write("Hi, $L!", name);
        }
    }
}
```


JordonPhillips (approved):

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeInterceptor.java
```diff
@@ -0,0 +1,191 @@
...
+     * <p>This method does not support custom filters on matched CodeSections. That
+     * functionality must be implemented by directly creating an Appender class.
+     *
+     * @param type The type of section to intercept.
+     * @param appender A BiConsumer that takes the writer and section and is expected to make write calls.
+     * @param <S> The type of section being intercepted.
+     * @param <W> The type of writer to use.
+     * @return Returns the created Appender.
+     */
+    static <S extends CodeSection, W extends AbstractCodeWriter<W>> CodeInterceptor<S, W> appender(
```
kstich, line 91: Any reason a concise version of `prepender` doesn't exist?

JordonPhillips (approved):

kstich (approved):

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeInterceptor.java
mtdowling: I thought about it, but I think it'll be significantly more rare than appending, and you can still create a prepender manually if needed. We can always add it later was my thought.

PR #1123, opened 2022-03-09
Fix CodeWriter backwards compatibility

adamthom-amzn:
*Description of changes:*
Adding a parent of CodeWriter that changed the return type of its methods, even
though those returned objects were always in practice the same type as they had
been before, broke binary backwards compatibility for consumers of CodeWriter.

This forks CodeWriter back to its previous state, mostly, and adds a new
default extension of AbstractCodeWriter called TypedCodeWriter. Current
consumers of CodeWriter will continue to work, and have access to some new
features, such as the C formatter (though their Consumer will receive an
AbstractCodeWriter, not CodeWriter), and will keep onSectionPrepend and
onSectionAppend, but not have access to typed sections.

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/DocWriter.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/DocWriter.java
@@ -15,7 +15,7 @@
 
 package software.amazon.smithy.codegen.core.writer;
 
-import software.amazon.smithy.utils.CodeWriter;
+import software.amazon.smithy.utils.AbstractCodeWriter;
 import software.amazon.smithy.utils.SmithyUnstableApi;
 
 /**
@@ -36,16 +36,11 @@ import software.amazon.smithy.utils.SmithyUnstableApi;
  * }
  * }</pre>
  *
- * @param <T> The type of {@code CodegenWriter} being written to.
- * @deprecated prefer {@link DocWriter}. This will be removed in a future release.
+ * @param <T> The type of {@code AbstractCodeWriter} being written to.
  */
-@Deprecated
 @FunctionalInterface
 @SmithyUnstableApi
-public interface DocumentationWriter<T extends CodeWriter> {
-    // Implementer's note: this class is not tied to CodegenWriter; it can be
-    // used with any kind of CodeWriter, allowing any kind of CodegenWriters to
-    // be used but also making this type more general-purpose.
+public interface DocWriter<T extends AbstractCodeWriter<T>> {
 
     /**
      * Writes documentation comments.
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/JavaStyleDocumentationWriterBuilder.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/JavaStyleDocumentationWriterBuilder.java
@@ -16,7 +16,7 @@
 package software.amazon.smithy.codegen.core.writer;
 
 import java.util.function.Function;
-import software.amazon.smithy.utils.AbstractCodeWriter;
+import software.amazon.smithy.utils.CodeWriter;
 import software.amazon.smithy.utils.SmithyUnstableApi;
 
 /**
@@ -27,7 +27,10 @@ import software.amazon.smithy.utils.SmithyUnstableApi;
  * closing documentation comment (i.e., star (*) followed by a forward slash
  * (/)). This should also work for JavaScript, PHP, and other languages that
  * use Java-style comments.
+ *
+ * @deprecated this class uses CodeWriter, which is deprecated.
  */
+@Deprecated
 @SmithyUnstableApi
 public final class JavaStyleDocumentationWriterBuilder {
 
@@ -62,7 +65,7 @@ public final class JavaStyleDocumentationWriterBuilder {
      * @param <T> The type of writer to create.
      * @return Returns the created documentation writer.
      */
-    public <T extends AbstractCodeWriter<T>> DocumentationWriter<T> build() {
+    public <T extends CodeWriter> DocumentationWriter<T> build() {
         Function<String, String> function = resolveMappingFunction();
         String sectionName = namedDocumentationSection;
 
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
@@ -15,12 +15,17 @@
 
 package software.amazon.smithy.utils;
 
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
 /**
- * Helper class for generating code.
- *
- * <p>This class is a general purpose implementation of {@link AbstractCodeWriter}.
+ * @deprecated prefer {@link SimpleCodeWriter} or a custom subclass of {@link AbstractCodeWriter}.
  */
+@Deprecated
 public class CodeWriter extends AbstractCodeWriter<CodeWriter> {
+
     /**
      * Creates a default instance of a CodeWriter that uses "\n" for newlines,
      * flattens multiple successive blank lines into a single blank line,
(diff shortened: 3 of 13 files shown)


smithy-utils/src/main/java/software/amazon/smithy/utils/TypedCodeWriter.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package software.amazon.smithy.utils;
+
+/**
+ * Helper class for generating code.
+ *
+ * <p>This class is a general purpose implementation of {@link AbstractCodeWriter}.
+ */
+public class TypedCodeWriter extends AbstractCodeWriter<TypedCodeWriter> {
```
mtdowling, line 23: nit: Can we call this `SimpleCodeWriter` or something similar? We alread have `SimpleParser`, so it's consistent.

smithy-utils/src/main/java/software/amazon/smithy/utils/TypedCodeWriter.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Creates a default instance of a TypedCodeWriter that uses "\n" for newlines,
+     * flattens multiple successive blank lines into a single blank line,
+     * and adds a trailing new line if needed when converting the TypedCodeWriter
+     * to a string.
+     *
+     * @return Returns the created and configured TypedCodeWriter.
+     */
+    public static TypedCodeWriter createDefault() {
```
mtdowling, line 32: I don't think we need this method

mtdowling (changes requested):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/DocWriter.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * <p>Implementations are expected to write out the beginning of a documentation
+     * comment, set any necessary prefix for each line written while writing docs,
+     * then invoke the given {@code runnable}, then finally write the closing
+     * characters for documentation.
+     *
+     * @param writer   Writer to configure for writing documentation.
+     * @param runnable Runnable that handles actually writing docs with the writer.
+     */
+    void writeDocs(T writer, Runnable runnable);
```
mtdowling, line 56: Since this is brand new... should we add another method that takes a `Consumer<T>` and writes docs rather than a runnable? Then the one that takes a Runnable could provide a default implementation that calls the method that takes a `Consumer<T>`. These historically didn't take a `T` because there was no `T` available.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/CodegenWriter.java
```diff
@@ -80,10 +80,12 @@
  *
  * @param <T> The concrete type, used to provide a fluent interface.
  * @param <U> The import container used by the writer to manage imports.
+ * @deprecated prefer {@link SimpleCodegenWriter}
  */
 @SmithyUnstableApi
+@Deprecated
```
mtdowling, line 86: Let's call out that we'll remove this in a future release since it's unstable.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/SimpleCodegenWriter.java
```diff
@@ -0,0 +1,341 @@
...
+ * that have a concept of namespaces can use {@link #setRelativizeSymbols} to
+ * the namespace of the CodegenWriter, and then the default symbol formatter
+ * will relativize symbols against that namespace using {@link Symbol#relativize}
+ * when writing the symbol as a string.
+ *
+ * @param <T> The concrete type, used to provide a fluent interface.
+ * @param <U> The import container used by the writer to manage imports.
+ */
+@SmithyUnstableApi
+public class SimpleCodegenWriter<T extends SimpleCodegenWriter<T, U>, U extends ImportContainer>
```
mtdowling: Maybe a better name here is `SymbolWriter` since the key thing here is that this subclass writes and tracks Symbols

mtdowling (changes requested):

smithy-utils/src/main/java/software/amazon/smithy/utils/SimpleCodeWriter.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Helper class for generating code.
+ *
+ * <p>This class is a general purpose implementation of {@link AbstractCodeWriter}.
+ */
+public class SimpleCodeWriter extends AbstractCodeWriter<SimpleCodeWriter> {
+    public SimpleCodeWriter() {
+        super();
+        trimTrailingSpaces();
```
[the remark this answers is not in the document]
adamthom-amzn: yes, tests fail without it

smithy-utils/src/main/java/software/amazon/smithy/utils/CodeWriter.java
```diff
@@ -67,13 +73,284 @@ public final CodeWriter onSectionPrepend(String sectionName, Runnable writeBefor
...
-    public final CodeWriter onSectionAppend(String sectionName, Runnable writeAfter) {
+    public CodeWriter onSectionAppend(String sectionName, Runnable writeAfter) {
         return onSection(sectionName, contents -> {
             writeWithNoFormatting(contents);
             writeAfter.run();
         });
     }
+
+    /**
+     * @see AbstractCodeWriter#copySettingsFrom(AbstractCodeWriter)
```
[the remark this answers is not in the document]
adamthom-amzn: This one isn't actually an override - yes, I know it looks like one, but it's not.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/AbstractDocumentationWriter.java
```diff
@@ -0,0 +1,57 @@
...
+ *         runnable.run();
+ *     }
+ * }
+ * }</pre>
+ *
+ * @param <T> The type of {@code AbstractCodeWriter} being written to.
+ */
+@FunctionalInterface
+@SmithyUnstableApi
+public interface AbstractDocumentationWriter<T extends AbstractCodeWriter<T>> {
```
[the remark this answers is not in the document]
adamthom-amzn: Done

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/DocWriter.java
mtdowling: Actually, nevermind

mtdowling (approved):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/DocumentationWriter.java
```diff
@@ -37,10 +37,12 @@
  * }</pre>
  *
  * @param <T> The type of {@code CodegenWriter} being written to.
+ * @deprecated prefer {@link AbstractDocumentationWriter}
  */
+@Deprecated
 @FunctionalInterface
 @SmithyUnstableApi
-public interface DocumentationWriter<T extends AbstractCodeWriter<T>> {
+public interface DocumentationWriter<T extends CodeWriter> {
```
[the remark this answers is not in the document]
adamthom-amzn: I marked it as @Deprecated, adding more to the comments.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/JavaStyleDocumentationWriterBuilder.java
```diff
@@ -62,7 +65,7 @@ public static String escapeAtSignWithEntity(String contents) {
      * @param <T> The type of writer to create.
      * @return Returns the created documentation writer.
      */
-    public <T extends AbstractCodeWriter<T>> DocumentationWriter<T> build() {
+    public <T extends CodeWriter> DocumentationWriter<T> build() {
```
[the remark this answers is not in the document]
adamthom-amzn: I think we have to change this class for this to compile, but I have left out the new variant of this class.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/CodegenWriter.java
adamthom-amzn: Done.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/writer/SimpleCodegenWriter.java
adamthom-amzn: Done

PR #1124, opened 2022-03-09
Reset ModelAssembler's validationEventListener to default

gosar:
The `resetDoesNotBarf` test would throw NPE before the change and now passes.


smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
```diff
@@ -684,4 +684,27 @@ public void transientTraitsAreNotValidated() {
...
+
+    @Test
+    public void resetDoesNotBarf() {
+        ModelAssembler assembler = new ModelAssembler();
+        assembler.assemble();
+        assembler.reset();
+        assembler.assemble();
+    }
+
+    // reset() affects multiple properties of the ModelAssembler. The test asserts one of them (shapes).
```
gosar, line 696: I didn't think it was worth asserting each property, by exposing a package-private access to each property, or a separate behavior test like below. LMK if you think otherwise.

PR #1127, opened 2022-03-10
(docs): Remove use of sensitive trait on members in doc examples

gosar:
Since sensitive trait on member is going to be disallowed in IDL 2.0, updating the 1.0 docs to remove examples with such usage.

### DIFF
--- a/docs/source/1.0/spec/core/idl.rst
+++ b/docs/source/1.0/spec/core/idl.rst
@@ -806,11 +806,9 @@ Traits can be applied to the set shape and its members:
 
         @deprecated
         set StringSet {
-            member: SensitiveString
+            @pattern("\\w+")
+            member: String
         }
-
-        @sensitive
-        string SensitiveString
 
     .. code-tab:: json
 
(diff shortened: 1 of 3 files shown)


docs/source/1.0/spec/core/idl.rst
```diff
@@ -1031,10 +1029,9 @@ The following example defines a union shape with several members:
         union MyUnion {
             i32: Integer,
 
-            stringA: String,
+            string: String,
 
-            @sensitive
```
kstich, line 1036: This was to show that traits can also be applied to members of unions, can one be added here?

docs/source/1.0/spec/core/model.rst
```diff
@@ -828,10 +828,9 @@ The following example defines a union shape with several members:
         union MyUnion {
             i32: Integer,
 
-            stringA: String,
+            string: String,
 
-            @sensitive
```
kstich, line 833: This was to show that traits can also be applied to members of unions, can one be added here?

docs/source/1.0/spec/core/idl.rst
gosar: oh makes sense that's why there were 2 strings. added a length trait to the string.

kstich (approved):

PR #1136, opened 2022-03-16
Update streaming target validation for mixins

kstich:
This commit updates the validation of streaming targets to allow
for a shape with a streaming member to be mixed in to input and
output shapes.


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/StreamingTraitValidator.java
```diff
@@ -104,8 +104,32 @@
 
         for (Shape shape : streamingStructures) {
             for (Relationship rel : provider.getNeighbors(shape)) {
+                // Mixins that contain streams are allowed to be mixed into
+                // input and output shapes, but nowhere else.
+                boolean isInputOutputMixin = false;
```
mtdowling, line 109: Seems like a good thing to break out into a method for readability

mtdowling: Can we just ignore MIXIN relationships in this validator, and it also solves the problem?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/StreamingTraitValidator.java
kstich: Broken out to its own method.

Just ignoring MIXIN allows [an input shape with a nested structure to mixin a structure containing a streaming member](https://github.com/awslabs/smithy/pull/1136/files#diff-bf97606a099a629dd90fcc4b1f2b51aca2f11622126e9387c751b28bbcbc5d31R108), which needs to be disallowed.

mtdowling (approved):

PR #1137, opened 2022-03-16
Disallow sensitive trait on members

gosar:
Allowing @sensitive trait on members can lead to insecure usage
patterns, like forgetting to put @sensitive on ALL members that target
the same shape, which can lead to sensitive data not getting handled
properly, e.g., not getting redacted from logs, etc.

Instead marking the target shape as sensitive prevents modeling
mistakes by ensuring every reference to data types that are inherently
sensitive are always considered sensitive.

TODO: Doc updates. Currently `idl-2.0` branch's `docs` still references version 1.0 mostly. Will update docs separately later.


smithy-jsonschema/src/test/java/software/amazon/smithy/jsonschema/JsonSchemaConverterTest.java
```diff
@@ -551,6 +552,25 @@ public void supportsMapPatternPropertiesWithDefaultPattern() {
         assertThat(schema.getPatternProperties().get(".+").getType().get(), equalTo("string"));
     }
 
+    @Test
+    public void sensitiveTraitHasNotImpact() {
```
kstich, line 556: One of these strings shouldn't have the trait applied.

kstich (changes requested):

mtdowling:
The spec probably needs to be updated to reflect this change. The selector shouldn't allow members for sensitive. Our implementation of Smithy will need to allow it since we support v1 and v2 simultaneously, but the extra validation we're doing will cover us to not actually allow it on members.

smithy-jsonschema/src/test/java/software/amazon/smithy/jsonschema/JsonSchemaConverterTest.java
gosar: haha, copy paste error. fixing test name too.

docs/source/1.0/spec/core/traits/documentation-traits.rst.template
```diff
@@ -391,9 +391,9 @@ Summary
     Indicates that the data stored in the shape or member is sensitive
```
kstich, line 391: rm "or member"

gosar: fixed

kstich (approved):

mtdowling (approved):

PR #1143, opened 2022-03-21
Add in a migration tool

JordonPhillips:
This adds a basic migration tool. The actual user interface is empty right now, but the implementation is there.

### DIFF
--- a/smithy-upgrade-tool/build.gradle
+++ b/smithy-upgrade-tool/build.gradle
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import org.apache.tools.ant.taskdefs.condition.Os
+
+plugins {
+    id "application"
+    id "org.beryx.runtime" version "1.12.7"
+}
+
+description = "This module implements the Smithy command line interface."
+
+ext {
+    displayName = "Smithy :: CLI"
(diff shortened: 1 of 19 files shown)


smithy-upgrade-tool/src/main/java/software/amazon/smithy/upgrade/Upgrader.java
```diff
@@ -0,0 +1,389 @@
...
+            return modelString;
+        }
+
+        @Override
+        protected Void getDefault(Shape shape) {
+            if (shape.hasTrait(BoxTrait.class)) {
+                eraseTrait(shape, shape.expectTrait(BoxTrait.class));
+            }
+            // handle members in reverse definition order
+            shape.members().stream()
```
kstich, line 133: I think this will attempt to modify list/set/map shapes, since they return their standard members via `.members()`. It's worth proving out that those are safe in the test cases.

smithy-upgrade-tool/src/main/java/software/amazon/smithy/upgrade/Upgrader.java
```diff
@@ -0,0 +1,389 @@
...
+            }
+            return modelString;
+        }
+
+        @Override
+        protected Void getDefault(Shape shape) {
+            if (shape.hasTrait(BoxTrait.class)) {
+                eraseTrait(shape, shape.expectTrait(BoxTrait.class));
+            }
+            // handle members in reverse definition order
```
kstich, line 132: nit: chunk of comments that start with lower case words.

smithy-upgrade-tool/src/main/java/software/amazon/smithy/upgrade/Upgrader.java
JordonPhillips: Yep, that's intentional since the box trait can technically show up there too. Updated to add cases for all the collection types.

smithy-upgrade-tool/build.gradle
```diff
@@ -0,0 +1,98 @@
...
+
+plugins {
+    id "application"
+    id "org.beryx.runtime" version "1.12.7"
+}
+
+description = "This module implements the Smithy command line interface."
+
+ext {
+    displayName = "Smithy :: CLI"
```
mtdowling, line 26: The Smithy CLI stuff in this section aren't right since this isn't the Smithy CLI

mtdowling (changes requested):

PR #1144, opened 2022-03-21
Add support for conditions and loops to AbstractCodeWriter

mtdowling:
AbstractCodeWriter now supports Mustache-like behavior by allowing conditions and loops to be evaluated in templates. Note that this does disallow using '/', '?', '^', and '#' as formatters.

Conditions are defined using `${?foo}` where "foo" is the writer named parameter to evaluate. `?` is used to check if a value is truthy.

The following template with a "foo" value set to "Noid":

```
${?foo}
Hello, ${foo:L}!
${/foo}
```

Evaluates to:

```
Hello, Noid!
```

`^` can be used to check if a value is falsey. The following template with no set "baz" value:

```
${?baz}
  Yes baz
${/baz}
${^baz}
  No baz
${/baz}
```

Evaluates to:

```
  No baz
```

A variable is falsey if it is Boolean false, an empty Iterable, an empty Optional, an empty Map, or an empty String. A variable is truthy if it is not falsey.

If a conditional expression or loop expression is the non-whitespace text on a line and is immediately followed by a `\r` or `\n`, then the entire line is omitted from the output.

Loops can be created using `${#foo}`. Each value contained in the referenced named parameter is sent through the contained template, one after the other.

The following template with a "foo" value of ["a", "b"]:

```
${#foo}
- ${currentKey:L}: ${currentValue:L}
${/foo}
```

Evaluates to:

```
- 0: a
- 1: b
```

Each iteration of the loop creates a new state in the writer that sets a named parameter named `key` that contains the current index of the loop or the current key of a map entry; `value` containing the current value of the iterable or current value of a map entry; `key.first` is set to true when the first value is being evaluated, and `key.last` is set to true when the last value is being evaluated.

```
SimpleCodeWriter writer = new SimpleCodeWriter()
        .trimTrailingSpaces(false)
        .insertTrailingNewline(false);
writer.putContext("foo", Arrays.asList("a", "b", "c"));
writer.write("""
        ${#foo}
        ${?key.first}
        [
        ${/key.first}
            ${value:L}${^key.last},${/key.last}
        ${?key.last}
        ]
        ${/key.last}
        ${/foo}""");

assertThat(writer.toString(), equalTo("""
        [
            a,
            b,
            c
        ]
        """);
```

A custom variable name can be used in loops. For example:

```
${#foo as key1, value1}
    - ${key1:L}: ${value1:L} (first: ${key1.first:L}, last: ${key.last:L}
${/foo}
```

Whitespace that comes before an expression removed by putting `~` at the beginning of an expression.

Assuming that the first positional argument is "hi":

```
Greeting:
    ${~L}
```

Expands to:

```
Greeting:hi
```

Whitespace that comes after an expression can be removed by adding `~` to the end of the expression:

```
${L~}

.    
```

Expands to:

```
hi.
```

Leading whitespace cannot be removed when using inline block alignment ('|'). The following is *invalid*:

```
    ${~C|}
```

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/AbstractCodeWriter.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/AbstractCodeWriter.java
@@ -31,13 +31,13 @@ import java.util.regex.Pattern;
 /**
  * Helper class for generating code.
  *
- * <p>A CodeWriter can be used to write basically any kind of code, including
+ * <p>An AbstractCodeWriter can be used to write basically any kind of code, including
  * whitespace sensitive and brace-based.
  *
  * <p>The following example generates some Python code:
  *
  * <pre>{@code
- * CodeWriter writer = new CodeWriter();
+ * SimpleCodeWriter writer = new SimpleCodeWriter();
  * writer.write("def Foo(str):")
  *       .indent()
  *       .write("print str");
@@ -52,7 +52,7 @@ import java.util.regex.Pattern;
  * {@code write}:
  *
  * <pre>{@code
- * CodeWriter writer = new CodeWriter();
+ * SimpleCodeWriter = new SimpleCodeWriter();
  * writer.write("Hello, $L", "there!");
  * String code = writer.toString();
  * }</pre>
@@ -60,7 +60,7 @@ import java.util.regex.Pattern;
  * <p>In the above example, {@code $L} is interpolated and replaced with the
  * relative argument {@code there!}.
  *
- * <p>A CodeWriter supports three kinds of interpolations: relative,
+ * <p>An AbstractCodeWriter supports three kinds of interpolations: relative,
  * positional, and named. Each of these kinds of interpolations pass a value
  * to a <em>formatter</em>.</p>
  *
@@ -68,7 +68,7 @@ import java.util.regex.Pattern;
  *
  * <p>Formatters are named functions that accept an object as input, accepts a
  * string that contains the current indentation (it can be ignored if not useful),
- * and returns a string as output. The {@code CodeWriter} registers two built-in
+ * and returns a string as output. {@code AbstractCodeWriter} registers two built-in
  * formatters:
  *
  * <ul>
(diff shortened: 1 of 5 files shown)


adamthom-amzn:
I wonder if you could trim down some of the bloat from `currentIsFirst` and `currentKey`, etc. Maybe something overly complicated and bad like

```
SimpleCodeWriter writer = new SimpleCodeWriter()
        .trimTrailingSpaces(false)
        .insertTrailingNewline(false);
writer.putContext("foo", Arrays.asList("a", "b", "c"));
writer.write("""
        ${#foo|k,v}
        ${?kIsFirst}
        [
        ${/kIsFirst}
            ${v:L}${^kIsLast},${/kIsLast}
        ${?kIsLast}
        ]
        ${/kIsLast}
        ${/foo}""");

assertThat(writer.toString(), equalTo("""
        [
            a,
            b,
            c
        ]
        """);
```

mtdowling:
I pushed an update that allows loops to define a custom key and value variable name. It now defaults to `key`, `key.first`, `key.last`, and `value`.

adamthom-amzn (approved):

smithy-utils/src/main/java/software/amazon/smithy/utils/AbstractCodeWriter.java
```diff
@@ -426,6 +426,142 @@
...
+ * </ul>
+ *
+ * <p>Values that are not falsey are considered truthy.
+ *
+ * <h4>Loops</h4>
+ *
+ * <p>Loops can be created to repeat a section of a template for each key value pair
+ * stored in a named property. Loops are created using "#".
+ *
+ * <p>The following template with a "foo" value of ["a", "b", "c"]:
```
JordonPhillips, line 485: You say key-value pair but then immediately follow up with a list. Either this first example should use a map, or the explanation that iterables use the index as a pseudo-key needs to come before this example.

smithy-utils/src/main/java/software/amazon/smithy/utils/AbstractCodeWriter.java
```diff
@@ -426,6 +426,142 @@
...
+ * ${#foo as key1, value1}
+ *     - ${key1:L}: ${value1:L} (first: ${key1.first:L}, last: ${key1.last:L})
+ * ${/foo}
+ * }</pre>
+ *
+ * <h4>Whitespace control</h4>
+ *
+ * <p>Conditional blocks that occur on lines that only contain whitespace are not written
+ * to the template output. For example, if the condition in the following template evaluates
+ * to falsey, then the template expands to nothing:
```
JordonPhillips, line 523:
```suggestion
 * to falsey, then the template expands to an empty string:
```

smithy-utils/src/main/java/software/amazon/smithy/utils/AbstractCodeWriter.java
```diff
@@ -426,6 +426,142 @@
...
+ *
+ * <pre>{@code
+ * ${?foo}
+ * Foo is set: ${foo:L}
+ * ${/foo}
+ * }</pre>
+ *
+ * <p>Whitespace that comes before an expression removed by putting "~" at the beginning of an expression.
```
JordonPhillips, line 531:
```suggestion
 * <p>Whitespace that comes before an expression can be removed by putting "~" at the beginning of an expression.
```

JordonPhillips (changes requested):

PR #1146, opened 2022-03-22
Fix trait source locations

JordonPhillips:
This fixes trait source locations from the IDL to start at the trait
itself rather than at the start of the trait's node definition. This
is particularly important for annotation traits where the source
location previously pointed to the column *after* the trait, which
doesn't necessarily even exist.

The following shows where the source location used to point:

```
@annotation
           ^

@annotationWithParens()
                       ^

@listTrait(["foo"])
           ^

@structuredTrait({
                 ^
    "foo": "bar"
})

@structuredTraitWithoutBraces(
    foo: "bar"
    ^
)

@primitiveTrait(true)
                ^
```

And this shows where they point now:

```
@annotation
^

@annotationWithParens()
^

@listTrait(["foo"])
^

@structuredTrait({
^
    "foo": "bar"
})

@structuredTraitWithoutBraces(
^
    foo: "bar"
)

@primitiveTrait(true)
^
```

### DIFF
--- a/smithy-model/src/test/java/software/amazon/smithy/model/loader/IdlModelLoaderTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/loader/IdlModelLoaderTest.java
@@ -35,9 +35,11 @@ import software.amazon.smithy.model.shapes.ResourceShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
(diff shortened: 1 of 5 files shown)


smithy-model/src/test/java/software/amazon/smithy/model/loader/IdlModelLoaderTest.java
```diff
@@ -68,6 +70,26 @@ public void loadsAppropriateSourceLocations() {
         });
     }
 
+    @Test
+    public void loadsAppropriateTraitSourceLocations() {
```
mtdowling, line 74: This test is genius

mtdowling (approved):

Issue #1116, opened 2022-03-02
Server behavior with `httpResponseCode` outside valid range


david-perez:
As of now, [`httpResponseCode` can only be applied to `integer` shapes](https://github.com/awslabs/smithy/issues/1115).

What should the behavior of a server implementation be when it attempts to serialize as an HTTP response code a value outside the 100-999 range? In the Rust sSDK we simply fail to serialize the response, but perhaps another behavior is desired (e.g. [falling back](https://github.com/awslabs/smithy/issues/1098#issuecomment-1051019501) to the `http` trait) .

mtdowling:
IMO this seems like an HTTP spec thing and not a Smithy thing that needs to be explicitly documented in Smithy. I recommend failing to send the response.

david-perez:
@mtdowling  Why shouldn't it be explicitly documented? It's a source of possible discrepancies among server implementations.

mtdowling:
It's up to the server to set the httpResponseCode. It's not an arbitrary value, so it seems overly in the weeds to talk about this in the Smithy spec (and it's already covered in the HTTP, there's no need to repeat this in Smithy's spec). The client reads the value of a member marked with this trait to know what the HTTP status code returned by the server was.

PR #1168, opened 2022-04-01
Remove invalid but extant base64 protocol test case

david-perez:
While `YmxvyG=` is _technically_ invalid base64 according to the spec
(its Unicode code point length is not divisible by 4), it can always be
unambiguously decoded, even over a network protocol that concatenates
base64-encoded strings and that would hence require padding characters.

Some off the shelf base 64 implementations, like the canonical `base64`
Rust crate (https://docs.rs/base64/latest/base64/), decode this case
successfully.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-blob.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-blob.smithy
@@ -34,7 +34,7 @@ apply MalformedBlob @httpMalformedRequestTests([
             }
         },
         testParameters: {
-            value: ["blob", "\"xyz\"", "\"YmxvYg=\"", "[98, 108, 11, 98]",
+            value: ["blob", "\"xyz\"", "[98, 108, 11, 98]",
                     "[\"b\", \"l\",\"o\",\"b\"]", "981081198", "true", "[][]", "-_=="]
         }
     },
--- a/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-string.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/malformedRequests/malformed-string.smithy
@@ -37,8 +37,6 @@ apply MalformedString @httpMalformedRequestTests([
             value: [
                 // Insufficient padding
                 "xyz",
-                // Extant, but also insufficient padding
-                "YmxvYg=",
                 // Invalid characters
                 "[][]",
                 // Invalid characters which are commonly used as filename-safe


mtdowling (approved):

adamthom-amzn:
@david-perez if we're going to be more accepting of base64 than the tests allowed previously, we have the same sort of portability problem, but now undocumented. Can you add a positive request test forcing implementations that are strict about padding to correct the padding before decoding?

david-perez:
@adamthom-amzn A positive request test might break other off the shelf implementations.

```
>> atob("YmxvYg=");
Uncaught DOMException: String contains an invalid character
```

Does Smithy want to take a stance on this issue and enforce being lenient in what a server accepts?

adamthom-amzn:
It has to pick one or the other. If Rust is more lenient than TS, then the same smithy model cannot host a server with the same behavior on both backends.

david-perez:
The same argument goes for https://github.com/awslabs/smithy/issues/1116 and it was closed; it's another source of possible discrepancies among server implementations.

Personally, if Smithy has to take a stance on this, I'd prefer that it adopt [being lenient in what it accepts](https://en.wikipedia.org/wiki/Robustness_principle).

adamthom-amzn:
I don't agree that #1116 is the same - it's controlled by the HTTP spec, so there's no need for Smithy to additionally constrain it, whereas base64 in the contexts we accept it in is not controlled by any particular spec I am aware of.

I also don't really like the robustness principle - after all, it's also an excuse for some of our RPC protocols to accept "y", "n", 1, and 0 for booleans. That's now a requirement for every implementation of those protocols even though very few JSON parsers do that out of the box. It also means that every implementation of our HTTP binding protocols is backwards-incompatibly stricter than our existing implementation, because it will accept any date in any format in any location.

david-perez:
I don't understand how #1116 is not the same in the strict sense that it is a source of possible discrepancies among server implementations. The HTTP spec is unrelated to the fact that Smithy accepts binding any integer as an HTTP response code without prescribing what to do at runtime when a said integer falls outside the allowed range.

I'm also not fond of Postel's law being cited arbitrarily. In any case, if Smithy team decides it wants to be strict in accepting base64, this PR should be reverted and the documentation should be expanded. If we want to be lenient, I'll submit a positive test case on Monday.

david-perez:
My commit message is incorrect: it cites `Ymxvyg=`, but the case I removed was `YmxvYg=`.

david-perez:
After reading https://eprint.iacr.org/2022/361, I am now convinced that this test case should not be removed and that base64 decoding implementations should be strict about padding.

PR #1192, opened 2022-04-15
Implement comparable, add static compare method

srchase:
This CR implements Comparable on SourceLocation and adds a `compare` static method to FromSourceLocation for comparing SourceLocations.

This can be used to compare and sort shapes or traits, replacing the `SourceLocationSorter` and `TraitSorter` added in https://github.com/awslabs/smithy/pull/1175. It will also be used to sort shapes in the Smithy Language Server (https://github.com/awslabs/smithy-language-server/pull/35).

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/FromSourceLocation.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/FromSourceLocation.java
@@ -30,13 +30,13 @@ public interface FromSourceLocation {
     }
 
     /**
-     * Compares two SourceLocations.
+     * Compares two FromSourceLocations.
      *
-     * @param s1 the first SourceLocation to compare.
-     * @param s2 the second SourceLocation to compare.
+     * @param s1 the first FromSourceLocation to compare.
+     * @param s2 the second FromSourceLocation to compare.
      * @return the value 0 if s1 == s2; a value less than 0 if s1 < s2; and a value greater than 0 if s1 > s2.
      */
-    static int compare(SourceLocation s1, SourceLocation s2) {
-        return s1.compareTo(s2);
+    static int compare(FromSourceLocation s1, FromSourceLocation s2) {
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/SourceLocation.java
```diff
@@ -97,4 +99,23 @@ public int hashCode() {
 
         return h;
     }
+
+    public static class Sorter implements Comparator<FromSourceLocation>, Serializable {
```
mtdowling, line 103: Why implement Serializable? Why not add this functionality to SourceLocation directly and implement Comparable?

mtdowling: And what about moving this helper method to a static method on`FromSourceLocation` called `compare` to match how stuff like `Integer.compare` already works in Java?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/FromSourceLocation.java
```diff
@@ -28,4 +28,15 @@ public interface FromSourceLocation {
...
     }
+
+    /**
+     * Compares two SourceLocations.
+     *
+     * @param s1 the first SourceLocation to compare.
+     * @param s2 the second SourceLocation to compare.
+     * @return the value 0 if s1 == s2; a value less than 0 if s1 < s2; and a value greater than 0 if s1 > s2.
+     */
+    static int compare(SourceLocation s1, SourceLocation s2) {
```
mtdowling, line 39: This should take two FromSourceLocations and compare them by getting the source location of s1 and calling compareTo with the sourcelocaiton of s2

srchase: Fixed in  bb54034

mtdowling (approved):

Issue #1062, opened 2022-01-19
Silent execution/warnings as errors


lunaris:
Currently, it's not (I don't believe) possible to run `smithy build` in a manner that is truly quiet/silent -- one always receives some output. There is the `--logging` flag, which has an `OFF` value, but this does not apply to the output of the `Validator`, which always outputs statistics about how the model fared against various validations, etc.

A use case we have for truly silent execution is calling Smithy as part of a larger build process (e.g. Bazel). Here Smithy's output generates noise we'd rather not see. Of course, this only pertains to the "happy path", so the ideal would be to hook in the Validator's output to correspond to the logging level -- if the level is warning, Smithy outputs warnings; if the level if error, Smithy outputs errors and so on. Coupled with this it would be useful if Smithy had a flag like `-Werror` for promoting warnings to errors.

There are a few ways I can see about implementing this in the Smithy codebase (with https://github.com/awslabs/smithy/blob/7ea2682a5ab3c5cd6794c826c33121fa6dd40bf2/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/BuildCommand.java#L109 being a starting point), but wanted to raise an issue first to see a. if this is of interest/the use case is a valid one that makes sense and b. whether or not there are other considerations to be made that I have missed. Thanks in advance for your thoughts!

mtdowling:
Yeah,`-- logging` right now doesn't impact other informational output.

Both the `build` and `validate` commands have a `--severity` parameter which I think is a clearer way to control Smithy validation output (that is, showing the issues that were detected while validating the model, not the logging output). The default logging level I think is INFO, and that's controlled with `--logging`.

We have a `--debug` option which turns everything up all the way (--logging, --severity, etc). Or at least it should if it doesn't already.

Since we have a `--debug` flag, maybe we should add a mutually exclusive `--silent` flag? We'd check if the silent flag is set before writing to stdout/stderr for informational stuff. It could automatically set a default value of `--logging` to `OFF`, and `--severity` to `DANGER` so that you'd see when anything fails the build. Both `--logging` and `--severity` could be explicitly configured when using `--silent` too; for example, you could use `--silent` to turn off informational output, but set `--loging` to SEVERE to see more error information.

Would that work?

As for promoting warnings to errors, we have a feature request for that, but need to figure out a good place for it:  https://github.com/awslabs/smithy/issues/856

lunaris:
Thanks for the response!

[...]

PR #1195, opened 2022-04-20
Implement `--silent` for `smithy build`

lunaris:
*Issue #, if available:* #1062

*Description of changes:*

This commit adds the `--silent` flag to `smithy build`, which suppresses all non-error output during build. This includes validation statistics and information about plugins and projections. In the event that an error is encountered, output is printed and execution halted as normal.

For now, this PR only extends the `build` command. Validation is used in several places, such as `validate` and `ast`, but in these instances it is already passed a feature set containing `QUIET`. The code is a little hacky but I've started by making the changes as minimally invasive as possible. Additionally some test cases have been added to try and guard the functionality of the need case. Feedback welcome!

### DIFF
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/SmithyCli.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/SmithyCli.java
@@ -30,6 +30,7 @@ public final class SmithyCli {
     public static final String DISCOVER_CLASSPATH = "--discover-classpath";
     public static final String ALLOW_UNKNOWN_TRAITS = "--allow-unknown-traits";
     public static final String SEVERITY = "--severity";
+    public static final String SILENT = "--silent";
 
     private ClassLoader classLoader = getClass().getClassLoader();
     private boolean configureLogging;
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/BuildCommand.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/BuildCommand.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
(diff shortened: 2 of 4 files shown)


DavidOgunsAWS:
A better way to handle this flag would be to configure the stdout `Consumer<String>`  
member in https://github.com/awslabs/smithy/blob/main/smithy-cli/src/main/java/software/amazon/smithy/cli/Cli.java to a no-op when the `--silent` flag is picked up. After setting that up, the rest of the command execution can stay the same. All of the `Cli.stdout(...)` statements can remain without check and it'll still go nowhere.

lunaris:
Ah of course, I can't believe I missed that. However, this would disable error and danger outputs in the validator, wouldn't it? We'd want to change that logic there so that it used stderr, or something else?

DavidOgunsAWS:
I agree, and think we need associated changes needed to what is going to stdout and stderr in smithy-build. Errors observed in the smithy build logic should go into stderr.

lunaris:
So the plot thickens on this. I can make the necessary changes, but because the `setStdout` and associated methods are static on `Cli`, other tests start to fail because the change persists throughout the suite. Two options I can see here:

* Have `BuildCommand` clean up after itself, restoring `Cli`'s `stdout` property when it has `execute`d.
* Look at redesigning `Cli` so that we create instances of it, in which case the `stdout` change will be localised to the (presumably singleton) instance used in a single CLI execution (as in each test case, for instance).

Thoughts on this/any other solutions?

mtdowling:
Awe, sorry this is more work than it first seemed. The only public API surface area of the CLI we need to not change are the arguments and how it's invoked. All the innards of it can be completely changed if needed. We can also add this to our backlog of things to work on if it's more than you can commit to right now.

mtdowling:
I started working on https://github.com/awslabs/smithy/pull/1257 which should make this easier to implement.

PR #1215, opened 2022-05-05
Add `@nullable` trait and update proposal

mtdowling:
* Move `aws.api#clientOptional` to `smithy.api#nullable`. In order
  for diff tools to understand backward compatible changes, this trait has to
  be a core part of Smithy that tooling can reason about. The trait has
  utility outside of AWS too as it allows any model to define their own
  backward compatibility guarantees for a required member and does not
  require teams to ignore Smithy's built-in diff support that checks for
  allowed `@required` changes.
* Remove Smithy 2.0 traits and semantics from 1.0. Adding `@default` and
  allowing `@nullable` in 1.0 made an already complex nullability story
  impossibly complex. `@default` and `@nullable` is now only be allowed in
  Smithy 2.0.


designs/default-zero-value.md
```diff
@@ -263,24 +249,15 @@ structure Message {
...
+    selector: """
+        structure > member
         :not(> :test(union, structure > :test([trait|required])))"""",
-       conflicts: [required, box])
+    conflicts: [nullable, required]
+)
 structure default {}
 ```
 
-The `@default` trait conflicts with the `@box` trait and `@required` trait. The
```
JordonPhillips, line 273: What do these mean when used together?

designs/default-zero-value.md
```diff
@@ -376,22 +353,92 @@ optional members that are not marked with `@default` or `@required` implicitly
...
+## Backward compatibility rules
+
+Backward compatibility rules of the `@default`, `@required`, and `@nullable`
+traits are as follows:
 
-- The `@default` trait can never be removed from a member.
-- The `@default` trait can only be added to a member if the member was
-  previously marked as `@required`.
 - The `@required` trait cannot be added to a member. This would transition the
   member from nullable to non-nullable in generated code.
```
JordonPhillips, line 428: What if you add the required trait to a member that is nullable?

JordonPhillips (changes requested):

designs/default-zero-value.md
mtdowling: Yeah that would be fine. Will update.

mtdowling: Pushed up a clarification.

designs/default-zero-value.md
```diff
@@ -376,22 +353,96 @@ optional members that are not marked with `@default` or `@required` implicitly
...
+    @nullable
+    baz: String
+}
+```
+
+The primary use case of the `@nullable` trait is to indicate that while a
+member is _currently_ defined as `@required`, the service reserves the right to
+later remove the `@required` trait and make the member optional in the future.
+
+For example, `foo` in the following structure is considered a validation
```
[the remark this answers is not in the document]
mtdowling: Not sure I follow. It says: "For example, `foo` in the following structure is considered a validation constraint rather than a type refinement trait"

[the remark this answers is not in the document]
kstich: `foo` in this case is the member, but the _trait_ is the validation constraint and not the member.

>For example, the `@required` trait applied to `foo` in the following structure...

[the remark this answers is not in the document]
kstich: Similar change suggested for the nullable trait spec.

designs/default-zero-value.md
```diff
@@ -376,22 +357,96 @@ optional members that are not marked with `@default` or `@required` implicitly
...
+```
+structure Message {
+    @default    // ERROR: this trait conflicts with @nullable.
+    @nullable
+    title: String
+}
+```
+
+
+## Backward compatibility rules
```
kstich, line 426: These should be reflected in the evolving models guide.

mtdowling: Done. I also rewrote the nullability specs to be, I hope, more clear and define backward compatibility in the spec instead of just in the non-normative guide.

JordonPhillips (approved):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedNullability.java
```diff
@@ -0,0 +1,130 @@
...
+        NullableIndex oldIndex = NullableIndex.of(differences.getOldModel());
+        NullableIndex newIndex = NullableIndex.of(differences.getNewModel());
+
+        List<ValidationEvent> events = new ArrayList<>();
+
+        Stream<ChangedShape<MemberShape>> changed = Stream.concat(
+            // Get members that changed.
+            differences.changedShapes(MemberShape.class),
+            // Get members of structures that added/removed the input trait.
+            changedInputMembers(differences)
```
JordonPhillips, line 53: Do members not count as having changed if their traits change?

mtdowling: Yeah, any change to a member gets picked up

PR #1217, opened 2022-05-10
Improved Camel Case linter to automatically detect if upper or lower …

rchache:
…camel should be enforced for consistency

*Issue #, if available:*
Users who consistently use upper camel for member shape member names will otherwise be flooded with extraneous violations that complain their names are not lower camel.

*Description of changes:*
With this change, we will add a new (default) mode to automatically detect whether the model uses upper or lower camel case more, and enforce that pattern. 


smithy-linters/src/main/java/software/amazon/smithy/linters/CamelCaseValidator.java
```diff
@@ -135,11 +151,27 @@ public List<ValidationEvent> validate(Model model) {
...
+        Pattern memberNamePattern;
+        if (MemberNameHandling.AUTO.equals(config.getMemberNames())) {
+            long upperMemberNameCount = model.shapes(MemberShape.class)
+                    .filter(shape -> !shape.getContainer().getNamespace().equals(Prelude.NAMESPACE))
+                    .filter(shape -> MemberNameHandling.UPPER.getRegex().matcher(shape.getMemberName()).find())
+                    .count();
+            long lowerMemberNameCount = model.shapes(MemberShape.class)
+                    .filter(shape -> !shape.getContainer().getNamespace().equals(Prelude.NAMESPACE))
+                    .filter(shape -> MemberNameHandling.LOWER.getRegex().matcher(shape.getMemberName()).find())
+                    .count();
```
JordonPhillips, line 163: It would be better to use a single for-loop here. You can grab a cached set of members with `model.toSet(MemberShape.class)`. You're also going to want to filter out the following namespaces:

* `aws.apigateway`
* `aws.cloudformation`
* `aws.iam`
* `aws.api`
* `aws.auth`
* `aws.customizations`
* `aws.protocols`
* `smithy.test`

JordonPhillips (changes requested):

mtdowling:
I wonder of instead of needing to explicitly list namespaces to ignore (which is kind of breaking abstractions because this package shouldn't know about aws traits), we could instead validate services closures and potentially per/namespace?

So first we validate service closures by iterating over each service shape and use a `Walker` to find the closure of shapes connected to the service. We then do consistent casing on those.

Then, if we really wanted to and maybe we do this later, we could track shapes that weren't in any service closure and then validate per/namespace. That would catch things like inconsistencies across traits in a single namespace.

smithy-linters/src/main/java/software/amazon/smithy/linters/CamelCaseValidator.java
```diff
@@ -87,6 +94,21 @@ Pattern getRegex() {
...
+         */
+        AUTO {
+            Pattern getRegex() {
+                // No Pattern associated with auto setting
+                return null;
+            }
+
+            @Override
+            public String toString() {
+                return "consistent";
```
mtdowling, line 110: This should return "auto"

smithy-linters/src/main/java/software/amazon/smithy/linters/CamelCaseValidator.java
```diff
@@ -135,14 +157,66 @@ public List<ValidationEvent> validate(Model model) {
...
                 .forEach(events::add);
 
-        Pattern isValidMemberName = config.getMemberNames().getRegex();
-        model.shapes(MemberShape.class)
-                .filter(shape -> !isValidMemberName.matcher(shape.getMemberName()).find())
-                .map(shape -> danger(shape, format(
-                        "Member shape member name, `%s`, is not %s camel case",
-                        shape.getMemberName(), config.getMemberNames())))
-                .forEach(events::add);
+        // First validate all member shapes connected to services
```
mtdowling, line 160: This is close, but you should group by service closures not validate that any shape in any service closure uses consistent casing. For example, a model might contain two services: one service uses lower case members, and the other uses upper case. We would want to emit correctly for both of these.

smithy-linters/src/main/java/software/amazon/smithy/linters/CamelCaseValidator.java
```diff
@@ -135,14 +157,66 @@ public List<ValidationEvent> validate(Model model) {
...
+        Set<Shape> serviceClosure = new HashSet<>();
+        for (ServiceShape serviceShape : model.getServiceShapes()) {
+            Walker walker = new Walker(model);
+            walker.iterateShapes(serviceShape).forEachRemaining(serviceClosure::add);
```
mtdowling, line 164: You should get the set of shapes in the closure and validate those shapes. Then also add all the shapes found in the closure to `serviceClosure` to exclude it from the namespace checks.

smithy-linters/src/main/java/software/amazon/smithy/linters/CamelCaseValidator.java
```diff
@@ -135,14 +157,66 @@ public List<ValidationEvent> validate(Model model) {
...
+                    ? nonUpperCamelMemberShapes : nonLowerCamelMemberShapes;
+        } else if (MemberNameHandling.UPPER.equals(config.getMemberNames())) {
+            violatingMemberShapes = nonUpperCamelMemberShapes;
+        } else if (MemberNameHandling.LOWER.equals(config.getMemberNames())) {
+            violatingMemberShapes = nonLowerCamelMemberShapes;
+        }
+
+        return violatingMemberShapes.stream()
+                .map(shape -> danger(shape, format(
+                        "Member shape member name, `%s`, is not %s camel case. "
```
mtdowling, line 217: I think the error message should be more specific. For example, if set to AUTO, then say in this error message what AUTO resolved to whether it was uppercase or lowercase. It would be nice if we also said if this was a service closure or namespace too. For example:

For a service:

```
`UpperCamel` is not lower camel case; members in the FooService must all use lower camel case.
```

And for a namespace:

```
`UpperCamel` is not lower camel case; members in the com.baz namespace must all use lower camel case.
```

mtdowling (changes requested):

mtdowling (approved):

kstich (approved):

JordonPhillips (approved):

PR #1231, opened 2022-05-19
Add ability to elide targets for mixins/resources

JordonPhillips:
This adds in syntactic sugar that allows modelers to elide targets for members that are inherited from mixins or defined in resources.

This necessarily also adds the ability for a structure to indicate a resource to pull potential targets from.

This is all syntactic sugar for the IDL, so the JSON AST isn't impacted.

This is accomplished by refactoring the code used for mixins to instead be a generic system for applying modifications to a shape after some set of dependencies has been resolved.

Note that we'll also need to update the syntax highlighter in a follow-up pr.


smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
```diff
@@ -285,10 +260,106 @@ private <S extends Shape, B extends AbstractShapeBuilder<? extends B, S>> Option
...
             }
             return Optional.of(builder.build());
+        } catch (IllegalStateException e) {
+            if (builder.getShapeType() == ShapeType.MEMBER && ((MemberShape.Builder) builder).getTarget() == null) {
+                events.add(ValidationEvent.builder()
+                        .severity(Severity.ERROR)
+                        .id(Validator.MODEL_ERROR)
+                        .shapeId(builder.getId())
+                        .sourceLocation(builder.getSourceLocation())
+                        .message("Member target was elided, but no parent shape contained a matching member name or "
```
kstich, line 270: This error message should have resource identifier first, since it's checked first.

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
```diff
@@ -285,10 +260,106 @@ private <S extends Shape, B extends AbstractShapeBuilder<? extends B, S>> Option
...
+        }
+
+        @Override
+        public void modifyShape(
+                AbstractShapeBuilder<?, ?> builder,
+                Map<String, MemberShape.Builder> memberBuilders,
+                TraitContainer resolvedTraits,
+                Map<ShapeId, Shape> shapeMap
+        ) {
+            Shape mixinShape = shapeMap.get(mixin);
```
kstich, line 324: Lost the comment here when this was broken out.

kstich (changes requested):

docs/source/1.0/spec/core/idl.rst
```diff
@@ -190,12 +190,14 @@ The Smithy IDL is defined by the following ABNF:
...
-    shape_member_kvp       :`trait_statements` `identifier` `ws` ":" `ws` `shape_id`
+    shape_members          :"{" `ws` *(`trait_statements` `shape_member` `ws`) "}"
+    shape_member           :`shape_member_kvp` / `shape_member_elided`
+    shape_member_kvp       :`identifier` `ws` ":" `ws` `shape_id`
+    shape_member_elided    :"$" `identifier`
     list_statement :"list" `ws` `identifier` [`mixins`] `ws` `shape_members`
     set_statement :"set" `ws` `identifier` [`mixins`] `ws` `shape_members`
     map_statement :"map" `ws` `identifier` [`mixins`] `ws` `shape_members`
-    structure_statement     :"structure" `ws` `identifier` [`mixins`] `ws` `shape_members`
+    structure_statement     :"structure" `ws` `identifier` ["from" `shape_id`] [`mixins`] `ws` `shape_members`
```
mtdowling, line 200: we've been saying "for" in mockups, and I can't help but think "for" is better than "from".

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
```diff
@@ -285,10 +260,106 @@ private <S extends Shape, B extends AbstractShapeBuilder<? extends B, S>> Option
                 builder.addTrait(trait);
             }
             return Optional.of(builder.build());
+        } catch (IllegalStateException e) {
+            if (builder.getShapeType() == ShapeType.MEMBER && ((MemberShape.Builder) builder).getTarget() == null) {
```
mtdowling, line 264: nit - should we throw a more specific exception for this case?

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
```diff
@@ -285,10 +260,106 @@ private <S extends Shape, B extends AbstractShapeBuilder<? extends B, S>> Option
...
+            }
+            throw e;
         } catch (SourceException e) {
             events.add(ValidationEvent.fromSourceException(e, "", builder.getId()));
             resolvedTraits.clearTraitsForShape(builder.getId());
             return Optional.empty();
         }
     }
+
+    private static class ApplyMixin implements PendingShapeModifier {
```
mtdowling, line 283: I really like this refactor :)

It might be nice to just move this to the upper level and keep it package-private to make the code easier to read.

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
```diff
@@ -285,10 +260,106 @@ private <S extends Shape, B extends AbstractShapeBuilder<? extends B, S>> Option
...
+        private ShapeId mixin;
```
mtdowling, line 284: nit: can this be final too?

smithy-model/src/test/resources/software/amazon/smithy/model/loader/invalid/elide-nonexistent-id.smithy
```diff
@@ -0,0 +1,10 @@
+// Member target was elided, but no parent shape contained a matching member name or resource identifier name.
```
mtdowling, line 1: instead of "parent shape" can we be more explicit and say "no mixin or resource contained a matching"...

smithy-model/src/test/resources/software/amazon/smithy/model/loader/invalid/elided-type-conflicts.smithy
```diff
@@ -0,0 +1,19 @@
+// smithy.example#Invalid$id: Member conflicts with an inherited mixin member: smithy.example#MixinStructure$id
```
mtdowling, line 1: This is a good test case! But to be clear -- this validation is knows nothing about elision right? It's like an orthogonal validation step for ensuring mixins are sound?

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -532,25 +541,33 @@ private void parseMember(ShapeId parent, Set<String> allowed, Set<String> define
         }
 
         ShapeId memberId = parent.withMember(memberName);
+
+        if (isTargetElided && !modelFile.getVersion().supportsTargetElision()) {
```
mtdowling, line 545: why is this not right after detecting elision?

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -948,4 +996,45 @@ private String peekDebugMessage() {
 
         return result.length() == 0 ? "[EOF]" : result.toString();
     }
+
+    private static final class SetResourceBasedTargets implements PendingShapeModifier {
```
mtdowling, line 1000: nit: Move to outer class for readability? This class is already 1,000 lines.

mtdowling (changes requested):

docs/source/1.0/spec/core/idl.rst
```diff
@@ -1438,6 +1445,97 @@ For example:
...
+    }
+
+Note that the ``UserSummary`` structure does not attempt to define the
+``uuid`` identifier. When referencing a resource in this way, only the
+identifiers that are explicitly referenced are added to the structure. This
+allows structures to define subsets of identifiers, which can be useful for
+operations like create operations where some of those identifiers may be
+generated by the service.
+
+The target for a resource identifier can also be elided when the member's name
```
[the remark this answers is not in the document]
JordonPhillips: > The current approach requires that traits are resolved before we can make sense of the resource binding (i.e., we need to know that resourceIdentifier is smithy.api#resourceIdentifier and not some other overloaded trait).

The resourceIdentifier trait will already have been resolved before we start looking at it, so this isn't adding any additional complexity. And we know it's `smithy.api#resourceIdentifier` because we're checking for that exact shape id.

The difference between having this feature and not having it is 3 lines of code, and omitting it makes it feel incomplete.

> We automatically inject a resourceIdentifier trait for the member

What is the benefit of doing this?

docs/source/1.0/spec/core/idl.rst
```diff
@@ -1438,6 +1445,97 @@ For example:
     string SensitiveText with [SensitiveString]
 
 
+.. _idl-target-elision:
+
+Target Elision
```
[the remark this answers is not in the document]
JordonPhillips: A structure referencing a resource isn't useful by itself though, so I don't think it can reasonably have its own section. And starting with the mixin case here seems more natural to me since it follows the mixin section.

smithy-model/src/test/resources/software/amazon/smithy/model/loader/invalid/elided-type-conflicts.smithy
JordonPhillips: Yes. By the time we get to this validator, the member looks just like a normal member, whether the target was set based on a mixin or a resource.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
JordonPhillips: Using the member id gives a much better error message

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
JordonPhillips: The comment omitted no longer makes sense. There's javadoc on the `ApplyMixin` class that serves the same purpose.

smithy-model/src/main/java/software/amazon/smithy/model/loader/AbstractMutableModelFile.java
JordonPhillips: I'm not sure adding any more detail would actually be more clear. We could list all the bound shapes, but that just seems like noise

mtdowling: I more meant that you didn’t wouldn’t have to do as much introspection of the exception if there was a more specific type to try/catch. But it’s fine.

docs/source/1.0/spec/core/idl.rst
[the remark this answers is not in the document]
mtdowling: Even if it’s a few lines, I think being able to use indirection to refer to properties of a resource adds a lot complexity for readers, potentially other tools implemented differently, and in IDEs that are trying to jump to the definition too.

Injecting resourceProperty traits and eventually property traits would be nice because “$” is adding these bindings explicitly through syntax, so rather than rely on implicit bindings later, we add the explicit traits. This makes it so that you can use selectors to look for these traits. Before doing this, I’d want to make sure this would work with resource properties when they ship though.

kstich (approved):

docs/source/1.0/spec/core/idl.rst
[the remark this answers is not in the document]
mtdowling: This was resolved. What ended up being the associated change?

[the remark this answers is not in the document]
JordonPhillips: You can see the rest int he most recent diff

mtdowling (changes requested):
Minor changes, but shipit after addressing.

PR #1235, opened 2022-05-23
Remove trait definitions as part of default codegen transforms

gosar:
IIUC trait definitions itself aren't needed by codegen (clients, server, types). And current code generators like typescript, ruby, go are using `getModelWithoutTraitShapes` currently anyways.

If it is better to leave the model with traits (note this includes traits like `aws.cloudformation#cfnAdditionalIdentifier` in the assembled model, even if the model file does not reference it), we can update individual code generators to use appropriate walker/transform in the specific places as needed.

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
@@ -79,13 +79,13 @@ public final class CodegenDirector<
      *
      * <ul>
      *     <li>Flattens error hierarchies onto every operation.</li>
-     *     <li>Converts enum strings to enum shapes.</li>
      *     <li>Flattens mixins</li>
+     *     <li>Converts enum strings to enum shapes.</li>
+     *     <li>Removes trait definition shapes.</li>
      * </ul>
      *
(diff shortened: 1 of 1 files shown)


mtdowling:
Is this necessary though? Directed codegen invokes the methods to perform codegen for you anyways, so you don’t need to filter them out. I think each directive also provides methods to get the shapes connected to the service. Removing trait shapes limits the kinds of selectors you can perform too, like finding protocol bearing traits.

gosar:
Thanks @mtdowling. When discussing with some other folks in the team, it seemed (maybe I misunderstood or the context wasn't right) that there isn't a need to have the trait definition shapes in the model. Yes, `Directive.connectedShapes` provides the Walker usage, but the current codegen logic explicitly uses [`getModelWithoutTraitShapes` in the Walker](https://github.com/awslabs/smithy-typescript/blob/cde5684b442488d53818fa59362d79a153da6ec6/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/CodegenVisitor.java#L118) making it seem necessary.

I wasn't sure overall and hence made this PR as a way to discuss and check if this is ok. Thanks for pointing out about the protocol traits - it indeed fails to find the protocols with this change and that fails elsewhere.

PR #1236, opened 2022-05-24
Encapsulate Walker inside generateShapesInService

gosar:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
@@ -266,18 +266,22 @@ public final class CodegenDirector<
         preprocessModelWithIntegrations(integrations);
 
         ServiceShape serviceShape = model.expectShape(service, ServiceShape.class);
-        Set<Shape> shapes = new Walker(model).walkShapes(serviceShape);
 
         SymbolProvider provider = createSymbolProvider(integrations, serviceShape);
 
         C context = createContext(serviceShape, provider);
+
+        // After the context is created, it holds the model that should be used for rest of codegen. So `model` should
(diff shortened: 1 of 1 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
```diff
@@ -266,7 +266,6 @@ public void run() {
         preprocessModelWithIntegrations(integrations);
 
         ServiceShape serviceShape = model.expectShape(service, ServiceShape.class);
-        Set<Shape> shapes = new Walker(model).walkShapes(serviceShape);
```
gosar, line 269: moving this inside `generateShapesInService` should not affect anything, as none of the logic in between updates the `model`.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
```diff
@@ -367,8 +366,9 @@ private void registerInterceptors(C context, List<I> integrations) {
         context.writerDelegator().setInterceptors(interceptors);
     }
 
-    private void generateShapesInService(C context, ServiceShape serviceShape, Set<Shape> shapes) {
+    private void generateShapesInService(C context, ServiceShape serviceShape) {
         LOGGER.fine(() -> "Generating shapes for " + directedCodegen.getClass().getName());
+        Set<Shape> shapes = new Walker(context.model()).walkShapes(serviceShape);
```
gosar, line 371: Technically, before `context.model()` could be different than `model` because `createContext` _could_ modify the model. But it shouldn't. Using a different model object to walk v/s topological index doesn't make sense. As such could probably use `model` in both places here, but `context.model()` is more appropriate I think.

kstich (approved):

mtdowling (approved):

mtdowling (approved):

PR #1237, opened 2022-05-25
Make SmithyIntegrations available from CodegenContext

gosar:
This PR has 2 commits.

The 1st commit, only made `SmithyIntegration`s available from CreateContextDirective, so subclasses
of CodegenContext can access it. But it likely can be made part of parent CodegenContext interface.

The 2nd commit, made it part of CodegenContext interface. ~~This requires another generic parameter to be added in multiple places.~~ Avoided adding generic parameter to CodegenContext by defining the method as `List<? extends SmithyIntegration<S, W, ?>> integrations();` where subclasses can return more specific `List<MyIntegration> integrations();`

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/DirectedCodegen.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/DirectedCodegen.java
@@ -30,8 +30,9 @@ import software.amazon.smithy.codegen.core.WriterDelegator;
  *
  * @param <C> Smithy {@link CodegenContext} to use in directed methods.
  * @param <S> Settings object passed to directed methods as part of the context.
+ * @param <I> {@link SmithyIntegration} type to use in directed methods.
  */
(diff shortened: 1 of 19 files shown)


mtdowling (approved):

mtdowling (approved):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/DirectedCodegen.java
```diff
@@ -30,8 +30,9 @@
  *
  * @param <C> Smithy {@link CodegenContext} to use in directed methods.
  * @param <S> Settings object passed to directed methods as part of the context.
+ * @param <I> {@link SmithyIntegration} type to use in directed methods.
  */
-public interface DirectedCodegen<C extends CodegenContext<S, ?>, S> {
+public interface DirectedCodegen<C extends CodegenContext<S, ?, I>, S, I extends SmithyIntegration<S, ?, ?>> {
```
gosar, line 35: hmm, this can be `I extends SmithyIntegration<S, ?, C>>` since `C` is defined here too

PR #1238, opened 2022-05-25
Sort events in NodeDiff for ease of use

kstich:


smithy-model/src/main/java/software/amazon/smithy/model/node/NodeDiff.java
```diff
@@ -32,10 +32,11 @@ final class NodeDiff {
 
     static List<String> diff(ToNode actual, ToNode expected) {
         return new NodeDiff().findDifferences(actual.toNode(), expected.toNode(), "")
+                .sorted()
                 .collect(Collectors.toList());
     }
 
-    Stream<String> findDifferences(Node actual, Node expected, String prefix) {
+    private Stream<String> findDifferences(Node actual, Node expected, String prefix) {
```
kstich, line 39: This is package private but unused elsewhere, so locking it down seems safe and ensures sorted events still.

mtdowling (approved):

PR #1248, opened 2022-06-01
Rule-Engine: Experimental Context Traits

skmcgrail:
Adds support for new AWS context traits `clientContextParams`, `staticContextParams`, and `contextParam`.

* `smithy.rules#clientContextParams` is a trait that may be applied to a Service shape. It is used to indicate that the named context parameters that should be configurable on a top-level API client. The trait is a mapping of the context parameter name to their definition containing the Smithy shape type used to represent it, documentation string for the client, and context usages. Valid usages are `endpoint`, and valid shape types supported are `string` and `boolean`.
* `smithy.rules#staticContextParams` is a trait that may be applied to an Operation shape. It is a mapping of context parameter names to their definition containing the static value to be configured for the parameter, and context usages. Valid usages are `endpoint`.
* `smithy.rules#contextParam` is a trait that may be applied to a Member shape. The trait indicates the target context parameter name that the member's value should be bound to as, as well as the context usages. Valid usages are `endpoint`. Per specification this trait will only be supported on the top-level operation input shape.

* `smithy.rules#endpointRuleSet` is a trait that is used to embed a rule-set document.

## Example Model
```smithy
$version: "1.0"

namespace ns.foo

use smithy.rules#clientContextParams
use smithy.rules#staticContextParams
use smithy.rules#contextParam
use smithy.rules#endpointRuleSet

@clientContextParams(
    stringParam: {type: "string", documentation: "a client string parameter"},
    booleanParam: {type: "boolean", documentation: "a client boolean parameter"}
)
service ExampleService {
    version: "2022-01-01",
    operations: [GetThing]
}

apply ExampleService @endpointRuleSet({
    version: "1.0",
    serviceId: "example",
    parameters: {
        stringParam: {type: "string"},
        booleanParam: {type: "boolean"}
    },
    rules: []
})

@readonly
@staticContextParams(
    stringParam: {value: "some value"},
    booleanParam: {value: true}
)
operation GetThing {
    input: GetThingInput
}

@input
structure GetThingInput {
    fizz: String,

    @contextParam(name: "stringParam")
    buzz: String,
}

```


mtdowling:
I’m tempted to make this part of Smithy core rather than something AWS-specific. Seems simple enough so far, but I guess the decision hinges on the associated endpoint rules configuration and whether they can be generic enough too.

rcoh:
My recommendation would be to have this be in Smithy core—this would influence code generation of endpoint parameter structures / configuration wiring.

For the rules, I'd put them in something like `smithy.experimental.endpoints`—I think it's too early to move them into core

smithy-aws-traits/src/main/resources/META-INF/smithy/aws.experimental.smithy
```diff
@@ -0,0 +1,70 @@
+$version: "1.0"
```
kstich, line 1: Each shape/member in this file should have documentation. Will need specification updates for the traits as well.

kstich (changes requested):

skmcgrail:
Implemented all feedback, except for the final documentation pieces within the model.

skmcgrail:
Rebased commits on top of 1.22.0.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/StaticContextParamsTrait.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        @Override
+        public Trait createTrait(ShapeId target, Node value) {
+            NodeMapper mapper = new NodeMapper();
+            Map<String, StaticContextParamDefinition> parameters = new LinkedHashMap<>();
+            value.expectObjectNode().getMembers().forEach((stringNode, node) -> {
+                parameters.put(stringNode.getValue(), mapper.deserialize(node, StaticContextParamDefinition.class));
+            });
+            return builder()
```
mtdowling, line 100: Would be good to set the nodeCache here too. Example: https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/traits/HttpTrait.java#L60

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/StaticContextParamsTrait.java
```diff
@@ -0,0 +1,140 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public Map<String, StaticContextParamDefinition> getParameters() {
+        return parameters;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 55: equals and hashCode should be covered by AbstractTrait

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
```diff
@@ -0,0 +1,53 @@
...
+
+namespace smithy.rules
+
+@unstable
+@trait(selector: "service")
+document endpointRuleSet
+
+@unstable
+@trait(selector: "service")
+document endpointTests
```
mtdowling, line 11: Is the format of endpointTests possible to model?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTrait.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public SmithyBuilder<EndpointTestsTrait> toBuilder() {
+        return builder()
+                .sourceLocation(getSourceLocation())
+                .tests(tests);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 60: equals and hashCode should be handled by AbstractTrait

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointRuleSetTrait.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public SmithyBuilder<EndpointRuleSetTrait> toBuilder() {
+        return builder()
+                .sourceLocation(getSourceLocation())
+                .ruleSet(ruleSet);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 60: equals and hashCode are covered by AbstractTrait

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextParamTrait.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    public SmithyBuilder<ContextParamTrait> toBuilder() {
+        return new Builder()
+                .sourceLocation(getSourceLocation())
+                .definition(definition);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 67: equals and hashCode are covered by AbstractTrait

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextParamDefinition.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+package software.amazon.smithy.rulesengine.traits;
+
+import java.util.Objects;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+@SmithyUnstableApi
+public final class ContextParamDefinition {
```
mtdowling, line 23: should implement ToSmithyBuilder

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ClientContextParamDefinition.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private ClientContextParamDefinition(Builder builder) {
+        this.type = SmithyBuilder.requiredState("type", builder.type);
+        this.documentation = builder.documentation;
+    }
+
+    public ShapeType getType() {
+        return type;
+    }
+
```
mtdowling, line 38: nit - blank lines

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ClientContextParamsTrait.java
```diff
@@ -0,0 +1,141 @@
...
+        super(ID, builder.getSourceLocation());
+        this.parameters = builder.parameters.copy();
+    }
+
+    public Map<String, ClientContextParamDefinition> getParameters() {
+        return parameters;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 52: equals and hashCode are covered by AbstractTrait

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ClientContextParamsTrait.java
```diff
@@ -0,0 +1,141 @@
...
+        @Override
+        public Trait createTrait(ShapeId target, Node value) {
+            NodeMapper mapper = new NodeMapper();
+
+            Map<String, ClientContextParamDefinition> parameters = new LinkedHashMap<>();
+            value.expectObjectNode().getMembers().forEach((stringNode, node) -> {
+                parameters.put(stringNode.getValue(), mapper.deserialize(node, ClientContextParamDefinition.class));
+            });
+
+            return builder()
```
mtdowling, line 102: would be good to set the nodeCache here too (see other comment)

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextParamTrait.java
```diff
@@ -0,0 +1,118 @@
...
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+/**
+ * Binds a Structure member shape to an AWS endpoint parameter.
+ */
+@SmithyUnstableApi
+public final class ContextParamTrait extends AbstractTrait implements ToSmithyBuilder<ContextParamTrait> {
+    public static final ShapeId ID = ShapeId.from("smithy.rules#contextParam");
+
+    private final ContextParamDefinition definition;
```
mtdowling, line 36: Any reason why this trait isn't providing access to this property?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextParamDefinition.java
mtdowling: Why make this a class (and the other Definition classes)? Why not just use ContextParamTrait and provide access to `name` through there instead of adding this layer? We typically use the traits themselves to carry their own information.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextIndex.java
```diff
@@ -0,0 +1,103 @@
...
+        for (OperationShape operation : model.getOperationShapes()) {
+            operation.getTrait(StaticContextParamsTrait.class).ifPresent(endpointParamsTrait -> {
+                operationParams.put(operation.toShapeId(), endpointParamsTrait.getParameters());
+            });
+            operation.getInput().ifPresent(shapeId -> {
+                StructureShape input = model.expectShape(shapeId, StructureShape.class);
+                Map<MemberShape, ContextParamDefinition> parameterReferenceMap = new LinkedHashMap<>();
+                input.members().forEach(memberShape -> {
+                    memberShape.getTrait(ContextParamTrait.class).ifPresent(contextParamTrait -> {
+                        parameterReferenceMap.put(memberShape, ContextParamDefinition.builder()
```
mtdowling, line 58: I think this is a kind of deficiency with the current design that separates traits from the data they contain

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextIndex.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Gets the mapping of context parameter names and corresponding {@link ClientContextParamDefinition} to be
+     * generated on the service client's configuration.
+     *
+     * @param service The service shape.
+     * @return The mapping of context parameter names to {@link ClientContextParamDefinition}.
+     */
+    public Map<String, ClientContextParamDefinition> getClientContextParams(ToShapeId service) {
```
mtdowling, line 79: I think for this index, maybe we save the overhead and memory of precomputing this and just return the trait for a `Shape` rather than a `ToShapeId`. Same comment for getStaticContextParams since there's no actual computation.

mtdowling (changes requested):

smithy-rules-engine/src/test/resources/software/amazon/smithy/rulesengine/traits/traits-test-model.smithy
```diff
@@ -0,0 +1,49 @@
...
+    rules: []
+})
+
+apply ExampleService @endpointTests({
+    endpoint_tests: []
+})
+
+@readonly
+@staticContextParams(
+    stringParam: {value: "some value"},
```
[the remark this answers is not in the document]
skmcgrail: Good question, I know we had discussed this, but I might have gotten lost on what our final decision was given that this originally still had the `usages` concept tied to it. I'm inclined to say, despite having removed `usages` for the time being, we should keep this as defined right now. Largely if we wish to adopt the `usages` pattern at a later date, then this trait can simply be updated to include that property. The other reason would be that the other two traits are already extensible if we require additional properties in the future, which would make this asymmetrical if we aren't able to easily extend the feature set here as well.

But I will go with your best judgement here if you feel like we should further simplify this.

skmcgrail:
Good for follow-up review.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/ContextIndex.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Resolves an indexes the context parameters in the model.
+ */
+@SmithyUnstableApi
+public final class ContextIndex implements KnowledgeIndex {
+    private final Model model;
+
+    public ContextIndex(Model model) {
+        this.model = model;
```
mtdowling, line 38: Make this a WeakReference to avoid a cyclical reference

mtdowling (changes requested):

mtdowling (approved):

kstich (changes requested):

kstich (approved):

PR #1262, opened 2022-06-09
Update zero value proposal and implementation

mtdowling:
* Remove default zero values from structures. Assigning zero values to
  structures adds complexity to serializers that try to omit members set
  to their zero values (they would have needed to recursively inspect
  structure member values or generate some kind of "isZeroValue" flag).
  Rather than add default zero values to structures to aid in migrating
  required structures to default structures, services can choose to apply
  the `@clientOptional` trait to the member to ensure that clients generate
  the member as optional, and the service reserves the right to remove the
  `@required` trait in the future.
* Rename `@nullable` to `@clientOptional`. `@nullable` caused confusion;
  some assumed it meant that a member accepts explicit null values when it
  actually is just intended to make clients generate the member as optional.


JordonPhillips (approved):

designs/default-zero-value.md
```diff
@@ -761,6 +751,19 @@ See https://github.com/awslabs/smithy/blob/main/designs/operation-input-output-a
 
 ## Updates
 
+* 2022-06-8
```
gosar, line 754: nit: 2022-06-08

gosar: Fixing in https://github.com/awslabs/smithy/pull/1267

PR #1275, opened 2022-06-15
Add option to enable prelude serialization

srchase:
This PR adds an option to enable serializing prelude shapes.

By default, prelude shapes are filtered from the output of `SmithyIdlModelSerializer.serialize()`.


smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -302,6 +309,16 @@ public Builder basePath(Path basePath) {
             return this;
         }
 
+        /**
+         * Enables serializing shapes in the Smithy prelude.
+         * Defaults to false.
+         * @return Returns the builder.
+         */
+        public Builder enableSerializingPrelude() {
```
mtdowling, line 317: nit: I think something like `serializePrelude` would be a much more succinct name

srchase: fixed in 28dc5ac9b7879f02a778303931c750541b04ce98.

mtdowling:
What's the motivation of this change?

srchase:
> What's the motivation of this change?

SmithyIdlModelSerializer will be used by the Hover capability of the Language Server to show the information about shapes/traits in a model. I want to make this available for showing the same information about prelude shapes.

For example, if `@input` is hovered on in a client, the following will be shown:
```
/// Specializes a structure for use only as the input of a single operation.
@trait(
    selector: "structure",
    conflicts: [output, error],
    breakingChanges: [{change: "presence"}]
)
structure input {}
```

mtdowling (approved):

PR #1278, opened 2022-06-17
Deprecate sets in favor of uniqueItems

mtdowling:
This commit is a reversal of a previous decision to deprecate
uniqueItems. This reversal is due to an observation that sets
aren't all that useful or desirable for client/server scenarios,
which is the primary use case for Smithy.

First, we added a requirement that sets maintain insertion order.
This was to guarantee that validation errors returned by servers
are actionable by clients. If a client and server don't agree on
what element `0` means, then a client doesn't know which value in
their unordered set is incorrect.

Next, we limited what kinds of values can be stored in a set because
some types like float makes equality checks in languages like Rust
difficult. We also realized that clients with imperfect knowledge
of a model (which is basically all clients) often drop unknown
structure members. If a server sends two structures to a client
and one of the structures uses a member the client doesn't know
about, the client will drop the member, and then drop the second
structure because it considers it a duplicate of the first.

So at this point, we realized that sets are more confusing than
useful. The ideal state is that clients ignore unique-ness
constraints and leave that to servers to perform as _validation_
and not as part of the type system. It turns out, that's exactly
the use case of the @uniqueItems trait.

So instead of removing the @uniqueItems trait, we're doubling
down on it. The trait can once again target any kind of type
(except for float, double, and document), we're defining how to
determine equality, order is implicit because the trait targets a
list, and the fact that this is _validation_ and not type
refinement is implicit because it is a "constraint trait".

Sets will be removed from IDL v2.

*Issue #, if available:*

*Description of changes:*


kstich (approved):

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/unique-items-floats.errors
```diff
@@ -0,0 +1,6 @@
+[WARNING] smithy.example#Floats: This shape marked with @uniqueItems contains a float in its closure (`smithy.api#Float`). Equality comparisons for floating point numbers or document types that can contain floating point numbers can be problematic. | UniqueItems
+[WARNING] smithy.example#Doubles: This shape marked with @uniqueItems contains a double in its closure (`smithy.api#Double`). Equality comparisons for floating point numbers or document types that can contain floating point numbers can be problematic. | UniqueItems
+[WARNING] smithy.example#Documents: This shape marked with @uniqueItems contains a document in its closure (`smithy.api#Document`). Equality comparisons for floating point numbers or document types that can contain floating point numbers can be problematic. | UniqueItems
+[WARNING] smithy.example#Structures: This shape marked with @uniqueItems contains a document in its closure (`smithy.api#Document`). Equality comparisons for floating point numbers or document types that can contain floating point numbers can be problematic. | UniqueItems
```
kstich, line 4: Should this be pointing to the member or is it a duplicate?

mtdowling: I made it so warnings for the same shape are grouped now.

kstich (approved):

PR #1280, opened 2022-06-17
Copy maven metadata on local publish

srchase:
This PR adds a build task that copies the maven metadata to `maven-metadata.xml` in the `.m2` repository.

Currently, the `publishToMavenLocal` task writes maven metadata to `maven-metadata-local.xml`. Some downstream tools, including Coursier only look for `maven-metadata.xml` when resolving dependencies from the `.m2` repository.

This task will run after `publishToMavenLocal` to make the same metadata available in both `maven-metadata-local.xml` and `maven-metadata.xml`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,6 +178,17 @@ subprojects {
         }
     }
 
+    task copyMavenMetadataForDevelopment(type: Copy) {
+        from('build/tmp/publishMavenJavaPublicationToMavenLocal') {
+            rename 'module-maven-metadata.xml', 'maven-metadata.xml'
+        }
+
+        def wdir = System.getProperty("user.home") + '/.m2/repository/software/amazon/smithy/' + project.name
+        into(wdir)
+    }
+
+    publishToMavenLocal.finalizedBy(copyMavenMetadataForDevelopment)
+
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -178,6 +178,17 @@ subprojects {
         }
     }
 
+    task copyMavenMetadataForDevelopment(type: Copy) {
+        from('build/tmp/publishMavenJavaPublicationToMavenLocal') {
+            rename 'module-maven-metadata.xml', 'maven-metadata.xml'
```
mtdowling, line 183: is this actually renaming the file? Is that necessary if we just want to copy it?

srchase: No, it isn't renaming the file in the source directory, but renames only during the copy. This is necessary, otherwise the copy task would write `module-maven-metadata.xml` to the target destination.

After this task runs, the original file in build/tmp/publishMavenJavaPublicationToMavenLocal still exists, e.g.`smithy-model/build/tmp/publishMavenJavaPublicationToMavenLocal/module-maven-metadata.xml`.

mtdowling (approved):

PR #1291, opened 2022-06-29
Do not map required members that are only used for output

sugmanue:
*Description of changes:*  This change excludes from the CFN schema `required` element all shape members that even though are marked as `@required` in the model are *only* used in output operations. 


kstich (approved):

sugmanue:
I will create a new PR to remove the extra commits.

Issue #1265, opened 2022-06-10
OpenAPI does not output multiple errors for the same status code


eduardomourar:
Whenever we have errors shapes that are mapped to the same HTTP status code, only one of them appears in the OpenAPI document generated by the Smithy plugin.

<details>
  <summary>Input Smithy model</summary>

```
namespace smithy.example

use aws.protocols#restJson1
use smithy.framework#ValidationException

@restJson1
service Weather {
    version: "2006-03-01",
    operations: [GetCurrentTime],
    errors: [
        QuotaExceededException,
        ThrottlingException,
        ValidationException
    ]
}

@error("client")
@retryable(throttling: true)
@httpError(429)
structure QuotaExceededException {
    @required
    message: String,
}

@error("client")
@retryable(throttling: true)
@httpError(429)
structure ThrottlingException {
    @required
    message: String,
}

@readonly
@http(uri: "/time", method: "GET")
operation GetCurrentTime {
    input: GetCurrentTimeInput,
    output: GetCurrentTimeOutput
}

@input
structure GetCurrentTimeInput {}

@output
structure GetCurrentTimeOutput {
    @required
    time: Timestamp
}
```

</details>

You can see below that only the `ThrottlingException` has been placed in the OpenAPI components section:

<details>
  <summary>OpenAPI output</summary>

```json
{
    "openapi": "3.0.2",
    "info": {
        "title": "Weather",
        "version": "2006-03-01"
    },
    "paths": {
        "/time": {
            "get": {
                "operationId": "GetCurrentTime",
                "responses": {
                    "200": {
                        "description": "GetCurrentTime 200 response",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/GetCurrentTimeResponseContent"
                                }
                            }
                        }
                    },
                    "400": {
                        "description": "ValidationException 400 response",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/ValidationExceptionResponseContent"
                                }
                            }
                        }
                    },
                    "429": {
                        "description": "ThrottlingException 429 response",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/ThrottlingExceptionResponseContent"
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "components": {
        "schemas": {
            "GetCurrentTimeResponseContent": {
                "type": "object",
                "properties": {
                    "time": {
[...]

PR #1304, opened 2022-07-14
Adds support for [multiple error response objects with same HTTP status code] to Smithy2OpenAPI.

sichanyoo:
*Issue #, if available:*
#1265 

*Description of changes:*
- Add new OpenAPI config setting ``disableOneOf``.
- Combines response objects that share same HTTP status code before adding them to ``Map<String, ResponseObject> responses`` of ``OperationObject`` to prevent overwrites from happening.
- Test case + input / output files.
- Add documentation to existing OpenAPI converter guide regarding new config setting ``disableOneOf``.

_Internal:_
[Design doc](https://quip-amazon.com/BOTiANUdbNNu/Smithy2OpenAPI-Improvement-Project#temp:C:aMfab7e781fe5474397ab83a5d64)

### DIFF
--- a/docs/source/1.0/guides/converting-to-openapi.rst
+++ b/docs/source/1.0/guides/converting-to-openapi.rst
@@ -518,6 +518,56 @@ useIntegerType (``boolean``)
             }
         }
 
+.. _generate-openapi-setting-onErrorStatusConflict:
+
+onErrorStatusConflict (``String``)
+    Specifies how to resolve multiple error responses that share a same HTTP status code.
+    This behavior can be customized using the following values for the ``onErrorStatusConflict`` setting:
+
+    ``oneOf``
+        Use OpenAPI's ``oneOf`` keyword to combine error responses with same HTTP status code.
+
+
+    ``properties``
+        Use ``properties`` field of OpenAPI schema object to combine error responses with same HTTP status code.
+
+    .. note::
+            ``oneOf`` keyword is not supported by Amazon API Gateway.
+
+    Both options generate a single combined response object called "UnionError XXX Response" in the
+    OpenAPI model output, where "XXX" is the status code shared by multiple errors. Both options drop
+    the ``@required`` trait from all members of conflicting error structures, making them optional.
+
+    ``oneOf`` option wraps schemas for contents of conflicting errors responses schemas into a synthetic union schema
+    using OpenAPI's ``oneOf`` keyword.
+
+    ``properties`` option combines the conflicting error structure shapes into one union error shape that contains
+     all members from each and every conflicting error.
+
+    .. warning::
+            When using ``properties`` option, make sure that conflicting error structure shapes do not have member(s)
+            that have same name while having different target shapes. If member shapes with same name
+            (in conflicting error structures) target
+            different shapes, error shapes will not be able to be merged into one union error shape, and
+            an exception will be thrown.
+
+    .. warning::
+            Regardless of the setting, an exception will be thrown if any one of conflicting error structure shape
+            has a member shape with ``@httpPayload`` trait.
+
+    By default, this setting is set to ``oneOf``.
+
+    .. code-block:: json
+
+        {
+            "version": "1.0",
+            "plugins": {
+                "openapi": {
+                    "service": "smithy.example#Weather",
+                    "onErrorStatusConflict": "properties"
+                }
+            }
+        }
 
 ----------------------------------
 JSON schema configuration settings
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
@@ -48,6 +48,28 @@ public class OpenApiConfig extends JsonSchemaConfig {
         WARN
     }
 
+    /** Specifies how to resolve multiple error responses with same error codes. */
+    public enum ErrorStatusConflictHandlingStrategy {
+        /** The default setting that uses OpenAPI's oneOf keyword to combine multiple schemas for same media type. */
+        ONE_OF("oneOf"),
+        /**
+         * The custom setting that combines multiple schemas for same media type using properties field of
+         * OpenAPI Schema object.
+         */
+        PROPERTIES("properties");
+
+        private String stringValue;
+
+        ErrorStatusConflictHandlingStrategy(String stringValue) {
+            this.stringValue = stringValue;
+        }
+
+        @Override
+        public String toString() {
+            return stringValue;
+        }
+    }
+
     /** The JSON pointer to where OpenAPI schema components should be written. */
     private static final String SCHEMA_COMPONENTS_POINTER = "#/components/schemas";
 
@@ -80,6 +102,7 @@ public class OpenApiConfig extends JsonSchemaConfig {
     private boolean forbidGreedyLabels;
     private boolean removeGreedyParameterSuffix;
     private HttpPrefixHeadersStrategy onHttpPrefixHeaders = HttpPrefixHeadersStrategy.FAIL;
+    private ErrorStatusConflictHandlingStrategy onErrorStatusConflict = ErrorStatusConflictHandlingStrategy.ONE_OF;
     private boolean ignoreUnsupportedTraits;
     private Map<String, Node> substitutions = Collections.emptyMap();
     private Map<String, Node> jsonAdd = Collections.emptyMap();
@@ -237,6 +260,19 @@ public class OpenApiConfig extends JsonSchemaConfig {
      */
     public void setOnHttpPrefixHeaders(HttpPrefixHeadersStrategy onHttpPrefixHeaders) {
         this.onHttpPrefixHeaders = Objects.requireNonNull(onHttpPrefixHeaders);
+    }
+
+    public ErrorStatusConflictHandlingStrategy getOnErrorStatusConflict() {
+        return onErrorStatusConflict;
+    }
+
+    /**
+     * Specifies what to do when multiple error responses share the same HTTP status code.
+     *
+     * @param onErrorStatusConflict Strategy to use for multiple errors with same status code.
+     */
+    public void setOnErrorStatusConflict(ErrorStatusConflictHandlingStrategy onErrorStatusConflict) {
+        this.onErrorStatusConflict = Objects.requireNonNull(onErrorStatusConflict);
     }
 
     public boolean getIgnoreUnsupportedTraits() {
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
@@ -38,13 +38,20 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.ServiceIndex;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
 import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.AuthTrait;
 import software.amazon.smithy.model.traits.DeprecatedTrait;
 import software.amazon.smithy.model.traits.DocumentationTrait;
+import software.amazon.smithy.model.traits.ErrorTrait;
+import software.amazon.smithy.model.traits.HttpErrorTrait;
+import software.amazon.smithy.model.traits.HttpHeaderTrait;
+import software.amazon.smithy.model.traits.HttpPayloadTrait;
+import software.amazon.smithy.model.traits.RequiredTrait;
 import software.amazon.smithy.model.traits.TitleTrait;
 import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.model.transform.ModelTransformer;
@@ -168,6 +175,10 @@ public final class OpenApiConverter {
     }
 
     private ConversionEnvironment<? extends Trait> createConversionEnvironment(Model model) {
+        if (config.getOnErrorStatusConflict() == OpenApiConfig.ErrorStatusConflictHandlingStrategy.PROPERTIES) {
+            model = replaceConflictingErrorsWithUnionErrors(model);
+        }
+
         ShapeId serviceShapeId = config.getService();
 
         if (serviceShapeId == null) {
@@ -234,6 +245,120 @@ public final class OpenApiConverter {
                 openApiProtocol, document, securitySchemeConverters);
 
         return new ConversionEnvironment<>(context, extensions, components, composedMapper);
+    }
+
+    private Model replaceConflictingErrorsWithUnionErrors(Model model) {
+        // Get mapping of status code to error shapes.
+        Map<String, List<StructureShape>> statusCodeToErrors = new HashMap<>();
+        for (StructureShape error : model.getStructureShapesWithTrait(ErrorTrait.class)) {
+            statusCodeToErrors.computeIfAbsent(
+                    error.hasTrait(HttpErrorTrait.ID)
+                            ? Integer.toString(error.getTrait(HttpErrorTrait.class).get().getCode())
+                            : Integer.toString(error.getTrait(ErrorTrait.class).get().getDefaultHttpStatusCode()),
+                    (k) -> new ArrayList<>()
+            ).add(error);
+        }
+
+        // This is the list of union error shapes.
+        List<StructureShape> unions = new ArrayList<>();
+        // This is the mapping used for renaming conflicting errors in the model to corresponding union error.
+        Map<ShapeId, ShapeId> renameMapping = new HashMap<>();
+
+        Map<String, ShapeId> members = new HashMap<>();
+        Map<String, String> memberNameToErrorName = new HashMap<>();
+
+        // Create union errors for each group of conflicting errors (errors with same status code).
+        for (Map.Entry<String, List<StructureShape>> statusCodeToConflictingErrors : statusCodeToErrors.entrySet()) {
+            if (statusCodeToConflictingErrors.getValue().size() > 1) {
+                // Synthesize the union error structure for conflicting errors.
+                StructureShape.Builder union = StructureShape.builder();
+                String unionId = config.getService().getNamespace()
+                                    + "#Union"
+                                    + statusCodeToConflictingErrors.getKey()
+                                    + "Error";
+
+                for (StructureShape conflictingError : statusCodeToConflictingErrors.getValue()) {
+                    for (MemberShape eachMember : conflictingError.getAllMembers().values()) {
+                        // Check if members with same name target different shapes.
+                        if (members.containsKey(eachMember.getMemberName())
+                                && members.get(eachMember.getMemberName()) != eachMember.getTarget()) {
+                            String message =
+                                    "Error structures %s and %s with status code %s both have a member named \"%s\" "
+                                            + "that target different shapes, and the config setting for "
+                                            + "\"onErrorStatusConflict\" is set to \"properties\". Failed to combine "
+                                            + "error structures into one union error structure.";
+                            // Throw exception.
+                            String memberName = eachMember.getMemberName();
+                            String error1 = conflictingError.toShapeId().toString();
+                            String error2 = memberNameToErrorName.get(memberName);
+                            throw new OpenApiException(String.format(message, error1, error2, memberName));
+                        }
+                        members.put(eachMember.getMemberName(), eachMember.getTarget());
+                        memberNameToErrorName.put(eachMember.getMemberName(), conflictingError.getId().toString());
+
+                        // If the member has @httpPayload trait, throw an exception.
+                        if (eachMember.hasTrait(HttpPayloadTrait.ID)) {
+                            throw new OpenApiException(
+                                    createConflictingErrorHttpPayloadExceptionMessage(
+                                            statusCodeToConflictingErrors.getKey(),
+                                            conflictingError.getId().toString(),
+                                            eachMember.toShapeId().toString()
+                                    )
+                            );
+                        }
+
+                        // Drop @required trait and @httpPayload traits from members that are put into union error.
+                        eachMember = eachMember
+                                        .toBuilder()
+                                        .removeTrait(RequiredTrait.ID)
+                                        .build();
+                        // Change header id to [unionId + originalHeaderName] in order to match over-the-wire data.
+                        if (eachMember.hasTrait(HttpHeaderTrait.ID)) {
+                            String originalHeaderName = eachMember.toShapeId().getName();
+                            eachMember = eachMember.toBuilder().id(unionId + "$" + originalHeaderName).build();
+                        }
+                        union.addMember(eachMember);
+                    }
+                    // Add rename mapping from conflicting error to newly synthesized union error.
+                    renameMapping.put(conflictingError.toShapeId(), ShapeId.from(unionId));
+                }
+
+                // Add @httpError and/or @error trait(s) to union error shape.
+                if (statusCodeToConflictingErrors.getKey().equals("400")
+                        || statusCodeToConflictingErrors.getKey().equals("500")) {
+                    union.addTrait(new ErrorTrait(statusCodeToConflictingErrors.getKey()));
+                } else {
+                    if (statusCodeToConflictingErrors.getKey().charAt(0) == '4') {
+                        union.addTrait(new ErrorTrait("client"));
+                    } else {
+                        union.addTrait(new ErrorTrait("server"));
+                    }
+                    union.addTrait(new HttpErrorTrait(Integer.parseInt(statusCodeToConflictingErrors.getKey())));
+                }
+
+                // Set ShapeID on union error, build, and add to list of union errors.
+                unions.add(union.id(unionId).build());
+            }
+        }
+
+        // Add synthesized union errors to model, build, rename conflicting errors to union errors, and return.
+        return ModelTransformer.create().renameShapes(
+                (model.toBuilder().addShapes(unions).build()),
+                renameMapping
+        );
+    }
+
+    public static String createConflictingErrorHttpPayloadExceptionMessage(
+            String code,
+            String error,
+            String memberName
+    ) {
+        String message =
+                "The status code %s of %s is shared by two or more error shapes, and %s has a "
+                        + "member %s that has a @httpPayload trait. Failure in combining error shapes into "
+                        + "one union error. Remove @httpPayload trait to resolve the exception";
+
+        return String.format(message, code, error, error, memberName);
     }
 
     private static OpenApiMapper createComposedMapper(
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
@@ -17,6 +17,7 @@ package software.amazon.smithy.openapi.fromsmithy.protocols;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -42,12 +43,15 @@ import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
 import software.amazon.smithy.model.traits.HttpChecksumRequiredTrait;
+import software.amazon.smithy.model.traits.HttpPayloadTrait;
 import software.amazon.smithy.model.traits.HttpTrait;
 import software.amazon.smithy.model.traits.TimestampFormatTrait;
 import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.openapi.OpenApiException;
 import software.amazon.smithy.openapi.fromsmithy.Context;
+import software.amazon.smithy.openapi.fromsmithy.OpenApiConverter;
 import software.amazon.smithy.openapi.fromsmithy.OpenApiProtocol;
+import software.amazon.smithy.openapi.model.LinkObject;
 import software.amazon.smithy.openapi.model.MediaTypeObject;
 import software.amazon.smithy.openapi.model.OperationObject;
 import software.amazon.smithy.openapi.model.ParameterObject;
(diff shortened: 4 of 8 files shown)


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -436,9 +437,97 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+        // handle error code collision
+        if (responses.containsKey(statusCode)) {
+            newResponse =
+                    handleErrorCodeCollision(context.getConfig().isDisableOneOf(), statusCode, responses, newResponse);
+        }
+
+        responses.put(statusCode, newResponse);
+    }
+
+    private ResponseObject handleErrorCodeCollision(boolean disableOneOf,
```
sugmanue, line 452: I think we can improve this if we handle this case in a different code path instead of patching up whatever we had before, that will make the code a lot easier to understand and maintain in the long run. In particular since we need to handle correctly the following cases:

1. Properties with `@HttpHeader` and `@required`: given that we will create a response having both responses those cannot be "required" anymore, thus we will need to drop it from the output (I tested this with your change and the required is emitted)
2. Responses with different Media Types, we cannot handle this case in particular and we need to error out instead of generating an incorrect response (I tested this with your change and the output is empty). I don't have a good story on how to do that, in particular an exception this deep in the code will lead to a bad UX, feel free to ask around about a good way to signal this issue to the client.

I think that we can simplify the code if we process the errors twice, one to aggregate them by status code and then we branch for the single response vs multi response cases. I think we can do that in the [`createResponses` method](https://github.com/awslabs/smithy/blob/main/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java#L409), such that:

```
    private Map<String, ResponseObject> createResponses(
            Context<T> context,
            HttpBindingIndex bindingIndex,
            EventStreamIndex eventStreamIndex,
            OperationShape operation
    ) {
        Map<String, ResponseObject> result = new TreeMap<>();
        OperationIndex operationIndex = OperationIndex.of(context.getModel());

        StructureShape output = operationIndex.expectOutputShape(operation);
        updateResponsesMapWithResponseStatusAndObject(
                context, bindingIndex, eventStreamIndex, operation, output, result);

        Map<String, List<StructureShape>> errors = new LinkedHashMap<>();
        // Aggregate errors by status code to handle correctly those responses
        // that map to the same status
        for (StructureShape error : operationIndex.getErrors(operation)) {
            String statusCode = context.getOpenApiProtocol().getOperationResponseStatusCode(context, error);
            errors.computeIfAbsent(statusCode, (k) -> new ArrayList<>())
                    .add(error);
        }
        for (Map.Entry<String, List<StructureShape>> entry : errors.entrySet()) {
            List<StructureShape> shapes = entry.getValue();
            if (shapes.size() == 1) {
                updateResponsesMapWithResponseStatusAndObject(
                        context, bindingIndex, eventStreamIndex, operation, shapes.get(0), result);
            } else {
                updateResponsesMapWithResponseStatusAndMultipleObjects(
                        context, bindingIndex, eventStreamIndex, operation, shapes, result);
            }
        }
        return result;
    }

    // Handle the case with different responses for the same status code
    private void updateResponsesMapWithResponseStatusAndMultipleObjects(
            Context<T> context,
            HttpBindingIndex bindingIndex,
            EventStreamIndex eventStreamIndex,
            OperationShape operation,
            List<StructureShape> shapes,
            Map<String, ResponseObject> responses
    ) {
        // TODO
    }
```

sugmanue (commented):
Please also add test cases for:

`httpHeader` with `required`, e.g.,
```
@error("client")
@retryable(throttling: true)
@httpError(429)
structure ThrottlingException {
    @httpHeader("throttlingHeader")
    @required
    header: String,

    @required
    message: String,
}
```
Errors with `httpPayload` annotations, e.g.,
```
@error("client")
@retryable(throttling: true)
@httpError(429)
structure ThrottlingException {
    @httpHeader("throttlingHeader")
    header: String,

    @required
    @httpPayload
    message: String,
}
```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sugmanue: > 2\. Responses with different Media Types, we cannot handle this case in particular and we need to error out instead of generating an incorrect response (I tested this with your change and the output is empty). I don't have a good story on how to do that, in particular an exception this deep in the code will lead to a bad UX, feel free to ask around about a good way to signal this issue to the client.

I was confused, we can actually handle those using different media types keys, e.g.,

```
 "429": {
        ⋯
        "content": {
              "application/json": { ⋯ }
              "text/plain": { ⋯ }
```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -87,6 +87,8 @@ public enum HttpPrefixHeadersStrategy {
             "Homepage", "API Reference", "User Guide", "Developer Guide", "Reference", "Guide");
     private boolean useIntegerType;
 
+    private boolean disableOneOf;
```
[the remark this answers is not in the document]
sichanyoo: changed to using enum.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -330,6 +332,14 @@ public void setUseIntegerType(boolean useIntegerType) {
         this.useIntegerType = useIntegerType;
     }
 
+    public boolean isDisableOneOf() {
+        return disableOneOf;
+    }
+
+    public void setDisableOneOf(boolean disableOneOf) {
```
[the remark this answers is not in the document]
sichanyoo: added javadoc for setter for enum setting. this particular setter is now gone.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -330,6 +332,14 @@ public void setUseIntegerType(boolean useIntegerType) {
         this.useIntegerType = useIntegerType;
     }
 
+    public boolean isDisableOneOf() {
```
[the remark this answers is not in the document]
sichanyoo: this getter is deleted, due to changing config setting from boolean to enum.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+        ResponseObject.Builder combinedResponse = ResponseObject.builder();
+        String statusCode = context.getOpenApiProtocol().getOperationResponseStatusCode(context, shapes.get(0));
+        OpenApiConfig.ErrorStatusConflictHandlingStrategy strategy = context.getConfig().getOnErrorStatusConflict();
+        Map<String, List<MediaTypeObject>> contents = new LinkedHashMap<>();
+
+        // make response object for each error response & add to list
+        for (StructureShape shape : shapes) {
+            errors.add(createResponse(context, bindingIndex, eventStreamIndex, statusCode, shape,
+                    true));
+        }
```
sugmanue, line 478: You can use streams for this to make this code more compact and idiomatic, e.g.,

```
        List<ResponseObject> errors = shapes.stream()
                .map(shape -> createResponse(context, bindingIndex, eventStreamIndex, statusCode, shape, true))

```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+            // get mapping of contents
+            for (Map.Entry<String, MediaTypeObject> content : error.getContent().entrySet()) {
+                contents.computeIfAbsent(content.getKey(), (k) -> new ArrayList<>()).add(content.getValue());
+            }
+
+            // combine links
+            for (Map.Entry<String, Ref<LinkObject>> link : error.getLinks().entrySet()) {
+                combinedResponse.putLink(link.getKey(), link.getValue());
+            }
+        }
```
sugmanue, line 496: You can also use streams for this, e.g.,

```
        Map<String, List<MediaTypeObject>> contents =
                errors.stream()
                        .peek(error -> error.getHeaders().forEach(combinedResponse::putHeader))
                        .peek(error -> error.getLinks().forEach(combinedResponse::putLink))
                        .map(ResponseObject::getContent)
                        .map(Map::entrySet)
                        .flatMap(Collection::stream)
                        .collect(Collectors.groupingBy(Map.Entry::getKey,
                                LinkedHashMap::new,
                                Collectors.mapping(Map.Entry::getValue, Collectors.toList())));

```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+                                                         List<MediaTypeObject> contents) {
+        MediaTypeObject.Builder result = MediaTypeObject.builder();
+        Schema.Builder schema = Schema.builder();
+
+        if (strategy == OpenApiConfig.ErrorStatusConflictHandlingStrategy.USE_ONEOF) {
+            List<Schema> schemas = new ArrayList<>();
+            for (MediaTypeObject content : contents) {
+                schemas.add(content.getSchema().get());
+            }
+            schema.oneOf(schemas);
```
sugmanue, line 523: Same here, use streams, e.g.,

```
                schema.oneOf(contents.stream()
                        .map(MediaTypeObject::getSchema)
                        .map(Optional::get)
                        .collect(Collectors.toList()));
```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+        } else if (strategy == OpenApiConfig.ErrorStatusConflictHandlingStrategy.USE_PROPERTIES) {
+            int i = 1;
+            for (MediaTypeObject content : contents) {
+                schema.putProperty("Schema_" + i++, content.getSchema().get());
+            }
+        } else {
+            throw new OpenApiException("\"onErrorStatusConflict\" OpenAPI config setting has an invalid value.");
```
sugmanue, line 530: The setting is valid by definition, i.e., is part of the enum, probably a better wording would be "unknown" (at least to this code), also probably better to use `IllegalArgumentException`, but definitely include the value that you cannot handle in the exception message,  e.g.,

```
                throw new IllegalArgumentException("Unknown strategy: " + strategy);
```

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/protocols/AwsRestJson1ProtocolTest.java
```diff
@@ -266,4 +266,53 @@ public OperationObject updateOperation(
...
+        config.setOnErrorStatusConflict(OpenApiConfig.ErrorStatusConflictHandlingStrategy.USE_PROPERTIES);
+        ObjectNode result = OpenApiConverter.create()
+                .config(config)
+                .convertToNode(model);
+        InputStream openApiStream = getClass()
+                .getResourceAsStream("error-code-collision-test-use-properties.openapi.json");
+
+        if (openApiStream == null) {
+            throw new RuntimeException("OpenAPI model not found for test case: "
+                    + "error-code-collision-test-use-properties.openapi.json");
```
sugmanue, line 288: I think it would be more idiomatic to assert that the `InputStream` is not null instead, such as:


```
    InputStream openApiStream = getClass()
        .getResourceAsStream("error-code-collision-test-use-oneof.openapi.json");

    Assertions.assertNotNull(openApiStream);
    Node expectedNode = Node.parse(IoUtils.toUtf8String(openApiStream));
    Node.assertEquals(result, expectedNode);
```

Or at least to keep the code consistent use `fail` as is done [above in this file](https://github.com/awslabs/smithy/blob/98990e88585b2f55e75012ef41dbc04d9e934dd3/smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/protocols/AwsRestJson1ProtocolTest.java#L60).

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -80,6 +90,8 @@ public enum HttpPrefixHeadersStrategy {
     private boolean forbidGreedyLabels;
     private boolean removeGreedyParameterSuffix;
     private HttpPrefixHeadersStrategy onHttpPrefixHeaders = HttpPrefixHeadersStrategy.FAIL;
+
```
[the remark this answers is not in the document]
sichanyoo: mentioned in previous comment, resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
mtdowling: the for loop is more readable IMO

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sichanyoo: for loop kept, following discussion.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sichanyoo: for-loop kept, following discussion.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sichanyoo: for-loop kept, following discussion.

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/protocols/AwsRestJson1ProtocolTest.java
sichanyoo: runtime exception kept, following discussion.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
     }
 
+    private void updateResponsesMapWithResponseStatusAndMultipleObjects(
+            Context<T> context,
+            HttpBindingIndex bindingIndex,
+            EventStreamIndex eventStreamIndex,
+            List<StructureShape> shapes,
+            Map<String, ResponseObject> responses
+    ) {
+        List<ResponseObject> errors = new ArrayList<>();
```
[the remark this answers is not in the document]
sichanyoo: fixed as suggested.

docs/source/1.0/guides/converting-to-openapi.rst
```diff
@@ -518,6 +518,34 @@ useIntegerType (``boolean``)
...
         }
 
+.. _generate-openapi-setting-onErrorStatusConflict:
+
+onErrorStatusConflict (``String``)
+    Specifies how to resolve multiple error responses that share the same HTTP status code.
+    This behavior can be customized using the following values for the ``onErrorStatusConflict`` setting:
+
+    * USE_ONEOF
+    Use OpenAPI's ``oneOf`` keyword to combine schemas of errors with same HTTP status code.
```
mtdowling, line 528: Are these rendering correctly as bulleted lists? I always thought the contents had to be indented like this:
```
* USE_ONEOF
   Use OpenAPI's ....
* USE_PROPERTIES
   blah blah
```

Using a definition list might work better here though:

```
USE_ONEOF
    Use OpenAPI's ``oneOf`` keyword to combine schemas of errors with same HTTP status code.
    ...
USE_PROPERTIES
    .....
```

Also, for naming, in the `unionStrategy` option, we used "oneOf", "object", and "structure". Can we use "oneOf" and "properties" here instead of "USE_ONEOF" and "USE_PROPERTIES"?

docs/source/1.0/guides/converting-to-openapi.rst
```diff
@@ -518,6 +518,34 @@ useIntegerType (``boolean``)
...
+
+    .. note::
+            ``oneOf`` keyword is not supported by Amazon API Gateway.
+
+    * USE_PROPERTIES
+    Use ``properties`` field of OpenAPI schema object to combine schemas of errors with same HTTP status code.
```
mtdowling, line 534: Can you expand here? All of the errors that share a status code are merged together, right? What happens if properties conflict? Or if they're required in one error but not another?

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -48,6 +48,16 @@ public enum HttpPrefixHeadersStrategy {
         WARN
     }
 
+    /** Specifies how to resolve multiple error responses with same error codes. */
+    public enum ErrorStatusConflictHandlingStrategy {
+        /** the default setting that uses OpenAPI's oneOf keyword to combine multiple schemas for same media type. */
+        USE_ONEOF,
```
mtdowling, line 54: We used "ONE_OF" in other similar enums. Let's change to match. https://github.com/awslabs/smithy/blob/main/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConfig.java#L44

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -48,6 +48,16 @@ public enum HttpPrefixHeadersStrategy {
...
+        /** The custom setting that combines multiple schemas for same media type using properties field of
+         *  OpenAPI Schema object.
+         */
+        USE_PROPERTIES
```
mtdowling, line 58: Let's just use "PROPERTIES" here.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -48,6 +48,16 @@ public enum HttpPrefixHeadersStrategy {
         WARN
     }
 
+    /** Specifies how to resolve multiple error responses with same error codes. */
+    public enum ErrorStatusConflictHandlingStrategy {
+        /** the default setting that uses OpenAPI's oneOf keyword to combine multiple schemas for same media type. */
```
mtdowling, line 53:
```suggestion
        /** The default setting that uses OpenAPI's oneOf keyword to combine multiple schemas for same media type. */
```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -48,6 +48,16 @@ public enum HttpPrefixHeadersStrategy {
...
+        USE_ONEOF,
+        /** The custom setting that combines multiple schemas for same media type using properties field of
```
mtdowling, line 55: nit - expand this javadoc to use the normal form since it doesn't fit on a single line:
```
/**
 * The custom ....
 */
```

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,104 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+        if (strategy == OpenApiConfig.ErrorStatusConflictHandlingStrategy.USE_ONEOF) {
+            List<Schema> schemas = new ArrayList<>();
+            for (MediaTypeObject content : contents) {
+                schemas.add(content.getSchema().get());
+            }
+            schema.oneOf(schemas);
+        } else if (strategy == OpenApiConfig.ErrorStatusConflictHandlingStrategy.USE_PROPERTIES) {
+            int i = 1;
+            for (MediaTypeObject content : contents) {
+                schema.putProperty("Schema_" + i++, content.getSchema().get());
```
mtdowling, line 526: I think there's a misunderstanding here. We need to create a synthetic response that combines all of the schemas into a single top-level schema. We can't create a synthetic response that adds one layer of wrapping properties like this because it doesn't actually match what's being sent on the wire. Let me know if I'm missing something.

mtdowling (changes requested):

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
sichanyoo: resolved.

docs/source/1.0/guides/converting-to-openapi.rst
sichanyoo: - Yeah, the bulleted list does render correctly when I tested back then as well as now, using `make serve` command in docs directory. However, definition list was used to replace it anyways, as it does look better. 
- Naming in doc modified to resemble `unionStrategy` section as per the comment & `ErrorStatusConflictHandlingStrategy` enum updated in code to reflect this change.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
sichanyoo: resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
sichanyoo: resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sichanyoo: oh, I think I understand now. For option where we don't use `oneOf` keyword, what you were thinking [in this comment to the issue](https://github.com/awslabs/smithy/issues/1265#issuecomment-1153064368) was combining the conflicting response objects into one, resulting in only one schema for that response existing in the first place. Is that correct?

mtdowling: We discussed in person -- we need to merge the schemas. We can't introduce synthetic members or wrapping since that would make it impossible for clients to properly parse the response.

sichanyoo: Now top-level schema is created, and when combining error shapes with same status code, `@required` trait and `@httpPayload` traits are removed from each `MemberShape`.

docs/source/1.0/guides/converting-to-openapi.rst
sichanyoo: Resolved. When they conflict, exception is thrown. Also, the `@required` trait will be dropped from all member shapes within conflicting error structure shapes.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
sichanyoo: Resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,105 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+            }
+        } else {
```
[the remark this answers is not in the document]
sichanyoo: out of date commit.

docs/source/1.0/guides/converting-to-openapi.rst
```diff
@@ -518,6 +518,57 @@ useIntegerType (``boolean``)
...
+    Both options generate a single combined response object called "UnionError XXX Response" in the
+    OpenAPI model output, where "XXX" is the status code shared by multiple errors. Both options drop
+    the ``@required`` trait from all members of conflicting error structures, making them optional.
+
+    ``oneOf`` option wraps schemas for contents of conflicting errors responses schemas into a synthetic union schema
+    using OpenAPI's ``oneOf`` keyword.
+
+    ``properties`` option pre-processes the Smithy model before OpenAPI conversion takes place, during which
+    conflicting error structure shapes are combined into one union error shape with all members from each and every
+    conflicting error.
```
sugmanue, line 546: About this one, I think that the implementation details can be left off and rather just explain what it does. Instead something like:

" all the properties from the conflicting structures are combined in a single structure ..."

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,119 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+                                        .getOperationResponseStatusCode(context, error)
+                                    + " of "
+                                    + error.getId()
+                                    + " is shared by two or more error shapes, and "
+                                    + error.getId()
+                                    + " has a member "
+                                    + member.getValue().toShapeId().toString()
+                                    + " that has a @httpPayload trait."
+                                    + " Failure in combining error shapes into one union error."
+                                    + " Remove @httpPayload trait to reoslve the exception.";
```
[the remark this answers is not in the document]
sichanyoo: The message content are different, so I just used `String.format` to condense this down. Turns out the reason I had for not using `String.format` was wrong.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
```diff
@@ -236,6 +247,118 @@ private ConversionEnvironment<? extends Trait> createConversionEnvironment(Model
...
+                                            + statusCodeToConflictingErrors.getKey()
+                                            + " of "
+                                            + conflictingError.getId()
+                                            + " is shared by two or more error shapes, and "
+                                            + conflictingError.getId()
+                                            + " has a member "
+                                            + eachMember.toShapeId().toString()
+                                            + " that has a @httpPayload trait."
+                                            + " Failure in combining error shapes into one union error."
+                                            + " Remove @httpPayload trait to reoslve the exception.";
```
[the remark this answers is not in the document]
sichanyoo: Resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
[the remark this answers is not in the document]
sichanyoo: Oh, never-mind, I see where the repetition occurs. Will go ahead and make a method and call that instead.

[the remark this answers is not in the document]
sichanyoo: Resolved.

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/OpenApiConverter.java
```diff
@@ -236,6 +247,118 @@ private ConversionEnvironment<? extends Trait> createConversionEnvironment(Model
...
+                        if (eachMember.hasTrait(HttpPayloadTrait.ID)) {
+                            String message =
+                                    "The status code "
+                                            + statusCodeToConflictingErrors.getKey()
+                                            + " of "
+                                            + conflictingError.getId()
+                                            + " is shared by two or more error shapes, and "
+                                            + conflictingError.getId()
+                                            + " has a member "
+                                            + eachMember.toShapeId().toString()
```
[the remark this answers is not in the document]
sichanyoo: Resolved from one of other comments above.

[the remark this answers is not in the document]
sichanyoo: (Method was created, and the code this comment is for has been deleted)

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -437,23 +453,119 @@ private void updateResponsesMapWithResponseStatusAndObject(
...
+            Shape operationOrError,
+            boolean dropRequiredTrait
     ) {
         ResponseObject.Builder responseBuilder = ResponseObject.builder();
         String contextName = context.getService().getContextualName(operationOrError);
         String responseName = stripNonAlphaNumericCharsIfNecessary(context, contextName);
+        if (responseName.startsWith("Union")) {
+            // Delete redundant status code from name.
+            responseName = "UnionError";
+        }
```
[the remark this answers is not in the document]
sichanyoo: The only reason for the if-block being here was to make the output look a bit more cleaner, but I did not think of the possibility that a response just begins with `Union` from user's input. If-block is deleted.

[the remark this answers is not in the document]
sichanyoo: I don't think there'll need a further change, given the output without the if-block is ok anyways. The only change is for name of response to go from "UnionError 429 Response" with if-block, to "Union429Error 429 Response" without the if-block.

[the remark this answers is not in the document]
sichanyoo: Resolved.

docs/source/1.0/guides/converting-to-openapi.rst
sichanyoo: Modified as suggested by comment.

sugmanue (approved):

PR #1312, opened 2022-07-21
Handle unknown model versions for built-shapes

mtdowling:
The handling of unknown model versions for built shapes needs
to make no assumptions about the version of provided shapes.
Instead, it should use the UNKNOWN version which allows for any
kind of 1.0 or 2.0 trait/shape, but still attempts to convert
1.0 -> 2.0 compatible models when loading shapes. This was a
little haphazard before and we exposed controls to set the
assumed parsed version that have now been removed. I don't think
there's a reasonable use case where automated tooling will set
that right now.

*Issue #, if available:*

*Description of changes:*


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -651,8 +635,7 @@ private List<ModelFile> createModelFiles() {
             // Since we're pulling from a loaded model, we know that it has been converted to the latest
```
kstich, line 635: This comment is no longer accurate.

kstich (approved):

PR #1313, opened 2022-07-22
Various IDL 2.0 fixes and tweaks

mtdowling:


smithy-model/src/main/java/software/amazon/smithy/model/shapes/MapShape.java
```diff
@@ -83,32 +90,60 @@ public MemberShape getKey() {
     }
 
     @Override
-    public Collection<MemberShape> members() {
```
kstich, line 86: Both `members()` and `getMember(name)` are in the current public interface.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/CollectionShape.java
```diff
@@ -51,18 +59,13 @@ public final MemberShape getMember() {
     }
 
     @Override
-    public final Optional<MemberShape> getMember(String name) {
```
kstich, line 54: Both `members()` and `getMember(name)` are in the current public interface.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/NamedMembersShapeBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public B clearMembers() {
+        members.clear();
+        return (B) this;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public B addMember(MemberShape member) {
```
kstich, line 70: Missing docs.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/NamedMembersShapeBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+    public B removeMember(String member) {
+        if (members.hasValue()) {
+            members.get().remove(member);
+        }
+        return (B) this;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public B addMixin(Shape shape) {
```
kstich, line 128: Mixin related methods are missing docs as well.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
```diff
@@ -17,21 +17,55 @@
...
+            ShapeId.from("smithy.api#PrimitiveDouble"));
+
+    private static final Set<ShapeType> V1_INHERENTLY_BOXED = SetUtils.of(
+            ShapeType.STRING,
+            ShapeType.BLOB,
+            ShapeType.TIMESTAMP,
+            ShapeType.BIG_DECIMAL,
+            ShapeType.BIG_INTEGER,
+            ShapeType.LIST,
+            ShapeType.SET,
```
kstich, line 63: Was surprised to see this without a suppression for `ShapeType.SET` being deprecated, is that missing?

kstich (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/shapes/CollectionShape.java
mtdowling: These are both on and implemented in Shape now

smithy-model/src/main/java/software/amazon/smithy/model/shapes/NamedMembersShapeBuilder.java
mtdowling: This was missing docs before too so no net change

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
mtdowling: I decided not to deprecate this enum members. We can deprecate the SetShape and pretend it doesn't exist because it extends from ListShape. You can't do that with enum members in Java, so you still need to write type handling code that accounts for SET. I didn't see any value in deprecating it.

kstich (approved):

PR #1327, opened 2022-08-04
Fix grammar for IDL v2

### LINKED ISSUE #1324: Issues with the ABNF grammar for Smithy IDL
I was working through the ABNF grammar for the Smithy IDL and
encountered several minor issues.

### Some rules MUST require at least one whitespace

A structure statement is defined as:

```
structure_statement = "structure" ws identifier ws shape_members
```

And the `ws` rule is defined as:

```
ws = *(sp / newline / comment) ; whitespace
```

which matches zero or more whitespace.  This means something like this is a valid
`structure_statement`:

```
structureMyStructure{}
```

### Some rules should explicitly allow newlines

The `node_string_value` rule does not allow for new lines in the `quoted_char`
rule.  For example, the Smithy parser accepts this as valid:

```
@documentation("""

    """)
string NewlineString
```

But the grammar for `text_block` is:

```
text_block =
    three_dquotes br *quoted_char three_dquotes

three_dquotes =
    DQUOTE DQUOTE DQUOTE

quoted_char =
    %x20-21        ; space - "!"
  / %x23-5B        ; "#" - "["
  / %x5D-10FFFF    ; "]"+
  / escaped_char
  / preserved_double

```

These rules do not permit any newline characters (`%x0A / %x0D.0A`)
after the initial `three_dquotes br` new line.

### `trait_body_value` can be optional

A `trait_body` must contain a `trait_body_value`:

```
trait =
    "@" shape_id [trait_body]

trait_body =
    "(" ws trait_body_value ws ")"

trait_body_value =
    trait_structure / node_value
```

However, the Smithy parser excepts an empty `trait_body_value`,

```
structure GetCityInput {
    @required()
    cityId: CityId,
```

<br />

*(the remaining stuff is nit-picky and maybe just worth calling out where
you deviate from RFC 5234 instead of fixing)*

#### Rule names use underscores

From [Section 2.1 or RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234#section-2.1):

> The name of a rule is simply the name itself, that is, a sequence of
   characters, beginning with an alphabetic character, and followed by a
   combination of alphabetics, digits, and hyphens (dashes).

Note that underscores are not allowed.  This is also shown in the ABNF of ABNF
in [Section 4](https://datatracker.ietf.org/doc/html/rfc5234#page-10):

```
         rulelist       =  1*( rule / (*c-wsp c-nl) )

         rule           =  rulename defined-as elements c-nl
                                ; continues if next line starts
                                ;  with white space

         rulename       =  ALPHA *(ALPHA / DIGIT / "-")
```

I encountered this because I have tools that take as input a valid ABNF grammar
and were either failing or producing unexpected results.

#### Literal strings aren't case insensitive

From [Section 2.3](https://datatracker.ietf.org/doc/html/rfc5234#section-2.3):

> Literal text strings are interpreted as a concatenated set of
>    printable characters.
> 
>    NOTE:
> 
>       ABNF strings are case insensitive and the character set for these
>       strings is US-ASCII.
> 
>    Hence:
> 
>          rulename = "abc"
> 
>    and:
> 
>          rulename = "aBc"
> 
>    will match "abc", "Abc", "aBc", "abC", "ABc", "aBC", "AbC", and
>    "ABC".

Literal strings are used in the various statements for the `shape_body` rule,
e.g. `structure`, `resource`, etc.

mtdowling:
Closes #1324 


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/EnumShapeValidator.java
```diff
@@ -82,25 +84,48 @@
...
+            }
+
+            long longValue = number.getValue().longValue();
+            if (longValue > Integer.MAX_VALUE || longValue < Integer.MIN_VALUE) {
+                events.add(error(member, trait, "intEnum must fit within an integer, but found: " + longValue));
+                continue;
+            }
+
+            if (!values.add(number.getValue().intValue())) {
+                events.add(error(member, String.format("Multiple enum members found with duplicate value `%s`",
```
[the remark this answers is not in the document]
mtdowling: Fixed

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/EnumShapeValidator.java
```diff
@@ -23,6 +23,8 @@
 import java.util.regex.Pattern;
```
mtdowling, line 23: Added validation here that was previoulsy only performed in the IDL and not in the AST loader.

smithy-model/src/main/java/software/amazon/smithy/model/loader/LoadOperationProcessor.java
```diff
@@ -169,7 +169,7 @@ private void detectAndEmitForwardReference(LoadOperation.ForwardReference refere
         } else {
             // Try to find a prelude shape by ID if no ID exists in the namespace with this name.
             ShapeId preludeId = ShapeId.fromOptionalNamespace(Prelude.NAMESPACE, reference.name);
-            if (prelude.getShapeIds().contains(preludeId)) {
+            if (prelude != null && prelude.getShapeIds().contains(preludeId)) {
```
syall, line 172: Is there a case where prelude will not be available?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/dupe-list-member-names.errors
```diff
@@ -1 +1 @@
-[ERROR] com.foo#List: Parse error at line 7, column 11 near `: `: Duplicate member of com.foo#List: 'member' | Model
+[ERROR] -: Parse error at line 7, column 5 near `member`: Expected: '}', but found 'm' | Model
```
syall, line 1: Will this be added back later? This seems like a pretty important error to display.

There seems to be a lot of tests files with similar changes.

syall (changes requested):
Looks good, mostly minor changes.

Some other notes (couldn't comment on lines not included in the diff...)

- `docs/source-2.0/spec/service-types.rst`:
  - Line 20: `service_statement` should probably be `ServiceStatement`
  - Line 255: `operation_statement` should probably be `OperationStatement`
  - Line 351: `resource_statement` should probably be `ResourceStatement`

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/dupe-list-member-names.errors
mtdowling: This is the same test case just caught in the parser rather than after parsing and through logic. Previoulsy we parsed a bunch of members into a map and then validation. Now we can just parse known members one at a time.

docs/source-2.0/spec/idl.rst
```diff
@@ -88,136 +88,157 @@ Lexical notes
...
+    ServiceStatement        :%s"service" `SP` `Identifier` [`Mixins`] *`WS` `NodeObject`
+    ResourceStatement       :%s"resource" `SP` `Identifier` [`Mixins`] *`WS` `NodeObject`
+    OperationStatement      :%s"operation" `SP` `Identifier` [`Mixins`] *`WS` `OperationBody`
+    OperationBody           :"{" *`WS`
+                            :    [`OperationInput`]
+                            :    [`OperationOutput`]
+                            :    [`OperationErrors`]
+                            : *`WS` "}"
+    OperationInput          :%s"input" *WS (`InlineStructure` / `Identifier`) `BR`
+    OperationOutput         :%s"output" *WS (`InlineStructure` / `ShapeId`) `BR`
```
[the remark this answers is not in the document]
mtdowling: Yes good catch!

smithy-model/src/main/java/software/amazon/smithy/model/loader/LoadOperationProcessor.java
mtdowling: Good catch. Added.

mtdowling:
>Line 20: service_statement should probably be ServiceStatement
>Line 255: operation_statement should probably be OperationStatement
>Line 351: resource_statement should probably be ResourceStatement

Good question -- those links refer to anchors that still exist and not grammar rules, so we're ok there.

syall (approved):

PR #1330, opened 2022-08-08
Add transformation to exclude shapes based on ShapeType

michael-mml:
Add a new transformation to support excluding shapes based on their ShapeType.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesByType.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesByType.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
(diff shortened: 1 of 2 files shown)


mtdowling:
What's the use case for this?

michael-mml:
Our data model is populated from a variety of sources and clients are only interested in the data model itself, they do not care about the services, errors, or operations defined in our Smithy model. This transformation makes it easier to filter these out in a projection rather than tagging the unwanted shapes and using `ExcludeShapesByTag` transformation.

mtdowling:
I think we should add a transforms called `ExcludeShapesBySelector` and `IncludeShapesBySelector`, which would allow us to use [Smithy selectors](https://awslabs.github.io/smithy/2.0/spec/selectors.html) to filter out or include shapes based on type, traits, members, and more complex logic. For example, you could exclude by type using `:is(service, operation)`. You could exclude shapes that aren't part of a specific service closure using `:not([id=some.id#Service] ~> *)`. We could potentially make existing related transforms just wrappers around the selector based transform (e.g., ExcludeShapeByTrait).

sugmanue:
A more general approach was implemented in this [PR](https://github.com/awslabs/smithy/pull/1534). You can look at the docs for how to do either [inclusion](https://smithy.io/2.0/guides/building-models/build-config.html#includeshapesbyselector) or [exclusion](https://smithy.io/2.0/guides/building-models/build-config.html#excludeshapesbyselector) of shapes using a selector.

PR #1333, opened 2022-08-10
Validator for missing sensitive trait

rchache:
*Issue #, if available:*

*Description of changes:*
Based on various existing services for sensitive data detection and current AWS sensitive trait usage, generated a default list of words and phrases that likely indicate the data stored inside is sensitive. It is configurable the way the non inclusive terms validator is.  

### DIFF
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.linters;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.node.NodeMapper;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationUtils;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SetUtils;
+
+/**
+ * <p>Validates that shapes and members that possibly contain sensitive data are marked with the sensitive trait.
+ */
+public final class MissingSensitiveTraitValidator extends AbstractValidator {
+    public static final String SENSITIVE = "sensitive";
+    static final Set<String> DEFAULT_SENSITIVE_WORDS = SetUtils.of(
+            "authorization",
+            "bank",
+            "billing",
+            "birth",
+            "credential",
+            "email",
+            "ethnicity",
+            "insurance",
+            "license",
+            "passphrase",
+            "passport",
+            "password",
+            "phone",
+            "private",
+            "secret",
+            "sensitive",
+            "telephone",
+            "token",
+            "username",
+            "zip"
+    );
+    // "Phrases" will be searched for without regard to word boundaries. (This is so that user configuration )
+    static final Set<String> DEFAULT_SENSITIVE_PHRASES = SetUtils.of(
+            "accesskey",
+            "creditcard",
+            "firstname",
+            "lastname",
+            "plaintext",
+            "taxpayer"
+    );
+
+    public static final class Provider extends ValidatorService.Provider {
+        public Provider() {
+            super(MissingSensitiveTraitValidator.class, node -> {
+                NodeMapper mapper = new NodeMapper();
+                return new MissingSensitiveTraitValidator(
+                        mapper.deserialize(node, MissingSensitiveTraitValidator.Config.class));
+            });
+        }
+    }
+
+    /**
+     * MissingSensitiveTrait configuration.
+     */
+    public static final class Config {
+        private List<String> phrases = ListUtils.of();
+        private boolean excludeDefaults;
+
+        public List<String> getPhrases() {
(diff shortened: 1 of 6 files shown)


smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+                .flatMap(shape -> shape.members().stream())
+                // filter out members that target a sensitive shape
+                .filter(memberShape ->
+                        model.getShape(memberShape.getTarget())
+                        .map(shape -> !shape.hasTrait(SENSITIVE))
+                        .orElse(false))
+                .filter(memberShape -> containsSensitiveWord(memberShape.getMemberName())
+                        || containsSensitivePhrase(memberShape.getMemberName()))
+                .map(this::emit)
+                .collect(Collectors.toList());
```
kstich, line 172: nit: It would be nice if these `scan` functions used a consistent methodology - either loops and a list to add events to OR streaming with collectors.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+import software.amazon.smithy.model.validation.ValidationUtils;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SetUtils;
+
+/**
+ * <p>Validates that shapes and members that possibly contain sensitive data are marked with the sensitive trait.
+ */
+public final class MissingSensitiveTraitValidator extends AbstractValidator {
+    public static final String SENSITIVE = "sensitive";
```
kstich, line 40: Should use the `SensitiveTrait` class or its `ID` member instead.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private MissingSensitiveTraitValidator(Config config) {
+        if (!config.getExcludeDefaults()) {
+            Set<String> phrasesInit = new HashSet<>(DEFAULT_SENSITIVE_PHRASES);
+            phrasesInit.addAll(config.getPhrases()
+                    .stream()
+                    .map(phrase -> phrase.toLowerCase(Locale.US))
+                    .collect(Collectors.toSet()));
+            sensitivePhrases = Collections.unmodifiableSet(phrasesInit);
+            sensitiveWords = Collections.unmodifiableSet(DEFAULT_SENSITIVE_WORDS);
```
kstich, line 118: Since the docs aren't finished above, I'm not sure why these need to be separate. Should this be configurable?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+            "authorization",
+            "bank",
+            "billing",
+            "birth",
+            "credential",
+            "email",
+            "ethnicity",
+            "insurance",
+            "license",
+            "passphrase",
```
kstich, line 51: `passkey` as well.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+            "secret",
+            "sensitive",
+            "telephone",
+            "token",
+            "username",
+            "zip"
+    );
+    // "Phrases" will be searched for without regard to word boundaries. (This is so that user configuration )
+    static final Set<String> DEFAULT_SENSITIVE_PHRASES = SetUtils.of(
+            "accesskey",
```
kstich, line 65: `accesstoken` as well.

kstich (changes requested):
This will need documentation in the spec [in this file](https://github.com/awslabs/smithy/blob/main/docs/source-2.0/guides/model-linters.rst#linters-in-smithy-linters)

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+                .filter(shape -> !shape.hasTrait(SENSITIVE))
+                .filter(shape -> containsSensitiveWord(shape.toShapeId().getName())
+                        || containsSensitivePhrase(shape.toShapeId().getName()))
+                .forEach(shape -> validationEvents.add(emit(shape)));
+
+        return validationEvents;
+    }
+
+    private List<ValidationEvent> scanMemberNames(Model model) {
+        return model.shapes()
```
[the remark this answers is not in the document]
rchache: Hmm, the reason I didn't go that route was that I need to filter out members where their enclosing shape is already marked sensitive, but the member's target is not. Otherwise we get false positives

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private boolean containsSensitiveWord(String name) {
+        return ValidationUtils.splitCamelCaseWord(name)
+                .stream()
+                .map(word -> word.toLowerCase(Locale.US))
+                .anyMatch(sensitiveWords::contains);
+    }
+
+    private boolean containsSensitivePhrase(String name) {
+        String lowerCasedName = name.toLowerCase(Locale.US);
```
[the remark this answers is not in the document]
rchache: Yeah, I wanted to be able to detect when a model has weird or nonstandard word "breaks". Ex: this would catch both "FirstName" and "Firstname"

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,196 @@
...
+                        || containsSensitivePhrase(shape.toShapeId().getName()))
+                .forEach(shape -> validationEvents.add(emit(shape)));
+
+        return validationEvents;
+    }
+
+    private List<ValidationEvent> scanMemberNames(Model model) {
+        return model.shapes()
+                // filter out members with an already sensitive enclosing shape
+                .filter(shape -> !shape.hasTrait(SENSITIVE))
```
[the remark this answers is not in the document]
rchache: I did some testing and this did not work for me. I forgot which, but it either did not correctly recognize when the outer shape had the sensitive trait, or did not recognize when the targeted shape had the sensitive trait. So in the end I had to do the logic as shown.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: Made both configurable separately and added details in the spec about when to use each one

rchache:
due to git rebasing issues, have to abandon this PR and make a new one:
https://github.com/awslabs/smithy/pull/1364

PR #1337, opened 2022-08-10
Update CodegenDirector to handle mixin and enum shape

gosar:
Use transforms to flatten mixins and change string enums to enum shapes. Added test assertions for these behaviors.

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
@@ -34,6 +34,7 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.neighbor.Walker;
 import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.node.NodeMapper;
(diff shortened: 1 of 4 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
```diff
@@ -455,14 +455,11 @@ public Void stringShape(StringShape shape) {
             return null;
         }
 
-        /*
```
mtdowling, line 458: Hah, I had a nagging feeling that there was a TODO item here for us :)

mtdowling (approved):

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
```diff
@@ -455,14 +455,11 @@ public Void stringShape(StringShape shape) {
...
-        TODO: uncomment in idl-2.0 and remove the stringShape method
```
gosar, line 459: Note, this comment said `remove the stringShape method`, but I think it needs to remain, because a code generator could chose to not `performDefaultCodegenTransforms` and thus need to handle string enums.

PR #1359, opened 2022-08-18
Update `@httpResponseCode` to not be applicable to `@input` structures

### LINKED ISSUE #1105: Make `httpResponseCode` not worked with `httpError`
The following Smithy model builds without any warnings:

```smithy
$version: "1.0"

namespace com.amazonaws.simple

use aws.protocols#restJson1

@restJson1
@title("SimpleService")
service SimpleService {
    operations: [
        AnOperation,
    ],
}

@http(uri: "/operation", method: "GET")
operation AnOperation {
    input: AnOperationInput,
    output: AnOperationOutput,
    errors: [MyError]
}

structure AnOperationInput {
    @httpResponseCode
    responseCode: Integer
}

structure AnOperationOutput { }

@error("client")
@httpError(404)
structure MyError {
    @httpResponseCode
    responseCode: Integer
}
```

I'd expect it to be fail with errors:
1. ~[The spec](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpresponsecode-trait) says `httpResponseCode` is ignored if applied to operation input member. Why not enforce this by failing the build? It's clearly an error by the user.~
2. What response code should `MyError` set? Does the dynamic `httpResponseCode` one take precedence over the static `httpError` one if provided? I'd prefer it if [we made `httpResponseCode` coupled with `required`](https://github.com/awslabs/smithy/issues/1098), and outright rejected `httpResponseCode` applied to error structure members.

syall:
*Issue #, if available:*

#1105

---

*Description of changes:*

Update `@httpResponseCode` to not be applicable to `@input` structures:

- Changed the selector of `@httpResponseCode` to not select structures
  with the `@input` trait applied
- Updated the docs for `@httpResponseCode`

---

*Testing:*

Ran `./gradlew clean build check assemble --parallel --no-build-cache --no-daemon`

---


smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-response-code-input-conflicts.json
```diff
@@ -0,0 +1,19 @@
+{
```
DavidOgunsAWS, line 1: In general, prefer to write new unit test data in IDL over JSON.

docs/source-1.0/spec/core/http-traits.rst
```diff
@@ -1157,12 +1158,12 @@ to the `code` set by the :ref:`http-trait`.
 
 .. rubric:: ``httpResponseCode`` is only used on output
 
-``httpResponseCode`` is ignored when resolving the HTTP bindings of an
-operation's input structure. This means that if a structure that contains
-members marked with the ``httpResponseCode`` trait is used as the top-level
-input structure of an operation, then those members are sent as part of the
-:ref:`protocol-specific document <http-protocol-document-payloads>` sent in
-the body of the request.
+``httpResponseCode`` is ignored when resolving the HTTP bindings of any
```
DavidOgunsAWS, line 1161: This change would be backwards incompatible and potentially break existing models. It's a 180 on the prior posture of ignoring it, to making it an error. It seems like we should have a new validator and make it a warning instead.

syall: The blast radius of existing models is probably 0 for input structures with `@httpResponseCode` applied.

Having stronger constraints helps to avoid misusing the trait in future models which is helpful for API developers.

---

The documentation change rewords existing and intended behavior to avoid misunderstandings about input / error structures.

DavidOgunsAWS (approved):

PR #1364, opened 2022-08-18
Sensitive validator branch

rchache:
*Issue #, if available:*

*Description of changes:*
Based on various existing services for sensitive data detection and current AWS sensitive trait usage, generated a default list of words and phrases that likely indicate the data stored inside is sensitive. It is configurable the way the non inclusive terms validator is.

### DIFF
--- a/docs/source-2.0/guides/model-linters.rst
+++ b/docs/source-2.0/guides/model-linters.rst
@@ -129,7 +129,8 @@ This validator scans shape or member names and identifies ones that look like th
 sensitive information but are not marked with the ``@sensitive`` trait. This does not apply to
 shapes where the ``@sensitive`` trait would be invalid. Users may also configure this validator
 with a custom list of terms, and choose to ignore the built-in defaults. The defaults terms include
-categories of personal information such as 'birth day', 'billing address', 'zip code', or 'gender'.
+types of personal information such as 'birth day', 'billing address', 'zip code', or 'gender',
+as well as information that could be maliciously exploited such as  'password', 'secret key', or 'credit card'.
 
 Rationale
     Sensitive information often incurs legal requirements regarding the handling and logging
@@ -169,7 +170,7 @@ Example:
         name: "MissingSensitiveTrait"
         configuration: {
             excludeDefaults: false,
-            terms: ["social security number"]
+            terms: ["home planet"]
         }
     }]
 
@@ -400,7 +401,7 @@ be specified.
 .. _words-boundaries:
 
 Words boundary matching
---------------------------------
+-----------------------
 
 Word boundaries can be used to find terms of interest. Word boundary search
 text consists of one or more alphanumeric words separated by a single
--- a/docs/source-2.0/guides/model-linters.rst
+++ b/docs/source-2.0/guides/model-linters.rst
@@ -128,7 +128,8 @@ MissingSensitiveTrait
 This validator scans shape or member names and identifies ones that look like they could contain
 sensitive information but are not marked with the ``@sensitive`` trait. This does not apply to
 shapes where the ``@sensitive`` trait would be invalid. Users may also configure this validator
-with custom lists of words or phrases, and choose to ignore the built-in defaults.
+with a custom list of terms, and choose to ignore the built-in defaults. The defaults terms include
+categories of personal information such as 'birth day', 'billing address', 'zip code', or 'gender'.
 
 Rationale
     Sensitive information often incurs legal requirements regarding the handling and logging
@@ -146,22 +147,17 @@ Configuration
        * - Property
          - Type
          - Description
-       * - words
+       * - terms
          - [ ``string`` ]
-         - A list of words to either be added to or replace the default
-           words. These are matched to without regards to casing.
-           This property defaults to an empty list.
-       * - phrases
-         - [ ``string`` ]
-         - A list of phrases to either be added to or replace the default
-           phrases. These are matched to without regards to casing or word boundaries.
-           Use this property when trying to match word fragments or multi-words.
-           This property defaults to an empty list.
+         - A list of search terms that match shape or member names
+           case-insensitively based on word boundaries (for example, the term
+           "access key id" matches "AccessKeyId", "access_key_id", and
+           "accesskeyid"). See :ref:`words-boundaries` for details.
        * - excludeDefaults
          - ``boolean``
          - A flag indicating whether or not to disregard the default set
-           of words and phrases. This property is not required and defaults to false.
-           If set to true, either ``words`` or ``phrases`` must be provided.
+           of terms. This property is not required and defaults to false.
+           If set to true, ``terms`` must be provided.
 
 Example:
 
@@ -173,8 +169,7 @@ Example:
         name: "MissingSensitiveTrait"
         configuration: {
             excludeDefaults: false,
-            words: ["confidential"],
-            phrases: ["MyPrivateInfo"]
+            terms: ["social security number"]
         }
     }]
 
@@ -284,7 +279,7 @@ Configuration
           - A list of search terms that match shape or member names
             case-insensitively based on word boundaries (for example, the term
             "access key id" matches "AccessKeyId", "access_key_id", and
-            "accesskeyid"). See :ref:`reserved-words-boundaries` for details.
+            "accesskeyid"). See :ref:`words-boundaries` for details.
         * - selector
           - ``string``
           - Specifies a selector of shapes to validate for this configuration.
@@ -402,12 +397,12 @@ be specified.
       * - **Codename**
         - Match
 
-.. _reserved-words-boundaries:
+.. _words-boundaries:
 
-Reserved words boundary matching
+Words boundary matching
 --------------------------------
 
-Word boundaries can be used to find reserved words. Word boundary search
+Word boundaries can be used to find terms of interest. Word boundary search
 text consists of one or more alphanumeric words separated by a single
 space. When comparing against another string, the contents of the string
 are separated into words based on word boundaries. Those words are
@@ -436,7 +431,7 @@ demonstrates how comparison text is parsed into words.
     * - access_keyID
       - access key id
 
-The following table shows matches for a reserved term of ``secret id``,
+The following table shows matches for a search term of ``secret id``,
 meaning the word "secret" needs to be followed by the word "id". Word
 boundary searches also match if the search terms concatenated together with
 no spaces is considered a word in the search text (for example,
--- a/docs/source-2.0/guides/model-linters.rst
+++ b/docs/source-2.0/guides/model-linters.rst
@@ -120,6 +120,63 @@ Example:
         {name: "CamelCase"}
     ]
 
+.. _MissingSensitiveTrait:
+
+MissingSensitiveTrait
+=====================
+
+This validator scans shape or member names and identifies ones that look like they could contain
+sensitive information but are not marked with the ``@sensitive`` trait. This does not apply to
+shapes where the ``@sensitive`` trait would be invalid. Users may also configure this validator
+with custom lists of words or phrases, and choose to ignore the built-in defaults.
+
+Rationale
+    Sensitive information often incurs legal requirements regarding the handling and logging
+    of it. Mistakenly not marking sensitive data accordingly carries a large risk, and it is
+    helpful to have an automated validator to catch instances of this rather than rely on best efforts.
+
+Default severity
+    ``WARNING``
+
+Configuration
+    .. list-table::
+       :header-rows: 1
+       :widths: 20 20 60
+
+       * - Property
+         - Type
+         - Description
+       * - words
+         - [ ``string`` ]
+         - A list of words to either be added to or replace the default
+           words. These are matched to without regards to casing.
+           This property defaults to an empty list.
+       * - phrases
+         - [ ``string`` ]
+         - A list of phrases to either be added to or replace the default
+           phrases. These are matched to without regards to casing or word boundaries.
+           Use this property when trying to match word fragments or multi-words.
+           This property defaults to an empty list.
+       * - excludeDefaults
+         - ``boolean``
+         - A flag indicating whether or not to disregard the default set
+           of words and phrases. This property is not required and defaults to false.
+           If set to true, either ``words`` or ``phrases`` must be provided.
+
+Example:
+
+.. code-block:: smithy
+
+    $version: "2"
+
+    metadata validators = [{
+        name: "MissingSensitiveTrait"
+        configuration: {
+            excludeDefaults: false,
+            words: ["confidential"],
+            phrases: ["MyPrivateInfo"]
+        }
+    }]
 
 .. _NoninclusiveTerms:
 
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
@@ -16,22 +16,16 @@
 package software.amazon.smithy.linters;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.node.NodeMapper;
+import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.traits.SensitiveTrait;
 import software.amazon.smithy.model.validation.AbstractValidator;
-import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
-import software.amazon.smithy.model.validation.ValidationUtils;
 import software.amazon.smithy.model.validation.ValidatorService;
 import software.amazon.smithy.utils.ListUtils;
 import software.amazon.smithy.utils.SetUtils;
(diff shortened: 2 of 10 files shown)


smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+import software.amazon.smithy.model.validation.ValidationUtils;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SetUtils;
+
+/**
+ * <p>Validates that shapes and members that possibly contain sensitive data are marked with the sensitive trait.
+ */
+public final class MissingSensitiveTraitValidator extends AbstractValidator {
+    static final Set<String> DEFAULT_SENSITIVE_WORDS = SetUtils.of(
```
mtdowling, line 43: If we used wildcard syntax similar to what we did with ReservedWords, we could even put these in a newline delimited file instead to make it easier to manage.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            "authentication",
+            "authorization",
+            "bank",
+            "billing",
+            "birth",
```
mtdowling, line 48: "birthday"?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            "credential",
+            "email",
```
mtdowling, line 50: would this match "emailAddress" too?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            "ethnicity",
+            "insurance",
+            "license",
+            "passkey",
+            "passphrase",
+            "passport",
+            "password",
+            "phone",
```
mtdowling, line 58: Would this match phoneNumber?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            "private",
+            "secret",
+            "sensitive",
+            "telephone",
```
mtdowling, line 62: would this match telephoneNumber?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            "token",
+            "username",
+            "zip"
```
mtdowling, line 65: would this match zipCode?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+    );
+    static final Set<String> DEFAULT_SENSITIVE_PHRASES = SetUtils.of(
+            "accesskey",
+            "accesstoken",
+            "auth",
+            "creditcard",
+            "firstname",
+            "lastname",
+            "plaintext",
```
mtdowling, line 74: Why this one?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    /**
+     * MissingSensitiveTrait configuration.
+     */
+    public static final class Config {
+        private List<String> phrases = ListUtils.of();
+        private List<String> words = ListUtils.of();
+
```
mtdowling, line 94: nit extra newline

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+    private Optional<ValidationEvent> detectSensitivePhrase(String name, Shape shape) {
+        String lowerCasedName = name.toLowerCase(Locale.US);
+        return sensitivePhrases.stream()
+                .filter(lowerCasedName::contains)
+                .findAny()
+                .map(phrase -> emit(shape, phrase));
+    }
+
+    private ValidationEvent emit(Shape shape, String word) {
+        return ValidationEvent.builder()
```
mtdowling, line 225: You could just use the `warning` method of AbstractValidator instead.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+    private Optional<ValidationEvent> detectSensitiveWord(String name, Shape shape) {
+        return ValidationUtils.splitCamelCaseWord(name)
+                .stream()
+                .map(word -> word.toLowerCase(Locale.US))
+                .filter(sensitiveWords::contains)
+                .findAny()
+                .map(word -> emit(shape, word));
+    }
+
+    private Optional<ValidationEvent> detectSensitivePhrase(String name, Shape shape) {
```
mtdowling, line 216: Could this not match multiple phrases?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+                .map(memberShape -> detectSensitiveWord(memberShape.getMemberName(), memberShape).map(Optional::of)
+                        .orElse(detectSensitivePhrase(memberShape.getMemberName(), memberShape)))
+                .flatMap(o -> o.map(Stream::of).orElseGet(Stream::empty))
+//                Once we finally upgrade from Java 8, above line can be simplified to
+//                .flatMap(Optional::stream)
+                .collect(Collectors.toList());
+    }
+
+    private Optional<ValidationEvent> detectSensitiveWord(String name, Shape shape) {
+        return ValidationUtils.splitCamelCaseWord(name)
```
mtdowling, line 208: Seems like this could match multiple words too but doesn't support that, right?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+        return model.shapes()
+                .filter(shape -> !shape.isMemberShape()
+                        && !shape.isOperationShape()
+                        && !shape.isServiceShape()
+                        && !shape.isResourceShape())
+                .filter(shape -> !shape.hasTrait(SensitiveTrait.ID))
+                .map(shape -> detectSensitiveWord(shape.toShapeId().getName(), shape).map(Optional::of)
+                        .orElse(detectSensitivePhrase(shape.toShapeId().getName(), shape)))
+                .flatMap(o -> o.map(Stream::of).orElseGet(Stream::empty))
+//                Once we finally upgrade from Java 8, above line can be simplified to
```
mtdowling, line 184: I also think this would be much more readable with just normal for loops... which also removes the need for this comment :)

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+//                .flatMap(Optional::stream)
+                .collect(Collectors.toList());
+    }
+
+    private List<ValidationEvent> scanMemberNames(Model model) {
+        return model.shapes()
```
mtdowling, line 190: This and the other method are pretty inefficient. You could filter up front all shapes marked with the sensitive trait, put that in a set, then consult that when iterating over shapes. In this case, you'd only want to iterate over members too.

```
// This method is cached and might have already have been called.
Set<Shape> sensitiveShapes = model.getShapesWithTrait(SensitiveTrait.class);
```

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+                    .map(phrase -> phrase.toLowerCase(Locale.US))
+                    .collect(Collectors.toSet()));
+            wordsInit.addAll(config.getWords()
+                    .stream()
+                    .map(word -> word.toLowerCase(Locale.US))
+                    .collect(Collectors.toSet()));
+            sensitivePhrases = Collections.unmodifiableSet(phrasesInit);
+            sensitiveWords = Collections.unmodifiableSet(wordsInit);
+        } else {
+            if (config.getPhrases().isEmpty() && config.getWords().isEmpty()) {
```
mtdowling, line 140: nit: make this an else if branch, then move the rest that follows as an else branch

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+    private final Set<String> sensitivePhrases;
+
+    private MissingSensitiveTraitValidator(Config config) {
+        if (!config.getExcludeDefaults()) {
+            Set<String> phrasesInit = new HashSet<>(DEFAULT_SENSITIVE_PHRASES);
+            Set<String> wordsInit = new HashSet<>(DEFAULT_SENSITIVE_WORDS);
+            phrasesInit.addAll(config.getPhrases()
+                    .stream()
+                    .map(phrase -> phrase.toLowerCase(Locale.US))
+                    .collect(Collectors.toSet()));
```
mtdowling, line 132: you could just `.forEach(phrasesInit::add)` so you don't create an unnecessary intermediate set

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,233 @@
...
+            wordsInit.addAll(config.getWords()
+                    .stream()
+                    .map(word -> word.toLowerCase(Locale.US))
+                    .collect(Collectors.toSet()));
```
mtdowling, line 136: Same as above (use forEach with add to avoid intermediate set)

mtdowling (changes requested):

mtdowling:
I created a PR here that is used by ReservedWords to support word boundary based term matching. I made an abstraction for this so that it could be used by this linter as well. Let me know what you think. https://github.com/awslabs/smithy/pull/1461

docs/source-2.0/guides/model-linters.rst
```diff
@@ -120,6 +120,63 @@ Example:
...
+
+       * - Property
+         - Type
+         - Description
+       * - words
+         - [ ``string`` ]
+         - A list of words to either be added to or replace the default
+           words. These are matched to without regards to casing.
+           This property defaults to an empty list.
+       * - phrases
```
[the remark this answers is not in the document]
mtdowling: Based on the work done in #1461, I think we should remove "words" and "phrases", and add "terms" to match the changes being made to the reserved words validator to match based on word boundaries (you can use the WordBoundaryMatcher too). Given this validator validates member names and shape names, I don't think we actually need substring matching if we support word boundary matching.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: Yeah, that's definitely true actually

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: My thought was that "plaintext" usually implies that something had been decrypted, but it's true that this could lead to mostly false-positives. I'll remove it

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: It would yes, but I'll be reworking this with terms and WordBoundary

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: It would yes, but I'll be reworking this with terms and WordBoundary

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: It would yes, but I'll be reworking this with terms and WordBoundary

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: good catch, I'll add it to the new terms list

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+                        detectSensitiveTerms(memberShape.getMemberName(), memberShape);
+                optionalValidationEvent.ifPresent(validationEvents::add);
+            }
+        }
+
+        return validationEvents;
+    }
+
+    private Optional<ValidationEvent> detectSensitiveTerms(String name, Shape shape) {
+        Optional<String> matchedTerm = wordMatcher.getAllMatches(name).stream().findAny();
```
mtdowling, line 170: This looks like another place where I'd just use a for loop and avoid the streaming API since this gets call for tons of times and the streaming API isn't really needed.

```
Set<String> matches = wordMatcher.getAllMatches(name);
return !matches.isEmpty() ? Optional.of(emit(shape, matches)) : Optional.empty();
```

Not sure why `emit` doesn't accept all the matches though in the current implementation. Is it just emitting for the first right now?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        List<ValidationEvent> validationEvents = new ArrayList<>();
+        validationEvents.addAll(scanShapeNames(model));
+        validationEvents.addAll(scanMemberNames(model));
+        return validationEvents;
+    }
+
+    private List<ValidationEvent> scanShapeNames(Model model) {
```
mtdowling, line 133: Since you have to scan over every shape in the model already, I'd look to combine scanShapeNames and scanMemberNames into a single method that iterates over the shapes in the model once. Then do the `if (shape.isMemberShape()) {` check first to scan member names, followed by the other checks to test shape names.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+        public void setExcludeDefaults(boolean excludeDefaults) {
+            this.excludeDefaults = excludeDefaults;
+        }
+    }
+
+    private MissingSensitiveTraitValidator(Config config) {
+        wordMatcher = new WordBoundaryMatcher();
+        if (config.getExcludeDefaults() && config.getTerms().isEmpty()) {
+            //This configuration combination makes the validator a no-op.
+            throw new IllegalArgumentException("Cannot set 'excludeDefaults' to true and leave "
```
mtdowling, line 107: Can you add a test for this?

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+            "mailing address",
+            "passport",
+            "phone",
+            "religion",
+            "sexual orientation",
+            "social security",
+            "ssn",
+            "tax payer",
+            "telephone",
+            "user name",
```
mtdowling, line 63: I'm not sure this is sensitive

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+ * <p>Validates that shapes and members that possibly contain sensitive data are marked with the sensitive trait.
+ */
+public final class MissingSensitiveTraitValidator extends AbstractValidator {
+    static final Set<String> DEFAULT_SENSITIVE_TERMS = SetUtils.of(
+            "account number",
+            "bank",
+            "billing address",
+            "birth day",
+            "birth",
+            "citizen ship",
```
mtdowling, line 43: Not sure I've ever seen this as anything but "citizenship". The current form might catch false-positives.

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
```diff
@@ -0,0 +1,184 @@
...
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidatorService;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SetUtils;
+
+/**
+ * <p>Validates that shapes and members that possibly contain sensitive data are marked with the sensitive trait.
+ */
+public final class MissingSensitiveTraitValidator extends AbstractValidator {
+    static final Set<String> DEFAULT_SENSITIVE_TERMS = SetUtils.of(
```
mtdowling, line 37: Can add more here?
* `secret`
* `pass phrase` (will matchPassPhrase, pass_phrase, and passphrase)
* `pass word` (will match password, passWord, and pass_word)
* `credentials`
* `maiden name`
* `csc`

docs/source-2.0/guides/model-linters.rst
```diff
@@ -120,6 +120,58 @@ Example:
...
+.. _MissingSensitiveTrait:
+
+MissingSensitiveTrait
+=====================
+
+This validator scans shape or member names and identifies ones that look like they could contain
+sensitive information but are not marked with the ``@sensitive`` trait. This does not apply to
+shapes where the ``@sensitive`` trait would be invalid. Users may also configure this validator
+with a custom list of terms, and choose to ignore the built-in defaults. The defaults terms include
+categories of personal information such as 'birth day', 'billing address', 'zip code', or 'gender'.
```
mtdowling, line 132: Nice. I'd expand this a bit to also include password, secret key, and credit card since it's more than just PII being caught here.

mtdowling (changes requested):

smithy-linters/src/main/java/software/amazon/smithy/linters/MissingSensitiveTraitValidator.java
rchache: Since we decided to short-circuit, I'll change this to just getting the first match (as a new method) and emitting that alone

mtdowling: Ok, in that case, the method would need a different name than getAllMatches (getFirstMatch or something).

rchache: Yep, I wrote that method 👍 Current plan is to also leave in getAllMatches as well, unless you think I should remove it since it's not currently used

mtdowling: Sure yeah, remove it if it isn't used any longer

mtdowling (approved):

PR #1385, opened 2022-09-04
Add method to check if NumberNode is set to zero

mtdowling:

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
@@ -16,6 +16,7 @@
 package software.amazon.smithy.model.node;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -31,7 +32,7 @@ import software.amazon.smithy.model.SourceLocation;
 public final class NumberNode extends Node {
 
     private final Number value;
-    private String stringCache;
+    private final String stringCache;
 
     public NumberNode(Number value, SourceLocation sourceLocation) {
         super(sourceLocation);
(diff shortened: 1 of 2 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
```diff
@@ -94,6 +95,38 @@ public Optional<NumberNode> asNumberNode() {
...
+        // Do a cheap test based on the serialized value of the number first.
+        if (toString().equals("0")) {
+            return true;
+        } else if (value instanceof BigDecimal) {
+            return value.equals(BigDecimal.ZERO);
+        } else if (value instanceof BigInteger) {
+            return value.equals(BigInteger.ZERO);
+        } else if (value instanceof Float) {
+            // Using Math.signum to avoid epsilon issues.
+            return Math.signum((Float) value) == 0;
```
gosar, line 115: I'm unsure if we want isZero to return true of false for negative zero. From signum javadoc:
>Special Cases:
    If the argument is NaN, then the result is NaN.
    If the argument is positive zero or negative zero, then the result is the same as the argument. 

So if we want isZero false, this seems fine. But if we want isZero true, then we may need to add another `||` condition here for negative zero.

Either ways, we should add javadoc comment and test case for negative zero.

smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
```diff
@@ -94,6 +95,38 @@ public Optional<NumberNode> asNumberNode() {
...
+    public boolean isZero() {
+        // Do a cheap test based on the serialized value of the number first.
+        if (toString().equals("0")) {
+            return true;
+        } else if (value instanceof BigDecimal) {
+            return value.equals(BigDecimal.ZERO);
+        } else if (value instanceof BigInteger) {
+            return value.equals(BigInteger.ZERO);
+        } else if (value instanceof Float) {
+            // Using Math.signum to avoid epsilon issues.
```
gosar, line 114: If there is a test case that shows the epsilon issue that would have broken without this behavior, that'd be great.

smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
mtdowling: I think -0 and +0 are fine. I'll add tests.

smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
```diff
@@ -94,6 +95,38 @@ public Optional<NumberNode> asNumberNode() {
...
+            return true;
+        } else if (value instanceof BigDecimal) {
+            return value.equals(BigDecimal.ZERO);
+        } else if (value instanceof BigInteger) {
+            return value.equals(BigInteger.ZERO);
+        } else if (value instanceof Float) {
+            return value.floatValue() == 0;
+        } else {
+            return value.doubleValue() == 0;
+        }
```
gosar, line 127:
```suggestion
        } else if (value instanceof Float) {
            return value.floatValue() == 0f;
        } else {
            return value.doubleValue() == 0d;
        }
```

gosar (commented):
ship it after this last change requested

smithy-model/src/main/java/software/amazon/smithy/model/node/NumberNode.java
mtdowling: updated

gosar (approved):

Issue #1398, opened 2022-09-12
Content-Type without modeled body in input


82marbag:
A content-type must not be set and the body empty when there is no modeled input. Is it the same when the input is modeled with only members with `@httpHeader` traits?

https://github.com/awslabs/smithy-rs/pull/1723#discussion_r967033105

jvschneid:
Related PR has been merged, closing issue.

PR #1399, opened 2022-09-12
Test content-type modeled inputs without body

82marbag:
*Issue #, if available:* #1398

*Description of changes:* Add a test to validate the body must be empty and the content-type not set when the input is modeled with only members with `@httpHeader` traits


JordonPhillips (changes requested):
The CI failure doesn't seem to be related, but can you rebase to re-kick it off? I would do it manually, but GitHub isn't letting me.

JordonPhillips (approved):

PR #1410, opened 2022-09-16
Add test for 1.0 roundtrip and serialization

syall:
*Issue #, if available:*

N/A

---

*Description of changes:*

Current test cases for roundtrip and serialization:

- 2.0 AST -> 2.0 In-memory Model -> 2.0 AST
- 2.0 AST -> 2.0 In-memory Model -> 1.0 AST

This commmit adds the following test case for enums:

- 1.0 AST -> 2.0 In-memory Model -> 2.0 AST
- 1.0 AST -> 2.0 In-memory Model -> 1.0 AST

Also, added a string shape with the `@enum` trait for the 2.0 enum roundtrip test case.

- [ ] Add tests for root level box shapes in 1.0
- [ ] Add tests for member level box shapes in 1.0

---


smithy-model/src/main/java/software/amazon/smithy/model/transform/DowngradeToV1.java
```diff
@@ -119,12 +118,8 @@ private Model removeUnnecessaryDefaults(ModelTransformer transformer, Model mode
     private Model removeOtherV2Traits(ModelTransformer transformer, Model model) {
         Set<Shape> updates = new HashSet<>();
 
-        for (StructureShape structure : model.getStructureShapes()) {
-            for (MemberShape member : structure.getAllMembers().values()) {
-                if (member.hasTrait(ClientOptionalTrait.class)) {
-                    updates.add(member.toBuilder().removeTrait(ClientOptionalTrait.ID).build());
-                }
-            }
+        for (MemberShape member : model.getMemberShapesWithTrait(ClientOptionalTrait.class)) {
```
mtdowling, line 121: This trait can only be applied to structure member, so why check all members?

syall: Reverted the change.

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
```diff
@@ -122,9 +122,16 @@ public void removesUnnecessaryDefaults() {
...
                 + "integer MyInteger\n"
+                + "@default(0)\n"
+                + "integer ZeroInteger\n"
+                + "integer BoxedInteger\n"
                 + "structure Struct {\n"
                 + "  @default(10)\n"
                 + "  foo: MyInteger\n"
+                + "  defaultZero: ZeroInteger = 0\n"
+                + "  boxedZero: ZeroInteger = null\n"
+                + "  boxedInteger: BoxedInteger = 0\n"
```
gosar, line 133: `boxedInteger` having a default 0 and `boxedMemberInteger` being boxed seems confusing.
maybe?
```
boxedIntegerWithMemberDefault: BoxedInteger = 0
boxedInteger: BoxedInteger = null
```

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
```diff
@@ -122,9 +122,16 @@ public void removesUnnecessaryDefaults() {
...
+                + "  boxedMemberInteger: BoxedInteger = null\n"
```
gosar, line 134: hmm, I didn't realize we could add `= null` to a member even if the target had no default. seems unnecessary but seems ok to make it explicit that member is boxed.

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
```diff
@@ -145,10 +152,38 @@ public void removesUnnecessaryDefaults() {
...
+
+        ShapeId boxedIntegerShape = ShapeId.from("smithy.example#BoxedInteger");
+        assertThat(downgraded.expectShape(boxedIntegerShape).hasTrait(DefaultTrait.class), Matchers.is(false));
+        assertThat(downgraded.expectShape(boxedIntegerShape).hasTrait(BoxTrait.class), Matchers.is(true));
+
         StructureShape dStruct = downgraded.expectShape(ShapeId.from("smithy.example#Struct"), StructureShape.class);
         assertThat(dStruct.getAllMembers().get("foo").hasTrait(DefaultTrait.class), Matchers.is(false));
+        assertThat(dStruct.getAllMembers().get("foo").hasTrait(BoxTrait.class), Matchers.is(false));
+
+        assertThat(dStruct.getAllMembers().get("defaultZero").hasTrait(DefaultTrait.class), Matchers.is(false));
```
gosar, line 168: `// ZeroInteger keeps the default trait since it matches the zero value of a 1.0 model` above and this seems inconsistent. It is implying we retain default zero value for root shape but not member - is this expected behavior?

I see `// Members with a null default are considered boxed. Keep the trait to retain consistency with other indexes and checks.` in the source code for `null` default to be retained. But doesn't say about retain default zero on root/member. _If_ the current behavior is expected, we should at least explain that in the code comment as it is not intuitive.

smithy-model/src/test/resources/software/amazon/smithy/model/shapes/ast-serialization/cases/box-member-1.0-to.json
```diff
@@ -0,0 +1,157 @@
+{
```
gosar, line 1: are these file names correct?

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -52,15 +52,18 @@ public class ModelSerializerTest {
...
     }
 
     public void testRoundTrip(Path path) {
         Model model = Model.assembler().addImport(path).assemble().unwrap();
         ModelSerializer serializer = ModelSerializer.builder().build();
         ObjectNode actual = serializer.serialize(model);
-        ObjectNode expected = Node.parse(IoUtils.readUtf8File(path)).expectObjectNode();
+        // Use explicit 2.0 comparison file if present; otherwise, default to original path
+        Path potentialExpectedFile2 = Paths.get(path.toString().replace(".json", ".2.0.json"));
+        Path expectedFile2 = Files.exists(potentialExpectedFile2) ? potentialExpectedFile2 : path;
```
gosar, line 65: i'm a bit unsure about all this logic and the file names. Seems like there should be a better way to do this.

smithy-model/src/test/resources/software/amazon/smithy/model/shapes/ast-serialization/cases/box-member-1.0-to.json
syall: Will rename once we find a better way to load in expected file names for 1.0 roundtrip serialization.

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
syall: Will rename the added members.

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
syall: This will be fixed in the next revision to keep the default zero value for members that have a default zero value and a target shape that has a default zero value.

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
syall: Will be refactored into more verbose / separate test factories.

smithy-model/src/test/java/software/amazon/smithy/model/transform/DowngradeToV1Test.java
```diff
@@ -120,15 +120,34 @@ public void removesResourceProperties() {
...
+                // StructureShape still exists
                 + "structure Struct {\n"
+                // A Member that targets a shape with a 1.0 non-zero value drops the default value and are not boxed
                 + "  @default(10)\n"
                 + "  foo: MyInteger\n"
+                // A Member that targets a shape with a 1.0 zero value and the member default value matches keeps the default value
+                + "  zeroTargetZeroMember: ZeroInteger = 0\n"
+                // A Member that has a default value of null keeps the default value and is boxed
+                + "  zeroTargetBoxedMember: ZeroInteger = null\n"
+                // A Member that has doesn't have the same default value on the member and trait have the default value dropped
```
gosar, line 140: This sentence doesn't make sense. It seems like
```
// A Member that has a zero default value and targets a shape without a default value...
```

Which also tells me that we should add a case for 
```
// A Member that has a non-zero default value and targets a shape without a default value...
boxedTargetZeroMember: BoxedInteger = 20\n"
```

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
 
 public class ModelSerializerTest {
     @TestFactory
-    public Stream<DynamicTest> generateTests() throws IOException, URISyntaxException {
+    public Stream<DynamicTest> generateTestsV2RoundTrip() throws IOException, URISyntaxException {
```
gosar, line 52: nit: generateV2RoundTripTests

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
...
         return Files.list(Paths.get(
-                        SmithyIdlModelSerializer.class.getResource("ast-serialization/cases").toURI()))
+                        SmithyIdlModelSerializer.class.getResource("ast-serialization/cases/v2").toURI()))
                 .filter(path -> !path.toString().endsWith(".1.0.json"))
-                .map(path -> DynamicTest.dynamicTest(path.getFileName().toString(), () -> testRoundTrip(path)));
+                .map(path -> DynamicTest.dynamicTest(path.getFileName().toString(), () -> testRoundTripV2(path)));
     }
 
-    public void testRoundTrip(Path path) {
+    public void testRoundTripV2(Path path) {
```
gosar, line 59: nit: private

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
...
+        testV2Serialization(path, path);
+        testV1DowngradeSerialization(path, Paths.get(path.toString().replace(".json", ".1.0.json")));
+    }
+
+    @TestFactory
+    public Stream<DynamicTest> generateTestsV1RoundTrip() throws IOException, URISyntaxException {
```
gosar, line 65: nit: generateV1RoundTripTests

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
...
+        return Files.list(Paths.get(
+                        SmithyIdlModelSerializer.class.getResource("ast-serialization/cases/v1").toURI()))
+                .filter(path -> !path.toString().endsWith(".2.0.json"))
+                .map(path -> DynamicTest.dynamicTest(path.getFileName().toString(), () -> testRoundTripV1(path)));
+    }
+
+    public void testRoundTripV1(Path path) {
```
gosar, line 72: nit: private

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
...
+        testV2Serialization(path, Paths.get(path.toString().replace(".json", ".2.0.json")));
+        testV1DowngradeSerialization(path, path);
+    }
+
+    public void testV2Serialization(Path path, Path expectedV2Path) {
```
gosar, line 77: nit: private

smithy-model/src/test/java/software/amazon/smithy/model/shapes/ModelSerializerTest.java
```diff
@@ -49,24 +49,43 @@
...
-        ObjectNode expected = Node.parse(IoUtils.readUtf8File(path)).expectObjectNode();
+        ObjectNode expected = Node.parse(IoUtils.readUtf8File(expectedV2Path)).expectObjectNode();
 
         Node.assertEquals(actual, expected);
+    }
 
-        // Now validate the file is serialized correctly when downgraded to 1.0.
-        Path downgradeFile = Paths.get(path.toString().replace(".json", ".1.0.json"));
-        ObjectNode expectedDowngrade = Node.parse(IoUtils.readUtf8File(downgradeFile)).expectObjectNode();
+    public void testV1DowngradeSerialization(Path path, Path expectedV1Path) {
```
gosar, line 86: nit: private

gosar (approved):
My comments are non-blocking, so marking this as approved anyways.

PR #1412, opened 2022-09-17
Revise default value design doc to match recent updates

mtdowling:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/designs/defaults-and-model-evolution.md
+++ b/designs/defaults-and-model-evolution.md
@@ -135,8 +135,7 @@ This proposal:
 1. Introduces a `@default` trait
 2. Introduces a `@clientOptional` trait
 3. Removes the `@box` trait
-4. Deprecates the `Primitive*` shapes from the prelude
-5. Make the optionality of a structure completely controlled by members rather
+4. Make the optionality of a structure completely controlled by members rather
    than the shape targeted by a member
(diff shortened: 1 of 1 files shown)


srchase (approved):

designs/defaults-and-model-evolution.md
```diff
@@ -277,6 +320,11 @@ The following shapes have restrictions on their default values:
 
 ### Updating default values
 
+The default value of a member MAY be changed. The default value of a root-level
```
gosar, line 323: It is a SHOULD NOT below. The MAY sentence can be removed here.

designs/defaults-and-model-evolution.md
```diff
@@ -277,6 +320,11 @@ The following shapes have restrictions on their default values:
...
+shape MUST NOT be changed since that would break any shape that refers to
+the shape, and could break other models that refer to a shape defined in a
+shared model.
+
 The default value of a member SHOULD NOT be changed. However, it MAY be
```
gosar, line 328: It is implied that this applies only when the member has a default without the target root level shape having a default, but maybe worth calling it out.
`The default value of a member (that targets a shape without a default value) SHOULD NOT be changed.`

PR #1414, opened 2022-09-19
Fix apply statement parsing and ABNF

milesziemer:
This fixes an issue where `applySomeShape @someTrait` was valid, without any space between the apply and shape name. The fix makes it so spacing is required, not line breaks as was the case in the old ABNF, to be consistent with other shapes grammar (see [SimpleShapeStatement](https://awslabs.github.io/smithy/2.0/spec/idl.html#grammar-token-smithy-SimpleShapeStatement))

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -1126,7 +1126,7 @@ final class IdlModelParser extends SimpleParser {
         expect('p');
         expect('l');
         expect('y');
-        sp();
+        rsp();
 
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -1126,7 +1126,7 @@ private void parseApplyStatement() {
         expect('p');
         expect('l');
         expect('y');
-        sp();
+        rsp();
```
sugmanue, line 1129: Too bad that the names of the methods have diverged from the name of the grammar productions, but that's not related to this change.

sugmanue (approved):

PR #1415, opened 2022-09-19
Correct the Resource shape properties Type entry

pcolby:
*Issue #, if available:*

*Description of changes:*

Prior to this change, the docs incorrectly lists the [(JSON AST) Resource shape Type](https://awslabs.github.io/smithy/2.0/spec/json-ast.html#resource-shape) as "AST shape reference", but as per the Description column, and the [Model docs for Resource Properties](https://awslabs.github.io/smithy/2.0/spec/service-types.html#resource-properties), as well as the [Introducing Smithy IDL 2.0 blog post](https://aws.amazon.com/blogs/developer/introducing-smithy-idl-2-0/), this Type should be a map of strings to shape ID.

ie the following blue-highlighted cell is incorrect:

image

This PR changes it to `Map<String, shape ID>`, which matches the Model description and AWS blog post examples.

There's also two other very minor grammar corrections including (dropping two instances of `of`).

Cheers.

### DIFF
--- a/docs/source-2.0/spec/json-ast.rst
+++ b/docs/source-2.0/spec/json-ast.rst
@@ -69,7 +69,7 @@ a ``type`` property to define the shape type or ``apply``.
     }
 
 All entries in the ``shapes`` map can contain a ``traits`` property that
-defines the traits attached to the shape. ``traits`` is a map of where
+defines the traits attached to the shape. ``traits`` is a map where
 each key is the absolute shape ID of a trait shape and each value is
 the value to assign to the trait.
 
--- a/docs/source-2.0/spec/json-ast.rst
+++ b/docs/source-2.0/spec/json-ast.rst
@@ -496,7 +496,7 @@ shapes defined in JSON support the same properties as the Smithy IDL.
       - Map<String, :ref:`AST shape reference <ast-shape-reference>`>
       - Defines identifier names and shape IDs used to identify the resource.
     * - :ref:`properties <resource-properties>`
-      - Map<String, :ref:`shape ID <shape-id>`>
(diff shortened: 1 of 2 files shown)


AndrewFossAWS (approved):

docs/source-2.0/spec/json-ast.rst
```diff
@@ -496,7 +496,7 @@ shapes defined in JSON support the same properties as the Smithy IDL.
       - Map<String, :ref:`AST shape reference <ast-shape-reference>`>
       - Defines identifier names and shape IDs used to identify the resource.
     * - :ref:`properties <resource-properties>`
-      - :ref:`AST shape reference <ast-shape-reference>`
+      - Map<String, :ref:`shape ID <shape-id>`>
```
kstich, line 499: This should be a map of AST shape references instead like `identifiers`, [as can be seen here](https://github.com/awslabs/smithy/blob/c8e8ae3c4e667feecc4205a21f3773c26943b70d/smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/elided-members.json#L183).

```
Map<String, :ref:`AST shape reference <ast-shape-reference>`>
```

kstich (changes requested):

docs/source-2.0/spec/json-ast.rst
pcolby: Thanks @kstich, seeing that example helps, since the use of ref-to-id indirection is not present in the IDL (its an AST-specific concept). I'll update the PR accordingly :)

kstich (approved):

Issue #1277, opened 2022-06-15
Run built-in plugins before others


gosar:
When smithy build fails on a plugin that is not [built-in](https://awslabs.github.io/smithy/1.0/guides/building-models/build-config.html#plugins) (i.e., not `model`, `build-info` or `sources`), even though I think the model itself is good, depending on the order of plugins, the [`model` plugin](https://awslabs.github.io/smithy/1.0/guides/building-models/build-config.html#model-plugin) may not have been run yet. Having the `model` plugin run before would allow seeing the full model serialized even when other plugins fail.

I found that the plugins end up in a [TreeMap here](https://github.com/awslabs/smithy/blob/5c319a22c177dd89989c27c435c12cf64c69550b/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java#L412) making it natural order. It uses LinkedHashMap in the [SmithyBuildConfig here](https://github.com/awslabs/smithy/blob/5c319a22c177dd89989c27c435c12cf64c69550b/smithy-build/src/main/java/software/amazon/smithy/build/model/SmithyBuildConfig.java#L179), so it seems if we use LinkedHashMap instead of TreeMap in SmithyBuildImpl, we could make sure the BUILTIN_PLUGINS always get run first.

milesziemer:
It looks like `BUILTIN_PLUGINS` are added right before `SmithyBuildConfig` is built [here](https://github.com/awslabs/smithy/blob/319389fd289a70514d983efd052a0df7d6d3119a/smithy-build/src/main/java/software/amazon/smithy/build/model/SmithyBuildConfig.java#L199), so the built-in plugins are always after any other plugins present in `SmithyBuildConfig`. I think we would need a combination of using `LinkedHashMap` in `SmithyBuildImpl`, and some way to make sure that `BUILTIN_PLUGINS` come before any other plugins when `SmithyBuildConfig` is built.

[...]

PR #1416, opened 2022-09-20
Defer projection failure until after all plugins run

milesziemer:
*Issue*: #1277 

This changes `SmithyBuildImpl` to continue applying plugins after one fails, throwing the error (if present) after they have all completed. This is useful for example when you want to see the serialized output of a model you know is valid but a plugin is causing the whole build to fail. A new test plugin was added that throws a runtime exception, causing the projection to fail purposefully, and a new test case was added to ensure that artifacts produced by valid plugins are still created despite the build failing.


milesziemer:
Opening this as a draft PR because it's pretty ugly, but not clear to me how to clean it up.

smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
```diff
@@ -283,13 +283,24 @@ private ProjectionResult applyProjection(
                 .model(projectedModel)
                 .events(modelResult.getValidationEvents());
 
+        // Keep track of the first error created by plugins to fail the build after all plugins have run
+        Throwable firstPluginError = null;
```
srchase, line 287: Will this only keep track of a single plugin error? This might be confusing if multiple plugins fail, but only the error for the first (or last?) ends up being thrown at the end.

milesziemer: This only keeps track of the first plugin error. My initial thinking was to keep track of all plugin errors, but errors are handled per-projection (in [`SmithyBuild`](https://github.com/awslabs/smithy/blob/513f05fc34b236d40bcf0c1be849f5e25b90eaf4/smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuild.java#L117) and [`BuildCommand`](https://github.com/awslabs/smithy/blob/513f05fc34b236d40bcf0c1be849f5e25b90eaf4/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/BuildCommand.java#L206)). In order to not change these, we would need to aggregate all plugin errors into one error to throw as the projection error, but I wanted to defer making a decision on what exactly this would look like. Every plugin error is logged however, see [here](https://github.com/awslabs/smithy/pull/1416/commits/1fc7385fd15da01bd98b1463fbdd97e8e680f3ed#diff-e2603e6b410f984d3b14021df3820bac9de49da2db254ff5834522279283fba0R294), but I think it would be nice to throw one big exception with all plugin errors

srchase (approved):

smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
srchase: That sounds like a bigger change that could be done later, but this still seems like a good improvement for now.

PR #1418, opened 2022-09-21
Warn when default used with union member target

mtdowling:
This NOTE warning makes it clear to modelers that default values on the target shape of a union member are ignored.


smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/defaults/warn-when-union-targets-default.smithy
```diff
@@ -0,0 +1,9 @@
+$version: "2.0"
+
+namespace smithy.example
+
+union Example {
+    foo: PrimitiveInteger, // warn
```
gosar, line 6: adding a 
```
foo2: Integer, // don't warn
```
would make it clearer

gosar (approved):

Issue #1373, opened 2022-08-29
`CreateDedicatedInputAndOutput` leaves orphaned shapes in some cases


jdisanti:
Consider the following Smithy model:
```
        namespace test
        operation GenerateSpeech {
            output: GenerateSpeechOutput,
            input: GenerateSpeechOutput
        }

        structure GenerateSpeechOutput {
            data: BlobStream
        }

        @streaming
        blob BlobStream
```
The `CreateDedicatedInputAndOutput` model transformer (with suffixes “Input” and “Output”) will create `GenerateSpeechInput` and `GenerateSpeechOperationOutput` structures to replace the input/output on that operation. However, when it does so, it does NOT remove `GenerateSpeechOutput` from the model when it should because it gets confused into thinking it’s not a singular reference due to the original unmutated operation’s input referencing it. However, it _is_ a singular reference because it replaced that input with `GenerateSpeechInput`.

PR #1419, opened 2022-09-22
Fix dedicated io transform leaving unused shapes

milesziemer:
This commit changes the way `CreateDedicatedInputOutput` transform checks for singular references to a shape being used as input or output in an operation. It now checks to make sure the shape is only referenced by the specified operation. This fixes an edge case where a shape was being used as both the input and output for an operation, and would be left unused in the model after the transformation was applied.  Two new test cases were added to verify the behavior when a shape is used as both input and output.

*Issue #, if available:* #1373

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
@@ -22,9 +22,7 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.neighbor.NeighborProvider;
-import software.amazon.smithy.model.neighbor.Relationship;
 import software.amazon.smithy.model.neighbor.RelationshipDirection;
-import software.amazon.smithy.model.neighbor.RelationshipType;
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeId;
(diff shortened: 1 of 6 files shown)


smithy-model/src/test/resources/software/amazon/smithy/model/transform/dedicated-input-output/handles-dangling-single-reference-before.smithy
```diff
@@ -0,0 +1,13 @@
...
+
+// We have a shape being used by both input and output
+// so there will be an extra or "dangling" reference
+// to `GetFooOutput` from the input
+operation GetFoo {
+    input: GetFooOutput
+    output: GetFooOutput
+}
+
+structure GetFooOutput {}
```
srchase, line 13: Can the naming of this output structure get updated so it's clear that it will be removed in the `-after.smithy` file?

smithy-model/src/test/resources/software/amazon/smithy/model/transform/dedicated-input-output/handles-dangling-single-reference-before.smithy
```diff
@@ -0,0 +1,13 @@
+$version: "2.0"
+
+namespace smithy.example
+
+// We have a shape being used by both input and output
+// so there will be an extra or "dangling" reference
```
srchase, line 6: nit: This comment is a little confusing without the context of the PR and original issue. Maybe it can just describe what this test is showing, that a shared input/output shape gets cleaned up and replaced by the appropriate input and output shapes.

smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
```diff
@@ -153,36 +156,22 @@ private StructureShape renameShapeIfNeeded(
...
-            if (rel.getRelationshipType().getDirection() == RelationshipDirection.DIRECTED) {
-                totalDirectedEdges++;
-                if (rel.getRelationshipType() != expected) {
-                    return false;
-                }
-            }
-        }
-
-        return totalDirectedEdges == 1;
+        // edges to the expected shape
```
srchase, line 171: nit: missing period.

smithy-model/src/test/resources/software/amazon/smithy/model/transform/dedicated-input-output/handles-dangling-single-reference-before.smithy
milesziemer: I added a second test that verifies that the unused shape will be removed, I'll leave this open to make sure that looks ok

srchase (approved):

PR #1436, opened 2022-10-01
Add protocol tests for Content-Type and Content-Length for empty body

gosar:
https://github.com/awslabs/smithy-typescript/issues/552 is reported in smithy-typescript with proposed fix https://github.com/awslabs/smithy-typescript/pull/553. We are missing protocol tests that specify these behaviors. 

I've added a bunch of test cases. I've just written them quickly without building the package, so there may be typos, bad naming, bad documentation, mistakes in these. And some might be repetitive. But I wanted to get started on these, so pushing this PR out so someone can pick it up. I think having these would require the fix from https://github.com/awslabs/smithy-typescript/pull/553 and maybe more, and maybe across other smithy generators too.

For reference, earlier PR that added the forbidHeader cases in http-content-type, for GET only. The Content-Length assertions in these tests are based on https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2

>   A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body.  For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).  A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

** This could change existing behavior of some SDKs. Double check if this is all good.

I also found this related open PR in our repo - https://github.com/awslabs/smithy/pull/1399.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy
@@ -27,7 +27,13 @@ apply NoInputAndNoOutput @httpRequestTests([
         protocol: restJson1,
         method: "POST",
         uri: "/NoInputAndNoOutput",
-        body: ""
+        body: "",
+        headers: {
+            "Content-Length": "0"
+        },
+        forbidHeaders: [
+            "Content-Type"
+        ],
     },
     {
         id: "RestJsonNoInputAllowsAccept",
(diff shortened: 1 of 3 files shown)


gosar:
Closing this for now. We are tracking a backlog item to pick this up later.

PR #1442, opened 2022-10-11
Allow transforms to remove enum members

JordonPhillips:
This fixes a bug that was preventing enum members from being removable from transforms, such as removeShapesByTag.


srchase (approved):

JordonPhillips:
> what would happen if all of the members are removed

If all the members are removed, the shape should be removed since I'm making use of the code that does just that for unions/structures. I can add a test case tomorrow to cover that.

PR #1453, opened 2022-10-18
Fix resource identifier collision and validation

mtdowling:
When loading a model with multiple members bound to the same resource identifier, Smithy failed with an error in IdentifierBindingIndex due to Collectors.toMap:

```
Duplicate key X (attempted merging values y and z)
```

This change updates IdentifierBindingIndex to ignore collisions. However this means that colliding resource identifier bindings would be ignored (which is arguably worse than failing to load the model since it wouldn't ever flag these issues, and retroactively enforcing things is always challenging). To address this, I added validation to detect when members with the resourceIdentifier trait conflict with members of the same binding name or members that have the same resourceIdentifier trait.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/IdentifierBindingIndex.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/IdentifierBindingIndex.java
@@ -21,8 +21,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.SortedSet;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.MemberShape;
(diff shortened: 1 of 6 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/knowledge/IdentifierBindingIndex.java
```diff
@@ -155,25 +152,29 @@ private void processResource(ResourceShape resource, OperationIndex operationInd
     }
 
     private boolean isCollection(ResourceShape resource, ToShapeId operationId) {
-        return resource.getCollectionOperations().contains(operationId)
+        return resource.getCollectionOperations().contains(operationId.toShapeId())
```
mtdowling, line 155: Note: I changed this too because this was raising a warning because `operationId` is a `ToShapeId` but the set contains `ShapeId`.

kstich (approved):

PR #1454, opened 2022-10-18
Improve ChangedNullability diff evaluator

mtdowling:
This validator now does not emit trait change warnings when the required trait is changed, and instead relies on ChangedNullability (this matches how the box trait is handled too). When the required trait is removed from a member that targets a structure or union, a warning is emitted instead of an error to indicate that this is often allowed in Smithy generators that treat these members as always nullable regardless of the required trait.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedNullability.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedNullability.java
@@ -25,11 +25,13 @@ import software.amazon.smithy.diff.Differences;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NullableIndex;
 import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ClientOptionalTrait;
 import software.amazon.smithy.model.traits.DefaultTrait;
 import software.amazon.smithy.model.traits.InputTrait;
 import software.amazon.smithy.model.traits.RequiredTrait;
+import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
 
 /**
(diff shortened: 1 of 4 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedNullability.java
```diff
@@ -91,37 +93,60 @@ private ValidationEvent createError(
...
                     && !newMember.hasTrait(DefaultTrait.ID)
                     && !oldMember.hasTrait(ClientOptionalTrait.ID)) {
-                joiner.add("The @required trait was removed and not replaced with the @default trait.");
+                if (newTarget.isStructureShape() || newTarget.isUnionShape()) {
+                    severity = severity == null ? Severity.WARNING : severity;
+                    joiner.add("The @required trait was removed from a member that targets a " + newTarget.getType()
+                               + ". This is backward compatible in generators that always treat structures and "
+                               + "unions as optional (e.g., AWS generators)");
+                } else {
+                    joiner.add("The @required trait was removed and not replaced with the @default trait.");
```
syall, line 135: Does the message also need `@addedDefault`?

```suggestion
                    joiner.add("The @required trait was removed and not replaced with the @default and @addedDefault traits.");
```

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/ChangedNullabilityTest.java
```diff
@@ -305,4 +308,27 @@ public void roundTrippedV1ModelHasNoEvents() {
...
         assertThat(events, empty());
     }
+
+    @Test
+    public void specialHandlingForRequiredStructureAndUnionMembers() {
+        String originalModel =
+                "$version: \"2.0\"\n"
+                + "namespace smithy.example\n"
+                + "structure Baz {}\n"
+                + "structure Foo {\n"
```
syall, line 318: nit: add a member that targets a union shape

syall (changes requested):

syall (approved):

Issue #1456, opened 2022-10-21
Default trait applied to streaming in version 1


82marbag:
The [data](https://github.com/awslabs/smithy-rs/blob/e74ecf348b67134a8f0a9abde5cf5b6f964a1327/codegen-server-test/python/model/pokemon.smithy#L39) member in this v1 model is being applied the `@default` trait. 

In the v1 spec there is no mention of this behavior.

82marbag:
Related: https://github.com/awslabs/smithy/issues/1380

mtdowling:
You're seeing that in the in memory semantic model when using the Smithy libraries. The Smithy reference implementation automatically transforms all models to be valid for v2 semantics. It does that for anything that previously had a default zero value too. Tools that don't use or know about the default trait can simply ignore it. A code generator that supports both v1 and v2 should not ignore it. In Smithy v1, these members were implicitly considered to have a default by most server generators because it's not really possible to know if a streaming member was supplied but empty, so we decided to make that explicit in v2.

PR #1458, opened 2022-10-21
Clarify streaming trait values and semantics

### LINKED ISSUE #1380: Confusion with the streaming example in section 13.1 of IDL v2
[Link](https://awslabs.github.io/smithy/2.0/spec/streaming.html#data-streams). Example as of writing:

```
operation StreamingOperation {
    input: StreamingOperationInput
    output: StreamingOperationOutput
}

@input
structure StreamingOperationInput {}

@output
structure StreamingOperationOutput {
    @required
    streamId: String
    output: StreamingBlob
}

@streaming
blob StreamingBlob
```

I'm confused because:

1. Smithy IDL v2 introduces a [change](https://awslabs.github.io/smithy/2.0/guides/migrating-idl-1-to-2.html#add-the-default-trait-to-streaming-blobs) where members targeting streaming blobs now [MUST have `@default` or `@required`](https://awslabs.github.io/smithy/2.0/spec/streaming.html#streaming-trait). However, `output` in the example above does not have any of the two traits attached.
2. In IDL v1 there is a restriction on [the `streaming` trait](https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html#streaming-trait) that says: 

> If a service supports a protocol that supports the [httpPayload](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httppayload-trait) trait, any member that targets a streaming blob must also be marked as `@httpPayload`.

However, this restriction does not appear [in IDL v2](https://awslabs.github.io/smithy/2.0/spec/streaming.html#streaming-trait). Is that intentional? The Smithy library is still enforcing it for IDL v2 models:

> = Member `com.amazonaws.simple#AnOperationInput$streamingBlob` referencing @streaming shape `com.amazonaws.simple#StreamingBlob` must have the @httpPayload trait, as service `com.amazonaws.simple#SimpleService` has a protocol that supports @httpPayload.

3. The example is confusing because you can't send both a string (`streamId`) and a streaming blob (`output`) in the same HTTP message _body_ (?). I acknowledge that in another application-level protocol sending both pieces of data might be possible (which one?), but if you want to keep the example as-is, I think it would be better to call out immediately below it that it's not realizable in HTTP.

mtdowling:
* Clarify that streaming trait members need httpPayload when using an http binding protocol.
* Clarify that servers should interpret streaming members as having a default value of zero bytes.
* Mention this in the v1 docs too.
* Fixed a doc linking issue for the streaming docs where we were using the wrong casing.

Closes #1380
Addresses #1456

### DIFF
--- a/docs/source-1.0/spec/core/stream-traits.rst
+++ b/docs/source-1.0/spec/core/stream-traits.rst
@@ -36,12 +36,12 @@ Validation
     * The ``streaming`` trait is *structurally exclusive by target*, meaning
       only a single member of a structure can target a shape marked as
       ``streaming``.
-    * If a service supports a protocol that supports the :ref:`httpPayload-trait`,
-      any member that targets a ``streaming`` ``blob`` must also be marked as
-      ``@httpPayload``.
+    * If a service supports the :ref:`httpPayload-trait`, any member that
+      targets a ``streaming`` shape must also be marked as ``@httpPayload``.
 
 .. code-block:: smithy
 
+    @http(method: "GET", uri: "/streaming-operation")
     operation StreamingOperation {
         input: StreamingOperationInput,
         output: StreamingOperationOutput,
@@ -58,6 +58,20 @@ Validation
 
     @streaming
     blob StreamingBlob
+
+
+Deserializing streaming blobs
+=============================
+
+It is often impossible for a server to know if a zero-length payload was sent
+by a client, or if no payload was explicitly sent by a client. As a result,
+required streaming members SHOULD be interpreted by deserializers as if they
+have a default zero-length value when not present.
+
+.. note::
+    In Smithy IDL version 2, members that target a streaming blob MUST be
+    marked as ``required`` or be given an explicit default value of
+    zero bytes.
 
 
 .. smithy-trait:: smithy.api#requiresLength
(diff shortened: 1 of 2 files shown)


docs/source-1.0/spec/core/stream-traits.rst
```diff
@@ -60,6 +60,20 @@ Validation
     blob StreamingBlob
 
 
+Deserializing streaming blobs
+=============================
+
+It is often impossible for a server to know if a payload was sent by the
+client that contains zero bytes, or if no payload was explicitly sent by the
+client. As a result, required streaming members SHOULD be interpreted by
+deserializers as if they have a default value of zero bytes.
```
DavidOgunsAWS, line 69: What does it mean to have a default value of zero bytes? Zero bytes refers to a length of data, but an interpreted member value would have a non-zero length so does this mean interpret the value as if all (necessary) bytes for the type were set to zero? I think this needs to be re-worded to be unambiguous.

docs/source-2.0/spec/streaming.rst
```diff
@@ -483,24 +480,33 @@ Summary
...
+      targets a ``streaming`` shape must also be marked as ``@httpPayload``.
+
+
+Deserializing streaming blobs
+-----------------------------
+
+It is often impossible for a server to know if a payload was sent by the
+client that contains zero bytes, or if no payload was explicitly sent by the
+client. As a result, required streaming members SHOULD be interpreted by
+deserializers as if they have a default value of zero bytes.
```
DavidOgunsAWS, line 509: Same issue here.

DavidOgunsAWS (approved):

PR #1459, opened 2022-10-21
Fix dropping doc comments after default values

### LINKED ISSUE #1457: Default value syntax sugar causes subsequent trait to not be parsed
See https://github.com/adamthom-amzn/smithy/commit/325926992a1de516885326f22ca732c7266731da for a simple reproduction. The test fails as-is, and passes if you remove `= false` from the definition of `foo`.

kstich:
This commit fixes an issue where documentation comments were dropped for members that follow a member with a default value. Pending docs are only cleared when a default isn't encountered or the container is complete.

*Issue #, if available:*
Fixes #1457 

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
@@ -802,13 +802,10 @@ final class IdlModelParser extends SimpleParser {
 
             parseMember(op, definedMembers, memberParsing);
 
-            // Clears out any previously captured documentation
-            // comments that may have been found when parsing the member.
-            clearPendingDocs();
-
             ws();
         }
 
+        clearPendingDocs();
         expect('}');
     }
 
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -797,22 +797,29 @@ private void parseMembers(LoadOperation.DefineShape op, MemberParsing memberPars
...
-            // comments that may have been found when parsing the member.
-            clearPendingDocs();
+            // Having loaded a default value means we've run the `br()`
+            // parse that will pre-fetch documentation comments. That's
+            // fine, we should just avoid clearing them so the next
+            // member can have them applied properly.
+            if (!consumedPendingDocs) {
+                // Clears out any previously captured documentation
+                // comments that may have been found when parsing the member.
+                clearPendingDocs();
```
mtdowling, line 813: Can you just do this in parseMember? I don't think it's called by anything but this method. Then the big explanation can go away too.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelParser.java
```diff
@@ -797,22 +797,29 @@ private void parseMembers(LoadOperation.DefineShape op, MemberParsing memberPars
 
         while (!eof()) {
             if (peek() == '}') {
+                clearPendingDocs();
```
mtdowling, line 800: nit: I think would be easier to read if it was moved before the `expect('}')` at the end of the method after the break.

mtdowling (changes requested):

mtdowling (approved):

PR #1460, opened 2022-10-21
Correct precedence in NullableIndex and fix diffs

mtdowling:
Smithy-Diff was flagging acceptable changes of removing `@default(null)` from member when the trait had no effect because the target shape was already nullable. NullableIndex was also giving the `@default(null)` trait precedence over the `@required` trait which is incorrect. Required members are non-nullable, and the `@default(null)` trait doesn't override this. It simply means that there is no default value for the member (so coupled with the required trait, the member is still always present).

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedDefault.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedDefault.java
@@ -59,10 +59,11 @@ public final class ChangedDefault extends AbstractDiffEvaluator {
             DefaultTrait oldTrait,
             DefaultTrait newTrait
     ) {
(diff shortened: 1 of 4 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedDefault.java
```diff
@@ -61,8 +61,10 @@ private void validateChange(
     ) {
         // Removing the default trait is always a breaking change.
```
sugmanue, line 62: Nit, this comment also needs update.

sugmanue (approved):

sugmanue (approved):

syall (approved):

PR #1461, opened 2022-10-22
Add ability to lint based on word boundaries

mtdowling:
Add ability to lint based on word boundaries

This commit introduces a new syntax for matching words with the
ReservedWords linter and is intended to be used with the upcoming
sensitive words linter defined in #1364.

In addition to supporting wildcard searches ("*" prefix, suffix,
and contains), we now support matching based on word boundaries.

This commit introduces the "terms" keyword for word boundary
searches and adds dedicated abstractions for word boundary and
wildcard matching.

For example, "access key id" will match "AccessKeyId",
"access_key_id", "accessKeyID", "access_key_id100", "AccesKeyIDValue".
It will also match when all the words are concatenated together:
"accesskeyid". However, it will not match "accesskey_id" because it
only has two word boundaries ("accesskey" and "id").

### DIFF
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/WordBoundaryMatcher.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/WordBoundaryMatcher.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.linters;
+
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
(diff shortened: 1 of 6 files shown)


JordonPhillips (approved):

smithy-linters/src/main/java/software/amazon/smithy/linters/WordBoundaryMatcher.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private final Set<String> words = new HashSet<>();
+
+    // Use an LRU cache that stores up to 128 canonicalized search strings (e.g. don't parse "member" over and over).
+    private final Map<String, String> searchCache = new LinkedHashMap<String, String>(128, 1.0f, true) {
+        @Override
+        protected boolean removeEldestEntry(Map.Entry<String, String> eldest) {
+            return this.size() > 127;
+        }
+    };
```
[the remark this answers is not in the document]
mtdowling: It's not currently thread safe or used in multiple threads. I added documentation for this to the class.

smithy-linters/src/main/java/software/amazon/smithy/linters/WordBoundaryMatcher.java
```diff
@@ -0,0 +1,168 @@
...
+            addLowerCaseStringToBuilder(result, str, tokenStart, str.length() - tokenStart);
+        } else {
+            // Since the last segment is ignored, remove the trailing space.
+            result.setLength(result.length() - 1);
+        }
+
+        return result.toString();
+    }
+
+    private static void addLowerCaseStringToBuilder(StringBuilder result, String str, int start, int count) {
```
sugmanue, line 163: Small nit, I think that we can simplify the callers and a bit the function if we instead take the `endIndex` instead of the `count`.

mtdowling: I was mimicking what java.lang.String's constructor does here:

```
public String(char value[], int offset, int count) {
```

It's internal to the class, so probably not worth changing IMO

PR #1464, opened 2022-10-24
Update identifier ABNF and parser

syall:
*Issue #, if available:*

N/A

---

*Description of changes:*

Previously, the ABNF did not match the identifier parser.

This commit updates the ABNF to match current parsing behavior, and updates the identifier parser to avoid out of bounds exceptions.

Also, a set of tests are added for ONLY parsing identifiers rather than being mixed with namespace tests.

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
@@ -131,13 +131,17 @@ public final class ShapeId implements ToShapeId, Comparable<ShapeId> {
             return -1;
         }
 
-        // Parse the required identifier_start production.
+        // Parse the required IdentifierStart production.
         char startingChar = identifier.charAt(offset);
         if (startingChar == '_') {
-            while (identifier.charAt(offset) == '_') {
+            while (offset < identifier.length() && identifier.charAt(offset) == '_') {
                 offset++;
             }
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeId.java
```diff
@@ -131,21 +131,25 @@ private static int parseIdentifier(CharSequence identifier, int offset) {
...
         if (startingChar == '_') {
-            while (identifier.charAt(offset) == '_') {
+            while (offset < identifier.length() && identifier.charAt(offset) == '_') {
                 offset++;
             }
-            if (!ParserUtils.isValidIdentifierCharacter(identifier.charAt(offset))) {
+            if (offset == identifier.length()) {
+                return -1;
+            }
+            if (!ParserUtils.isAlphabetic(identifier.charAt(offset))
```
mtdowling, line 143: super minor nit, given this method is called thousands of times: move the current character to a local variable to avoid bounds checks multiple times with charAt.

e.g.:
```
            char current = identifier.charAt(offset);
            if (!ParserUtils.isAlphabetic(current) && !ParserUtils.isDigit(current)) {
                return -1;
            }
```

mtdowling (changes requested):

mtdowling (approved):

PR #1466, opened 2022-10-25
Add support for hierarchical event IDs

mtdowling:
Many of the validators in Smithy and Smithy Diff today emit events for various reasons. A single validator might event different events with different severities, and there is currently no way to suppress just one specific type of event for a given validator. For example, the ChangedNullability SmithyDiff validator emits events for various reasons, some of which different tools may choose to suppress. However, without inspecting the message of the event, there's no way to know what you're suppressing.

To give existing validators the ability to emit more granular validation events without breaking current validators, this change introduces event ID hierarchies and hierarchical suppression IDs using dots (.). For example, ChangedNullability.RemovedRequiredTrait is now emitted, and tools can check for "ChangedNullability" to match all events, or more granularly only match "ChangedNullability.RemovedRequiredTrait". This can also be used in suppressions so that existing Smithy validators can be updated to emit more granular events.

### DIFF
--- a/smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/TestHelper.java
+++ b/smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/TestHelper.java
@@ -24,7 +24,7 @@ import software.amazon.smithy.model.validation.ValidationEvent;
 public final class TestHelper {
     public static List<ValidationEvent> findEvents(List<ValidationEvent> events, String eventId) {
         return events.stream()
-                .filter(event -> event.getId().equals(eventId))
+                .filter(event -> event.containsId(eventId))
                 .collect(Collectors.toList());
     }
 
--- a/smithy-model/src/test/java/software/amazon/smithy/model/validation/suppressions/MetadataSuppressionTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/validation/suppressions/MetadataSuppressionTest.java
@@ -1,0 +1,65 @@
+package software.amazon.smithy.model.validation.suppressions;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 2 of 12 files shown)


smithy-model/src/test/java/software/amazon/smithy/model/validation/suppressions/MetadataSuppressionTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class MetadataSuppressionTest {
+
+    @Test
+    public void doesNotMatchWhenUsingDifferentNamespace() {
+        ObjectNode node = Node.objectNodeBuilder()
+                .withMember("id", "Foo")
+                .withMember("namespace", "smithy.example.nested")
+                .build();
+        Shape s = StringShape.builder().id("smithy.example#String").build();
```
kstich, line 26: Doesn't look like it's used, which means this test probably isn't doing a namespace comparison.

smithy-model/src/test/java/software/amazon/smithy/model/validation/suppressions/TraitSuppressionTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    public void doesNotMatchWhenUsingDifferentShape() {
+        Shape s = StringShape.builder()
+                .id("smithy.example#String")
+                .addTrait(SuppressTrait.builder().values(ListUtils.of("Foo")).build())
+                .build();
+        Suppression suppression = Suppression.fromSuppressTrait(s);
+        ValidationEvent event = ValidationEvent.builder()
+                .id("Foo")
```
kstich, line 29: Technically correct in that _no_ shape on the event is different from `s`, but clearer with a specified shape.

kstich (changes requested):

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/TestHelper.java
```diff
@@ -24,7 +24,7 @@
 public final class TestHelper {
     public static List<ValidationEvent> findEvents(List<ValidationEvent> events, String eventId) {
         return events.stream()
-                .filter(event -> event.getId().equals(eventId))
+                .filter(event -> event.getId().contains(eventId))
```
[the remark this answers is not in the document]
mtdowling: Oh whoops, I should have used containsId

kstich (approved):

Issue #1086, opened 2022-02-09
Questions about streaming and `Content-Length`


david-perez:
What should `Content-Length` be set to for streaming operations in the different AWS protocols?

The [restJson1 spec](https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html) does not prescribe anything. However, the protocol test suite contains this test:

```
{
    id: "RestJsonStreamingTraitsRequireLengthWithBlob",
    documentation: "Serializes a blob in the HTTP payload with a required length",
    protocol: restJson1,
    code: 200,
    body: "blobby blob blob",
    bodyMediaType: "application/octet-stream",
    headers: {
        "X-Foo": "Foo",
        "Content-Type": "application/octet-stream"
    },
    requireHeaders: [
        "Content-Length"
    ],
    params: {
        foo: "Foo",
        blob: "blobby blob blob"
    }
}
```

Under what circumstances should streaming operations include the `Content-Length` header?

Here it's clear it should be set to 16 (which by the way, the test does not assert, it only asserts the _presence_ of the header, why?), but for an arbitrary stream the client / server might not know in advance the size of the payload. I therefore think a better name for this test would be something like `RestJsonStreamingTraitsRequireLengthWithBlobOfKnownSize` (unless of course the header should **always** be set in restJson1, in which case I'd like to know what value it should be set to if the stream's size is not known in advance).

mtdowling:
This particular test targets a blob that has the `@requiresLength` trait: https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/streaming.smithy#L186-L196. This means that blob is streaming and not something you'd fit into memory, but the size must be known before streaming. Content-Length should be set when possible, and in this case it's possible because you know the length.

I agree the name could be misleading. Would you be able to send a PR to change the name? Also feel free to make the content-length check more specific rather than a presence check if you like.

gosar:
My understanding is that for the more general case where length won't be known in advance and the server is ok to support that case, the model just won't have `requiresLength`.

gosar:
Re: your follow ups - Yeah I think 1 and 2 is Yes. 4 also is Yes - why allow a service to not send Content-Length when it has modeled it as `requiresLength`. 3 I'm a bit unsure - I'm debating if it is too harsh for a client to reject a response if it could still process it anyways, though I think it would be better to fail since the consumer of the client code may care to know the size.

gosar:
[...]

PR #1471, opened 2022-10-27
Restrict `@requiresLength` request tests to clients

jjant:
*Issue #1086*

*Description of changes:*
As discussed in the linked issue, these tests only apply to clients.

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/streaming.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/streaming.smithy
@@ -123,7 +123,8 @@ apply StreamingTraitsRequireLength @httpRequestTests([
         params: {
             foo: "Foo",
             blob: "blobby blob blob"
-        }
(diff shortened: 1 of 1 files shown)


mtdowling:
IIRC, we want the server to validate that there's a content-length I thought, which means these wouldn't be client only.

jjant:
@mtdowling We do want to validate that, but that should be a [`httpMalformedRequestTests` suite](https://github.com/awslabs/smithy/issues/1472).
The request in these tests don't actually carry the `Content-Length` header, so we can't really test anything on the server.

JordonPhillips:
From the server perspective this tests that whatever you're doing under the hood with content length doesn't fail on a valid length. It can also test that you've exposed it somewhere. In the typescript ssdk, for example, this could be exposed in the context object they pass along. Or it could be a property of whatever you wrap the http request body with.

PR #1474, opened 2022-10-28
Lower resource identifier collisions to warning

kstich:
This commit updates the validation for resource identifier collisions to warn instead of error. When an explicit and implicit binding for the same resource identifier are found, the explicit binding is preferred.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceIdentifierBindingValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceIdentifierBindingValidator.java
@@ -18,7 +18,6 @@ package software.amazon.smithy.model.validation.validators;
 import static java.lang.String.format;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceIdentifierBindingValidator.java
```diff
@@ -66,7 +66,7 @@ private void validateResourceIdentifierTraits(Model model, List<ValidationEvent>
             Map<String, Set<String>> bindings = computePotentialStructureBindings(model, container);
             for (Map.Entry<String, Set<String>> entry : bindings.entrySet()) {
                 if (entry.getValue().size() > 1) {
-                    events.add(error(model.expectShape(container), String.format(
+                    events.add(warning(model.expectShape(container), String.format(
```
mtdowling, line 69: We should not allow explicit binding conflicts. Warning should only warn when explicit bindings are used and there's a viable implicit binding of the same name that isn't used (since it's confusing). Explicit binding collisions should be error.

e.g., this must be an error:

```
structure Foo {
    @required
    @resourceIdentifier("foo")
    a: String

    @required
    @resourceIdentifier("foo")
    b: String
}

```

mtdowling (changes requested):

mtdowling (approved):

PR #1508, opened 2022-11-21
Add warnings for private access on traits

milesziemer:
Previously, it was possible to refer to a trait marked with `@private` from another namespace. This commit makes doing this cause a warning. A warning is used instead of an error to maintain backward compatibility, but it will be upgraded to an error in the future.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PrivateAccessValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PrivateAccessValidator.java
@@ -22,6 +22,7 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.neighbor.NeighborProvider;
 import software.amazon.smithy.model.neighbor.Relationship;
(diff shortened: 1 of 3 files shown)


milesziemer:
I want to try adding a check to see if the invalid trait relationship is via a mixin. That way we can suggest adding `localTraits` on the mixin.

milesziemer:
Take this model for example:
```
namespace example.private

// Mixin with the private trait
@private
@mixin
structure PrivateMixin {}

// Trait uses mixin, inheriting the private trait
@trait
structure someTrait with [PrivateMixin] {}
```
```
namespace example.main

use example.private#someTrait

// Using the trait from another namespace
@someTrait
structure ExampleStructure {}
```
The warning message this PR would produce is:
```
[WARNING] example.main#ExampleStructure: This shape has an invalid trait relationship that targets a private shape, `example.private#someTrait`, in another namespace.
```
The issue I see here is that it isn't clear how `someTrait` is getting the `@private` trait. In a simple example like this, it's not a big deal, but in a more complex model it could be much more confusing.

Do we want to try to handle this case in this PR, or open a separate issue?

syall (approved):

mtdowling (approved):

PR #1516, opened 2022-11-28
Add validation of operationInput operationParams

rcoh:
*Description of changes:* This adds validation that `operationInputs` on the EndpointTestsTrait match the actual operation shape input.


smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
```diff
@@ -122,8 +122,9 @@ list OperationInputs {
 @unstable
 @private
 structure OperationInput {
-    /// The name of the service operation targeted by the test.
+    /// The service operation targeted by the test.
     @required
+    @idRef(failWhenMissing: true, selector: "operation")
```
alextwoods, line 127: does this work when the operationName is not the full shape name? Are we counting on a tool to transform this?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -28,6 +39,28 @@
...
+                    .flatMap(endpointTestCase -> endpointTestCase.getOperationInputs().stream())
+                    .flatMap(operationInput -> validateOperationInput(operationInput, service, model))
+                    .collect(Collectors.toList()));
+        }
+        return result;
+    }
+
+    private Stream<ValidationEvent> validateOperationInput(EndpointTestOperationInput testOperationInput, ServiceShape serviceShape, Model model) {
+        OperationIndex index = OperationIndex.of(model);
+        StructureShape input = index.expectInputShape(ShapeId.from(testOperationInput.getOperationName()));
```
alextwoods, line 56: Are you planning separate changes to validate the buildIn/clientParams (at least against the ruleset.  Both for required/type?)

alextwoods (approved):

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
rcoh: it doesn't—maybe it's better to do it manually so we can emit a warning instead

alextwoods: Yeah - I think a warning here might be preferable.

kstich: `idRef` is for absolute shape IDs, so just a name that doesn't resolve to a shape ID according to [the IDL resolution rules](https://smithy.io/2.0/spec/idl.html#relative-shape-id-resolution) won't work.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
kstich, line 56: Could this be a `ShapeId` from the start?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -28,6 +39,28 @@
 public final class EndpointTestsTraitValidator extends AbstractValidator {
     @Override
     public List<ValidationEvent> validate(Model model) {
-        return null;
+
+        List<ValidationEvent> result = new ArrayList<>();
+        for (ServiceShape service : model.getServiceShapesWithTrait(EndpointTestsTrait.class)) {
+            result.addAll(
```
kstich, line 45: This would be a lot cleaner with iteration instead of streaming operations, removing three intermediate streams and leading to more readable code. The below method could also return a collection instead of a `Stream`.

smithy-rules-engine/src/test/resources/software/amazon/smithy/rulesengine/traits/errorfiles/missing-required-value.smithy
```diff
@@ -0,0 +1,46 @@
...
+apply InvalidService @endpointTests({
+    "version": "1.0",
+    "testCases": [
+        {
+            "params": {
+                "stringFoo": "c d",
+                "boolFoo": true
+            },
+            "operationInputs": [{
+                                    "operationName": GetThing,
```
kstich, line 21: nit: These should only be one extra line indented.

kstich (changes requested):

kstich:
Closing due to age, having outstanding actions, and that this is going to be included in other work on `smithy-rules-engine`.

PR #1518, opened 2022-12-01
Tweak class caching in node (de)serializers.

milesziemer:
Prior to this PR, class caching mechanisms in DefaultNodeDeserializers
and DefaultNodeSerializers would cache classes the first time they were
loaded. If the class was loaded again using a new ClassLoader, and then
retrieved from the cache, there would be a mismatch in the loaded class
and the one retrieved from the cache because class equality is dependent
on the ClassLoader used. This caused class cast exceptions in the language
server.

Based on discussion with Michael offline, this PR introduces a modified
caching mechanism that associates a key-value pair with an instance
of a loaded class. When retrieving a value from the cache, the associated
class is also provided, and the cache will be invalidated if the associated
class is not referentially equal to the currently cached class. This modified
caching mechanism replaces caches in DefaultNodeSerializers and
DefaultNodeDeserializers.

Testing
------
- Test class for the new caching mechanism
- Added test case to ModelAssemblerTest for loading external jars with trait
implementations multiple times. The jar and model used for this test case
was taken from the language server.


milesziemer:
Opening this as a draft because it still needs a test added, but I still have to think about how exactly to test with loading classes in.

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
srchase, line 2: nit: 2022

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,59 @@
...
+
+package software.amazon.smithy.model.node;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
+
+/**
+ * Caches instances of loaded classes of type T using referential equality
+ * to compare keys.
```
mtdowling, line 24: super nit - we aren't really comparing keys, but the class of the returned value against the given class.

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,59 @@
...
+ * class. This ensures that the most recently loaded versions of classes are
+ * being used.
+ *
+ * @param <T> The type of the cached classes.
+ */
+final class IdentityClassCache<T> {
+
+    private final ConcurrentMap<String, T> cache;
+
+    IdentityClassCache() {
```
mtdowling, line 36: nit - you could just remove the ctor if there are no args and instantiate the cache on line 34

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,59 @@
...
+        cache = new ConcurrentHashMap<>();
+    }
+
+    /**
+     * Retrieves cached value associated with the given key, computing the
+     * value if the key is not present in the cache, or the cache is
+     * invalidated.
+     *
+     * @param forClass Key of value to retrieve from the cache.
```
mtdowling, line 45:
```suggestion
     * @param forClass The exact Class of value to retrieve from the cache.
```

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
milesziemer: Updated

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
milesziemer: Updated

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
milesziemer: True, the wording was kind of difficult. I updated various comments to make it more clear hopefully.

milesziemer:
Looking at this again and trying to update some of the comments, something started to seem not quite right. @mtdowling I was going off your suggestions, but either this implementation isn't exactly what we want, or it needs to be integrated into the node (de)serializers differently. See the diffs [DefaultNodeDeserializers:626](https://github.com/awslabs/smithy/pull/1518/files#diff-b97479d8946922af704f3004a94c1e7f7e147ff551690953364c254cf6665b16R624) and [DefaultNodeSerializers:259](https://github.com/awslabs/smithy/pull/1518/files#diff-4032099e1ccc0cea6c482715e50ed0b3b6ece8c0595f952d569f8eac27a873d6R259): the class that is being cached, ObjectMapper and ClassInfo respectively, do not have the same class as the keys of the cache, NodeType and Class<?> respectively. I added failing test case to show why this doesn't work.

Maybe we can use the previous implementation and this new one together, or we need to be able to use a different class as the key and value in IdentityClassCache.

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Retrieves cached instance of T associated with the given Class,
+     * computing the value to store if the given Class has no associated
+     * value, or if the cache is invalidated.
+     *
+     * @param forClass The exact Class of value to retrieve from the cache.
+     * @param supplierIfNotPresent Function to compute cached value.
+     * @return Cached value.
+     */
+    T get(Class<?> forClass, Supplier<? extends T> supplierIfNotPresent) {
```
milesziemer, line 44:
```suggestion
    T get(Class<? extends T> forClass, Supplier<? extends T> supplierIfNotPresent) {
```
I think we'd need something like this to make sure comparison to invalidate the cache works properly.

mtdowling: I don't think this will address your observation that the class key type != the cached return value. Instead, maybe track the original cache key in a Pair and then check the original cache key class vs the given class to invalidate the cache. Something like:

```
final class IdentityClassCache<T> {

    private final ConcurrentMap<String, Pair<Class<?>, T>> cache = new ConcurrentHashMap<>();

    T get(Class<?> forClass, Supplier<? extends T> supplierIfNotPresent) {
        return cache.compute(forClass.getName(), (name, current) -> {
            if (current == null || current.getLeft() != forClass) {
                return Pair.of(forClass, supplierIfNotPresent.get());
            } else {
                return current;
            }
        }).getRight();
    }
}
```

milesziemer: Yea this makes sense. My only question is, do we want to change the keys that are currently being used in [DefaultNodeDeserializers:626](https://github.com/awslabs/smithy/pull/1518/files#diff-b97479d8946922af704f3004a94c1e7f7e147ff551690953364c254cf6665b16L626)? It uses the node class *and* the target type to create a key, whereas these ideas would just be using the target type as a key. Is there a situation where a single node class may have multiple target types, with different ObjectCreators?

sugmanue (approved):

smithy-model/src/main/java/software/amazon/smithy/model/node/IdentityClassCache.java
```diff
@@ -0,0 +1,58 @@
...
+ * <p>Cached values are invalidated if the requested type is not referentially
+ * equal to the cached type. This ensures that the most recently loaded versions
+ * of classes are being used.
+ *
+ * @param <K> The type of the cache key.
+ * @param <V> The type of the cached value.
+ */
+final class IdentityClassCache<K, V> {
+
+    private final ConcurrentMap<K, Pair<Type, V>> cache = new ConcurrentHashMap<>();
```
kubukoz, line 36: Q: isn't this going to cause leaks if you keep reloading (e.g. in the language server)?

kubukoz: ah, I see, `compute` will overwrite the value using the same key - alright then

kubukoz (approved):

smithy-model/src/test/java/software/amazon/smithy/model/node/IdentityClassCacheTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test
+    public void invalidatesCacheForDifferentClassInstance() {
+        IdentityClassCache<String, String> cache = new IdentityClassCache<>();
+
+        String key = "key";
+        Type associatedClass = new Type() {};
+        String storedInstance = "value";
+        cache.getForClass("key", associatedClass, () -> storedInstance);
```
daddykotex, line 46: minor, I think you want to reuse `key` here, rather than `"key"` (they're the same right now, but... you know :P)

milesziemer: Good call. Didn't see this before I merged, opening a new PR for this.

kubukoz:
Hey @milesziemer, can you share an estimate of when this might be released? It'd be much appreciated, thanks!

PR #1521, opened 2022-12-01
Clarify that metadata has no namespace

### LINKED ISSUE #1352: metadata validation is stricter
Previously, we've had code like that:

```kotlin
metadata thing = [
  {
    value: ActorId,
  },
  {
    value: examples.movies#ActorId,
  }
]

namespace examples.movies

string ActorId
string MovieId
```

And it was fine. Both the short version (no namespace), and the fully qualified name resolved correctly. When updating to 1.23.0 (at least in my language-server), I see the following:

Screen Shot 2022-08-17 at 09 19 31

The specific error I see is: `SyntacticShapeIdTarget: Syntactic shape ID `ActorId` does not resolve to a valid shape ID: `smithy.api#ActorId`. Did you mean to quote this string? Are you missing a model file?`

Although this makes sense to me, I'm surprised that this shows up. Is this expected?

I think it's coming from: https://github.com/awslabs/smithy/commit/a93711da822bc0f0b5730ea340efdf473b9d0023

In all cases, I'm not asking for a fix (using the fully qualified name works), maybe just some explanation. This can serve as documentation for people who see this error.

mtdowling:
Closes #1352

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source-1.0/spec/core/idl.rst
+++ b/docs/source-1.0/spec/core/idl.rst
@@ -385,6 +385,28 @@ The following example defines metadata in the model:
             }
         }
 
+Metadata is not defined within a namespace. Unquoted object property values
+are considered :ref:`syntactic shape IDs <syntactic-shape-ids>` and resolve
+to the prelude namespace, ``smithy.api``.
+
+The following Smithy IDL model:
+
+.. code-block:: smithy
+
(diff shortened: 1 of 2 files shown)


sugmanue:
Today I learned.

PR #1523, opened 2022-12-05
Add Maven dependency resolution to the CLI

mtdowling:
This commit adds the ability to resolve dependencies in the Smithy CLI. This allows end users to author models, generate code, and other activities from the Smithy CLI without needing to use tools like Gradle or Maven. This lowers the up-front learning curve of most Smithy use cases to author models and generate code.

Writing Java library code, model build plugins, or transforms is out of scope of the Smithy CLI. Such use cases require the use of tools like Gradle or Maven.

To support dependency resolution in the CLI, this commit also introduces the ``--version`` option to display the current version.

The ``smithy clean`` command was added since it was very useful to test dependency resolution by removing the classpath-cache file to force them to be re-resolved.

Using the Smithy CLI directly means that tools like Gradle aren't automatically detecting the model directory and adding sources. To account for this, the `sources` property was added to smithy-build.json. It acts like `imports`, but the models contained in `sources` are considered part of the model being built rather than just a dependeny (i.e., the sources plugin will contain these models).

To actually test maven resolution, this commit also adds integration tests for the CLI (./gradlew :smithy-cli:integ).

Other changes:

1. The use of @SmithyUnstableApi and @SmithyInternalApi was removed from all smithy-cli classes because there is a blanket @SmithyUnstableApi annotation on the entire package, making the individual usage unnecessary.
2. Moved --severity setting from StandardOptions to BuildOptions. This setting is only needed when building Smithy models.
3. Made Smithy CLI commands and related options classes package-private since they aren't referenced outside of the commands package.
4. This commit also prints out more useful exceptions when --stacktrace is on since wrapping things in a DependencyRunnably obscured the underlying issue.
5. Command line options related to model discovery and discovery classpath are now hidden from help output. They are much less relevant now that dependency resolution is baked into the CLI, and may not even be necessary if Smithy's Ant and Gradle integration aren't using it (need to verify).

(Note: this commit supersedes #1486 because that commit kept returning 500 errors from GitHub's UI)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,11 @@ jobs:
           distribution: 'zulu'
 
       - name: Clean and build
-        run: ./gradlew clean build -Plog-tests
+        run: ./gradlew clean build -Plog-tests --stacktrace
+
+      - name: CLI integration tests
+        if: matrix.java >= 17
+        run: ./gradlew :smithy-cli:integ -Plog-tests --stacktrace
 
       - uses: actions/upload-artifact@v3
         if: failure()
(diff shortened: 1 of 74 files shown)


mtdowling:
More 500 errors in the UI. GH does not like rebases.

PR #1526, opened 2022-12-05
Add Maven dependency resolution to the CLI

mtdowling:
This commit adds the ability to resolve dependencies in the Smithy CLI. This allows end users to author models, generate code, and other activities from the Smithy CLI without needing to use tools like Gradle or Maven. This lowers the up-front learning curve of most Smithy use cases to author models and generate code.

Writing Java library code, model build plugins, or transforms is out of scope of the Smithy CLI. Such use cases require the use of tools like Gradle or Maven.

To support dependency resolution in the CLI, this commit also introduces the ``--version`` option to display the current version.

The ``smithy clean`` command was added since it was very useful to test dependency resolution by removing the classpath-cache file to force them to be re-resolved.

Using the Smithy CLI directly means that tools like Gradle aren't automatically detecting the model directory and adding sources. To account for this, the `sources` property was added to smithy-build.json. It acts like `imports`, but the models contained in `sources` are considered part of the model being built rather than just a dependeny (i.e., the sources plugin will contain these models).

To actually test maven resolution, this commit also adds integration tests for the CLI (./gradlew :smithy-cli:integ).

Other changes:

1. The use of @SmithyUnstableApi and @SmithyInternalApi was removed from all smithy-cli classes because there is a blanket @SmithyUnstableApi annotation on the entire package, making the individual usage unnecessary.
2. Moved --severity setting from StandardOptions to BuildOptions. This setting is only needed when building Smithy models.
3. Made Smithy CLI commands and related options classes package-private since they aren't referenced outside of the commands package.
4. This commit also prints out more useful exceptions when --stacktrace is on since wrapping things in a DependencyRunnably obscured the underlying issue.
5. Command line options related to model discovery and discovery classpath are now hidden from help output. They are much less relevant now that dependency resolution is baked into the CLI, and may not even be necessary if Smithy's Ant and Gradle integration aren't using it (need to verify).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source-2.0/guides/building-models/build-config.rst
+++ b/docs/source-2.0/guides/building-models/build-config.rst
@@ -41,14 +41,15 @@ The configuration file accepts the following properties:
         added as sources. Sources are relative to the configuration file.
     * - imports
       - ``[string]``
-      - Provides a list of relative imports to combine into a single model.
-        Imports are a kind of local dependency: they aren't considered part of
-        model being built, but are required to build the model. When a
-        directory is encountered, all files in the entire directory tree are
-        imported. Note that build systems MAY choose to rely on other
-        mechanisms for importing models and forming a composite model.
-        Imports defined at the top-level are used in every projection. Imports
-        are relative to the configuration file.
+      - Provides a list of model files and directories to load when validating
+        and building the model. Imports are a local dependency: they are not
+        considered part of model package being built, but are required to build
+        the model package. Models added through ``imports`` are not present in
+        the output of the built-in ``sources`` plugin.
+
+        When a directory is encountered, all files in the entire directory
+        tree are imported. Imports defined at the top-level are used in every
+        projection. Imports are relative to the configuration file.
     * - projections
       - ``map<string, object>``
       - A map of projection names to projection configurations.
@@ -162,7 +163,7 @@ Maven picks the highest version of each project that satisfies all the hard
 requirements of the dependencies on that project. If no version satisfies
 all the hard requirements, dependency resolution fails.
 
-The following table defines version requirement syntax as defined in
+The following table demonstrates version requirement syntax as defined in
 the `official Maven documentation`_:
 
 .. list-table:: Dependency version syntax
@@ -231,7 +232,7 @@ the following configuration:
             file. Instead, use :ref:`environment variables <build_envars>` to
             keep credentials out of source control.
 
-.. code-block::
+.. code-block:: json
 
     {
         "version": "1.0",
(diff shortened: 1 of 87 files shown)


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/WarmupCommand.java
```diff
@@ -0,0 +1,97 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        return 0;
+    }
+
+    private void resolve(DependencyResolver resolver) {
+        resolver.addRepository(MavenRepository.builder().url("https://repo.maven.apache.org/maven2").build());
+        resolver.addDependency("software.amazon.smithy:smithy-model:1.23.0");
```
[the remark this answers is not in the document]
mtdowling: I'll do `SmithyCli.getVersion()` so that the dependency is resolved, but ignored, thereby not adding extraneous files to the CDS archive

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/CleanCommand.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "clean";
+    }
+
+    @Override
+    public String getSummary() {
+        return "Removes Smithy build artifacts";
```
JordonPhillips, line 49: nit
```suggestion
        return "Removes Smithy build artifacts.";
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/WarmupCommand.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "warmup";
+    }
+
+    @Override
+    public String getSummary() {
+        return "Creates caches for faster subsequent executions";
```
JordonPhillips, line 48:
```suggestion
        return "Creates caches for faster subsequent executions.";
```

JordonPhillips (changes requested):

docs/source-2.0/guides/building-models/build-config.rst
```diff
@@ -33,14 +33,22 @@ The configuration file accepts the following properties:
...
-      - ``string[]``
+      - ``[string]``
       - Provides a list of relative imports to combine into a single model.
-        When a directory is encountered, all files and all files within all
-        subdirectories are imported. Note that build systems MAY choose to rely
-        on other mechanisms for importing models and forming a composite model.
-        These imports are used in every projection. Note: imports are relative
-        to the configuration file.
+        Imports are a kind of local dependency: they aren't considered part of
+        model being built, but are required to build the model. When a
```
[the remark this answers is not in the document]
mtdowling: I'll make this more clear:

>Provides a list of relative imports to load when validating and building the model. Imports are a kind of local dependency: they are not considered part of model package being built, but are required to build the model package. Models added through ``imports`` are not present in the output of the ``sources`` plugin.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/CleanCommand.java
mtdowling: We are inconsistent with ending in a period, so I'll add a period to all command summaries.

JordonPhillips (approved):

sugmanue (approved):

PR #1527, opened 2022-12-07
Extend validation event ids to make them unique

rchache:
*Description of changes:*
We have a need to record and store "approvals" on a per-smithy-violation basis. The idea is that once approved, the same violation will be hidden as long as an "approval" can be matched to it. However, we realized an issue where ValidationEvents do not currently contain the granularity we need. Currently, some linters/validators raise multiple violations on the same shape, and the only difference is the message/source location (Enum casing validation or unstable trait validation for ex). However if we key or hash on the message/source location, we risk being broken if the message or source location changes slightly. 

To accomplish this goal, we can use hierarchical eventIds to pack distinguishing info into the eventId.


mtdowling:
Thanks! We'll go through and review each of these. I do think though that the JMESPath related IDs are too granular and won't be that useful considering it's fine to change JMESPath expressions in waiters as APIs evolve, bugs are fixed, or they're simplified. Let's look out for other cases like that too in this PR.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/operation/input-output-traits/invalid-multiple-operations-same-shape.errors
```diff
@@ -1,4 +1,4 @@
-[WARNING] smithy.example#GetFoo2: The input of this operation should target a shape that starts with the operation's name, 'GetFoo2', but the targeted shape is `smithy.example#GetFooInput` | OperationInputOutputName
-[WARNING] smithy.example#GetFoo2: The output of this operation should target a shape that starts with the operation's name, 'GetFoo2', but the targeted shape is `smithy.example#GetFooOutput` | OperationInputOutputName
+[WARNING] smithy.example#GetFoo2: The input of this operation should target a shape that starts with the operation's name, 'GetFoo2', but the targeted shape is `smithy.example#GetFooInput` | OperationInputOutputName.input.smithy.example#GetFooInput
```
mtdowling, line 1: I don't think the targeted input shape ID is useful in the event ID for this validator, so I'd say remove it.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-deeply-nested.errors
```diff
@@ -1,2 +1,2 @@
-[WARNING] ns.foo#DeeplyNested: paginated trait `items` contains a path with more than two parts, which can make your API cumbersome to use | PaginatedTrait
-[WARNING] ns.foo#DeeplyNested: paginated trait `outputToken` contains a path with more than two parts, which can make your API cumbersome to use | PaginatedTrait
+[WARNING] ns.foo#DeeplyNested: paginated trait `items` contains a path with more than two parts, which can make your API cumbersome to use | PaginatedTrait.items.DEEPLY_NESTED
```
mtdowling, line 1: Why DEEPLY_NESTED and not DeeplyNested? I'd also think it might be better to do this as `PaginatedTrait.DeeplyNested.items` so you could ignore all deeply nested violations for a single event.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-invalid.errors
```diff
@@ -5,7 +5,7 @@
 [ERROR] ns.foo#Invalid3: paginated trait `outputToken` member `OutputNotString` targets a integer shape, but must target one of the following: [`map`, `string`] | PaginatedTrait
 [ERROR] ns.foo#Invalid4: paginated trait `inputToken` member `nextToken` must not be required | PaginatedTrait
 [ERROR] ns.foo#Invalid4: paginated trait `outputToken` member `nextToken` must not be required | PaginatedTrait
-[WARNING] ns.foo#Invalid4: paginated trait `pageSize` member `pageSize` should not be required | PaginatedTrait
+[WARNING] ns.foo#Invalid4: paginated trait `pageSize` member `pageSize` should not be required | PaginatedTrait.pageSize.pageSize.SHOULD_NOT_BE_REQUIRED
```
mtdowling, line 8: Similar to the other paginated trait validator, I'd like to see the category first, and `pageSize` shouldn't be repeated:
`PaginatedTrait.ShouldNotBeRequired.pageSize`

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-map-tokens.errors
```diff
@@ -1,2 +1,2 @@
-[DANGER] ns.foo#MapTokens: paginated trait `inputToken` member `MapTokensInput` targets a map shape, but this is not recommended. One of [`string`] SHOULD be targeted. | PaginatedTrait
-[DANGER] ns.foo#MapTokens: paginated trait `outputToken` member `MapTokensOutput` targets a map shape, but this is not recommended. One of [`string`] SHOULD be targeted. | PaginatedTrait
+[DANGER] ns.foo#MapTokens: paginated trait `inputToken` member `MapTokensInput` targets a map shape, but this is not recommended. One of [`string`] SHOULD be targeted. | PaginatedTrait.inputToken.MapTokensInput.WRONG_SHAPE_TYPE.map
```
mtdowling, line 1: The input shape and the all the additional information beyond the fact that it's inputToken can be removed. They don't add to the uniqueness of the event.

`PaginatedTrait.WrongShapeType.inputToken`

This matches the format of the others, and you don't really need to know any more information.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/trait-value/extra-trait-property.errors
```diff
@@ -1 +1 @@
-[WARNING] smithy.example#MyInteger: Error validating trait `range`: Invalid structure member `maxx` found for `smithy.api#range` | TraitValue
+[WARNING] smithy.example#MyInteger: Error validating trait `range`: Invalid structure member `maxx` found for `smithy.api#range` | TraitValue.smithy.api#range.maxx
```
mtdowling, line 1: Maybe we should update the hierarchical suppression matching to also treat `#` like `.` so you could suppress an entire namespace?

smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/errorfiles/http-checksum-header-conflicts.errors
```diff
@@ -1,9 +1,9 @@
...
-[WARNING] smithy.example#HeaderConflicts: The `httpHeader` binding of `x-amz-checksum-crc32` on `HeaderConflictsOutput` starts with the prefix `x-amz-checksum-` used by the `httpChecksum` trait. | HttpChecksumTrait
-[DANGER] smithy.example#HeadersConflicts: The `httpPrefixHeaders` binding of `HeadersConflictError` uses the prefix `x-amz-checksum-` that conflicts with the prefix `x-amz-checksum-` used by the `httpChecksum` trait. | HttpChecksumTrait
-[DANGER] smithy.example#HeadersConflicts: The `httpPrefixHeaders` binding of `HeadersConflictsInput` uses the prefix `x-amz-checksum-` that conflicts with the prefix `x-amz-checksum-` used by the `httpChecksum` trait. | HttpChecksumTrait
-[DANGER] smithy.example#HeadersConflicts: The `httpPrefixHeaders` binding of `HeadersConflictsOutput` uses the prefix `x-amz-checksum-` that conflicts with the prefix `x-amz-checksum-` used by the `httpChecksum` trait. | HttpChecksumTrait
+[WARNING] smithy.example#HeaderConflicts: The `httpHeader` binding of `x-amz-checksum-crc32` on `HeaderConflictError` starts with the prefix `x-amz-checksum-` used by the `httpChecksum` trait. | HttpChecksumTrait.httpHeader.HeaderConflictError.x-amz-checksum-crc32
```
mtdowling, line 4: 1. The pattern isn't consistent - use UpperCamel for all sections: `HttpChecksumTrait.HttpHeader.HeaderConflictError`
2. You probably don't need to say `HeaderConflictError`. Could this just be shortened to `HttpChecksumTrait.HeaderConflict`?
3. Do we need to sanitize the headers we put into the event ID at all, or does that not matter? I guess it doesn't matter since "." can just be ignored.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/input-output-structure-reuse.errors
```diff
@@ -1,2 +1,2 @@
-[WARNING] smithy.example#GetFooInput: This structure is the input of `smithy.example#GetFoo`, but it is not marked with the @input trait. The @input trait gives operations more flexibility to evolve their top-level input members in ways that would otherwise be backward incompatible. | InputOutputStructureReuse
-[WARNING] smithy.example#GetFooOutput: This structure is the output of `smithy.example#GetFoo`, but it is not marked with the @output trait. | InputOutputStructureReuse
+[WARNING] smithy.example#GetFooInput: This structure is the input of `smithy.example#GetFoo`, but it is not marked with the @input trait. The @input trait gives operations more flexibility to evolve their top-level input members in ways that would otherwise be backward incompatible. | InputOutputStructureReuse.INPUT.smithy.example#GetFoo
+[WARNING] smithy.example#GetFooOutput: This structure is the output of `smithy.example#GetFoo`, but it is not marked with the @output trait. | InputOutputStructureReuse.OUTPUT.smithy.example#GetFoo
```
mtdowling, line 2: I don't see how the shape ID of the operation it's used by is that helpful here. The intent of the validator is that shapes used as input or output should have the corresponding input/output trait, so suppressing specific combinations of these won't help.

Can we also do `InputOutputStructureReuse.Output` and `InputOutputStructureReuse.Input` instead of using all caps?

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-return-types.errors
```diff
@@ -1,4 +1,4 @@
...
-[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 1: Waiter acceptors with a stringEquals comparator must return a `string` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem
-[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 2: Waiter acceptors with a allStringEquals comparator must return a `array` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem
-[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 3: Waiter acceptors with a anyStringEquals comparator must return a `array` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem
+[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 0: Waiter acceptors with a booleanEquals comparator must return a `boolean` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem.Invalid1.0.booleanEquals.number
```
mtdowling, line 1: Like the other jmespath based validators -- it's too specific to put the path into the event ID. These will be just as fickle as suppressing based on messages since they can and do change.

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-jmespath-syntax.errors
```diff
@@ -1,3 +1,3 @@
 [ERROR] smithy.example#A: Waiter `Invalid1`, acceptor 0: Invalid JMESPath expression (||): Syntax error | WaitableTrait
 [ERROR] smithy.example#A: Waiter `Invalid2`, acceptor 0: Problem found in JMESPath expression (length(`10`)): length function argument 0 error | WaitableTrait
-[DANGER] smithy.example#A: Waiter `Invalid2`, acceptor 0: Waiter acceptors with a booleanEquals comparator must return a `boolean` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem
+[DANGER] smithy.example#A: Waiter `Invalid2`, acceptor 0: Waiter acceptors with a booleanEquals comparator must return a `boolean` type, but this acceptor was statically determined to return a `number` type. | WaitableTraitJmespathProblem.Invalid2.0.booleanEquals.number
```
mtdowling, line 3: Same comment - Let's not put JMESPath into event IDs. Suppressing these events is super dangerous anyways and not something your team would ever want to do.

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-errorType.errors
```diff
@@ -1 +1 @@
-[WARNING] smithy.example#A: Waiter `A`, acceptor 0: errorType 'Nope' not found on operation. This operation defines the following errors: [smithy.example#OhNo] | WaitableTraitInvalidErrorType
+[WARNING] smithy.example#A: Waiter `A`, acceptor 0: errorType 'Nope' not found on operation. This operation defines the following errors: [smithy.example#OhNo] | WaitableTraitInvalidErrorType.A.0.Nope
```
mtdowling, line 1: Same comment - Let's not put JMESPath into event IDs

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/emits-danger-and-warning-typechecks.errors
```diff
@@ -1,3 +1,3 @@
-[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 0: Problem found in JMESPath expression (`10`.foo): Object field 'foo' extraction performed on number (1:6) | WaitableTraitJmespathProblem
-[DANGER] smithy.example#A: Waiter `Invalid2`, acceptor 0: Waiter acceptors with a booleanEquals comparator must return a `boolean` type, but this acceptor was statically determined to return a `null` type. | WaitableTraitJmespathProblem
+[DANGER] smithy.example#A: Waiter `Invalid1`, acceptor 0: Problem found in JMESPath expression (`10`.foo): Object field 'foo' extraction performed on number (1:6) | WaitableTraitJmespathProblem.Invalid1.0
```
mtdowling, line 1: Same comment - Let's not put JMESPath into event IDs.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
```diff
@@ -1,13 +1,13 @@
...
-[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {document/whitelist_doc_key_violation_1} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms
-[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {collection/2/key} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms
-[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {collection/1/blacklist_key} contains a non-inclusive term `blacklist`. Consider using one of the following terms instead: `denyList` | NoninclusiveTerms
-[WARNING] ns.foo#MyUnionTrait$int_whitelist: Member shape uses a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms
+[WARNING] ns.foo#MyMasterService: Service shape uses a non-inclusive term `Master`. Consider using one of the following terms instead: `Main`, `Parent`, `Primary` | NoninclusiveTerms.SHAPE.Master
```
mtdowling, line 1: Can we do `Shape` and `Trait` instead of all caps? If you need the word in the event too, you should normalize the casing to be either all lower or UpperCamelCase.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
```diff
@@ -1,13 +1,13 @@
...
+[WARNING] ns.foo#BlackListThings: Operation shape uses a non-inclusive term `BlackList`. Consider using one of the following terms instead: `DenyList` | NoninclusiveTerms.SHAPE.BlackList
+[WARNING] ns.foo#MyWhitelistTrait: Structure shape uses a non-inclusive term `Whitelist`. Consider using one of the following terms instead: `AllowList` | NoninclusiveTerms.SHAPE.Whitelist
+[WARNING] ns.foo#AInput$master_member_name: Member shape uses a non-inclusive term `master`. Consider using one of the following terms instead: `main`, `parent`, `primary` | NoninclusiveTerms.SHAPE.master
+[WARNING] ns.foo#NestedTraitStructure$master_key_violation: Member shape uses a non-inclusive term `master`. Consider using one of the following terms instead: `main`, `parent`, `primary` | NoninclusiveTerms.SHAPE.master
+[WARNING] ns.foo#AInput$foo: 'ns.foo#MySimpleValueTrait' trait has a value that contains a non-inclusive term `slave`. Consider using one of the following terms instead: `child`, `clone`, `replica`, `secondary` | NoninclusiveTerms.TRAIT.ns.foo#MySimpleValueTrait.slave
```
mtdowling, line 6: I'd expect the term to appear after the qualifier, just like on Shape. So `NoninclusiveTerms.Trait.<Term>.ns.foo#MySimpleValueTrait`

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
```diff
@@ -1,13 +1,13 @@
...
+[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {nested/master_key_violation} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms.TRAIT.ns.foo#MyWhitelistTrait.nested/master_key_violation.whitelist
+[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {string_value} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms.TRAIT.ns.foo#MyWhitelistTrait.string_value.whitelist
+[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {document/whitelist_doc_key_violation_1} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms.TRAIT.ns.foo#MyWhitelistTrait.document/whitelist_doc_key_violation_1.whitelist
+[WARNING] ns.foo#AOutput: 'ns.foo#MyWhitelistTrait' trait value at path {collection/2/key} contains a non-inclusive term `whitelist`. Consider using one of the following terms instead: `allowList` | NoninclusiveTerms.TRAIT.ns.foo#MyWhitelistTrait.collection/2/key.whitelist
```
mtdowling, line 11: The path to the instance of the term seems way too specific to me. I also don't think the pathing here matches up with pathing used elsewhere (/ vs .). I think we could get away with just the trait shape ID and not need the entire path to the term.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/annotation-trait-extra-properties.errors
```diff
@@ -1 +1 @@
-[WARNING] ns.foo#Baz: Error validating trait `sensitive`: Invalid structure member `test` found for `smithy.api#sensitive` | TraitValue
+[WARNING] ns.foo#Baz: Error validating trait `sensitive`: Invalid structure member `test` found for `smithy.api#sensitive` | TraitValue.smithy.api#sensitive.test
```
mtdowling, line 1: I'm not sure how useful this one is -- do you have a use case for suppressing this? How does pathing work here for nested things? Can you just include the shape ID and omit the path?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/deprecated-shape/deprecated-shape.errors
```diff
@@ -1,4 +1,4 @@
-[WARNING] smithy.example#Test$foo1: Member targets a deprecated shape, smithy.example#Foo1 | DeprecatedShape
-[WARNING] smithy.example#Test$foo2: Member targets a deprecated shape, smithy.example#Foo2 (since 2.0) | DeprecatedShape
-[WARNING] smithy.example#Test$foo3: Member targets a deprecated shape, smithy.example#Foo3. hello (since 2.0) | DeprecatedShape
-[WARNING] smithy.example#Test: Applies a deprecated mixin, smithy.example#DeprecatedMixin | DeprecatedShape
+[WARNING] smithy.example#Test$foo1: Member targets a deprecated shape, smithy.example#Foo1 | DeprecatedShape.MEMBER_TARGET.smithy.example#Foo1
```
mtdowling, line 1: Does the kind of member relationship matter here? Seems like it could just refer to the deprecated target instead. Exposing MEMBER_TARGET et al seems like it's exposing too much of an implementation detail in the ID.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
```diff
@@ -1,7 +1,8 @@
...
+[DANGER] ns.foo#M: This operation uses the `DELETE` method in the `http` trait, but the following members are sent as part of the payload of the request: `payload`. These members are sent as part of the payload because they are not explicitly configured to be sent in headers, in the query string, or in a URI segment. Many HTTP clients do not support payloads with DELETE requests. Consider binding these members to other parts of the HTTP request such as query string parameters using the `httpQuery` trait, headers using the `httpHeader` trait, or URI segments using the `httpLabel` trait. | HttpMethodSemantics.DELETE.payload
+[WARNING] ns.foo#G: This operation uses the `POST` method in the `http` trait, but is marked with the readonly trait | HttpMethodSemantics.POST.smithy.api#readonly
```
mtdowling, line 5: Few thoughts here:
1. You don't need the HTTP method in the event ID. That's specific to the operation already and doesn't vary across events for the same shape.
2. I think something like `HttpMethodSemantics.MissingReadonlyTrait` would be more readable. Same comment as `HttpMethodSemantics.MissingIdempotentTrait`.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
```diff
@@ -1,7 +1,8 @@
...
+[DANGER] ns.foo#L: This operation uses the `GET` method in the `http` trait, but the following members are sent as part of the payload of the request: `morePayload`. These members are sent as part of the payload because they are not explicitly configured to be sent in headers, in the query string, or in a URI segment. Many HTTP clients do not support payloads with GET requests. Consider binding these members to other parts of the HTTP request such as query string parameters using the `httpQuery` trait, headers using the `httpHeader` trait, or URI segments using the `httpLabel` trait. | HttpMethodSemantics.GET.morePayload
```
mtdowling, line 3: I think this should be `HttpMethodSemantics.UnexpectedPayload` with no members or anything. Those don't help you identify the error here and there could be multiple members for a single event. This also was missing a qualifier for the subtype of the event.

You could also combine this event ID with the one on line 1 (no need to separate these subtypes of explicit vs implicit payloads).

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
```diff
@@ -1,7 +1,8 @@
...
+[WARNING] ns.foo#G: This operation uses the `POST` method in the `http` trait, but is marked with the readonly trait | HttpMethodSemantics.POST.smithy.api#readonly
+[WARNING] ns.foo#H: This operation uses the `DELETE` method in the `http` trait, but is not marked with the idempotent trait | HttpMethodSemantics.DELETE.smithy.api#idempotent
+[WARNING] ns.foo#I: This operation uses the `GET` method in the `http` trait, but is not marked with the readonly trait | HttpMethodSemantics.GET.smithy.api#readonly
+[WARNING] ns.foo#Options: OPTIONS requests are typically used as part of CORS and should not be modeled explicitly. They are an implementation detail that should not appear in generated client or server code. Instead, tooling should use the `cors` trait on a service to automatically configure CORS on clients and servers. It is the responsibility of service frameworks and API gateways to automatically manage OPTIONS requests. For example, OPTIONS requests are automatically created when using Smithy models with Amazon API Gateway. | HttpMethodSemantics.OPTIONS
```
mtdowling, line 8: Even though it's all caps, this instance is good IMO

mtdowling (changes requested):

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-deeply-nested.errors
rchache: In general, I wanted to use constant/enum casing for these sort of things because in some of the validators, it appears ambiguous at first glance if it is supposed to be a shape Id or a signal-type thing as in the case of DEEPLY_NESTED. I also thought that it was better to scope any suppressions to each trait, but I'll change the ordering on the next commit

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/trait-value/extra-trait-property.errors
rchache: I guess that would be smithy team's call. Trebuchet won't be doing any partial matching on these eventIds, only exact matching

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/input-output-structure-reuse.errors
rchache: It's necessary for uniqueness. The same structure can be the input or output for multiple operations (bad design but too late to fix for many shipped services), and would raise a new violation for each of those. Also we don't want to aggregate that case or any similar ones, because then it would hide any future additions to that aggregation

I'll fix the casing after syncing with you about it throughout

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-return-types.errors
rchache: Hmm, I also don't like having the path in the eventId, but I couldn't think of another way to maintain uniqueness. I was under the impression that different JMES paths (for the same waiter and acceptor) could potentially have the exact same issue and therefore have identical eventIds

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
rchache: When you say normalize, do you mean including Shape/Trait? or just normalize the noninclusive terms?

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
rchache: Wouldn't that make it harder to do hierarchical suppressions if someone wanted to?

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
rchache: Actually as shown in this testcase, if we omitted the path then line 10 and line 11 would have duplicate eventIds

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/annotation-trait-extra-properties.errors
rchache: Well we need the path because one trait could have multiple invalid structure members, each as a different violation

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
mtdowling: I don't see how. I just am pointing out that the structure of the event is different for trait vs shape and I don't see a reason for that.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
mtdowling: Yeah just the term "Foo" vs "foo" shouldn't change the event ID

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-return-types.errors
mtdowling: Let's just remove it for now

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-deeply-nested.errors
mtdowling: I think a good pattern for us to follow would be Validator.[Category].[more-specific-categories].[specific shape ID or member], where everything after Validator is optional.

As for casing, I think exposing ENUM_CASING for some things but not others, including the main validator name, adds inconsistency to the event IDs and exposes the implementation detail that Java uses all caps enum member names. Is it difficult to get the casing right or something? Maybe you could just use the desired casing in the enum names rather than ALL_CAPS member names in the Java enums (if they're private enums it doesn't really matter).

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-invalid.errors
mtdowling: I don't think it appearing twice is useful though and instead you could just always use the terms that are defined by the paginated trait rather than the actual members used in the model (e.g., `items`, `pageSize`, `inputToken`, etc). So that could translate into `PaginatedTrait.ShouldNotBeRequired.pageSize`, which would be emitted event if the member bound to pageSize is MaxResults or something else.

> Are we sure that "PaginatedTrait.ShouldNotBeRequired.[member]" can NEVER appear twice for the same target shape?

This is addressed above.

> What if multiple traits mistakenly require members with the same name?

Not sure I follow, but I suspect this is addressed above.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-map-tokens.errors
mtdowling: > Is there not a case where one trait can have multiple members with the wrong target type?

The validator looks for a very specific set of members based on the trait. So if there are multiple issues, then there would be different events: `PaginatedTrait.WrongShapeType.inputToken`, `PaginatedTrait.WrongShapeType.outputToken`, `PaginatedTrait.WrongShapeType.pageSize`, etc.

> Also, "map" needs to be included as part of the eventId so that if they change it to target a different but also wrong shape, we will get a new non-deduped eventId

I don't think so. Event IDs don't have to be so unique that they account for breaking changes like changing the target shape of a paginator. We have bigger problems for changes like that, and we should err towards simplicity for these.

smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/errorfiles/http-checksum-header-conflicts.errors
mtdowling: (2) Sure yeah, appending to `HttpChecksumTrait.HeaderConflict` is fine -- `HttpChecksumTrait.HeaderConflict.x-amz-checksum-crc32`.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/input-output-structure-reuse.errors
mtdowling: Hm ok. It might be good to use just the name of the operation to help mitigate cases where a team changes the namespace for external use vs internal use (e.g., `InputOutputStructureReuse.Output.GetFoo`)

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
rchache: Well with the original way, they could suppress on NoninclusiveTerms.TRAIT.ns.foo#MySimpleValueTrait 
But if we switch the order, such that the noninclusive word is before the shape, they can't suppress everything for a single shape in one go. But also I could see an argument for being able to blanket suppress on the term itself as well.... So ultimately I don't mind changing the order on the next commit 👍

smithy-waiters/src/test/resources/software/amazon/smithy/waiters/errorfiles/invalid-return-types.errors
rchache: Ok so I can remove any JMES paths from event Ids, but 
1. what you see above is technically the waiter name and acceptor number and comparator name and statically-determined return type
2. Those sort of fields ARE necessary in other WaitableTrait violations in shipped models today. For example the following 3 are raised on a currently shipped model and would otherwise have conflicting eventIds: (sorry for the nonstandard formatting, it was from a script I was running for investigative purposes and goes [target] [eventid] [message] )

com.amazonaws.cloudformation#DescribeStacks WaitableTraitInvalidErrorType Waiter `StackDeleteComplete`, acceptor 1: errorType 'ValidationError' not found on operation. This operation defines the following errors: []

com.amazonaws.cloudformation#DescribeStacks WaitableTraitInvalidErrorType Waiter `StackCreateComplete`, acceptor 6: errorType 'ValidationError' not found on operation. This operation defines the following errors: []

com.amazonaws.cloudformation#DescribeStacks WaitableTraitInvalidErrorType Waiter `StackExists`, acceptor 1: errorType 'ValidationError' not found on operation. This operation defines the following errors: []

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/input-output-structure-reuse.errors
rchache: Ok sure, I'll make this change on next commit 👍

smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/errorfiles/http-checksum-header-conflicts.errors
rchache: Oh actually in this case "HeaderConflictError" is the name of the shape in the testcase:

operation HeadersConflicts {
    input: HeadersConflictsInput,
    output: HeadersConflictsOutput,
    errors: [HeadersConflictError,]
}

mtdowling: Ah ok. There are probably other categories of errors this validator reports right? Like missing a member binding I'd guess. If that's the case, a category before the shape ID would be nice.

smithy-linters/src/test/resources/software/amazon/smithy/linters/errorfiles/noninclusive-term-filter.errors
mtdowling: Ah I see. I think consistency is a nicer goal here, and I bet anyone suppressing non-inclusive terms in trait values could make do with either suppressing based on terms or based on a trait to where it wouldn't matter much. I'd probably rather they had to suppress terms individually though.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/operation/input-output-traits/invalid-multiple-operations-same-shape.errors
rchache: This was a case where I thought it necessary so that we don't unintendedly hide violations from a service team when something changes and the violation becomes slightly different. An example in this case would be where the operation changes to target another different shape that also does not start with the operation name. Ideally we _would_ show the violation in that case, even if the previous violation was approved or grandfathered in.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-invalid.errors
rchache: I still am not sure I get it. What if you had a scenario of:

[WARNING] ns.foo#Invalid4: paginated trait `pageSize` member `pageSize` should not be required
and
[WARNING] ns.foo#Invalid4: paginated trait `pageSize` member `someOtherMember` should not be required

Wouldn't those both map to "PaginatedTrait.ShouldNotBeRequired.pageSize"

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-map-tokens.errors
rchache: > Event IDs don't have to be so unique that they account for breaking changes like changing the target shape of a paginator. 

Well sure the semantic diff or other would catch it, but Tony and I were thinking that it would still be good to reraise the violation in that case so that they _do_ know they are doing something they shouldn't. Maybe it's a case where they are intentionally changing the target shape and have escalated to get approval to break backwards compatibility, but don't even know that they are (still) doing something not recommended.

> The validator looks for a very specific set of members based on the trait. So if there are multiple issues, then there would be different events: PaginatedTrait.WrongShapeType.inputToken, PaginatedTrait.WrongShapeType.outputToken, PaginatedTrait.WrongShapeType.pageSize, etc.

What about within just one of those, such as the inputToken? Can inputToken not have multiple members that all target wrong shape types?

smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/errorfiles/http-checksum-header-conflicts.errors
rchache: Actually it's only prefix conflicts on the headers and prefix headers. It's redundant to have a category AND "httpHeader"/"httpPrefixHeaders", but I could replace the latter with categories if you prefer

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/deprecated-shape/deprecated-shape.errors
rchache: That was just to make it easier to suppress all violations regarding deprecated members or all violations regarding deprecated mixins separately. But yeah there's maybe not too much value there so I'll remove those categories on the next commit

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
rchache: 1. Ah that's a good point. I'll remove it
2. Sure I'll change it to that on the next commit

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
rchache: The only concern I had with aggregating multiple members into one event was that once "approved", it would hide the warning for any new members being added to the GET payload. So a service team could end up making that change and complain that Trebuchet validations were incomplete and didn't give them proper guidance or warning

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-deeply-nested.errors
mtdowling: Since we already have validation event IDs, and they don't use SCREAM_CASE casing, I'd rather we avoided mixing it into event IDs.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/operation/input-output-traits/invalid-multiple-operations-same-shape.errors
mtdowling: That would get picked up by other validators, so I think it's a separate responsibility. This event is also demanding that the input/output has a name that matches the operation, so it seems very rare that you'd see this.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-method-semantics-validator.errors
mtdowling: It wouldn't matter IMO. The team already put things in the payload and approved it. Adding more stuff to something they approved shouldn't need human intervention (also keep in mind that we have a strict policy now of not allowing payloads on GET requests now so this doesn't matter in practice).

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-map-tokens.errors
mtdowling: I think we need to err towards simplicity here and in all events. There will be other Smithy implementations in the future, and they're going to have to reimplement all of this.

> What about within just one of those, such as the inputToken? Can inputToken not have multiple members that all target wrong shape types?

If I understand what you're asking, then no. The event is bound to the operation (ns.foo#MapTokens), and there is a single inputToken binding per/operation.

smithy-aws-traits/src/test/resources/software/amazon/smithy/aws/traits/errorfiles/http-checksum-header-conflicts.errors
mtdowling: I like keeping the category in case we add future categories here. But I don't feel that strongly about it.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/operation/input-output-traits/invalid-multiple-operations-same-shape.errors
rchache: Ok, will remove

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-deeply-nested.errors
rchache: ok, will change to camel case 👍

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-invalid.errors
rchache: Ah I understand now. I had assumed that these traits could have arbitrarily many members. Will remove the member name 👍

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/paginated/paginated-map-tokens.errors
rchache: ok, sounds fine to me

rchache:
Addressed all comments except the JMES one which does not yet have a clear path forward

smithy-model/src/main/java/software/amazon/smithy/model/validation/AbstractValidator.java
```diff
@@ -28,48 +28,104 @@ public String getName() {
...
+                .message(msg)
+                .shapeId(shape.getId())
+                .sourceLocation(loc.getSourceLocation())
+                .id(assembleFullEventId(additionalEventIdParts))
+                .build();
     }
 
-    private ValidationEvent createEvent(ValidationEvent.Builder builder) {
-        return builder.id(getName()).build();
+    protected final String assembleFullEventId(String... additionalEventIdParts) {
```
mtdowling, line 124: We should optimize this method since it'll be used a ton. First check if the length is 1, and if so, just return the first entry. If > 1, then use a StringBuilder and build up the result.

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/HttpChecksumTraitValidator.java
```diff
@@ -185,11 +185,13 @@ private List<ValidationEvent> validateHeaderConflicts(OperationShape operation,
                     .map(HttpPrefixHeadersTrait::getValue)
                     .ifPresent(headerPrefix -> {
                         if (HttpChecksumTrait.CHECKSUM_PREFIX.startsWith(headerPrefix)) {
+                            final String memberName = member.getId().getName();
```
sugmanue, line 188: `final` can be removed here, except for some very rare cases we don't use `final` for method scoped variables.
```suggestion
                            String memberName = member.getId().getName();
```

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/HttpChecksumTraitValidator.java
```diff
@@ -200,10 +202,12 @@ private List<ValidationEvent> validateHeaderConflicts(OperationShape operation,
                     .map(HttpHeaderTrait::getValue)
                     .ifPresent(headerName -> {
                         if (headerName.startsWith(HttpChecksumTrait.CHECKSUM_PREFIX)) {
+                            final String memberName = member.getId().getName();
```
sugmanue, line 205: Same as above.

```suggestion
                            String memberName = member.getId().getName();
```

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -130,70 +135,79 @@ private Collection<ValidationEvent> getValidationEvents(TextInstance instance) {
...
         }
         return events;
     }
 
-    private static String formatNonInclusiveTermsValidationMessage(
-            Map.Entry<String, List<String>> termEntry,
-            String matchedText,
-            TextInstance instance
+    /**
+     * Creates a single ValidationEvent.
```
sugmanue, line 145: Nit, missing a new line after first paragraph. Also, javadoc comments are usually for public API, not for private methods, so I'd rather remove it.

smithy-model/src/main/java/software/amazon/smithy/model/validation/AbstractValidator.java
rchache: optimized in latest commit

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PaginatedTraitValidator.java
```diff
@@ -165,20 +173,25 @@ private List<ValidationEvent> validateMember(
             if (validator.dangerTargets().contains(target.getType())) {
                 Set<ShapeType> preferredTargets = new TreeSet<>(validator.validTargets());
                 preferredTargets.removeAll(validator.dangerTargets());
+                final String traitName = validator.propertyName();
+                final String memberName = member.getId().getName();
+                final String targetType = target.getType().toString();
```
sugmanue, line 178: Nit, final can be removed here, as the code above shows.
```suggestion
                String traitName = validator.propertyName();
                String memberName = member.getId().getName();
                String targetType = target.getType().toString();
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/TargetValidator.java
```diff
@@ -204,7 +204,7 @@ private void validateDeprecatedTargets(
         deprecatedTrait.getMessage().ifPresent(message -> builder.append(". ").append(message));
         deprecatedTrait.getSince().ifPresent(since -> builder.append(" (since ").append(since).append(')'));
         events.add(ValidationEvent.builder()
-                           .id("DeprecatedShape")
+                           .id(String.format("DeprecatedShape.%s", target.getId()))
```
sugmanue, line 207: Use string concatenation instead, e.g., `"DeprecatedShape" + target.getId()`

```suggestion
                           .id("DeprecatedShape." + target.getId())
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/AbstractValidator.java
```diff
@@ -28,48 +28,106 @@ public String getName() {
...
+        String eventId = getName();
+        if (additionalEventIdParts.length == 1) {
+            eventId += "." + additionalEventIdParts[0];
+        } else if (additionalEventIdParts.length > 1) {
+            eventId += "." + String.join(".", additionalEventIdParts);
+        }
+        return eventId;
```
sugmanue, line 131: Probably minor but, `String.join` does a lot more than what we need.

```suggestion
    if (additionalEventIdParts.length > 0) {
        StringBuilder buf = new StringBuilder(getName());
        for (String part : additionalEventIdParts) {
            buf.append('.').append(part);
        }
        return buf.toString();
    }
    return getName();
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/OperationValidator.java
```diff
@@ -141,7 +141,7 @@ private ValidationEvent emitBadInputOutputName(Shape operation, String property,
         return ValidationEvent.builder()
                 .severity(Severity.WARNING)
                 .shape(operation)
-                .id(OPERATION_INPUT_OUTPUT_NAME)
+                .id(String.join(".", OPERATION_INPUT_OUTPUT_NAME, property))
```
sugmanue, line 144:
```suggestion
                .id(OPERATION_INPUT_OUTPUT_NAME + "." + property)
```

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
```diff
@@ -130,70 +135,79 @@ private Collection<ValidationEvent> getValidationEvents(TextInstance instance) {
...
-        final List<String> caseCorrectedEntryValue = termEntry.getValue().stream()
-            .map(replacement -> Character.isUpperCase(matchedText.charAt(0))
-                  ? StringUtils.capitalize(replacement)
-                  : StringUtils.uncapitalize(replacement))
-            .collect(Collectors.toList());
-        String replacementAddendum = !termEntry.getValue().isEmpty()
-                ? String.format(" Consider using one of the following terms instead: %s",
-                    ValidationUtils.tickedList(caseCorrectedEntryValue))
-                : "";
+        final String replacementAddendum = getReplacementAddendum(matchedText, replacements);
```
sugmanue, line 156: Nit, I know that this class in particular deviates from this, but in general we don't use final neither for arguments nor for local variables.

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
rchache: Removed in next commit

smithy-model/src/main/java/software/amazon/smithy/model/validation/AbstractValidator.java
rchache: Actually I can probably just use lines 126-130 in your suggested change in fact 👍

smithy-linters/src/main/java/software/amazon/smithy/linters/NoninclusiveTermsValidator.java
rchache: removed final from my changes in this class

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
```diff
@@ -138,18 +140,21 @@ private RuntimeType validatePath(LiteralExpression input, String path) {
...
         if (actual != RuntimeType.ANY && actual != expected) {
-            addEvent(Severity.DANGER, JMESPATH_PROBLEM, String.format(
+            addEvent(Severity.DANGER, String.format(
                     "Waiter acceptors with a %s comparator must return a `%s` type, but this acceptor was "
-                    + "statically determined to return a `%s` type.",
-                    comparator, expected, actual));
+                        + "statically determined to return a `%s` type.",
+                        comparator, expected, actual),
+                    JMESPATH_PROBLEM, waiterName, String.valueOf(acceptorIndex), comparator.toString(),
+                        actual.toString());
```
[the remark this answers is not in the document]
mtdowling: We should consistently use ID.Category, so that would mean `WaitableTraitJmespathProblem.InvalidReturnType.<waiter-name>.<acceptor-index>`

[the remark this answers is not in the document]
sugmanue: Nope, my bad.

PR #1529, opened 2022-12-08
Add github action for CI testing with SDK generators.

AndrewFossAWS:
*Description of changes:*

> In the past, Smithy changes have been released without integration testing with Smithy-based SDK code generators. In addition, SDK code generators often fall behind in Smithy versions. As a result, we don’t have visibility on code generators’ failures until they are upgraded to the latest smithy version. 

> Upgrading versions and building all the code generators manually is also a labor-intensive task. Hence, the goal is to design a continuous integration system for testing the latest Smithy implementation with the SDK code generators.

This PR adds a new Github action for integration testing with Smithy based SDKs. (See [example execution](https://github.com/AndrewFossAWS/smithy/actions/runs/3653031602))

Note:
- Most SDK generators' build or protocol tests are failing with the latest smithy changes. I have skipped github action triggering on push commit or scheduled event for now to reduce noise, and allow manual triggering via `workflow_dispatch`.
- Most SDKs generators' build and test commands are relatively concise. The Go SDK protocol test commands are more verbose/complicated. `smithy-ci` gradle task will be added in the Go SDK repo for running protocol tests.
- Rust SDK to be added in a subsequent PR.

### DIFF
--- a/.github/workflows/sdk-codegen-ci.yml
+++ b/.github/workflows/sdk-codegen-ci.yml
@@ -1,0 +1,179 @@
+name: sdk-codegen-ci
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-aws-sdk-js-v3:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+
+      - name: Increase Node Heap space
+        run: echo "NODE_OPTIONS=--max-old-space-size=16384" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v1
+      - id: smithy-version
+        uses: ./.github/workflows/checkout-and-build-smithy
+
+      - name: Add smithyVersion env variable
+        run: echo "smithyVersion=${{ steps.smithy-version.outputs.smithy-version }}" >> $GITHUB_ENV
+
+      - name: Checkout smithy-typescript
+        uses: actions/checkout@v2
+        with:
+          repository: awslabs/smithy-typescript
+          path: smithy-typescript
+
+      - name: Build smithy-typescript
+        run: cd smithy-typescript && ./gradlew -PsmithyVersion=$smithyVersion clean build publishToMavenLocal
+
+      - name: Checkout aws-sdk-js-v3
+        uses: actions/checkout@v2
+        with:
+          repository: aws/aws-sdk-js-v3
+          path: aws-sdk-js-v3
+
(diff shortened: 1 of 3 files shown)


.github/workflows/sdk-codegen-ci.yml
```diff
@@ -0,0 +1,179 @@
...
+        uses: actions/checkout@v2
+        with:
+          repository: aws/aws-sdk-go-v2
+          path: aws-sdk-go-v2
+
+      - name: Build aws-sdk-go-v2
+        run: cd aws-sdk-go-v2/codegen && ./gradlew -PsmithyVersion=$smithyVersion clean build -Plog-tests && ./gradlew clean
+
+      - name: Run aws-sdk-go-v2 protocol tests
+        # This is essentially the `ci-test` command in the Makefile
```
JordonPhillips, line 85: Then why aren't we just running that?

.github/workflows/sdk-codegen-ci.yml
```diff
@@ -0,0 +1,179 @@
+name: sdk-codegen-ci
+
+on:
+  workflow_dispatch:
```
JordonPhillips, line 4: You should add instructions in the README or somewhere else on how to actually trigger these.

JordonPhillips (changes requested):

.github/workflows/sdk-codegen-ci.yml
AndrewFossAWS: Because [ci-test](https://github.com/aws/aws-sdk-go-v2/blob/main/Makefile#L340) uses [generate](https://github.com/aws/aws-sdk-go-v2/blob/main/Makefile#L80) which uses [smithy-generate](https://github.com/aws/aws-sdk-go-v2/blob/main/Makefile#L84). 

The smithy version needs to be overriden via `./gradlew -PsmithyVersion=$smithyVersion clean build`,  the `smithy-generate` command doesn't allow that and hence I had to split up the commands from `ci-test`.

.github/workflows/sdk-codegen-ci.yml
AndrewFossAWS: Added a comment at the top of the workflow file.

JordonPhillips: Nobody is going around looking at workflow files - the docs need to be in the readme or in contributing.md

srchase (approved):

JordonPhillips (approved):

PR #1531, opened 2022-12-12
DEMO of the formatter

daddykotex:
**THIS SHOULD NOT BE MERGED**

This is a PR created after running the formatter on all files under: `src/test/resources/software/amazon/smithy/model/loader/valid/`.

It's far from perfect:

- it moves comment around sometimes
- it adds inconsistent spacing (especially when you have mixins and traits or some combination)

But it's just to show case that we can parse anything in there. We've made some changes to the parser and we diverge from the grammar a bit in some cases. Mostly related to required elements (that are not required after all) or spacing (or type of spacing) required around some elements.

The [first commit](https://github.com/awslabs/smithy/commit/2d9887c1c98aa6ae08c0e2298945238725b5d1b2) is me manually changing some files which contain things we could not parse. I manually fixed those instances because I believe they're not worth diverging from the grammar and I think those instances should be fixed (as I did) or moved to the `invalid` subdirectory.

This PR relates to https://github.com/awslabs/smithy-language-server/issues/79


smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/elided-members.smithy
```diff
@@ -101,6 +100,7 @@ operation ThrowAway {
         id: String
         property: String
     }
-    output := {}
+    output := {
```
mtdowling, line 103: This is going to be _very_ common, and I think readers would prefer it stay one line. Is that possible?

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/main-test.smithy
```diff
@@ -28,79 +32,93 @@ operation GetCurrentTime {
...
+    @required
+    name: smithy.api#String
+    @required
+    coordinates: CityCoordinates
 }
 
 @readonly
-@paginated(inputToken: "nextToken", outputToken: "nextToken",
-           pageSize: "pageSize", items: "items")
+@paginated(inputToken: "nextToken", outputToken: "nextToken", pageSize: "pageSize", items: "items")
```
mtdowling, line 79: This change is interesting. How do you decide when to use a single line vs multiple lines for a trait? Lots of traits tend to use a single line, so I like this (like http traits), but some like this one can have lots of properties.

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/mixins/loads-mixins.smithy
```diff
@@ -3,20 +3,32 @@ $version: "2.0"
...
-structure B with [A] {}
+}
+
+structure B with [A] {
+
+}
 
 @mixin
-structure C {}
+structure C {
```
mtdowling, line 15: Similar comment as inline input/ouptut I guess -- Is it possible to support single line structures?

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/service-with-errors.smithy
```diff
@@ -3,15 +3,17 @@ $version: "1.0"
 namespace ns.foo
 
 service EmptyService {
-    version: "2020-02-18",
-    errors: [
-        Common1,
-        Common2,
-    ],
+    version: "2020-02-18"
+    errors: [Common1, Common2]
```
mtdowling, line 7: Oh interesting. Is there a threshold for when this uses multiple lines? Service can have lots of errors.

daddykotex:
Thanks for the feedback Micheal, this is very nice of you. I think we're on the same page for most of the things.

> Oh interesting. Is there a threshold for when this uses multiple lines? Service can have lots of errors.

At first we used around with a length limit for a given line. But it turned out to be difficult to implement with the way the formatter is currently built. Then I switched over to a size limit: for instance, operation errors will use multiple lines if there are mort than two. It is pretty arbitrary and I think it's bad.

The line limit is much better so I'll revisit that soon. We're also considering configuration, so that you could tweak the formatter's behaviour.

> Similar comment as inline input/ouptut I guess -- Is it possible to support single line structures?

Yes, it would be preferable IMO as well. We'll implement.

> Not sure how you'd do it, but if members use traits, there should be a blank line between members

We also noticed that, structure can become very dense. We'll implement.

PR #1534, opened 2022-12-13
Add exclude/include tranforms using selectors

sugmanue:
*Description of changes:*

Added include/exclude transforms using selectors. Using selectors is a generic way to include or exclude shapes, instead of adding specific transforms such as `byShapeType` (e.g., [this PR](https://github.com/awslabs/smithy/pull/1330)) a selector can be used to replace that and other use cases.

### DIFF
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.build.transforms;
+
+import java.util.Set;
+import software.amazon.smithy.build.TransformContext;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.selector.Selector;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.transform.ModelTransformer;
+import software.amazon.smithy.utils.FunctionalUtils;
+
+/**
+ * {@code excludeShapesBySelector} excludes the shapes matching the given selector.
+ *
+ * <p>Prelude shapes are not removed by this transformer.
+ */
+public final class ExcludeShapesBySelector extends ConfigurableProjectionTransformer<ExcludeShapesBySelector.Config> {
+
+    /**
+     * {@code excludeShapesBySelector} configuration.
+     */
+    public static final class Config {
+        private Selector selector = null;
+
+        /**
+         * Gets the selector used to filter the shapes.
+         *
+         * @return The selector used to filter the shapes.
+         */
+        public Selector getSelector() {
+            return selector;
+        }
+
+        /**
+         * Sets the selector used to filter the shapes.
+         *
+         * @param selector The selector used to filter the shapes.
+         */
+        public void setSelector(Selector selector) {
+            this.selector = selector;
+        }
+    }
+
+    @Override
+    public Class<Config> getConfigType() {
+        return Config.class;
+    }
+
+    @Override
+    public String getName() {
+        return "excludeShapesBySelector";
+    }
+
+    @Override
+    protected Model transformWithConfig(TransformContext context, Config config) {
+        Selector selector = config.getSelector();
+        ModelTransformer transformer = context.getTransformer();
+        Model model = context.getModel();
+        Set<Shape> selected = selector.select(model);
+        return transformer.filterShapes(model, FunctionalUtils.not(selected::contains));
+    }
+}
--- a/smithy-build/src/main/java/software/amazon/smithy/build/transforms/IncludeShapesBySelector.java
+++ b/smithy-build/src/main/java/software/amazon/smithy/build/transforms/IncludeShapesBySelector.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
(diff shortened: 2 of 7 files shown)


docs/source-1.0/guides/building-models/build-config.rst
```diff
@@ -384,6 +426,48 @@ Removes shapes if they are marked with one or more specific traits.
...
+
+        {
+            "version": "1.0",
+            "projections": {
+                "exampleProjection": {
+                    "transforms": [
+                        {
+                            "name": "includeShapesBySelector",
+                            "args": {
+                                "selector": ":is([id=smithy.example#FooService], [id=smithy.example#FooService] ~> *)"
```
mtdowling, line 459: Same comment as above: a bit simpler to do this since you don't have to repeat the service ID:
```
[id=smithy.example#FooService] :is(*, ~> *)
```

docs/source-1.0/guides/building-models/build-config.rst
```diff
@@ -298,6 +298,48 @@ Only the following shape type changes are supported:
...
+.. list-table::
+    :header-rows: 1
+    :widths: 10 20 70
+
+    * - Property
+      - Type
+      - Description
+    * - selector
+      - ``string``
+      - A valid :ref:`selector <selectors>` used to exclude shapes
```
mtdowling, line 317:
```suggestion
      - A valid :ref:`selector <selectors>` used to exclude shapes.
```

docs/source-1.0/guides/building-models/build-config.rst
```diff
@@ -384,6 +426,48 @@ Removes shapes if they are marked with one or more specific traits.
...
+.. list-table::
+    :header-rows: 1
+    :widths: 10 20 70
+
+    * - Property
+      - Type
+      - Description
+    * - selector
+      - ``string``
+      - A valid :ref:`selector <selectors>` used to include shapes
```
mtdowling, line 445:
```suggestion
      - A valid :ref:`selector <selectors>` used to include shapes.
```

smithy-build/src/main/java/software/amazon/smithy/build/transforms/IncludeShapesBySelector.java
```diff
@@ -0,0 +1,80 @@
...
+import software.amazon.smithy.model.selector.Selector;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.transform.ModelTransformer;
+
+/**
+ * {@code includeShapesBySelector} includes the shapes matching the given selector.
+ *
+ * <p>Prelude shapes are not removed by this transformer.
+ */
+public final class IncludeShapesBySelector extends BackwardCompatHelper<IncludeShapesBySelector.Config> {
```
mtdowling, line 30: This is new so doesn't need to use BackwardCompatHelper

smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
```diff
@@ -0,0 +1,81 @@
...
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.transform.ModelTransformer;
+import software.amazon.smithy.utils.FunctionalUtils;
+
+/**
+ * {@code excludeShapesBySelector} excludes the shapes matching the given selector.
+ *
+ * <p>Prelude shapes are not removed by this transformer.
+ */
+public final class ExcludeShapesBySelector extends BackwardCompatHelper<ExcludeShapesBySelector.Config> {
```
mtdowling, line 31: This is new so doesn't need to use BackwardCompatHelper

smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
```diff
@@ -0,0 +1,81 @@
...
+        public String getSelector() {
+            return selector;
+        }
+
+        /**
+         * Sets the selector used to filter the shapes.
+         *
+         * @param selector The selector used to filter the shapes.
+         */
+        public void setSelector(String selector) {
```
mtdowling, line 53: I _think_ you can just require an actual Selector here and in the getter, and NodeMapper knows what to do.

smithy-build/src/main/java/software/amazon/smithy/build/transforms/IncludeShapesBySelector.java
```diff
@@ -0,0 +1,80 @@
...
+        return "selector";
+    }
+
+    @Override
+    protected Model transformWithConfig(TransformContext context, Config config) {
+        String selector = config.getSelector();
+        ModelTransformer transformer = context.getTransformer();
+        Model model = context.getModel();
+        Set<Shape> selected = Selector.parse(selector).select(model);
+        return transformer.filterShapes(model, selected::contains);
```
mtdowling, line 78: What happens if a selector returns, say, a list member? The list would be in an invalid state without the member. Does the ModelTransformer notice this, then remove the list, then remove any reference to the list?

mtdowling (changes requested):

smithy-build/src/main/java/software/amazon/smithy/build/transforms/IncludeShapesBySelector.java
sugmanue: I don't think so but I can test, I wondered about a similar idea for the change to allow to filter enum members using tags and an explicit test to validate that an exception is thrown when all the members are removed was part of [that change](https://github.com/awslabs/smithy/blob/c45d5dbcc404c9ff3af6a08cc24c83377fc918b2/smithy-model/src/test/java/software/amazon/smithy/model/transform/RemoveShapesTest.java#L120).

smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
sugmanue: Updated.

smithy-build/src/main/java/software/amazon/smithy/build/transforms/ExcludeShapesBySelector.java
sugmanue: That works.

docs/source-1.0/guides/building-models/build-config.rst
```diff
@@ -298,6 +298,48 @@ Only the following shape type changes are supported:
...
+
+        {
+            "version": "1.0",
+            "projections": {
+                "exampleProjection": {
+                    "transforms": [
+                        {
+                            "name": "excludeShapesBySelector",
+                            "args": {
+                                "selector": ":is([id=smithy.example#FooService], [id=smithy.example#FooService] ~> *)"
```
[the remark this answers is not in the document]
sugmanue: Simplified.

docs/source-1.0/guides/building-models/build-config.rst
sugmanue: Updated.

docs/source-1.0/guides/building-models/build-config.rst
```diff
@@ -298,6 +298,49 @@ Only the following shape type changes are supported:
         }
 
 
+.. _excludeShapesBySelector-transform:
```
syall, line 301: Missing the port to 2.0 docs.

sugmanue: 🤦 I'll fix that, it should only be for the 2.0 docs.

sugmanue: Updated, thanks!

jvschneid (approved):

AndrewFossAWS (approved):

Issue #1512, opened 2022-11-24
HttpUriConflict while defining multiple operations inside a service


protyay:
I'm defining multiple operations inside a resource and adding that resource to the service shape. 
During Smithy build, I get `HttpUriConflict` 
```
Operation URI, `/dc/labor/shift-planning/v1/input/{screenId}/tabs`, conflicts with other operation URIs in the same service: [`com.amazon.dclaborshiftplanningservice#FetchFilters` (/dc/labor/shift-planning/v1/input/{screenId}/{tabKey}/filters)]
```
Smithy Code
```
resources: [
        PickStrategyResource,
        ReportsPageResource,
        InputPageResource
    ]

resource InputPageResource {
    operations: [
        FetchTabs
        FetchInputPageContent
        SaveUserOverrides
    ]
}

@http(
    method: "GET",
    uri: "/dc/labor/shift-planning/v1/input/{screenId}/tabs"
)
operation FetchTabs {
// structures
}
@http(
    method: "GET",
    uri: "/dc/labor/shift-planning/v1/input/{screenId}/{tabKey}/filters"
)
operation FetchFilters {
}
```
Does that mean all resource operations in a given service closure should have the same URI ? 😕 

As a workaround, I'm not binding all my `resources` to the `service` shape and the build runs fine, but the downside is probably I have to repeat the `errors` property in each of the operation shapes. 
Do I understand these correctly ?

sugmanue:
> Does that mean all resource operations in a given service closure should have the same URI ? 😕

No, it does mean that the two patterns 

* `/dc/labor/shift-planning/v1/input/{screenId}/tabs`
* `/dc/labor/shift-planning/v1/input/{screenId}/{tabKey}/filters`

conflict according to the rules in the specification, see [here](https://smithy.io/2.0/spec/http-bindings.html?highlight=http%20bindings#pattern-validation-and-conflict-avoidance). In particular this bit

> A label and a literal SHOULD NOT both occupy the same segment in patterns which are equivalent to that point if they share a host.

You can workaround this by slightly changing one of the URI's, for instance by using `tabs` and `filters` as prefix rather than a suffix, not ideal but should unblock you.

[...]

Issue #1029, opened 2021-12-27
URI conflict validation is too constraining


Baccata:
#### Minimised repro 

```kotlin
namespace foo

service Foo {
  operations: [Hello1, Hello2]
}

structure Hello1Input {
  @httpLabel
  @required
  name: String,
}

@http(method: "GET", uri: "/hello")
@readonly
operation Hello2 {
}

@http(method: "GET", uri: "/{name}/greet")
operation Hello1 {
  input: Hello1Input,
}
```

Gives the following errors : 

![Screenshot 2021-12-27 at 12 06 39](https://user-images.githubusercontent.com/4439335/147465948-ac19726a-964a-4b40-a99c-82de8c47dc14.png)

#### Expected behaviour : 

These two URI paths should not be presenting a conflict, as there is always a possibility to distinguish between one and the other without additional informations. 

I believe the cause of the problem is [this condition](https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java#L135-L138) which rakes too widely.

mtdowling:
Yes, it is too constraining in several ways. We have a few proposals to make URI conflict resolution more flexible. I’m hoping we can finish those up and publish an RFC soon.

david-perez:
Is that RFC ready?

Quafadas:
Bumped into this whilst trying to smithy-translate an API...

kubukoz:
Another example from today:

```
GET /subscriptions/{id}
GET /subscriptions/device/{deviceId}
```

these shouldn't be conflicting as `{id}` isn't a greedy label. Is there any chance we can update this soon?

sugmanue:
We are aware that the rule is too restrictive and we are planning to publish an RFC in the coming weeks to allow these and URIs that _might conflict_ at runtime by specifying how the conflicts should be resolved by using the most specific URI, similar to what [API Gateway does](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-routes.html#http-api-develop-routes.evaluation). 

I will update this issue when the RFC is publish to get early feedback and make sure that the proposal covers your use cases.

sugmanue:
I have opened a [draft pull request](https://github.com/awslabs/smithy/pull/1539) for our current thinking about how to remove this restriction, if you have time we will appreciate your feedback on this proposal

lenguyenthanh:
[...]

PR #1539, opened 2022-12-14
URI Pattern Validation and Conflict Resolution design doc

sugmanue:
*Issue #, if available:* #1512 #1029

*Description of changes:*

Design proposal to change the way we validate URI patterns to allow conflicting ones while also specifying how the servers must route ambiguous requests.


designs/uri-conflict-resolution.md
```diff
@@ -0,0 +1,141 @@
...
+
+Given two ambiguous URI patterns `A` and `B` with segments
+`[A0, …, An]` and `[B0, …, Bm]` with query string literals
+`[AQ0, …, AQp]` and `[BQ0, …, BQq]` (with both `p` and `q` possibly
+zero, i.e., without query string literals), we use the following
+algorithm to compare them, for each index `x` from `0` to `min(n, m)`
+
+1. If `A[x]` and `B[x]` are both literals then continue (the literal values have to be equal otherwise the patterns would not be ambiguous)
+2. If `A[x]` is a literal and `B[x]` is a label then `A` is more specific than `B`, 
+3. If `A[x]` is a non-greedy label and `B[x]` is a greedy label than `A` is more specific than `B`
```
kubukoz, line 82:
```suggestion
3. If `A[x]` is a non-greedy label and `B[x]` is a greedy label then `A` is more specific than `B`
```

designs/uri-conflict-resolution.md
```diff
@@ -0,0 +1,141 @@
...
+
+* `/abc/{foo}/cde`
+* `/abc/{bar}/cde`
+
+## Proposal
+
+### Specificity Routing
+
+The proposal is to  accept any non-equivalent URI path patterns and
+also specifying the expected routing algorithm. The routing algorithm
```
kubukoz, line 62:
```suggestion
also specify the expected routing algorithm. The routing algorithm
```

kubukoz:
I think this addresses the problem well, thank you! I have no questions so far, just a couple spelling change suggestions.

guymguym:
hi @sugmanue this looks good, but I just wonder - wouldn't it be simple enough to assign an optional numeric priority to each conflicting route, so that api owners could solve any issue with the order of their routes. I would just say that the default priority is 0, and only if there is a collision on the same route and priority, then the api is invalid...

kubukoz:
I feel like doing priority with numbers is like good old `z-index` in CSS - at some point you're going to use the number `999` and regret this decision 😅 

having well-defined rules seems more principled.

sugmanue:
> having well-defined rules seems more principled.

+1 on that, and also opens up the possibility of subtle ways of introducing API breaking modes by changing this value.

guymguym:
I respect that you want to define rules, everyone loves rules, but with that approach I would just be worried that some api's will not be able to use it unless they add an `?x-id=OPNAME` query param to disambiguate between operations. See in S3 for example - https://github.com/awslabs/smithy-rs/issues/1012.

guymguym:
@sugmanue , @kubukoz  , thank you for the replies!

The point I still wanted to make is that using rules that reside "outside" the api definition, in a meta schema, is not going to be more backward compatible because changes to the rules could be implemented at different times for clients vs servers, which will introduce incompatibility. Adding  the routing order as information inside the api definition is actually more stable, and means the api definition is self-contained and described the api behavior as a configuration. So if I keep the api definition in my sources I can track the changes of routing as well. Which I see as a better api design than externalizing the routing as rules.

That said, I understand that a priority could sounds too simple and therefore not sophisticated enough, although I think that such simplicity is usually a good sign, and while z-index is criticized it still provides a pragmatic solution to those non common collision cases.

Two others ways to define such an order:
1. rely on the order of operations in the service.operations list itself.
2. a colliding operation could define an http binding property like routeAfter=OPNAME that would mean that this operation will always be after the other OPNAME and allow it to match first.

Thanks. Guy

mtdowling:
This looks great, Manuel! 

I looked at the S3 Smithy model, and I think all of it can be represented by this proposal _except_ for [CopyObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html) which appears to use headers for routing, and assigning number based priorities to routes wouldn't help here. IMO, header-based routing wouldn't be worth the complexity nor something we'd want to proliferate.

<details>
<summary>Smithy validation errors without x-id</summary>

```
ERROR: com.amazonaws.s3#AbortMultipartUpload (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
  49 |     "com.amazonaws.s3#AbortMultipartUpload": {
     |                                              ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#DeleteObject` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#CopyObject (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
1581 |     "com.amazonaws.s3#CopyObject": {
     |                                    ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#PutObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPartCopy` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPart` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#DeleteObject (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
3168 |     "com.amazonaws.s3#DeleteObject": {
     |                                      ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#AbortMultipartUpload` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#GetObject (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
4995 |     "com.amazonaws.s3#GetObject": {
     |                                   ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#ListParts` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#ListParts (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
8504 |     "com.amazonaws.s3#ListParts": {
     |                                   ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#GetObject` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#PutObject (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
11121 |     "com.amazonaws.s3#PutObject": {
     |                                   ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#CopyObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPartCopy` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPart` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#UploadPart (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
13681 |     "com.amazonaws.s3#UploadPart": {
     |                                    ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#CopyObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#PutObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPartCopy` (/{Bucket}/{Key+})]

ERROR: com.amazonaws.s3#UploadPartCopy (HttpUriConflict)
     @ /Users/dowling/projects/aws-sdk-js-v3/codegen/sdk-codegen/aws-models/s3.json
     |
13701 |     "com.amazonaws.s3#UploadPartCopy": {
     |                                        ^
     = Operation URI, `/{Bucket}/{Key+}`, conflicts with other operation URIs in the same service: [`com.amazonaws.s3#CopyObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#PutObject` (/{Bucket}/{Key+}), `com.amazonaws.s3#UploadPart` (/{Bucket}/{Key+})]
```

</details>

The thing we'd need to make more explicitly supported is placing dynamic query strings in the URI in addition to input parameters. This would remove the `x-id` thing for everything in the S3 model except CopyObject (TBD if/how there's a different solution for CopyObject, e.g., supporting `x-amz-copy-source` in the query string too). Here's a disambiguation example we could use instead of x-id.

```
AbortMultipartUpload: DELETE /{Bucket}/{Key+}?uploadId
DeleteObject:         DELETE /{Bucket}/{Key+}
```

Smithy validation allows this right now, but it's currently undefined behavior. It would be nice to add statements around if/how this works in the proposal.

guymguym:
@mtdowling you can see the summary of unsupported s3 routing cases in this comment - 
https://github.com/awslabs/smithy-rs/issues/1012#issuecomment-1023243176

for example in the case you mentioned about `?uploadId` we cannot match the route based on a query param with variable input value? Here is the longer description -

> These cases were added because a required query param cannot be specified in the route URI when its value is an actual input field (like ?uploadId=123 or ?partNumber=456), and not a fixed api value (like ?list-type=2). In the smithy spec you cannot specify just a key like ?uploadId in the uri to separate those cases because that's just like ?uploadId= and will fail to route and a value is provided. Maybe the spec can allow it with ?uploadId=* ?

mtdowling:
@guymguym: The way I think it could (and maybe even already) works is that servers would route based on static query string parameters regardless of if they have a value (e.g., `?acl` vs `?acl=true`). So in the `?uploadId` example, they could do the routing based on the presence of the query string parameter, and later, after an operation is matched, deserialize based on the query string binding of a member. We'd only ever want to do this for required input parameters since it would be used for routing. We'd also need to ensure that clients are able to handle this case.

zeyad001:
Any update on this?

AndrewFossAWS (approved):

PR #1569, opened 2023-01-05
Fix Upgrade1to2Command for Set shape

AndrewFossAWS:
*Description of changes:*
Smithy IDL 1 contains set shapes, but IDL 2 does not. When upgrading from IDL 1, set shapes should be converted to list shapes. The upgrade command currently does not perform this conversion.

### DIFF
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
@@ -46,6 +46,7 @@ import software.amazon.smithy.model.SourceLocation;
 import software.amazon.smithy.model.loader.ModelAssembler;
 import software.amazon.smithy.model.loader.Prelude;
 import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.SetShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.shapes.ShapeVisitor;
@@ -321,6 +322,14 @@ final class Upgrade1to2Command extends SimpleCommand {
             return null;
         }
 
+        @Override
+        public Void setShape(SetShape shape) {
+            getDefault(shape);
+            writer.erase(shape.getSourceLocation(), 3); // `set` has 3 characters
+            writer.insert(shape.getSourceLocation(), "@uniqueItems" + System.lineSeparator() + "list");
+            return null;
+        }
+
         private String serializeEnum(StringShape shape) {
             // Strip all the traits from the shape except the enum trait.
             // We're leaving the other traits where they are in the model
@@ -425,6 +434,13 @@ final class Upgrade1to2Command extends SimpleCommand {
             contents = String.join(System.lineSeparator(), lines);
         }
 
+        private void insert(SourceLocation from, String value) {
+            IdlAwareSimpleParser parser = new IdlAwareSimpleParser(contents);
+            parser.rewind(from);
+            int fromPosition = parser.position();
+            contents = contents.substring(0, fromPosition) + value + contents.substring(fromPosition);
+        }
+
         private void eraseLine(int lineNumber) {
             List<String> lines = new ArrayList<>(Arrays.asList(contents.split("\\r?\\n")));
             lines.remove(lineNumber - 1);
@@ -456,6 +472,16 @@ final class Upgrade1to2Command extends SimpleCommand {
             IdlAwareSimpleParser parser = new IdlAwareSimpleParser(contents);
             parser.rewind(from);
             int fromPosition = parser.position();
+            parser.rewind(to);
+            int toPosition = parser.position();
+            contents = contents.substring(0, fromPosition) + contents.substring(toPosition);
+        }
+
+        private void erase(SourceLocation from, int size) {
+            IdlAwareSimpleParser parser = new IdlAwareSimpleParser(contents);
+            parser.rewind(from);
+            int fromPosition = parser.position();
+            SourceLocation to = new SourceLocation(from.getFilename(), from.getLine(), from.getColumn() + size);
             parser.rewind(to);
             int toPosition = parser.position();
             contents = contents.substring(0, fromPosition) + contents.substring(toPosition);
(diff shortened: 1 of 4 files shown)


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
```diff
@@ -321,6 +323,18 @@ public Void stringShape(StringShape shape) {
             return null;
         }
 
+        @Override
+        public Void listShape(ListShape shape) {
+            if (shape instanceof SetShape) {
```
syall, line 328: Does the `uniqueItems` trait needed to be added for converting a `SetShape` to a `ListShape` here?

sugmanue: Yes, see [the migration guide here](https://smithy.io/2.0/guides/migrating-idl-1-to-2.html#convert-set-shapes-to-list-shapes).

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
```diff
@@ -321,6 +323,19 @@ public Void stringShape(StringShape shape) {
...
         }
 
+        @Override
+        public Void listShape(ListShape shape) {
+            getDefault(shape);
+            if (shape instanceof SetShape) {
+                int lineNumber = shape.getSourceLocation().getLine();
+                writer.eraseLine(lineNumber);
+                writer.insertLine(lineNumber, "list " + shape.toShapeId().getName() + " {");
+                writer.insertLine(lineNumber, "@uniqueItems");
```
sugmanue, line 333: I looked a bit closer into this as the `insertLine` at specific line with some new content seems a bit brittle, but I was able to break it using something like this (at the end of the file, unlikely but valid):

```
set SomeSet1 { member: NonBoxedInteger, }
```
the command fails with a cryptic message `Index: 35, Size: 34` (probably coming from trying to insert in an out of range index). If I have this (at the end of the file):

```
set SomeSet1 { member: NonBoxedInteger, }
set SomeSet2 { member: NonBoxedInteger, }
```
The command will output something like this 
```
@uniqueItems
list SomeSet1 {
@uniqueItems
list SomeSet2 {
```

This implementation will also remove comments like:

```
set SomeSet { // This comment is here
```
One way to solve it is just replacing the `set` part by `@uniqueItems\nlist` instead of the whole line, that will need some tweaking to the `ModelWriter` class to add the required `erase` and `insert` methods.

AndrewFossAWS: Good spot!

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
```diff
@@ -321,6 +324,16 @@ public Void stringShape(StringShape shape) {
             return null;
         }
 
+        @Override
+        public Void listShape(ListShape shape) {
+            getDefault(shape);
+            if (shape instanceof SetShape) {
+                writer.updateLine(shape.getSourceLocation().getLine(),
+                    line -> line.replace("set", "@uniqueItems\nlist"));
```
sugmanue, line 332: Using replace is brittle as some other occurrences of the characters "set" will be also replace, e.g.,

```
jshell> "set Fooset { // this set is awesome".replace("set", "@uniqueItems\nlist")
$1 ==> "@uniqueItems\nlist Foo@uniqueItems\nlist { // this @uniqueItems\nlist is awesome"
```

Using replace first might help but that might still fail for some other cases, e.g.,

```
jshell> "/* this set is awesome */ set Foo {".replaceFirst("set", "@uniqueItems\nlist")
$2 ==> "/* this @uniqueItems\nlist is awesome */ set Foo {"
```

syall, line 332: We need a test where there is "irregular" text / indentation before and after `set` on the same line, and where there are other occurrences of the letters `set` since [Java `String.replace()`](https://www.baeldung.com/string/replace) replaces all occurrences in a line, e.g.

Before:

```smithy
// "set" in "ResetResource"
resource ResetResource {
    properties: { member: String }
}

// 2 spaces, "set" in "@setter"
  @setter set for ResetResource {
    $member
}
```

After:

```smithy
// "set" in "ResetResource"
resource ResetResource {
    properties: { member: String }
}

// 2 spaces, "set" in "@setter"
  @setter @uniqueItems
list for ResetResource {
    $member
}
```

Maybe we need to add helper methods to `ModelWriter` for adding traits and renaming shapes based on source location?

sugmanue: The second one is wrong, we don't have block comments but the argument still stands, e.g.,

```
"@tags([\"set\"]) set Foo {".replaceFirst("set", "@uniqueItems\nlist")
$3 ==> "@tags([\"@uniqueItems\nlist\"]) set Foo {"
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/Upgrade1to2Command.java
```diff
@@ -321,6 +323,16 @@ public Void stringShape(StringShape shape) {
...
 
+        @Override
+        public Void listShape(ListShape shape) {
+            getDefault(shape);
+            if (shape instanceof SetShape) {
+                writer.erase(shape.getSourceLocation(), 3); // `set` has 3 characters
+                writer.insert(shape.getSourceLocation(), "@uniqueItems" + System.lineSeparator() + "list");
+            }
+            return null;
+        }
```
syall, line 334: nit: overriding the `setShape()` visitor method will reduce the need for the `instanceof SetShape` check I think.

```suggestion
        @Override
        public Void setShape(SetShape shape) {
            getDefault(shape);
            writer.erase(shape.getSourceLocation(), 3); // `set` has 3 characters
            writer.insert(shape.getSourceLocation(), "@uniqueItems" + System.lineSeparator() + "list");
            return null;
        }
```

syall (approved):

sugmanue (approved):

PR #1585, opened 2023-01-19
Fix params in `Ec2ComplexError` test

jdisanti:
The expected error body has a message, but the params did not.


jdisanti:
I just realized the `ComplexError` struct doesn't have a `Message` member, which could explain why it's missing here. This seems like one of those odd cases where generated errors have additional fields that may not be modeled, such as message and code.

The code my working on in smithy-rs is validating message/code error metadata during Smithy protocol tests, which is why this missing became a problem. I'm not really sure what the best path forward is for it.

jdisanti:
I don't need this change to be made after all. I ended up refactoring our protocol test generator to do field by field comparison for errors so that I can skip the error metadata fields.

PR #1595, opened 2023-01-26
fix referenced components removed in openapi schema

AndrewFossAWS:
*Issue #, if available:*

Customer reported referenced component being removed in openapi schema. 

In [RemoveUnusedComponents](https://github.com/awslabs/smithy/blob/main/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/mappers/RemoveUnusedComponents.java#L125), it only keeps referenced components if the field object size is 1.

However there are cases where the field object size is greater than 1.

For example:

__smithy:__
```
structure OperationInput {
    @default({})
    payload: StringMap
}

map StringMap {
    key: String
    value: String
}
```
__openapi:__

(`StringMap` is removed in the `schemas` object.)

```
    "schemas": {
        "OperationRequestContent": {
            "properties": {
                "payload": {
                    "$ref": "#/components/schemas/StringMap",
                    "default": {}
                }
            }
        }
    }
```

*Description of changes:*

Remove the field object size check when removing unused components

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/mappers/RemoveUnusedComponents.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/mappers/RemoveUnusedComponents.java
@@ -122,7 +122,7 @@ public final class RemoveUnusedComponents implements OpenApiMapper {
             public Set<String> objectNode(ObjectNode node) {
                 Set<String> result = new HashSet<>();
 
-                if (node.size() == 1 && node.getMember("$ref").isPresent()) {
+                if (node.getMember("$ref").isPresent()) {
                     node.getMember("$ref")
                             .flatMap(Node::asStringNode)
                             .map(StringNode::getValue)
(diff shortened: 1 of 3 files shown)


mtdowling:
I didn't think `$ref` objects allowed more than one property. Can you confirm?

AndrewFossAWS:
> I didn't think `$ref` objects allowed more than one property. Can you confirm?

@mtdowling From the reference object section of the OpenAPI [spec](https://spec.openapis.org/oas/v3.1.0#reference-object):

> This object cannot be extended with additional properties and any properties added SHALL be ignored.

So `default` property will be ignored in my example.
I was able to deploy an openapi json with the additional property `default` to API Gateway. The `default` property was ignored/removed.

I can add a WARN log to let users know the additional properties will be ignored.

mtdowling:
Thanks. Warning would be too noisy here since it looks like we add `default` values to these refs automatically.

mtdowling (approved):

mtdowling (approved):

Issue #1582, opened 2023-01-16
Examples trait requires output even in error example


kubukoz:
## What's happening

Hi! I think the validator of the examples trait isn't behaving correctly - it seems to require an `output` entry (and a valid one, at that) even if the example exhibits an error.

This doesn't seem intentional, looking at [the example from the PR](https://github.com/awslabs/smithy/pull/888/files#diff-25d3a8f46e169673c09176b5405f96d809ff05c58f5bf302244457b53cd5facbR245-R256) that introduced error examples.

## Reproduction

This can be reproduced by running the following with [scala-cli](https://scala-cli.virtuslab.org/):

```scala
//> using lib "software.amazon.smithy:smithy-model:1.27.1"
import software.amazon.smithy.model.Model

object demo extends App {
  val model = Model
    .assembler()
    .addUnparsedModel(
      "test.smithy",
      """
      $version: "2.0"

      namespace example.hello

      operation Hello {
          output: Greeting,
          errors: [YouShallNotPass]
      }

      structure Greeting {
          @required
          message: String
      }

      @error("client")
      structure YouShallNotPass {
          @required
          message: String
      }

      apply Hello @examples([
          {
              title: "My name is Jeff"
              output: {
                  message: "Hello Jeff"
              }
          },
          {
              title: "Unhappy path",
              error: {
                  shapeId: YouShallNotPass
                  content: {
                      "message": "Go back to the Shadow"
                  }
              }
          }
      ])
      """
    )
    .assemble()
    .unwrap()

  println("success")
}
```

Here's what you get:

```
software.amazon.smithy.model.validation.ValidatedResultException: Result contained ERROR severity validation events: 
[ERROR] example.hello#Hello: Example output of `Unhappy path`: Missing required structure member `message` for `example.hello#Greeting` | ExamplesTrait N/A:0:0
        at software.amazon.smithy.model.validation.ValidatedResult.validate(ValidatedResult.java:145)
        at software.amazon.smithy.model.validation.ValidatedResult.unwrap(ValidatedResult.java:132)
        at demo$.<clinit>(main.scala:9)
```

## Why I think this happens

I believe this code should consider `output` and `error` mutually exclusive, but it seems to always validate `output`:

https://github.com/awslabs/smithy/blob/f5a0893cd0f9983d0bd5697db7c226c1d379ef92/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ExamplesTraitValidator.java#L56-L75

## Extras

[...]

PR #1599, opened 2023-02-01
Modifies ExamplesTraitValidator to handle cases where both output and error are defined

jvschneid:
*Issue #, if available:* 
#1582 

*Description of changes:*
This change  modifies the behavior of the ExamplesTraitValidator when
both output and error are defined. This is achieved by treating the
output member as optional, only validating it if it's present and
failing when it's defined alongside the error member.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
@@ -133,8 +133,8 @@ public final class ExamplesTrait extends AbstractTrait implements ToSmithyBuilde
         /**
          * @return Gets the output object.
          */
-        public ObjectNode getOutput() {
-            return output;
+        public Optional<ObjectNode> getOutput() {
+            return Optional.ofNullable(output);
         }
 
         /**
@@ -154,7 +154,7 @@ public final class ExamplesTrait extends AbstractTrait implements ToSmithyBuilde
             if (!input.isEmpty()) {
                 builder.withMember("input", input);
             }
-            if (!output.isEmpty()) {
+            if (this.getOutput().isPresent()) {
                 builder.withMember("output", output);
             }
 
@@ -177,7 +177,7 @@ public final class ExamplesTrait extends AbstractTrait implements ToSmithyBuilde
             private String title;
             private String documentation;
             private ObjectNode input = Node.objectNode();
-            private ObjectNode output = Node.objectNode();
+            private ObjectNode output;
             private ErrorExample error;
 
             @Override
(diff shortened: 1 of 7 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ExamplesTraitValidator.java
```diff
@@ -53,23 +53,34 @@ private List<ValidationEvent> validateExamples(Model model, OperationShape shape
...
 
-            model.getShape(shape.getOutputShape()).ifPresent(output -> {
-                NodeValidationVisitor validator = createVisitor(
-                        "output", example.getOutput(), model, shape, example);
-                events.addAll(output.accept(validator));
-            });
+            boolean isOutputDefined = example.getOutput().isPresent();
+            boolean isErrorDefined = example.getError().isPresent();
+
+            if (isOutputDefined && isErrorDefined) {
```
AndrewFossAWS, line 59: If this is the behaviour we want, this needs to be mentioned in the doc as well.
https://smithy.io/2.0/spec/documentation-traits.html?highlight=example#examples-trait

jvschneid: Great point!

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ExamplesTraitValidator.java
```diff
@@ -53,23 +53,34 @@ private List<ValidationEvent> validateExamples(Model model, OperationShape shape
...
+                events.add(error(shape, trait, String.format(
+                        "Example: `%s` has both output and error defined, only one should be present.",
+                        example.getTitle())));
+            } else {
+                if (isErrorDefined) {
```
syall, line 64: nit: the `if` statements for `isErrorDefined` and `isOutputDefined` can be moved into the outer layer as `else if` statements.

syall (approved):

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
```diff
@@ -133,8 +133,8 @@ public ObjectNode getInput() {
         /**
          * @return Gets the output object.
          */
-        public ObjectNode getOutput() {
-            return output;
+        public Optional<ObjectNode> getOutput() {
```
mtdowling, line 136: A bit of a breaking change... but maybe using this trait in code is fringe enough to not matter

mtdowling (approved):

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
jvschneid: I thought about keeping it as is and checking if `example.getOutput().isEmpty()` to check if it was defined, but it opened the possibility of the example below being valid, which I think shouldn't be, if it's present it should conform, even if empty:

```
$version: "2.0"

namespace example.hello

@examples([
    {
        title: "Unhappy path",
        output: {},
        error: {
            shapeId: YouShallNotPass
            content: {
                "message": "Go back to the Shadow"
            }
        }
    }
])
operation Hello {
    output: Greeting,
    errors: [YouShallNotPass]
}

structure Greeting {
    @required
    message: String
}

@error("client")
structure YouShallNotPass {
    @required
    message: String
}
```

PR #1602, opened 2023-02-03
generate unique synthetic names from unnamed enums

AndrewFossAWS:
### Context

Smithy 1.0 defined enumerations using an @enum trait on a string shape. Smithy 2.0 introduces the enum and intEnum shapes and deprecates the @enum trait. Enum’s name has also been changed from _optional_ to _required_.

Smithy currently does not convert enum string into enum shape by default; code generators are handling both enum string and enum shape.

Code generators have the option of calling [CodegenDirector.changeStringEnumsToEnumShapes(boolean synthesizeEnumNames)](https://github.com/awslabs/smithy/blob/main/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java#L246) to convert all enum string to enum shape, and generate “synthetic” names from values if names are missing. This feature is currently NOT used by any code generators e.g. Swift, Kotlin, Typescript, Go and Rust.

When generating synthetic enum names, Smithy replaces some characters with underscore i.e. `replaceAll("[-.:/\\s]+", "_")`. However this is not sufficient as the generated names need to match the following regular expression: `^[A-Z]+[A-Z_0-9]*$` (See [enum validation in documentation](https://smithy.io/2.0/spec/simple-types.html?highlight=enum#enum-validation))

For example:
```
Linux/UNIX -> Linux_UNIX
foo:bar -> foo_bar
```
However the convertion fails for the following examples:
```
Linux/UNIX (Amazon VPC)
ns.foo#bar
```

### Description of changes:

* Generate synthetic names that match the following regular expression: `^[A-Z]+[A-Z_0-9]*$` 
* Generate synthetic names by replacing all non-alphabetical and non-numerical characters with `_`
* Generate unique synthetic names by padding a number as suffix to the name.

Examples:

1.
```
Linux/UNIX → LINUX_UNIX
Windows (Amazon VPC) → WINDOWS_AMAZON_VPC_
Linux/UNIX (Amazon VPC) → LINUX_UNIX_AMAZON_VPC_
```

2.
```
a:b → A_B_0
a=b → A_B_1
a_b → A_B_2
```

3.
```
:b → B_0
=b → B_1
_b → B_2
```

4.
```
a:b → A_B_0
a=b  → A_B_1
a_b_0 → A_B_0_2
```

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
@@ -15,14 +15,19 @@
 
 package software.amazon.smithy.model.shapes;
 
+import static software.amazon.smithy.model.validation.validators.EnumShapeValidator.RECOMMENDED_NAME_PATTERN;
+
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.logging.Logger;
-import java.util.regex.Pattern;
 import software.amazon.smithy.model.SourceException;
 import software.amazon.smithy.model.traits.DeprecatedTrait;
 import software.amazon.smithy.model.traits.DocumentationTrait;
@@ -33,11 +38,11 @@ import software.amazon.smithy.model.traits.TagsTrait;
 import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.model.traits.UnitTypeTrait;
 import software.amazon.smithy.model.traits.synthetic.SyntheticEnumTrait;
+import software.amazon.smithy.model.validation.validators.EnumShapeValidator;
 import software.amazon.smithy.utils.BuilderRef;
 
 public final class EnumShape extends StringShape {
 
-    private static final Pattern CONVERTABLE_VALUE = Pattern.compile("^[a-zA-Z-_.:/\\s]+[a-zA-Z_0-9-.:/\\s]*$");
     private static final Logger LOGGER = Logger.getLogger(EnumShape.class.getName());
 
     private final Map<String, MemberShape> members;
@@ -113,6 +118,9 @@ public final class EnumShape extends StringShape {
      * The result will be empty if the given shape doesn't have the {@link EnumTrait},
      * if the enum doesn't have names and name synthesization is disabled, or if a name
      * cannot be synthesized.
+     *
+     * Note: synthesized names are unique and match the regular expression: ^[A-Z]+[A-Z_0-9]*$
+     * See {@value EnumShapeValidator#RECOMMENDED_NAME_PATTERN}
      *
      * @param shape A base {@link StringShape} to convert.
      * @param synthesizeNames Whether names should be synthesized if possible.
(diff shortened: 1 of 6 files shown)


smithy-model/src/test/java/software/amazon/smithy/model/shapes/EnumShapeTest.java
```diff
@@ -495,4 +500,79 @@ public void canConvertToEnumWithNonConvertableNamelessEnumTrait() {
...
+    private static Stream<Arguments> testEnumValues() {
+        return Stream.of(
+            Arguments.of("a:b_", "a&b.", "a_b_0", true),
+            Arguments.of("Linux/UNIX", "Windows (Amazon VPC)", "Linux/UNIX (Amazon VPC)", true),
+            Arguments.of("a:b", "a&b", "a_b", true),
+            Arguments.of("a:b", "a&b", "a_b_0", true),
+            Arguments.of("a:b", "a_b_0", "a_b_0", true),
+            Arguments.of("a_b_0", "a_b_1", "a_b_2", true),
+            Arguments.of(":a", "&b", "a_b", true),
+            Arguments.of(":", "&", "c", false)
```
syall, line 562: Can we add a test case for numbers in strings, e.g. `"123"`?

mtdowling:
Is the synthesized enum name deterministic? Does the number padding follow the definition order? You provide this example:

```
a:b → A_B_0
a=b → A_B_1
a_b → A_B_2
```

What if they started out with this `a=b` and `a_b`, but later added `a:b`. Would that result in this?

```
a=b → A_B_0
a_b → A_B_1
a:b → A_B_2
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -355,9 +388,33 @@ public Builder setMembersFromEnumTrait(EnumTrait trait, boolean synthesizeNames)
...
+            if (synthesizeNames) {
+                Set<String> names = new HashSet<>();
+                members.forEach(m -> names.add(m.getMemberName()));
+                hasDuplicateMemberName = names.size() != members.size();
+            }
+
+
+            for (int i = 0; i < members.size(); i++) {
+                MemberShape member = members.get(i);
+                if (hasDuplicateMemberName) {
```
sugmanue, line 403: As this stands this change will penalize each name regardless of whether or not these  collide with another one. I think is better to just penalize those that do. For instance, if we have 

* `FOO`
* `FOO:BAR`
* `FOO=BAR`
* `BAR`

I think it would be better to get

* `FOO` → `FOO`
* `FOO:BAR` → `FOO_BAR_0`
* `FOO=BAR` → `FOO_BAR_1`
* `BAR` → `BAR`

Instead of

* `FOO` → `FOO_0`
* `FOO:BAR` → `FOO_BAR_1`
* `FOO=BAR` → `FOO_BAR_2`
* `BAR` → `BAR_3`

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -247,8 +255,16 @@ static Optional<MemberShape> memberFromEnumDefinition(
...
-        return definition.getName().isPresent()
-                || (withSynthesizedNames && CONVERTABLE_VALUE.matcher(definition.getValue()).find());
+        if (definition.getName().isPresent()) {
+            return true;
+        }
+
+        if (withSynthesizedNames) {
+            String syntheticName = synthesizeName(definition.getValue());
+            return RECOMMENDED_NAME_PATTERN.matcher(syntheticName).find();
+        }
```
sugmanue, line 265: I think we can do better than this, i.e., if `withSynthesizedNames` we should always be able to generate a valid name, I think if the name does not match the recommended pattern we should add a prefix to force it, e.g.,

* `0000` → `SYN_0000`
* `0001` → `SYN_0001`
* `0010`  → `SYN_0010`

I might be missing something but this is the only case I can think of.

PR #1615, opened 2023-02-13
Add shapes generation order in CodegenDirector

AndrewFossAWS:
*Description of changes:*
* Add shapes generation ordering in CodegenDirector
* This allows code generators to decide the ordering of the shapes which get passed to them. 
* E.g. `smithy-ruby` needs the shapes ordered alphabetically by their names because generated ruby code needs to be alphabetic in a ruby file.

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
@@ -16,6 +16,7 @@
 package software.amazon.smithy.codegen.core.directed;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.ServiceLoader;
@@ -26,6 +27,7 @@ import java.util.logging.Logger;
 import software.amazon.smithy.build.FileManifest;
 import software.amazon.smithy.codegen.core.CodegenContext;
 import software.amazon.smithy.codegen.core.ImportContainer;
+import software.amazon.smithy.codegen.core.ShapeGenerationOrder;
 import software.amazon.smithy.codegen.core.SmithyIntegration;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.codegen.core.SymbolWriter;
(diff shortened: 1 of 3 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapesGenerationOrdering.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Shapes ordered alphabetically by their names.
+     */
+    ALPHABETICAL,
+
+    /**
+     * Shapes ordered by default ordering.
+     */
+    DEFAULT,
```
mtdowling, line 38: Any need to have DEFAULT explicitly called out if DEFAULT == TOPOLOGICAL

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java
```diff
@@ -250,6 +253,22 @@ public void changeStringEnumsToEnumShapes(boolean synthesizeEnumNames) {
...
+     * Sets the shapes ordering for code generation.
+     *
+     * <p>CodegenDirector order the shapes appropriately before passing them to the code generators.
+     * The default ordering is topological, and can be overridden with this method
+     *
+     * @param ordering the ordering to use for the code generation process.
+     */
+    public void setShapesGenerationOrdering(ShapesGenerationOrdering ordering) {
+        if (Objects.isNull(ordering)) {
+            this.shapesGenerationOrdering = ShapesGenerationOrdering.NONE;
```
gosar, line 266: I'm not sure about this logic. Setting to null might imply someone wants the default behavior not necessarily NONE. There's no reason to allow setting this to null, I think. You can just not check for null here, similar to other "setter" methods and check in validateState that it is not null like the others. Also, following similar pattern to `model()`, etc. just name this method `shapesGeenerationOrdering`.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/ShapesGenerationOrdering.java
```diff
@@ -0,0 +1,39 @@
...
+
+package software.amazon.smithy.codegen.core;
+
+/**
+ * Shapes ordering for code generation.
+ *
+ * <p>CodegenDirector order the shapes appropriately before feeding them to the code generators. See {@link
+ * software.amazon.smithy.codegen.core.directed.CodegenDirector#setShapesGenerationOrdering(ShapesGenerationOrdering)}
+ */
+public enum ShapesGenerationOrdering {
```
gosar, line 24: naming is always hard. I'm thinking about, ordering v/s order. different people may have different opinions on this. I say consider `ShapesGenerationOrder`

gosar (approved):

Issue #1252, opened 2022-06-02
Support for generation of JSON Schema version 2020-12


ZackKanter:
The JSON Schema generator currently only [supports Draft 7](https://github.com/awslabs/smithy/blob/main/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/Schema.java#L41-L42). 

It would be great if it could also support [`2020-12`](https://json-schema.org/specification-links.html#published-drafts), which is the version we need. I would imagine others may need `2019-09`.

ZackKanter:
That would be fantastic.

ZackKanter:
Just checking back on this!

handrews:
JSON Schema spec co-author here: I definitely advise skipping 2019-09.  It has no advantages over 2020-12 and several disadvantages.  Plus 2020-12 is used by OpenAPI 3.1 which is another driver of adoption for this specific draft.

PR #1617, opened 2023-02-14
Add optional configuration parameter to select JSON schema version

hpmellema:
*Issue #, if available:* [issue #1252](https://github.com/awslabs/smithy/issues/1252)

*Description of changes*:
Adds an optional configuration parameter `jsonSchemaVersion` to select the version of JSON schema to use. The JSON schema version will still default to `draft07`, but can be set to `draft2020-12`. This new configuration parameter will also provide a mechanism to support future JSON schema changes.

Reviewing the JSON schema specifications, it appears that the only change needed was to update "definitions" -> "$defs" as Smithy jsonSchema does not support meta-schema generation, the use of plain-name fragments, tuples, or dynamic anchors.

*Cloudformation*
Cloudformation only supports jsonSchema `draft07` at this time (see [here](https://github.com/aws-cloudformation/cloudformation-template-schema/blob/25fdf51f246fff8129539fc8a19876bf9dc197cd/src/main/java/aws/cfn/codegen/json/SchemaDraft.java))
In order to prevent compatibility issues, a check was added to the cloudformation plugin to ensure only JSON Schema `draft07` is used in the configuration. I expect we will update this as Cloudformation and API Gateway add support for later versions.

*OpenAPI*
The OpenAPI plugin was updated to set the JSON Schema version parameter based on the latest version supported by the OpenAPI spec version. 
I.e. OpenAPI3.1 will use `draft2020-12` while OpenAPI3.0 will use `draft07`

*Testing* 
- manually tested that both OpenAPI and Cloudformation plugins correctly used the new configuration parameter and rejected unsupported versions/parameters
- Added a new integration test for `draft2020-12` in the JSON schema module
- Added test that OpenAPI config correctly associates with the right JSON schema version

*JSON Schema Change list*
The following two web pages detail the 
Schema Changes: 
- [draft07 -> draft2019-09](https://json-schema.org/draft/2019-09/release-notes.html)
- [draft2019-09 -> draft2020-12](https://json-schema.org/draft/2020-12/release-notes.html)


smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConfig.java
```diff
@@ -403,4 +404,24 @@ public boolean getUseIntegerType() {
     public void setUseIntegerType(boolean useIntegerType) {
         this.useIntegerType = useIntegerType;
     }
+
+    /**
+     * Set the JSON schema version to use when converting Smithy shapes into Json Schema.
```
srchase, line 409: nit: This looks looks like the text for the setter.

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaVersion.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
srchase, line 2: nit: Wrong year. [This](https://github.com/awslabs/smithy/pull/1614) updated checkstyle so we can use the short-form license header without the year now:

```
/*
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * SPDX-License-Identifier: Apache-2.0
 */
```

hpmellema: updated in cr 3f12c5b.

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaConfig.java
hpmellema: updated in cr 3f12c5b.

srchase (approved):

PR #1618, opened 2023-02-15
Fix OR condition in scoped attribute selector

AndrewFossAWS:
*Issue #, if available:*
A bug was discovered when using `,` (OR condition) in scoped attribute selector

```
enum Suit {
    DIAMOND = "diamond"
    CLUB = "club"
    HEART = "heart"
    SPADE = "spade"
}

enum Color {
    RED
    BLACK
}
```

`[@trait|enum|(values): @{name} ^= DIA, BLA]`  (return enums with names starting with `DIA` or `BLA`

Expected: `smithy.example#Suit`, `smithy.example#Color`

Acutal: `smithy.example#Suit`

*Description of changes:*
Fix OR condition in scoped attribute selector

### DIFF
--- a/smithy-model/src/test/resources/software/amazon/smithy/model/selector/enums.smithy
+++ b/smithy-model/src/test/resources/software/amazon/smithy/model/selector/enums.smithy
@@ -1,0 +1,17 @@
+$version: "1.0"
+
+namespace smithy.example
+
+@enum([
+    { name: "DIAMOND", value: "diamond"},
+    { name: "CLUB", value: "club"},
+    { name: "HEART", value: "heart"},
+    { name: "SPADE", value: "spade"},
+])
+string Suit
+
+@enum([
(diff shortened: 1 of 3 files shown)


smithy-model/src/test/resources/software/amazon/smithy/model/selector/enums.smithy
```diff
@@ -0,0 +1,15 @@
+$version: "2.0"
```
mtdowling, line 1: Since the test is using an `enum` trait, it would be more straightforward to use an enum trait in the model rather than rely on the synthetically created enum trait.

AndrewFossAWS: Updated it enum trait

mtdowling (approved):

PR #1626, opened 2023-02-21
Fix warnings in AST Loader for Resource and Operation Shapes with mixins

hpmellema:
*Description of changes:*
Updates the AST loader to allow both `Operation` and `Resource` shapes to have a `mixin` property without throwing a warning.

Previously the following warning was being emitted on building the source projection:
```
      |
50 |         "example.weather.operations#OperationWithMixin": {
      |                                                          ^
     = Expected an object with possible properties of `errors`, `input`, `output`, `traits`, `type`, but found additional properties: `mixins`
SUCCESS: Validated 299 shapes (WARNING: 1)
```


syall:
Looking at the [spec for mixins](https://smithy.io/2.0/spec/mixins.html), services and other shapes can also use mixins, but will need to make sure each of the shape types have the `addMixins()` call.

syall (approved):

PR #1628, opened 2023-02-22
Fix bug where mixin cycles were incorrectly detected

hpmellema:
**Description of bug**
A bug was discovered where mixin cycles would be incorrectly detected if a non structure or union shape had a trait that was removed by a transform.

*For example*:
The following smithy model
```smithy
// main.smithy
namespace example

@trait
@tags(["filtered"])
structure myTrait {}

@myTrait
operation OperationWithMixin with [ OperationMixin ] {
    input := {
        @required
        myInputField: String,
        other: String
    }
}
```
Using the following Transform: 
```json
{
  "version": "2.0",
  "projections": {
    "myProjection": {
      "transforms": [
       {
          "name": "excludeTraitsByTag",
          "args": {
            "tags": ["filtered"]
          }
        },
    ]}
  }
}
```
Would encounter a build exception like
```
Projection myProjection failed: software.amazon.smithy.model.loader.TopologicalShapeSort$CycleException: Mixin cycles detected among [example#OperationWithMixin]
software.amazon.smithy.model.loader.TopologicalShapeSort$CycleException: Mixin cycles detected among [example.#OperationWithMixin]
```

---
**Description of changes:**
- Adds new public methods to the Model class to access shapes using mixins and mixin shapes
- Uses the above methods to add all mixins and shapes with mixins to topological sorter in the ReplaceShapes class so that cycle detection can run correctly (previously only unions and structures were added)
- Adds tests for the specific error encountered above
- Adds simple unit test for methods added to Model class.


srchase (approved):

AndrewFossAWS (approved):

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -761,6 +762,26 @@ public <T extends Shape> Stream<T> shapes(Class<T> shapeType) {
...
+            .filter(shape -> !shape.getMixins().isEmpty());
+    }
+
+    /**
+     * Gets a stream of shapes defined as mixins.
+     *
+     * @return stream of shapes
+     */
+    public Stream<Shape> mixins() {
+        return shapeMap.values().stream().filter(shape -> shape.hasTrait(MixinTrait.class));
```
kstich, line 782: I don't think this needs a dedicated method, it's `getShapesWithTrait(MixinTrait.class).stream()`.

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -761,6 +762,26 @@ public <T extends Shape> Stream<T> shapes(Class<T> shapeType) {
         return toSet(shapeType).stream();
     }
 
+    /**
+     * Gets a stream of shapes that have mixins.
+     *
+     * @return stream of shapes
+     */
+    public Stream<Shape> shapesWithMixins() {
```
mtdowling, line 770: Given these are only used in one place and this class is already huge, I'd prefer to not add another utility method for this to Model.

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
```diff
@@ -761,6 +762,26 @@ public <T extends Shape> Stream<T> shapes(Class<T> shapeType) {
...
+            .filter(shape -> !shape.isMemberShape())
+            .filter(shape -> !shape.getMixins().isEmpty());
+    }
+
+    /**
+     * Gets a stream of shapes defined as mixins.
+     *
+     * @return stream of shapes
+     */
+    public Stream<Shape> mixins() {
```
mtdowling, line 781: I think a much better way for people to get this and any other set of shapes with a trait is: `model.getShapesWithTrait(MixinTrait.class)`. It's cached too. I'd rather we keep it to that single way to do it since it's already ingrained everywhere else.

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
hpmellema: Resolved in commit: 40775d9

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
hpmellema: Resolved in commit: 40775d9

smithy-model/src/main/java/software/amazon/smithy/model/Model.java
hpmellema: Resolved in commit: 40775d9

mtdowling (approved):

PR #1658, opened 2023-03-06
Do not return internal enum values

srchase:
This PR updates the enum validation protocol tests to exclude returning enum values that have been marked as internal with the `@internal` trait or tagged as `internal`.


gosar:
There was some concern on using `internal` tag as well in this (along with `@internal` trait), but we side towards using `internal` tags for this purpose too, as tags are also commonly used to mark shapes for internal use, and we'd rather not have a surprise for such users. If one wanted to use the `internal` tag for another purpose, they can use a different label.

PR #1672, opened 2023-03-11
Add process based plugins to Smithy build

mtdowling:
This commit adds the ability to integrate smithy build with process based plugins so that the Smithy CLI will be able to call other CLI tools through a process based plugin model.

To do this, I've introduced the `run` plugin which takes a `command` array containing the program to run and arguments to provide. The plugin will open a process and send the serialized Smithy model to the standard input of the process. The process can then perform its logic and write files in the working directory its called with.

A couple things are still missing here that we should figure out later:

1. The only way to see the output of a plugin is using --debug in the CLI. Should there be an output to somehow always show it? That will require some kind of hackery in the CLI I think.
2. There is no way currently to indicate the process is "serial" like how other Smithy Build plugins can. Serial indicates that a plugin requires some kind of global locking (like file locks) and can't be run in parallel with other plugins.

In addition to adding the `run` plugin, you can also apply the same plugin multiple times in a single projectiony by giving it a custom artifact name using the "plugin-name::artifact-name" syntax.

Example:

```
{
    "version": "1.0",
    "projections": {
        "source": {
            "plugins": {
                "run::hello": {
                    "command": ["hello.sh", "--arg", "arg-value"]
                }
            }
        }
    }
}
```

*Issue #, if available:*

*Description of changes:*


smithy-utils/src/main/java/software/amazon/smithy/utils/IoUtils.java
```diff
@@ -227,46 +228,90 @@ public static int runCommand(String command, Path directory, Appendable output,
...
+        }
+    }
+
+    private static void closeAndFlushStream(List<String> args, OutputStream outputStream) {
+        try {
+            outputStream.flush();
+            outputStream.close();
+        } catch (IOException e) {
+            // TODO: Why is this necessary?
+            LOGGER.log(Level.WARNING, "Unable to close and flush outputStream for " + args, e);
```
sugmanue, line 311: I'd say, if you decide to keep the flush, "flush and close" instead, as someone might think, oh, it's because you cannot flush after close :). But other than that I'm not sure why there's a `TODO` there, I guess it's necessary because any IO can throw, right?

smithy-utils/src/main/java/software/amazon/smithy/utils/IoUtils.java
```diff
@@ -227,46 +228,90 @@ public static int runCommand(String command, Path directory, Appendable output,
...
+
+            return process.waitFor();
         } catch (InterruptedException | IOException e) {
             throw new RuntimeException(e);
         }
     }
 
+    // Can be replaced with InputStream#transferTo in Java 9+.
+    private static void copyInputToOutput(InputStream source, OutputStream sink) throws IOException {
+        byte[] buffer = new byte[8192];
```
sugmanue, line 298: We already have a `BUFFER_SIZE` constant defined, can we use that instead here and update its value to be  8Kb?

docs/source-2.0/guides/building-models/build-config.rst
```diff
@@ -1665,6 +1719,115 @@ sources plugin. Lines that start with a number sign (#) are comments and are
...
+
+    * - Property
+      - Type
+      - Description
+    * - command
+      - ``[string]``
+      - **REQUIRED** The name of the program to run, followed by an optional
+        list of arguments. If the command uses a relative path, Smithy will
+        first check if the command can be found relative to the current working
+        directory. Otherwise, the program must be available on the ``$PATH`` or
```
sugmanue, line 1755: I think it will be worth adding a note along the lines of 
> ``$PATH`` (if using a Unix-like environment or its equivalent for other operating systems)

smithy-build/src/main/java/software/amazon/smithy/build/SmithyBuildImpl.java
```diff
@@ -48,7 +49,16 @@
 
 final class SmithyBuildImpl {
     private static final Logger LOGGER = Logger.getLogger(SmithyBuild.class.getName());
-    private static final Pattern PATTERN = Pattern.compile("^[A-Za-z0-9\\-_.]+$");
+
+    // Must start with one or more alphanumerics, followed by zero or more alphanumerics, "-", "_", or ".".
+    private static final String PATTERN_PART = "[A-Za-z0-9]+[A-Za-z0-9\\-_.]*";
```
sugmanue, line 54: We also need to update the pattern in [the docs](https://smithy.io/2.0/guides/building-models/build-config.html#build-info-plugin).

> Plugin names MUST match the following pattern: `^[A-Za-z0-9\-_.]+$.`

sugmanue (approved):
I left some minor notes.

smithy-utils/src/main/java/software/amazon/smithy/utils/IoUtils.java
```diff
@@ -227,46 +228,90 @@ public static int runCommand(String command, Path directory, Appendable output,
...
+        int read;
+        while ((read = source.read(buffer, 0, 8192)) >= 0) {
+            sink.write(buffer, 0, read);
+        }
+    }
+
+    private static void closeAndFlushStream(List<String> args, OutputStream outputStream) {
+        try {
+            outputStream.flush();
```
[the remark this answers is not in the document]
mtdowling: Ah, maybe the stream was already closed when I was calling flush which was the reason I needed this method in the first place

smithy-build/src/main/java/software/amazon/smithy/build/plugins/RunPlugin.java
```diff
@@ -0,0 +1,168 @@
...
+import software.amazon.smithy.model.shapes.ModelSerializer;
+import software.amazon.smithy.utils.IoUtils;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Runs a process-based plugin.
+ */
+public final class RunPlugin implements SmithyBuildPlugin {
+    private static final Logger LOGGER = Logger.getLogger(RunPlugin.class.getName());
+    private final Path root = Paths.get(".").toAbsolutePath().normalize();
```
sugmanue, line 45: We can use here instead `SmithyBuildUtils.getCurrentWorkingDirectory()`

smithy-build/src/test/java/software/amazon/smithy/build/SmithyBuildTest.java
```diff
@@ -658,4 +656,152 @@ public Node toNode() {
...
+    }
+
+    @Test
+    public void runsCommand() throws Exception {
+        Assumptions.assumeTrue(isPosix());
+
+        String output = runPosixTestAndGetOutput("run-plugin/run-plugin-test.json", "test-process");
+
+        assertThat(output, containsString("argv1: a"));
+        assertThat(output, containsString("SMITHY_ROOT_DIR: " + Paths.get(".").toAbsolutePath().normalize()));
```
sugmanue, line 695: Same a above, we can use instead `SmithyBuildUtils.getCurrentWorkingDirectory()`

smithy-build/src/test/java/software/amazon/smithy/build/model/SmithyBuildConfigTest.java
```diff
@@ -205,7 +205,7 @@ public void loadsFromNodeIgnoringBadSourceLocations() throws IOException {
                 .expectObjectNode();
         SmithyBuildConfig config = SmithyBuildConfig.fromNode(value);
 
-        Path cwd = Paths.get(".");
+        Path cwd = Paths.get(".").normalize().toAbsolutePath();
```
sugmanue, line 208: Same as above, we can use here `SmithyBuildUtils.getCurrentWorkingDirectory()`.

sugmanue (approved):
I just noted some duplication, not sure if those are per-design, i.e., part of the tests. I added a comments anyway.

smithy-build/src/main/java/software/amazon/smithy/build/plugins/RunPlugin.java
mtdowling: Can't use it here since it's package private right now. I'm ok with the duplication for now to spare needing to make it public.

smithy-build/src/test/java/software/amazon/smithy/build/model/SmithyBuildConfigTest.java
mtdowling: Was able to use it here. Updated.

smithy-build/src/test/java/software/amazon/smithy/build/SmithyBuildTest.java
mtdowling: Yeah keeping it package-private for now

sugmanue (approved):

PR #1693, opened 2023-03-24
Add includePreludeShapes in model plugin

AndrewFossAWS:
*Issue #, if available:*
https://github.com/awslabs/smithy/issues/947

*Description of changes:*
* Prelude serialization PR: https://github.com/awslabs/smithy/pull/955
* This PR allows prelude shapes to be included/excluded via the model plugin in smithy-build.json


mtdowling:
I think you shouldn’t be able to change the default model plugin. Instead, they can use a custom instance of the model plugin using the syntax introduced for process plugins “model::withPrelude” and then they can add the new property you introduced here.

smithy-build/src/main/java/software/amazon/smithy/build/plugins/ModelPlugin.java
```diff
@@ -34,10 +36,16 @@ public String getName() {
 
     @Override
     public void execute(PluginContext context) {
-        context.getFileManifest().writeJson("model.json", serializeModel(context.getModel()));
+        boolean includePrelude = false;
+        Optional<BooleanNode> node = context.getSettings().getBooleanMember("includePreludeShapes");
```
mtdowling, line 40: This seems off since you could set this to `false`. I'd use:

```
boolean includePrelude = context.getSettings().getBooleanMemberOrDefault("includePreludeShapes");
```

mtdowling (changes requested):

mtdowling (approved):

PR #1695, opened 2023-03-28
Improve CLI validation output

mtdowling:
This commit adds more colorful and easier to scan validation output for the CLI.

To support this change, I needed to introduce a new validation event formatter that uses ANSI codes. Updating the existing formatter would have added a ton of composition to it, making it hard to understand.

When breaking out a new validation event formatter, I noticed some duplicate code in handling: 1) creating a human-readable filename 2) loading a line of text from a source location.

To account for (1), I added getHumanReadableFilename() to ValidationEvent.

To account for (2), I created a new abstraction for grabbing source context from a SourceLocation.

I ended up needing an abstraction for emitting lots of content that's wrapped in ANSI color escapes, and I didn't want to add escapes over and over or have to buffer intermediate output to a string, so I added a method for wrapping a Consumer<Appendable> that handles adding ANSI styles before and after the consumer is called.

Style was updated to an interface in case we ever want to support theming and use HEX code colors.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-cli/src/test/java/software/amazon/smithy/cli/AnsiColorFormatterTest.java
+++ b/smithy-cli/src/test/java/software/amazon/smithy/cli/AnsiColorFormatterTest.java
@@ -1,0 +1,47 @@
+package software.amazon.smithy.cli;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class AnsiColorFormatterTest {
+    @Test
+    public void detectsIfColorIsEnabled() {
+        assertThat(AnsiColorFormatter.NO_COLOR.isColorEnabled(), is(false));
+        assertThat(AnsiColorFormatter.FORCE_COLOR.isColorEnabled(), is(true));
+    }
+
+    @Test
+    public void wrapsConsumerWithColor() {
+        ColorFormatter formatter = AnsiColorFormatter.FORCE_COLOR;
+        StringBuilder builder = new StringBuilder();
+
+        formatter.style(builder, b -> {
+            b.append("Hello");
+        }, Style.RED);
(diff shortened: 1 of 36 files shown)


kstich (approved):

smithy-model/src/main/java/software/amazon/smithy/model/loader/sourcecontext/DefaultSourceLoader.java
```diff
@@ -0,0 +1,229 @@
...
+    private List<Line> parseTraitAfterShape(SourceLocation location, Shape targetShape) {
+        List<Line> result = new ArrayList<>(2);
+        addLineIfValid(targetShape.getSourceLocation().getLine(), result);
+        addLineIfValid(location.getLine(), result);
+        return result;
+    }
+
+    private List<Line> parseMemberShape(SourceLocation location, MemberShape member) {
+        // Members should always crawl up to the defining shape in both the IDL and JSON.
+        Shape container = model.getShape(member.getContainer()).orElse(null);
```
JordonPhillips, line 195: Is there a reason you'd expect the container to not be present, or is this just in case?

JordonPhillips (changes requested):

smithy-cli/src/test/java/software/amazon/smithy/cli/AnsiColorFormatterTest.java
```diff
@@ -0,0 +1,47 @@
+package software.amazon.smithy.cli;
```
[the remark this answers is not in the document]
mtdowling: I think it's silly to include license headers on tests, but I'll add the license since I don't know what the policy here is.

smithy-model/src/main/java/software/amazon/smithy/model/loader/sourcecontext/DefaultSourceLoader.java
mtdowling: Yeah just in case. I added logging here and a comment.

PR #1708, opened 2023-04-01
Gracefully handle the case when additionSchema targets an invalid shape

sugmanue:
*Description of changes:*

Update the logic that gets process the shape of the `additionalSchema` property to avoid attempting to dereference a non-structure shape. Without this the `Optional::get` throws and exception when the shape is not an `structure` and hides out the actual error.

### DIFF
--- a/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceIndex.java
+++ b/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceIndex.java
@@ -150,11 +150,12 @@ public final class CfnResourceIndex implements KnowledgeIndex {
                         // setting, but gracefully handle if they're not.
                         model.getShape(additionalSchema)
                                 .map(Shape::asStructureShape)
-                                .map(Optional::get)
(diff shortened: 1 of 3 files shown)


smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceIndex.java
```diff
@@ -150,11 +150,12 @@ public CfnResourceIndex(Model model) {
                         // setting, but gracefully handle if they're not.
                         model.getShape(additionalSchema)
                                 .map(Shape::asStructureShape)
```
mtdowling, line 152: could this just be a flatMap?

mtdowling (changes requested):

smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourceIndex.java
sugmanue: Indeed, updated.

mtdowling (approved):

PR #1710, opened 2023-04-02
Use pretty validation output with smithy diff

mtdowling:
Smithy diff now uses the pretty validation output.

The old and new models are both validated, but only build-failing errors are shown. When the original model fails, a label is added before ERROR/DANGER to show it's on the OLD model. When the new model fails, it shows a NEW label.

When running the diff with two valid models, events have a label of DIFF. The events shown here use the --severity option and default to WARNING when not set.

image

image

image

With `--no-color`:

image


mtdowling:
Replaced with https://github.com/awslabs/smithy/pull/1712

PR #1716, opened 2023-04-04
Fix #1675

### LINKED ISSUE #1675: Model loading - determinism
Hi! My team has struggled with some false positives for `ModelDiff` saying there are changes, even when the input files haven't changed between two versions of the model.

We've narrowed it down to a difference in metadata merged from multiple files. It seems like, depending on the location on disk of the files being parsed (even if they're in the same relative directories, they may have a different base directory - e.g. in CI runs this may differ from run to run).

You can see this behavior here ([runnable online on Scastie](https://scastie.scala-lang.org/dcYX6sW3TeyuXGPH2RVnIQ)):

```scala

//> using scala "2.13.10"
//> using lib "software.amazon.smithy:smithy-model:1.28.0"
//> using lib "software.amazon.smithy:smithy-diff:1.28.0"
import scala.jdk.CollectionConverters._
import scala.util.Random
import software.amazon.smithy.diff.ModelDiff

def loadModel(prefix: String) = {
  val assembler = Model
    .assembler()

  assembler.addUnparsedModel(
    s"$prefix/smithy/first/a.smithy",
    """metadata items = ["item-2"]""",
  )
  assembler.addUnparsedModel(
    s"$prefix/smithy/second/b.smithy",
    """metadata items = ["item-3"]""",
  )
  assembler.addUnparsedModel(
    s"$prefix/smithy/a.smithy",
    """metadata items = ["item-1"]""",
  )

  assembler.assemble().unwrap()
}

def metas(model: Model) =
  model
    .getMetadata()
    .asScala
    .apply("items")
    .expectArrayNode()
    .getElements()
    .asScala
    .map(elem => elem.expectStringNode().getValue())
    .toList

ModelDiff.compare(loadModel("/Users/kubukoz/projects/demos"), loadModel("")).asScala.toList.head
// res0: software.amazon.smithy.model.validation.ValidationEvent = [WARNING] -: Metadata key `items` was changed from [
//     "item-1",
//     "item-2",
//     "item-3"
// ] to [
//     "item-3",
//     "item-2",
//     "item-1"
// ] | ChangedMetadata /smithy/a.smithy:1:18
```

I'm wondering if it's something that should be expected.

I suppose users shouldn't rely on metadata having any specific order - is that true?
If that's the case, should `ModelDiff` be adjusted to ignore changes in ordering of metadata on the top level, or should I adjust that on my side? Thanks!

miguel-vila:
Fixes #1675

*Description of changes:*

Makes it so source file names don't affect the order of the built metadata.

### DIFF
--- a/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
@@ -47,6 +47,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
@@ -405,8 +407,8 @@ public class ModelAssemblerTest {
                 .addUnparsedModel("b3.smithy", "metadata items = [3]")
                 .assemble()
                 .unwrap();
-        List<Number> metadata1 = model1.getMetadata().get("items").expectArrayNode().getElements().stream().map(s -> s.expectNumberNode().getValue()).toList();
-        List<Number> metadata2 = model2.getMetadata().get("items").expectArrayNode().getElements().stream().map(s -> s.expectNumberNode().getValue()).toList();
+        List<Number> metadata1 = model1.getMetadata().get("items").expectArrayNode().getElements().stream().map(s -> s.expectNumberNode().getValue()).collect(Collectors.toList());
+        List<Number> metadata2 = model2.getMetadata().get("items").expectArrayNode().getElements().stream().map(s -> s.expectNumberNode().getValue()).collect(Collectors.toList());
         assertThat(metadata1, is(metadata2));
     }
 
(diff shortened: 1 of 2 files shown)


smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
```diff
@@ -391,6 +391,26 @@ public void detectsMetadataConflicts() {
...
+                .addUnparsedModel("a3.smithy", "metadata items = [3]")
+                .assemble()
+                .unwrap();
+        Model model2 = new ModelAssembler()
+                .addUnparsedModel("b1.smithy", "metadata items = [1]")
+                .addUnparsedModel("b2.smithy", "metadata items = [2]")
+                .addUnparsedModel("b3.smithy", "metadata items = [3]")
+                .assemble()
+                .unwrap();
+        List<Number> metadata1 = model1.getMetadata().get("items").expectArrayNode().getElements().stream().map(s -> s.expectNumberNode().getValue()).toList();
```
mtdowling, line 408: The test is failing because we currently support Java 8 :(

This will need to use `.collect(Collectors.toList())`

mtdowling (changes requested):

mtdowling (approved):

PR #1717, opened 2023-04-04
Remove hidden --discover option from some commands

mtdowling:
The --discover and --discover-classpath options are only necessary for smithy build and smithy validate to support the Gradle pluing and the internal build tool within Amazon. This commit removes those options from ast, diff, migrate, and select, and removes / fixes tests that relied on them.

We should figure out if we can remove these options from build/validate or if we're comfortable with them as public options.

### DIFF
--- a/smithy-cli/src/test/java/software/amazon/smithy/cli/commands/MigrateCommandTest.java
+++ b/smithy-cli/src/test/java/software/amazon/smithy/cli/commands/MigrateCommandTest.java
@@ -109,11 +109,9 @@ public class MigrateCommandTest {
         SmithyCli.create().run(
                 "upgrade-1-to-2",
                 "--config", config.toString(),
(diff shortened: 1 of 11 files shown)


smithy-cli/src/test/java/software/amazon/smithy/cli/commands/MigrateCommandTest.java
```diff
@@ -109,11 +109,9 @@ public void testUpgradeDirectoryWithJar() throws Exception {
         SmithyCli.create().run(
                 "upgrade-1-to-2",
                 "--config", config.toString(),
-                "--discover-classpath", tempDir.toAbsolutePath().resolve("jar-import.jar").toString(),
```
mtdowling, line 112: Note: this should have actually used "imports" instead

JordonPhillips (approved):

PR #1721, opened 2023-04-05
Add --mode arbitrary|project to smithy diff

mtdowling:
This commit adds a --mode option to Smithy diff to support the arbitrary mode (the current mode used to compare two models), and the project mode used to compare the current project against another project or model. When running in project mode, the imports and sources of the current model are used to load the "new" model (the current project), but they aren't used to load the "old" model. If the --old argument points to a directory that contains a smithy-build.json file, its imports and sources are used when loading the old model, though its dependencies are ignored and it's loaded using the resolved classpath of the new model. This ensures that the models are comparable and won't cause comparison issues when trying to compare things like traits across class loaders.


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DiffCommand.java
```diff
@@ -51,7 +52,7 @@ public String getName() {
 
     @Override
     public String getSummary() {
-        return "Compares two Smithy models and reports differences.";
+        return "Compares two Smithy models and detects differences.";
```
sugmanue, line 55: The previous wording reads more accurate to me, as detecting does not imply reporting.

sugmanue (approved):

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DiffCommand.java
mtdowling: Changed it back to reports

PR #1723, opened 2023-04-05
Add --flatten flag to the AST command

sugmanue:
*Description of changes:*

Adds a new `--flatten` flag to the AST command that will flatten and remove the mixins from the model. This helps a lot with understanding how the model looks like without any mixins.


mtdowling (changes requested):
Can you add an integration test?

sugmanue:
> Can you add an integration test?

Done, added tests to verify that no flatten is done without the option and that the flatten is done with it.

mtdowling (approved):

PR #1733, opened 2023-04-14
Add rudimentary form of error recovery to parser

mtdowling:
When we encounter a parser error, we'll now attempt some basic error recovery by simply skipping token until we reach a token at the start of a line that is an identifier, documentation comment, or '@'. This should help with things like LSPs that still should be able to utilize jump to definition even with a broken or in-progress model.

(still need to test the experience out in the LSP before merging)

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
@@ -232,44 +232,48 @@ final class IdlModelLoader {
         Set<CharSequence> definedKeys = new HashSet<>();
 
         while (tokenizer.getCurrentToken() == IdlToken.DOLLAR) {
-            tokenizer.next();
-            tokenizer.expect(IdlToken.IDENTIFIER, IdlToken.STRING);
-            String key = tokenizer.internString(tokenizer.getCurrentTokenStringSlice());
+            try {
+                tokenizer.next();
+                tokenizer.expect(IdlToken.IDENTIFIER, IdlToken.STRING);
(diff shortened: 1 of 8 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
```diff
@@ -415,51 +415,84 @@ private void parseApplyStatement() {
...
+            if (tokenizer.getCurrentTokenColumn() == 1 && isErrorRecoveryToken(token)) {
+                break;
+            }
+        } while (tokenizer.hasNext());
+    }
+
+    // These tokens are good signals that the next shape is starting.
+    private boolean isErrorRecoveryToken(IdlToken token) {
+        return token == IdlToken.IDENTIFIER || token == IdlToken.DOC_COMMENT || token == IdlToken.AT;
+    }
```
JordonPhillips, line 494: I think `IdlToken.DOLLAR` should be added here too so we can recover in the middle of the control section

smithy-model/src/test/resources/software/amazon/smithy/model/loader/error-recovery.smithy
JordonPhillips, line 1: This should also have cases for recovering at the other possible "error recovery token" types

JordonPhillips (changes requested):

smithy-model/src/test/resources/software/amazon/smithy/model/loader/error-recovery.smithy
mtdowling: Added

PR #1739, opened 2023-04-18
Convert to internal trait from tag

srchase:
This PR updates how EnumShapes are converted from strings with the deprecated `@enum` trait by adding the `@internal` trait to the resulting Enum shape member when when the source enum value was tagged with `internal`.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
@@ -29,6 +29,7 @@ import software.amazon.smithy.model.traits.DocumentationTrait;
 import software.amazon.smithy.model.traits.EnumDefinition;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.EnumValueTrait;
+import software.amazon.smithy.model.traits.InternalTrait;
 import software.amazon.smithy.model.traits.TagsTrait;
 import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.model.traits.UnitTypeTrait;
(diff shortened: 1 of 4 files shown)


jvschneid (approved):

gosar:
We should document this behavior too. I think add javadoc to relevant methods in `EnumShape`, `ModelTransformer`.

I also noticed we have `ChangeStringEnumsToEnumShapes` but it is not documented in https://smithy.io/2.0/guides/building-models/build-config.html#transforms - it should be added there and this behavior documented there - this one can be separate PR.

gosar (approved):

mtdowling (approved):

srchase:
> I also noticed we have `ChangeStringEnumsToEnumShapes` but it is not documented in https://smithy.io/2.0/guides/building-models/build-config.html#transforms - it should be added there and this behavior documented there - this one can be separate PR.

Added separate PR in https://github.com/awslabs/smithy/pull/1740

smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -199,6 +200,12 @@ public static boolean canConvertToEnum(StringShape shape, boolean synthesizeEnum
     /**
      * Converts an enum definition to the equivalent enum member shape.
      *
+     * <p>If an enum definition is marked as deprecated, the DeprecatedTrait
+     * is applied to the converted enum member shape.
+     *
+     * <p>If an enum definition has an "internal" tag, the InternalTrait is
+     * applied ot the converted enum member shape.
```
gosar, line 207:
```suggestion
     * <p>If an enum definition has an "internal" tag, the InternalTrait is
     * applied to the converted enum member shape.
```

gosar (approved):

hpmellema (approved):

PR #1742, opened 2023-04-20
Revise how documentation comments are parsed

mtdowling:
This commit makes a minor change to how documentation comments work: rather than ignoring invalid documentation comments that don't appear before a shape, the parser will now fail. This simplifies parsing and helps modelers know when they mistakenly place documentation comments in the wrong place.

*Issue #, if available:*

*Description of changes:*


mtdowling:
This is a big improvement, but unfortunately it's also too big of a breaking change.

PR #1756, opened 2023-04-26
Fix inline io parsing with mixins

milesziemer:
When there was no space between the `:=` and `with` tokens when parsing an inline io statement, for example:
```
operation Foo {
  input :=with [Bar] {}
}
```
the parser would skip the `with` and so not realize its parsing a mixin, failing when it sees a `[` instead of a `{`. This was due to an extra call to `tokenizer.next()`, which skipped the `with`. Usually the `next()` would just skip the space before the `with`, but when the space wasn't present it skipped the `with`.

The inline-io test was updated to include this case.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
@@ -1083,7 +1083,6 @@ final class IdlModelLoader {
             }
             // Remove any pending, invalid docs that may have come before the inline shape.
             tokenizer.removePendingDocCommentLines();
-            tokenizer.next();
             // don't skip docs here in case there are docs on the inlined structure.
             tokenizer.skipWs();
             consumer.accept(parseInlineStructure(id.getName() + suffix, defaultTrait));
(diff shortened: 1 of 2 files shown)


AndrewFossAWS (approved):

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
```diff
@@ -1041,6 +1041,8 @@ private void parseOperationStatement(ShapeId id, SourceLocation location) {
                 case "input":
                     IdlToken nextInput = tokenizer.expect(IdlToken.COLON, IdlToken.WALRUS);
                     tokenizer.next();
+//                    tokenizer.skipSpaces();
```
hpmellema, line 1044: Remove?

mtdowling, line 1044: These two lines can be deleted no?

hpmellema (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
milesziemer: Oops, yes, thanks for catching that

milesziemer: Yes, forgot to remove them. Done

hpmellema (approved):

PR #1759, opened 2023-05-03
Add restXml protocol test for http label escaping

milesziemer:
*Issue #, if available:*

*Description of changes:*

Adds an httpRequestTests case to restXml protocol tests to verify that httpLabel bindings in the request URI are percent-encoded. restJson1 had an existing test case for this, but was updated to verify the space character is encoded. An operation was also added to the testing S3 service model with httpRequestTests to verify the same thing, but with S3's virtual host addressing. No other protocols had tests added, because restJson1 and restXml are the only protocols that support httpLabel bindings.

Changes were tested by generating and running protocol tests in aws-sdk-go-v2 and aws-sdk-js-v3, confirming the new test case is generated and runs successfully.

### DIFF
--- a/smithy-aws-protocol-tests/model/restXml/services/s3.smithy
+++ b/smithy-aws-protocol-tests/model/restXml/services/s3.smithy
@@ -17,6 +17,7 @@ use aws.api#service
 use aws.auth#sigv4
 use aws.customizations#s3UnwrappedXmlOutput
 use aws.protocols#restXml
+use aws.protocols#httpChecksum
 use smithy.test#httpRequestTests
 use smithy.test#httpResponseTests
 
@@ -42,6 +43,7 @@ service AmazonS3 {
     operations: [
         ListObjectsV2,
         GetBucketLocation,
+        DeleteObjectTagging
     ],
 }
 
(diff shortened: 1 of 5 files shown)


smithy-aws-protocol-tests/model/restXml/services/s3.smithy
```diff
@@ -272,6 +274,47 @@ operation ListObjectsV2 {
...
+        uri: "/my%20key.txt",
+        host: "s3.us-west-2.amazonaws.com",
+        resolvedHost: "mybucket.s3.us-west-2.amazonaws.com",
+        body: "",
+        queryParams: [
+            "tagging"
+        ],
+        params: {
+            Bucket: "mybucket",
+            Key: "my key.txt"
```
sugmanue, line 295: Can we add another test with a path in the key? Something like `foo/bar/my key.txt`. I'd expect this to be encoded as `foo/bar/my%20key.txt` and not `foo%2Fbar%2Fmy%20key.txt`, is that the case?

milesziemer: Your expectation is correct, I agree that's a good test to add, thanks for calling it out. I pushed an update including that test.

kstich (approved):

PR #1760, opened 2023-05-03
Avoid using String#chars() which is java 9+

sugmanue:
*Description of changes:*

~~Remove the use of the `chars()` method that is documented to be only for 9+ version.~~
(There is a default implementation for this on `CharSequence` in Java 8)

Possibly still relevant to remove the use of streams.


sugmanue:
🤦 the logic is not equivalent, is testing for `> 1` not `> 0`.

PR #1762, opened 2023-05-04
Defer projection failure until after plugins run

milesziemer:
*Issue #, if available:*
#1277 

*Description of changes:*
This changes `SmithyBuildImpl` to continue applying plugins after one fails, throwing the error (if present) after they have all completed. This is useful for example when you want to see the serialized output of a model you know is valid but a plugin is causing the whole build to fail. A new test case was added to ensure that artifacts produced by valid plugins are still created despite the build failing.

This was originally implemented in #1416, but was rolled back in #1429 as a precaution since we had an unrelated issue ocurring at the time.


mtdowling (approved):
LGTM. I think we could explore other options in the future, like: (a) making an aggregate exception to hold all encountered exceptions (b) making exceptions not throw but instead get added to the ProjectionResult, and maybe add an unwrap method to it.

PR #1767, opened 2023-05-11
Support `externalDocs` when converting operations to OpenAPI

Xtansia:
*Description of changes:*

Supports converting the `@externalDocumentation` trait on operations into the `externalDocs` property when converting to OpenAPI.


JordonPhillips:
Thanks for the contribution! I was curious why we had that method if we weren't using it, but it turns out we *do* use it... on basically every other shape type. Whoops!

JordonPhillips (approved):

PR #1774, opened 2023-05-17
Add validation events decorators

sugmanue:
*Description of changes:*

Adds a new interface `ValidationEventDecorator` that libraries can implement and plug into Smithy by making those discoverable using the java builtin [`ServiceLoader`](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html). All of the `ValidationEventDecorator` instances are loaded and each `ValidationEvent` is passed to each of the decorators. The decorators can make use of the `ValidationEvent#eventId()` method to quickly filter out the events that it knows how to decorate form the events it does not care about, and the `ValidationEvent#toBuilder()` to create a builder that can be used to mutate the instance.

The current use case is to add hints to a know set of events to point the user into the right solution for it (e.g., you need to pull X or Y dependency for this unresolved shape). Since often these solutions might require specific details about the environment in which Smithy is being used, those decorators cannot be part of Smithy itself but now can be implemented by libraries and loaded automatically by Smithy in a similar way in which validators are currently loaded.

For instance, a decorator that suggest the user to use or remove unused structures might look like this

```java
    @Override
    public ValidationEvent decorate(ValidationEvent ev) {
        if (ev.containsId("UnreferencedShape")) {
            if (ev.getMessage().contains("The structure ")) {
                return ev.toBuilder()
                    .hint("Consider using this structure in your service or remove it from the model")
                    .build();
            }
        }
        return ev;
    }
```


smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationEventDecorator.java
```diff
@@ -0,0 +1,34 @@
...
+package software.amazon.smithy.model.validation;
+
+/**
+ * Validation events decorators take validations events and transform them by adding more contextual information,
+ * usually adding a hint to let the user know what can it be done to solve the issue. This might add context specific
+ * information that is not relevant for all cases such as links to internal knowledge sites or explicit instructions
+ * relevant only to the context where Smithy is being used.
+ */
+@FunctionalInterface
+public interface ValidationEventDecorator {
```
kstich, line 25: Should this have a separate predicate method? That would enable a separation of concerns, make `decorate` only do actual decoration.

sugmanue: Makes sense, I changed it accordingly by adding a new `canDecorate(ValidationEvent)` method.

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationEventDecorator.java
```diff
@@ -0,0 +1,40 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.model.validation;
+
+/**
+ * Validation events decorators take validations events and transform them by adding more contextual information,
```
mtdowling, line 19:
```suggestion
 * Validation event decorators take validation events and transform them by adding more contextual information,
```

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -214,6 +245,39 @@ public Validator createValidator() {
...
+                    ValidationEvent decoratedEvent = event;
+                    for (ValidationEventDecorator decorator : decorators) {
+                        if (decorator.canDecorate(event)) {
+                            decoratedEvent = decorator.decorate(event);
+                        }
+                    }
+                    decorated.add(decoratedEvent);
+                }
+                return decorated;
+            } catch (Throwable e) {
```
mtdowling, line 273: Will this be a common issue? Why don't we expose these exceptions to the end user when they occur?

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -214,6 +245,39 @@ public Validator createValidator() {
...
 
+    /**
+     * Applies each of the given decorators to each of the validation events and returns the resulting list of
+     * validation events.
+     *
+     * @param decorators The decorators to apply
+     * @param events The events to decorate
+     * @return The decorated events
+     */
+    public static List<ValidationEvent> decorateEvents(
```
mtdowling, line 256: nit - this class is package private, so no need for public methods unless you're implementing an interface

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -214,6 +245,39 @@ public Validator createValidator() {
...
+                    for (ValidationEventDecorator decorator : decorators) {
+                        if (decorator.canDecorate(event)) {
+                            decoratedEvent = decorator.decorate(event);
+                        }
+                    }
+                    decorated.add(decoratedEvent);
+                }
+                return decorated;
+            } catch (Throwable e) {
+                LOGGER.log(Level.WARNING, e, () -> "A validation events decorator throw an exception, "
```
mtdowling, line 274:
```suggestion
                LOGGER.log(Level.WARNING, e, () -> "A validation event decorator threw an exception, "
```

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -93,6 +97,29 @@ public ModelValidator addValidator(Validator validator) {
         return this;
     }
 
+    /**
+     * Adds a custom {@link ValidationEventDecorator} to the ModelValidator.
+     *
+     * @param decorator Decorator to add.
+     * @return Returns the ModelValidator.
+     */
+    public ModelValidator addValidationEventDecorator(ValidationEventDecorator decorator) {
```
mtdowling, line 106: Since this class is package-private, you could probably get rid of this method or the other method and just keep the method that's actually used by ModelAssembler.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -214,6 +245,39 @@ public Validator createValidator() {
...
+     * @param events The events to decorate
+     * @return The decorated events
+     */
+    public static List<ValidationEvent> decorateEvents(
+        List<ValidationEventDecorator> decorators,
+        List<ValidationEvent> events
+    ) {
+        if (!decorators.isEmpty()) {
+            try {
+                List<ValidationEvent> decorated = new ArrayList<>(events.size());
```
mtdowling, line 262: Can we assume the array is mutable? If the list has a size > 0, I think we can, right? It's not like this method is getting called by something we don't control since it's package private.

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationEventDecoratorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceLoader;
+
+/**
+ * Creates {@link ValidationEventDecorator} instances.
+ */
+public interface ValidationEventDecoratorFactory {
```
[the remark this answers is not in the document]
sugmanue: That was my first take on this (e.g., the first [commit](https://github.com/awslabs/smithy/pull/1774/commits/110b1010d3826616fe5359e28b51d3131c7284f6#diff-fa50f59b51956a1b4f6c0286be3eab280dd0eae39c95ed753ac0c3437c07e466R42-R50) of this PR), however some of the validation events that we're interested in (e.g., undefined trait), are returned earlier [here](https://github.com/awslabs/smithy/blob/7342b2c5168dc868c49241e22632a18ca63438bb/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java#L574-L576) before the validation phase, thus I decoupled the decoration from the validation phase.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
sugmanue: I don't know if this will be common and we could expose the exceptions, however I'm not sure if breaking the build in potentially hard to solve way (e.g., the library is already part of the build dependency) is going to be a good user experience.

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationEventDecoratorFactory.java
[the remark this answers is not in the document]
mtdowling: I'm not sure that I follow how that impacts what creates the decorators.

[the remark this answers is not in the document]
sugmanue: It does not, we could use the `ValidationFactory`, it's just that we might need to pull up its default value into the assembler instead of having it inside the `ModelValidator` class [here](https://github.com/awslabs/smithy/blob/7342b2c5168dc868c49241e22632a18ca63438bb/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java#L167-L169), and then pass it down. I can make that change.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
sugmanue: Yeah, I think we can assume that, let me change the code to update the slot instead of copying the values.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
sugmanue: I think we need to have both if we want to allow the user to configure directly in the assembler the decorators since we need to pass down to the `ModelValidator` the configured values + the loaded ones, or none if we are willing to go without this. I could go with none for the time being and add it if and when the need arises.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java
```diff
@@ -587,23 +630,36 @@ public ValidatedResult<Model> assemble() {
...
-                .collect(Collectors.toList()));
+    private ValidatedResult<Model> returnOnlyErrors(
+        Model model,
+        List<ValidationEvent> events,
+        List<ValidationEventDecorator> decorators
+    ) {
+        List<ValidationEvent> filteredEvents = events.stream()
+                  .filter(event -> event.getSeverity() == Severity.ERROR)
+                  .collect(Collectors.toList());
+        return new ValidatedResult<>(model, ModelValidator.decorateEvents(decorators, filteredEvents));
```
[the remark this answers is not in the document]
sugmanue: done.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
sugmanue: I did remove methods to manually add decorators from the model assembler and validator for the time being, we can add those back if needed.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
sugmanue: Changed to use the original list overwriting the value in place.

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidatorFactory.java
```diff
@@ -63,18 +72,29 @@ public interface ValidatorFactory {
      * @param services ValidatorService instances to use to create validators.
      * @return Returns the created ValidatorFactory.
      */
-    static ValidatorFactory createServiceFactory(Iterable<Validator> validators, Iterable<ValidatorService> services) {
+    static ValidatorFactory createServiceFactory(
```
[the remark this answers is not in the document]
sugmanue: It is indeed but this method is only package private and only used within this class. I can make the change but I don't think we need/have to keep this backwards compatible but I might be missing something.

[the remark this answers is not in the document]
sugmanue: Sorry, my bad, is a public, I will make the change.

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidatorFactory.java
```diff
@@ -39,6 +39,15 @@ public interface ValidatorFactory {
      */
     List<Validator> loadBuiltinValidators();
 
+    /**
+     * Returns a list of decorators.
+     *
+     * @return Returns the loaded decorators.
+     */
+    default List<ValidationEventDecorator> loadBuiltinDecorators() {
```
mtdowling, line 47: I think this should just be `loadDecorators`. The "builtin" nomenclature was used to differentiate between validators that can be conditionally applied and configured vs validators that are always applied.

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidatorFactory.java
sugmanue: Updated in the latest commit.

sugmanue:
Approved offline after the last requested change to change `loadBuiltinDecorators` to `loadDecorators`.

PR #1781, opened 2023-05-21
New diff evaluator for added required members

rchache:
*Issue #, if available:*

*Description of changes:*
Validates that no members are newly created with the required trait (but no default trait) in existing structures.
Emits an event "AddedRequiredMember" for such cases

### DIFF
--- a/smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/AddedRequiredMemberTest.java
+++ b/smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/AddedRequiredMemberTest.java
@@ -1,0 +1,86 @@
+package software.amazon.smithy.diff.evaluators;
+
+import org.junit.jupiter.api.Test;
+import software.amazon.smithy.diff.ModelDiff;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.StringNode;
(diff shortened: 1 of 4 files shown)


smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/AddedRequiredMemberTest.java
```diff
@@ -0,0 +1,86 @@
...
+                .build();
+        Model model1 = Model.builder().addShapes(s, a).build();
+        Model model2 = Model.builder().addShapes(s, b).build();
+        ModelDiff.Result result = ModelDiff.builder().oldModel(model1).newModel(model2).compare();
+
+        assertThat(TestHelper.findEvents(result.getDiffEvents(), Severity.ERROR).size(), equalTo(1));
+        assertThat(TestHelper.findEvents(result.getDiffEvents(), "AddedRequiredMember").size(), equalTo(1));
+        assertThat(TestHelper.findEvents(result.getDiffEvents(), "AddedRequiredMember").get(0).getShapeId().get().toString(),
+                equalTo("smithy.example#A$foo"));
+        assertThat(TestHelper.findEvents(result.getDiffEvents(), "AddedRequiredMember").get(0).getMessage(),
```
sugmanue, line 34: Nit, I think asserting that specific messages is fragile, since improving it causes tests to fail.

sugmanue (approved):

PR #1782, opened 2023-05-22
Use structure member parsing for lists and maps

milesziemer:
This commit moves the validation of list and map members out of the parser. List and map members have special validation, because they can only have specific members, and those members can't have default values. Prior to this commit, we did this validation in the parser, throwing a parse error if default value syntax sugar was used, or incorrect member names were used. This meant we needed special parsing for list and map shape members. It also lead to situations like the following:
```
list Foo {
  member: String = "foo"
}
```
fails to parse, saying it expected a "}" but got a "=", while
```
list Foo {
  @default("foo")
  member: String
}
```
fails model validation, explaining you can't assign a default to a list member. Removing special handling of list and map members consolidates this logic and simplifies the parser.

Because we previously threw parse errors when encountering unexpected members, this PR had to update CollectionShape and MapShape to throw SourceExceptions, which become validation events, when the incorrect members were added to their respective builders. Since both shapes constructors require the builder as an argument, you can't construct the shapes with the wrong members.

For testing, cases were added for invalid map and list members, and for default value syntax on list and map members. Some old test cases that expected the parse error were removed or adjusted. The AST didn't previously throw errors when extra members were encountered, so this PR maintains that behavior.


smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/json-invalid-list-and-map-members.json
```diff
@@ -0,0 +1,77 @@
...
+    },
+    "com.test#MissingKeyMap": {
+      "type": "map",
+      "value": {
+        "target": "smithy.api#String"
+      }
+    },
+    "com.test#WrongMemberList": {
+      "type": "list",
+      "foo": {
```
mtdowling, line 24: Does this not cause an error because we first check for the required members before checking for extraneous members?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/idl-invalid-list-and-map-members.errors
```diff
@@ -0,0 +1,10 @@
+[ERROR] com.test#NoMemberList: Missing required member of shape `com.test#NoMemberList`: member | Model
+[ERROR] com.test#NoMemberMap: Missing required member of shape `com.test#NoMemberMap`: key | Model
+[ERROR] com.test#MissingKeyMap: Missing required member of shape `com.test#MissingKeyMap`: key | Model
+[ERROR] com.test#MissingValueMap: Missing required member of shape `com.test#MissingValueMap`: value | Model
+[ERROR] com.test#WrongMemberList: Collection shapes may only have a `member` member, but found `foo` | Model
```
mtdowling, line 5: Rather than `Collection`, can you say the actual shape type name as seen in the IDL / AST?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/MapShape.java
```diff
@@ -43,7 +44,19 @@ public final class MapShape extends Shape implements ToSmithyBuilder<MapShape> {
     private MapShape(Builder builder) {
         super(builder, false);
         key = getRequiredMixinMember(builder, builder.key, "key");
+        ShapeId expectedKeyId = getId().withMember("key");
+        if (!key.getId().equals(expectedKeyId)) {
+            throw new SourceException(String.format(
+                    "Expected member of `%s` to have an ID of `%s` but found `%s`",
```
mtdowling, line 50: Would it be possible to move this validation here and below into the builder, like what you did with CollectionShape?

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
```diff
@@ -669,60 +668,10 @@ private void parseMixins(LoadOperation.DefineShape operation) {
     private void parseCollection(ShapeId id, SourceLocation location, CollectionShape.Builder<?, ?> builder) {
```
mtdowling, line 668: Can we merge collection / map / and union into a single method that handles them all? Looks like the same grammar and code now.

mtdowling (changes requested):

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/json-invalid-list-and-map-members.json
milesziemer: No I think we check for extra members first: https://github.com/awslabs/smithy/blob/7342b2c5168dc868c49241e22632a18ca63438bb/smithy-model/src/main/java/software/amazon/smithy/model/loader/AstModelLoader.java#L237, but It just causes a warning: https://github.com/awslabs/smithy/blob/7342b2c5168dc868c49241e22632a18ca63438bb/smithy-model/src/main/java/software/amazon/smithy/model/loader/LoaderUtils.java#L44.

mtdowling: Ah I didn't see an error about the extra members in the test. Maybe it's not working as intended.

milesziemer: I stepped through the code and I think its this: https://github.com/awslabs/smithy/blob/7342b2c5168dc868c49241e22632a18ca63438bb/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelAssembler.java#L575
Since there were errors loading, the assembler doesn't return the warnings.

mtdowling: Can we make these errors?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/idl-invalid-list-and-map-members.errors
milesziemer: Done

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
milesziemer: Done

smithy-model/src/main/java/software/amazon/smithy/model/shapes/MapShape.java
milesziemer: I confused myself with this one. I did add these checks in the MapShape constructor, not the one in CollectionShape. They aren't necessary though because the constructor calls [validateMemberShapeIds](https://github.com/awslabs/smithy/blob/5636d15d067a63613cb9a1a069585fc81312df6b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java#L165) which does the same thing. I'll push an update to remove them.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -108,39 +109,69 @@ protected void validateMixins(Map<ShapeId, Shape> mixins, Map<ShapeId, Trait> in
         }
     }
 
-    protected MemberShape getRequiredMixinMember(
+    protected Map<String, MemberShape> getRequiredMembers(
```
mtdowling, line 112: I think this could be a `MemberShape[]` since we don't really need a dynamically sized List, nor do we need a Map since the caller of this methods knows the order of the provided member names.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -108,39 +109,69 @@ protected void validateMixins(Map<ShapeId, Shape> mixins, Map<ShapeId, Trait> in
...
+        return Optional.of(MemberShape.builder()
+                .id(getId().withMember(requiredMemberName))
                 .target(mostRecentMember.getTarget())
                 .source(getSourceLocation())
                 .addMixin(mostRecentMember)
-                .build();
+                .build());
+    }
+
+    protected SourceException missingRequiredMembersException(List<String> missingMembersNames) {
```
mtdowling, line 169: Can this be private?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/Shape.java
```diff
@@ -108,39 +109,69 @@ protected void validateMixins(Map<ShapeId, Shape> mixins, Map<ShapeId, Trait> in
...
+        }
+
+        if (missingMembersNames.size() > 0) {
+            throw missingRequiredMembersException(missingMembersNames);
+        }
+
+        return members;
+    }
+
+    protected Optional<MemberShape> getRequiredMixinMember(
```
mtdowling, line 136: Since this is all internal stuff, can it just have a nullable return value?

mtdowling (approved):

PR #1786, opened 2023-05-24
Validate that query literal do not conflict with query params

sugmanue:
*Issue #, if available:*

*Description of changes:*

Validate that query literal do not conflict with query params. Currently is possible to define a model like the following, that defines `code` as a query literal in the URI part of the `@http` trait and also the same query param in the `OperationOneShape`. This change will disallow this model that is not valid in practice.

```smithy
$version: "2"
namespace smithy.example

// OperationOne defines a query literal `code` that conflicts with the
// query param defined in the input structure.

service SmithyExample {
    operations: [
        OperationOne
    ]
}

@http(code: 200, method: "GET", uri: "/example?code")
@readonly
operation OperationOne {
    input: OperationOneInput
    output: OperationOneOutput
}

structure OperationOneInput {
    @httpQuery("code")
    code: String

    @httpQuery("state")
    state: String
}

structure OperationOneOutput {
    @required
    @httpHeader("Content-Type")
    contentType: String

    @required
    @httpPayload
    content: Blob
}
```

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
@@ -23,11 +23,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import software.amazon.smithy.model.Model;
-import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.pattern.UriPattern;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.OperationShape;
-import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.HttpQueryTrait;
@@ -94,9 +93,8 @@ public final class HttpQueryTraitValidator extends AbstractValidator {
                     String literalKey = literalEntry.getKey();
                     if (entry.getValue().containsKey(literalKey)) {
                         events.add(error(entry.getKey(), String.format(
-                            "`httpQuery` parameter name binding conflicts found for the `%s` parameter, the "
-                            + "http trait for the `%s` operation defines a query literal with the same name",
-                            literalKey, operation.getId())));
+                            "`httpQuery` name `%s` conflicts with the `http` trait of the `%s` operation: `%s`",
+                            literalKey, operation.getId(), pattern)));
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
```diff
@@ -77,8 +85,39 @@ private List<ValidationEvent> validateBindings(Map<StructureShape, Map<String, S
...
+            }
         }
 
         return events;
     }
+
+    private Map<StructureShape, List<OperationShape>> getStructureToOperations(Model model) {
+        Map<StructureShape, List<OperationShape>> structureToOperations = new HashMap<>();
+        for (ServiceShape service : model.getServiceShapes()) {
+            for (OperationShape operation : TopDownIndex.of(model).getContainedOperations(service)) {
```
hpmellema, line 111: My understanding is that _any_ operation in the model with the http trait could have this issue, so why do we need to traverse the model on a service-by-service basis? 
Then we could get rid of the outer loop and just do 
```java
for (OperationShape operation : model.getOperationShapesWithTrait(HttpTrait.class)) {
   ...
}
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
```diff
@@ -77,8 +85,39 @@ private List<ValidationEvent> validateBindings(Map<StructureShape, Map<String, S
...
+                if (operation.hasTrait(HttpTrait.class)) {
+                    model.expectShape(operation.getInputShape())
```
hpmellema, line 113: Have you considered using an `OperationIndex` here? 
I think that would look like: 
```
operationIndex.getInput(operation).ifPresent(inputStructure -> ...)
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
```diff
@@ -77,8 +85,39 @@ private List<ValidationEvent> validateBindings(Map<StructureShape, Map<String, S
...
+
+            List<OperationShape> operations = structureToOperations.getOrDefault(entry.getKey(),
+                                                                                 Collections.emptyList());
+            for (OperationShape operation : operations) {
+                UriPattern pattern = operation.expectTrait(HttpTrait.class).getUri();
+                for (Map.Entry<String, String> literalEntry : pattern.getQueryLiterals().entrySet()) {
+                    String literalKey = literalEntry.getKey();
+                    if (entry.getValue().containsKey(literalKey)) {
+                        events.add(error(entry.getKey(), String.format(
+                            "`httpQuery` parameter name binding conflicts found for the `%s` parameter, the "
```
mtdowling, line 97: Could this include the conflicting `http` trait URI value?

```
`httpQuery` name `%s` conflicts with the `http` trait of the `%s` operation: `%s`
```

Example:
> `httpQuery` name `foo` conflicts with the `http` trait of the `com.foo#PutFoo` operation: `/test?foo`

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpQueryTraitValidator.java
sugmanue: Sure, I made the change accordingly.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-query-literals-and-bindings-conflict1.smithy
```diff
@@ -0,0 +1,36 @@
+$version: "2"
+namespace smithy.example
+
+// OperationOne defines a query literal `code` that conflicts with the
+// query param defined in the input structure.
+
+service SmithyExample {
```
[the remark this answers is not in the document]
sugmanue: We don't have any branching in the validation code for operations bound to resources, so I'm not sure what else would this case be testing, can you please help me understand what this new case will be testing that is not already being tested?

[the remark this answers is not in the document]
hpmellema: Good point. I this should be fine without the test I mentioned.

hpmellema (approved):

mtdowling (approved):

PR #1787, opened 2023-05-24
Emit specific ids for ChangedEnumTrait ERROR events

rchache:
*Issue #, if available:*

*Description of changes:*
Differentiates between an inserted enum event, and the other ChangedEnumTrait events

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
@@ -39,7 +39,9 @@ import software.amazon.smithy.utils.Pair;
  * list of existing values.
  */
 public final class ChangedEnumTrait extends AbstractDiffEvaluator {
-    private static final String INSERTED = "Inserted";
+    private static final String ORDER_CHANGED = ".OrderChanged";
+    private static final String NAME_CHANGED = ".NameChanged";
+    private static final String REMOVED = ".Removed";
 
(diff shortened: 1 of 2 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
```diff
@@ -93,10 +96,17 @@ private List<ValidationEvent> validateEnum(ChangedShape<Shape> change, Pair<Enum
...
-                                    + "serialization, etc.", definition.getValue())));
+                    events.add(
+                            ValidationEvent.builder()
+                                    .severity(Severity.ERROR)
+                                    .message(String.format(
+                                    "Enum value `%s` was inserted before the end of the list of existing values. This "
+                                            + "can cause compatibility issues when ordinal values are used for "
+                                            + "iteration, serialization, etc.", definition.getValue()))
+                                    .shape(change.getNewShape())
+                                    .id(getEventId() + "." + INSERTED)
```
sugmanue, line 107: Nit, let's move the "." part to the INSERTED constant, and use here just `getEventId() + INSERTED`. Also I think that "OrderChanged" would be better.

Can we also add specificity to the other cases that error here, like `Removed` and `Changed`?

rchache: Ack, addressing all suggestions in next commit

sugmanue (approved):

PR #1791, opened 2023-05-25
Add guide on disabling authentication

hpmellema:
*Description of changes:*
Adds a short guide on how to disable authentication on a service-wide or per-operation basis. 

*Testing*: 
Built docs locally and confirmed all links worked correctly and all code blocks/notes rendered as expected.


docs/source-2.0/guides/disabling-auth.rst
```diff
@@ -0,0 +1,79 @@
+.. _disabling-auth:
+
+========================
+Disabling Authentication
+========================
+
+Services should, in general, have authentication configured. However, there are
+times when authentication may need to be disabled for the service or for a
+number of operations within a service. For example, you may want to expose
```
mtdowling, line 9: repeated "to expose"

mtdowling (changes requested):

docs/source-2.0/guides/disabling-auth.rst
```diff
@@ -0,0 +1,90 @@
+.. _disabling-auth:
+
+========================
+Disabling Authentication
```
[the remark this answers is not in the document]
hpmellema: Im not opposed to moving it into the spec as a subsection. The content is pretty limited so It probably doesnt justify a full guide. I'll update this in the next rev.

docs/source-2.0/spec/authentication-traits.rst
```diff
@@ -372,3 +372,87 @@ authentication scheme trait that is not applied to the service:
...
+        version: "2020-08-31",
+        operations: [
+            MyOperation,
+        ]
+    }
+
+    operation MyOperation {}
+
+
+Disabling Authentication for an Operation
```
gosar, line 414: I'm not sure if all this content from guide needs to be here as-is. The [auth trait](https://smithy.io/2.0/spec/authentication-traits.html#auth-trait) section already has "The following example defines all combinations in which auth can be applied to services and operations:" which covers multiple cases of auth, just missing an explicit case for disabling auth for an operation. It seems better to just add a case to that example, instead of this separate section just for disable?

Similarly, disable for a service seems like it could be covered under the https://smithy.io/2.0/spec/authentication-traits.html#authdefinition-trait section, by adding an example of service with no authentication.

gosar (approved):

PR #1797, opened 2023-05-30
Extended AddedEntityBinding/RemovedEntityBinding event ids with granularity

rchache:
*Issue #, if available:*

*Description of changes:*
Extended AddedResourceBinding, AddedOperationBinding, RemovedResourceBinding,  RemovedOperationBinding event ids with
1. the type of the parent entity
2. the id of the child shape being added/removed

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedEntityBinding.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedEntityBinding.java
@@ -23,6 +23,7 @@ import software.amazon.smithy.diff.ChangedShape;
 import software.amazon.smithy.diff.Differences;
 import software.amazon.smithy.model.shapes.EntityShape;
 import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
 
@@ -37,6 +38,8 @@ import software.amazon.smithy.model.validation.ValidationEvent;
 public final class AddedEntityBinding extends AbstractDiffEvaluator {
     private static final String ADDED_RESOURCE = "AddedResourceBinding";
     private static final String ADDED_OPERATION = "AddedOperationBinding";
+    private static final String TO_RESOURCE = ".ToResourceAdded.";
+    private static final String TO_SERVICE = ".ToServiceAdded.";
 
     @Override
     public List<ValidationEvent> evaluate(Differences differences) {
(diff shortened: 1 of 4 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedEntityBinding.java
```diff
@@ -37,6 +38,8 @@
 public final class AddedEntityBinding extends AbstractDiffEvaluator {
     private static final String ADDED_RESOURCE = "AddedResourceBinding";
     private static final String ADDED_OPERATION = "AddedOperationBinding";
+    private static final String TO_RESOURCE = ".ToResourceShapeBound.";
+    private static final String TO_SERVICE = ".ToServiceShapeBound.";
```
rchache, line 42: Open to suggestions if there are better wordings for this that don't also make the full id misleading

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -37,6 +38,8 @@
 public final class RemovedEntityBinding extends AbstractDiffEvaluator {
     private static final String REMOVED_RESOURCE = "RemovedResourceBinding";
     private static final String REMOVED_OPERATION = "RemovedOperationBinding";
+    private static final String FROM_RESOURCE = ".FromResourceRemoved.";
```
mtdowling, line 41: We already know it's removed from the name of the validator, so no need to repeat that in the secondary level. So I suggest ".Resource" and ".Service"

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/AddedEntityBinding.java
```diff
@@ -37,6 +38,8 @@
 public final class AddedEntityBinding extends AbstractDiffEvaluator {
     private static final String ADDED_RESOURCE = "AddedResourceBinding";
     private static final String ADDED_OPERATION = "AddedOperationBinding";
+    private static final String TO_RESOURCE = ".ToResourceAdded.";
```
mtdowling, line 41: Same thing here -- we know it wa added by the event ID, no need to repeat that here. I suggest ".Resource" and ".Service"

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/AddedEntityBindingTest.java
```diff
@@ -53,7 +53,7 @@ public void detectsAddedOperationToResource() {
         Model modelB = Model.assembler().addShapes(r1, o).assemble().unwrap();
         List<ValidationEvent> events = ModelDiff.compare(modelA, modelB);
 
-        assertThat(TestHelper.findEvents(events, "AddedOperationBinding").size(), equalTo(1));
+        assertThat(TestHelper.findEvents(events, "AddedOperationBinding.ToResourceAdded.foo.baz#Operation").size(), equalTo(1));
```
mtdowling, line 56: Can we just include the shape ID shape name rather than the full ID? A service can only have a single resource/operation in the entire closure with a given shape name, so the full ID actually isn't giving any more detail that we'd need to make these unique.

mtdowling (changes requested):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
rchache: I was worried that in the case of RemovedResourceBinding.Resource.MyResourceShape for example, it's a little misleading or unclear whether MyResourceShape is the parent or the child of the binding in question. What are your thoughts?

mtdowling: Ah ok. Sure maybe then "FromResource"?

rchache: Well actually the tricky part is that the shape I'm appending to the event id is the child, not the parent. I'm definitely open to suggestions for how to better word this though.

mtdowling (approved):

PR #1798, opened 2023-05-31
[Discussion] Quote dates in header tests

Baccata:
*Description of changes:*

The http-date timestamp values currently present in the protocol compliance tests are impossible (or much harder) to parse into a `List<String>`, because the commas present in the rendered date are not escaped. The values being parseable is desirable considering may http libraries represent headers as something similar to `MultiMap<String, String>`, which allows for a greater precision of comparison between expected values and received values. 

In other words, it's way harder  for `"Mon, 16 Dec 2019 23:48:18 GMT, Mon, 16 Dec 2019 23:48:18 GMT"` to be separated in `["Mon, 16 Dec 2019 23:48:18 GMT", "Mon, 16 Dec 2019 23:48:18 GMT"]`, because without the additional information of quotes the commas after `Mon` can be interpreted as a separating character. 

It makes it harder to use these tests specifications to automate tests of protocol implementations. It's also 
This PR attempts at alleviating this problem by quoting individual dates in the header value. 

I'm following the precedent set by the [StringList](https://github.com/awslabs/smithy/blob/c619a9a4a16f8cffbf4adcc06d32ecbe73ade669/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy#L57-L63) test. 

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
@@ -111,7 +111,7 @@ apply InputAndOutputWithHeaders @httpRequestTests([
         method: "POST",
         uri: "/InputAndOutputWithHeaders",
         headers: {
-            "X-TimestampList": "Mon, 16 Dec 2019 23:48:18 GMT, Mon, 16 Dec 2019 23:48:18 GMT"
+            "X-TimestampList": "\"Mon, 16 Dec 2019 23:48:18 GMT\", \"Mon, 16 Dec 2019 23:48:18 GMT\""
         },
         body: "",
         params: {
@@ -259,7 +259,7 @@ apply InputAndOutputWithHeaders @httpResponseTests([
         protocol: restJson1,
         code: 200,
         headers: {
-            "X-TimestampList": "Mon, 16 Dec 2019 23:48:18 GMT, Mon, 16 Dec 2019 23:48:18 GMT"
+            "X-TimestampList": "\"Mon, 16 Dec 2019 23:48:18 GMT\", \"Mon, 16 Dec 2019 23:48:18 GMT\""
         },
         params: {
             headerTimestampList: [1576540098, 1576540098]
@@ -452,11 +452,11 @@ apply TimestampFormatHeaders @httpRequestTests([
         uri: "/TimestampFormatHeaders",
         headers: {
             "X-memberEpochSeconds": "1576540098",
-            "X-memberHttpDate": "Mon, 16 Dec 2019 23:48:18 GMT",
+            "X-memberHttpDate": "\"Mon, 16 Dec 2019 23:48:18 GMT\"",
             "X-memberDateTime": "2019-12-16T23:48:18Z",
-            "X-defaultFormat": "Mon, 16 Dec 2019 23:48:18 GMT",
+            "X-defaultFormat": "\"Mon, 16 Dec 2019 23:48:18 GMT\"",
             "X-targetEpochSeconds": "1576540098",
-            "X-targetHttpDate": "Mon, 16 Dec 2019 23:48:18 GMT",
+            "X-targetHttpDate": "\"Mon, 16 Dec 2019 23:48:18 GMT\"",
             "X-targetDateTime": "2019-12-16T23:48:18Z",
         },
         body: "",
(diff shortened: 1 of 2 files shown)


JordonPhillips:
I sympathize with your frustration, this is definitely one of the most annoying aspects of implementing the AWS protocols. However, we can't remove or change these tests because they describe the actual behavior of the tooling that these protocols describe.

The most we could do would be to add additional tests to ensure Smithy-generated tooling can handle receiving either case. At some point in the future we could change the way we actually serialize them. Migrating to an entirely new protocol would be better though, because restJson1 for example has some other things we'd want to change (e.g. serializing big numbers and longs as strings in json).

Baccata:
> However, we can't remove or change these tests because they describe the actual behaviour of the tooling that these protocols describe. 
 
Okay, I can empathise with this. 

Would it be receivable to add an additional required field to the test specification to carry another, more tractable version of the `headers` field that would offer a `Map<String, List<String>>` representation ? 

Alternatively, to you have guidance on what we could do without resorting to convoluted logic ? 

#### Summary of the problem : 

Just to be absolutely clear : we are getting headers as `MultiMap<String, String>`, and we're trying to compare them with the expectations that are provided as `Map<String, String>`. 

* The values in the `Map<String, String>` cannot be parsed into `List<String>` because in the case of http-date timestamps, quotes aren't provided to know whether the comma characters are a separation between header values or not. 
* The vales in the `MultiMap<String, String>` cannot be collapsed into simple strings, because the logical thing to do would be to inspect each string to detect commas, and quote them when there are commas ... 
* ... unless we attempt to parse strings as http-date timestamps first, and use that information to decide whether to add quotes. It seems very ad-hoc and brittle.

JordonPhillips:
> Would it be receivable to add an additional required field to the test specification to carry another, more tractable version of the `headers` field that would offer a `Map<String, List<String>>` representation ?

I would actually love this. It's something I've been wanting to do for a few months, but just haven't been able to get the time for. The AWS SDK teams are making progress on a project that, amongst other things, seeks to more realistically model HTTP. One of those glaring holes is the reality that headers are a list of tuples rather than a map.

I don't think this really solves your problem though. You don't have to worry about serialization anymore, but you do still have to worry about intermediaries joining the headers with a comma anyway. So you still need to be able to parse that list as a single tuple.

As for how it's done, it can still be done deterministically, just a bit more complicated than other shapes that have better behavior. The important thing is that you *know* to expect a list of timestamps

```python
def split_header(header_value, is_http_date=false):
    entries = []
    i = 0
    while i < header_value.length:
        character = header_value[i]
        if character is whitespace:
            i += 1
        else if character is dquote:
            start = i + 1
            i = skip_until_dquote(start) + 1
            entries.append(header_value.substring(start, i - 2))
            i = skip_until_comma_or_end(i) + 1
        else:
            start = i
            i = skip_until_comma_or_end(start)
            if is_http_date:
                i = skip_until_comma_or_end(start)
            entry = header_value.substring(start, i - 1).strip()
            entries.append(entry)
            i += 1
    return entries
```

You might then only include this version of splitting in the AWS protocols if you want to save some bytes or not expose the extra complexity on other protocols. I can share some actual python code I wrote recently if you'd like.

Baccata:
> The important thing is that you know to expect a list of timestamps 

Right ... the problem is by the time we've "compiled" this test, we've already lost this contextual information. In particular, in an ideal world, I should be able to use the `httpRequestTest` and `httpResponseTest` traits regardless of the protocol, and therefore I shouldn't have to carry this contextual information to a layer that, in theory, shouldn't have to worry about it ... 

> You might then only include this version of splitting in the AWS protocols if you want to save some bytes or not expose the extra complexity on other protocols

> I sympathize with your frustration, this is definitely one of the most annoying aspects of implementing the AWS protocols. However, we can't remove or change these tests because they describe the actual behavior of the tooling that these protocols describe.

Wait, are you saying that the AWS-issued implementations of its protocols disregard the [http specification](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values). Specifically this paragraph : 

> Because commas (",") are used as the delimiter between members, they need to be treated with care if they are allowed as data within a member. This is true for both list-based and singleton fields, since a singleton field might be erroneously sent with multiple members and detecting such errors improves interoperability. Fields that expect to contain a comma within a member, such as within an [HTTP-date](https://www.rfc-editor.org/rfc/rfc9110.html#http.date) or [URI-reference](https://www.rfc-editor.org/rfc/rfc9110.html#uri.references) element, ought to be defined with delimiters around that element to distinguish any comma within that data from potential list separators.

> For example, a textual date and a URI (either of which might contain a comma) could be safely carried in list-based field values like these:

```
Example-URIs: "http://example.com/a.html,foo",
              "http://without-a-comma.example.com/"
Example-Dates: "Sat, 04 May 1996", "Wed, 14 Sep 2005"
``` 

(see how the dates are quoted) 

In other words, when calling an AWS service, do I have to expect to receive a list of timestamps without quotes ?

adamthom-amzn:
Unfortunately, we're not disregarding the HTTP spec. The example grammar given is not mandatory for all fields, it's just one that I think they wished was used from the beginning. Note the weasel-wording of "could" and "should" throughout the spec. In practice, there is an understood grammar for every standard header in HTTP, and many of them vary from the example grammar given in the RFC. (See `Set-Cookie` for a particularly disastrous example)

The problem we have is we are often not defining standard headers, but instead custom ones, and we never bothered to define a grammar for how these values should be serialized and deserialized. Unfortunately many of the existing protocol implementations, both client and server, do not tolerate multiple variations of specifying list values like you'd expect. So, unfortunately, yes, you should expect lists of timestamps without quotes. I spent a couple of months trying to see if I could wedge the RFC-recommended grammar into header values and limitations in the existing protocol implementations made it impractical.

Baccata:
Closing as this is not really something that can be addressed.

Baccata:
@adamthom-amzn, @JordonPhillips, thanks for sharing the context !

PR #1804, opened 2023-06-02
Add a Smithy IDL model formatter

mtdowling:
Add a Smithy IDL model formatter

This formatter uses Wadler's pretty printer algorithm to automatically format code based on line length (via prettier4j). Future commits should investigate shading the dependency on prettier4J.


smithy-syntax/src/main/java/software/amazon/smithy/syntax/Formatter.java
```diff
@@ -0,0 +1,667 @@
...
+                            .append(Doc.text("input"))
+                            .append(simpleTarget == null
+                                    ? visit(cursor.getFirstChild(TreeType.INLINE_AGGREGATE_SHAPE))
+                                    : Doc.text(": ")).append(visit(simpleTarget));
+                }
+
+                case OPERATION_OUTPUT: {
+                    TreeCursor simpleTarget = cursor.getFirstChild(TreeType.SHAPE_ID);
+                    return skippedComments(cursor, false).append(Doc.text("output"))
+                            .append(simpleTarget == null
```
JordonPhillips, line 266:
```suggestion
                    return skippedComments(cursor, false)
                            .append(Doc.text("output"))
                            .append(simpleTarget == null
```

JordonPhillips (approved):
This is great for reflowing what's already there, but I'd like to see some syntax changes that should probably be in a separate PR:

* Updating to target elision syntax where mixins / `for` already exists.
* Using ` = ` instead of the `@default` trait
* Using ` = ` instead of the `@enumValue` trait
* documentation trait becomes doc comment (except when used with apply)

I would consider all of those to be good default behavior. There's also some bigger refactorings that would be nice, but those are a different thing and should be exposed as such. These for sure don't belong here but just throwing out some stuff:

* Inlining IO shapes (different because usually that means applying a new trait)
* Shape sorting
* Detecting resource property definitions and injecting the `for` statement, then using target elision
* Detecting and applying use cases for mixins

PR #1807, opened 2023-06-04
Added enum values to ids for ChangedEnumTrait events

rchache:
*Issue #, if available:*

*Description of changes:*
Adds the enum indices to the event ids
eg. `ChangedEnumTrait.Removed.4`

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
@@ -73,7 +73,8 @@ public final class ChangedEnumTrait extends AbstractDiffEvaluator {
         List<ValidationEvent> events = new ArrayList<>();
         int oldEndPosition = oldTrait.getValues().size() - 1;
 
-        for (EnumDefinition definition : oldTrait.getValues()) {
+        for (int enumIndex = 0; enumIndex < oldTrait.getValues().size(); enumIndex++) {
+            EnumDefinition definition = oldTrait.getValues().get(enumIndex);
             Optional<EnumDefinition> maybeNewValue = newTrait.getValues().stream()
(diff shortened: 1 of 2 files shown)


jvschneid (approved):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
```diff
@@ -84,7 +84,7 @@ private List<ValidationEvent> validateEnum(ChangedShape<Shape> change, Pair<Enum
                                 .severity(Severity.ERROR)
                                 .message(String.format("Enum value `%s` was removed", definition.getValue()))
                                 .shape(change.getNewShape())
-                                .id(getEventId() + REMOVED)
+                                .id(getEventId() + REMOVED + definition.getValue())
```
mtdowling, line 87: Enum values support arbitrary content, so this seems problematic to include them in the ID. Ideally these events would use the `name` of an enum definition if available, and if not... I dunno. Maybe the index of the entry? Or a name derived from the value?

mtdowling (changes requested):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedEnumTrait.java
mtdowling: Yeah sounds good. The caveat I can think of is that what is, say index 2, today won't be index 2 in the future if index 2 is removed. Not sure if that matters for you.

rchache: Shouldn't be an issue since we only care about these events in the scope of a single Trebuchet feature and not after that

mtdowling (approved):

PR #1808, opened 2023-06-05
Assert trees are valid, fill in missing tests

milesziemer:
All TreeTypeTests now expect the entire tree to be valid. This helps us find ambiguity and inconsistencies in the parser for cases when parsing a child tree is invalid only due to the context of the parent. Fixes minor bugs that caused trees to be invalid:
- enumMember test didn't have a BR after VALUE_ASSIGNMENT

Fills in missing tests for:
- Operations
- Block apply statements
- Entity type name
- Aggregate type name

Other minor changes and improvements:
- Adds util methods for parsing optional for-resource, mixins, and value assignment productions.
- Adds util method for parsing any shape type and name.
- Updates peekPastSpaces to not skip the current token, in case it is not SP. This simplifies parsing value assignments, and has more intuitive behavior.
- Removes peekPastWs, which was only used for TRAIT_BODY_VALUE, inlining the implementation. peekPastWs skipped the current token like peekPastSpaces, which is necessary for TRAIT_BODY_VALUE.
- Add ENTITY_TYPE_NAME production.
- Updates name of some productions to be consistent with grammar.
- Fixed a few places where TOKEN was parsed as IDENTIFIER.

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
@@ -152,7 +152,7 @@ public enum TreeType {
         @Override
         void parse(CapturingTokenizer tokenizer) {
             tokenizer.withState(this, () -> {
-                tokenizer.next();
+                tokenizer.next(); // Skip "use"
                 SP.parse(tokenizer);
                 ABSOLUTE_ROOT_SHAPE_ID.parse(tokenizer);
                 BR.parse(tokenizer);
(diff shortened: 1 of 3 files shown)


smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
```diff
@@ -486,8 +469,7 @@ void parse(CapturingTokenizer tokenizer) {
         @Override
         void parse(CapturingTokenizer tokenizer) {
             tokenizer.withState(this, () -> {
-                tokenizer.expect(IdlToken.IDENTIFIER);
-                tokenizer.next(); // skip "input"
+                IDENTIFIER.parse(tokenizer); // skip "input"
```
mtdowling, line 472: This is `%s"input"` in the grammar, which would be a TOKEN, right? See also output and errors below this.

mtdowling (approved):

PR #1817, opened 2023-06-12
Rewrite bad doc comments to normal comments

mtdowling:
This doesn't account for every possible bad doc comment (e.g., doc comments that come after traits). That can be a followup.

*Issue #, if available:*

*Description of changes:*


smithy-syntax/src/main/java/software/amazon/smithy/syntax/FixBadDocComments.java
```diff
@@ -0,0 +1,130 @@
...
+
+            // Remove doc comments from all but the last use statement.
+            List<TreeCursor> useStatements = useSection.getChildrenByType(TreeType.USE_STATEMENT);
+            for (int i = 0; i < useStatements.size() - 1; i++) {
+                updateNestedChildren(useStatements.get(i));
+            }
+        }
+
+        // Doc comments are never allowed in NODE_VALUE values.
+        for (TreeCursor trait : shapeSection.findChildrenByType(TreeType.NODE_VALUE)) {
```
milesziemer, line 61: Nit: variable name `trait`

milesziemer (changes requested):

smithy-model/src/test/resources/software/amazon/smithy/model/loader/valid/doc-comments/doc-comments.json
milesziemer, line 1: Out of curiosity, was this behavior different before?

mtdowling: Yeah it didn't work thought technically it's allowed here

PR #1820, opened 2023-06-13
Add parse failure tests

milesziemer:
Adds test cases that validate the parser fails when expected by asserting that any descendant of the root is an error tree. There is a test per-production, if that production has terminal tokens, or represents an alternation of productions that are distinguished by terminal tokens. For example, SHAPE_OR_APPLY_STATEMENT chooses between shape and apply statements based on whether the current lexemme is "apply".

Only cases where the terminal tokens are invalid are tested because if a non-terminal (ie. another production) is invalid, the error tree is a child of the tree for that production. Since we also have tests that make sure the correct children are present for each production, we should have enough coverage to only test invalid non-terminals. For example, when parsing this shape statement:
```
structure Foo.Bar {}
```
we know the `Foo.Bar` is part of the IDENTIFIER production, we have tests that validate the the IDENTIFIER production is present, and we have tests that check for errors in IDENTIFIERs, so we don't need to specifically test this kind of error for shape statements.

One caveat is that the existing tests I'm referring to only ensure all expected children are present for valid trees, so there is a logical inconsistency in the argument that these tests provide full coverage, but I can't think of a case that could cause a problem. Either way, I think these tests provide _good enough_ coverage.

Some bug fixes:
- Fixed an issue where the error caused by not finding a "namespace" in NAMESPACE_STATEMENT was recovered by the parent tree.
- Expect IDENTIFIER for "with" token in MIXINS production.

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
@@ -116,19 +116,19 @@ public enum TreeType {
     NAMESPACE_STATEMENT {
         @Override
         void parse(CapturingTokenizer tokenizer) {
-            if (tokenizer.isCurrentLexeme("namespace")) {
-                tokenizer.withState(this, () -> {
+            tokenizer.withState(this, () -> {
+                if (tokenizer.isCurrentLexeme("namespace")) {
                     tokenizer.next(); // skip "namespace"
                     SP.parse(tokenizer);
                     NAMESPACE.parse(tokenizer);
                     BR.parse(tokenizer);
-                });
-            } else if (tokenizer.hasNext()) {
-                throw new ModelSyntaxException(
-                        "Expected a namespace definition but found "
-                        + tokenizer.getCurrentToken().getDebug(tokenizer.getCurrentTokenLexeme()),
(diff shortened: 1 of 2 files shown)


smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
```diff
@@ -532,6 +532,7 @@ void parse(CapturingTokenizer tokenizer) {
         @Override
         void parse(CapturingTokenizer tokenizer) {
             tokenizer.withState(this, () -> {
+                tokenizer.expect(IdlToken.IDENTIFIER);
```
mtdowling, line 535: Is this to capture "with"? In that case, we don't treat it as an IDENTIFIER since it's a literal string value.

mtdowling (changes requested):

smithy-syntax/src/main/java/software/amazon/smithy/syntax/TreeType.java
milesziemer: Yep, I confused myself with that one. Fixed it and removed the test case, since "with" is checked outside of the MIXINS production.

mtdowling (approved):

PR #1835, opened 2023-06-23
(doc): Minor rephrase for Timestamp description.

karthikbekalp:
*Issue #, if available:*

Minor timestamp description update. 

*Description of changes:*

The description was slightly confusing as I thought it was no (UTC offset and fractional milliseconds) instead of (no UTC) and (fractional milliseconds). 

### DIFF
--- a/docs/source-2.0/spec/model.rst
+++ b/docs/source-2.0/spec/model.rst
@@ -789,8 +789,8 @@ target from traits and how their values are defined in
       - number | string
       - If a number is provided, it represents Unix epoch seconds with optional
         millisecond precision. If a string is provided, it MUST be a valid
-        :rfc:`3339` string with no UTC offset and optional fractional
-        precision (for example, ``1985-04-12T23:20:50.52Z``).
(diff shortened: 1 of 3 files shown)


docs/source-2.0/spec/model.rst
```diff
@@ -789,7 +789,7 @@ target from traits and how their values are defined in
       - number | string
       - If a number is provided, it represents Unix epoch seconds with optional
         millisecond precision. If a string is provided, it MUST be a valid
-        :rfc:`3339` string with no UTC offset and optional fractional
+        :rfc:`3339` string with optional fractional precision but no UTC offset
```
mtdowling, line 792: Looks like the next line still has the word "precision", but should not

mtdowling (changes requested):

docs/source-2.0/spec/model.rst
karthikbekalp: Makes sense, updated. Thanks!

mtdowling (approved):

PR #1838, opened 2023-06-28
Fix basic HTTP authentication when resolving dependencies

denisrosca:
*Issue #1837 *

*Description of changes:*
Fixes broken http authentication.

Replace `MavenAuth` with existing aether builder to configure the repository auth context.
Modify the auth test to use a mock server to actually run the authentication flow.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1846, opened 2023-07-05
Skip unrecognized JSON files when loading models

### LINKED ISSUE #1841: CLI: ignore non-Smithy-model JSON files?
Hi!

I'm trying to use the CLI, and I've also [faced this problem before](https://github.com/kubukoz/smithy-playground/pull/290) using plain smithy-model. Perhaps there's something we can do in smithy itself to assist this usecase:

In my org's repo, I've seen some people put non-Smithy files (e.g. JSON body examples for a given endpoint) into the same directory as their Smithy specs. It probably makes sense to keep these close to the specs, rather than in a separate directory structure.

However, these make almost all the tooling fail validation: e.g. this file

```json
{
  "foo": "bar"
}
```

would fail because:

```
Missing expected member smithy.
```

For other files (non-JSON), it seems like the loader is ignoring them and just emitting a warning. Could we do the same for JSON files, if they _don't_ have that `smithy` member?

This could be an opt-in flag too, although for optimal UX I would suggest making it the default.

mtdowling:
*Issue #, if available:* #1841

*Description of changes:*

This commit updates the model loading process so that JSON files that are not objects or that don't contain a top-level "smithy" key are skipped. This can be useful when loading directories that contain Smithy models mixed with other JSON files. JAR files that contain Smithy models are not permitted to refer to unrecognized JSON files that lack a valid "smithy" version key/value pair.

This commit also only creates InputStreams for files that Smithy can actually load or attempt to load. Smithy would previously open an InputStream for all recursive files in a directory even if it was unable to load the file based on the filename. It now only creates an InputStream when the file has a .json, .smithy, or no file extension.

Closes #1841


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelLoader.java
```diff
@@ -55,52 +55,67 @@ private ModelLoader() {}
...
-        Version version = Version.fromString(versionNode.getValue());
-
-        if (version != null) {
-            new AstModelLoader(version, model).parse(operationConsumer);
-        } else {
-            throw new ModelSyntaxException("Unsupported Smithy version number: " + versionNode.getValue(), versionNode);
+    static boolean loadParsedNode(Node node, Consumer<LoadOperation> operationConsumer) {
+        if (node.isObjectNode()) {
+            ObjectNode model = node.expectObjectNode();
+            if (model.getStringMap().containsKey("smithy")) {
```
kstich, line 101: nit: use `containsMember` for brevity

smithy-model/src/test/java/software/amazon/smithy/model/JarUtils.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kstich, line 2: nit: use short header

smithy-model/src/test/java/software/amazon/smithy/model/JarUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+@SmithyInternalApi
+public final class JarUtils {
+    /**
+     * Creates a JAR in a temp directory on demand for test cases based on a directory.
+     *
+     * <p>This method is preferred over embedding JARs directly as resources when possible, because that JARs don't
```
kstich, line 37: s/because that JARs/because the JARs

kstich (changes requested):
Looks like issues with Windows CI, I had issues with `Paths.get` formatting Windows paths improperly because they needed to be URIs to encode correctly.

smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
```diff
@@ -1211,4 +1212,39 @@ public void versionTransformsAreAlwaysApplied() {
...
+
+        assertThat(result.getValidationEvents(Severity.DANGER), empty());
+        assertThat(result.getValidationEvents(Severity.ERROR), empty());
+
+        result.unwrap().expectShape(ShapeId.from("smithy.example#MyString"));
+    }
+
+    @Test
+    public void failsOnInvalidJarJsonFile() throws URISyntaxException, IOException {
+        Path jar = JarUtils.createJarFromDir(Paths.get(getClass().getResource("assembler-fail-invalid-jar").getFile()));
```
kstich, line 1242: I believe the CI is failing here because `.getFile()` should be `.toURI()`.

kstich (approved):

kubukoz:
@mtdowling do you have a timeline for the next release? This and #1838 would be really nice to have for my team.

mtdowling:
We're planning for a Monday release

kubukoz:
awesome, thanks!

PR #1853, opened 2023-07-12
Support SNAPSHOTs and fix dependency cache

### LINKED ISSUE #1850: Snapshot dependencies
Just updated to 1.34.0, and I bumped into this:

https://github.com/awslabs/smithy/blob/1ce05b2640eb006c737d8cb278a33ca2c31ad821/smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/MavenDependencyResolver.java#L155-L157

Snapshot dependencies not being supported seems like a very strange limitation - we used to rely on snapshots a lot for developing new features in a Smithy library, while being able to test out things like packaging, LSP support, etc. without committing to a stable release.

Why is this requirement in there? Could it be relaxed? After the recent change to the LSP, which now uses this custom resolver rather than Coursier (which allows snapshots), it's going to be a problem for orgs that share Smithy libraries (e.g. for custom protocols) between repositories.

mtdowling:
*Issue #, if available:* #1850

*Description of changes:*
We will no longer fail when a SNAPSHOT dependency is used. In order to make sure that SNAPSHOTs are periodically re-resolved, a TTL of 1 day was added to the cache file so that the cache is invalidated every 24 hours. The cache can be manually invalidated using `smithy clean`.

When implementing a TTL, I noticed that caching wasn't actually working and was always invalidated. That's fixed now by using the correct timestamp comparisons, and several new test cases were added.

Closes #1850

### DIFF
--- a/smithy-cli/src/it/java/software/amazon/smithy/cli/MavenResolverTest.java
+++ b/smithy-cli/src/it/java/software/amazon/smithy/cli/MavenResolverTest.java
@@ -117,7 +117,7 @@ public class MavenResolverTest {
                 RunResult result = IntegUtils.run(path, ListUtils.of("validate", "--debug", "model"));
 
                 assertThat(result.getExitCode(), is(0));
-                assertThat(result.getOutput(), containsString("Invalidating dependency cache"));
+                assertThat(result.getOutput(), containsString("Resolving Maven dependencies for Smithy CLI"));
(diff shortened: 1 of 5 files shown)


kubukoz (approved):
This is great, thank you so much!

kstich (approved):

PR #1863, opened 2023-07-18
Add documentationIdentifier value to aws.api#service trait

jvschneid:
*Description of changes:* Add documentationIdentifier value to aws.api#service trait. This value is used to implement linking between service and sdk documentation for AWS services.

### DIFF
--- a/docs/source-1.0/spec/aws/aws-core.rst
+++ b/docs/source-1.0/spec/aws/aws-core.rst
@@ -26,12 +26,12 @@ Value type
     * :ref:`service-cloudformation-name`
     * :ref:`service-arn-namespace`
     * :ref:`service-cloudtrail-event-source`
-    * :ref:`service-documentation-identifier`
+    * :ref:`service-doc-id`
     * :ref:`service-endpoint-prefix`
 
 The following example defines an AWS service that uses the default values of
 ``cloudFormationService``, ``arnNamespace``, ``cloudTrailEventSource`` and
-``documentationIdentifier``:
+``docId``:
 
 .. tabs::
 
@@ -80,7 +80,7 @@ The following example provides explicit values for all properties:
             cloudFormationName: "FooBaz",
             arnNamespace: "myservice",
             cloudTrailEventSource: "myservice.amazon.aws",
-            documentationIdentifier: "some-value-2018-03-17",
+            docId: "some-value-2018-03-17",
             endpointPrefix: "my-endpoint"
         )
         service FooBaz {
@@ -100,7 +100,7 @@ The following example provides explicit values for all properties:
                             "sdkId": "Some Value",
                             "cloudFormationName": "FooBaz",
                             "arnNamespace": "myservice",
-                            "documentationIdentifier": "some-value-2018-03-17",
+                            "docId": "some-value-2018-03-17",
                             "cloudTrailEventSource": "myservice.amazon.aws"
                         }
                     }
@@ -237,17 +237,17 @@ Amazon CloudWatch is ``monitoring.amazonaws.com``. Such services will
 need to explicitly configure the ``cloudTrailEventSource`` setting.
 
 
-.. _service-documentation-identifier:
+.. _service-doc-id:
 
-``documentationIdentifier``
+``docId``
 ===========================
 
-The ``documentationIdentifier`` property is a ``string`` value that is used to
-implement linking between service and SDK documentation for AWS services.
+The ``docId`` property is a ``string`` value that is used to implement linking
+between service and SDK documentation for AWS services.
 
 This value MAY be provided, but if it's missing it will default to the ``sdkId``
 value in lower case followed by the service ``version`` property, separated by
-dashes. For the example beloe the value for this property would default to
+dashes. For the example below the value for this property would default to
 ``some-value-2018-03-17``.
 
 .. code-block:: smithy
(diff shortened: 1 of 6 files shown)


mtdowling:
documentationIdentifier seems like it could be made shorter - docId?

docs/source-1.0/spec/aws/aws-core.rst
```diff
@@ -233,6 +237,32 @@ Amazon CloudWatch is ``monitoring.amazonaws.com``. Such services will
...
+``docId``
+===========================
+
+The ``docId`` property is a ``string`` value that is used to implement linking
+between service and SDK documentation for AWS services.
+
+This value MAY be provided, but if it's missing it will default to the ``sdkId``
+value in lower case followed by the service ``version`` property, separated by
+dashes. For the example below the value for this property would default to
+``some-value-2018-03-17``.
```
JordonPhillips, line 251:
```suggestion
This will default to the ``sdkId`` value in lower case followed by the
service ``version`` property, separated by dashes. For the example
below, the value for this property would default to
``some-value-2018-03-17``.
```

docs/source-2.0/aws/aws-core.rst
```diff
@@ -194,6 +197,33 @@ Amazon CloudWatch is ``monitoring.amazonaws.com``. Such services will
...
+``docId``
+===========================
+
+The ``docId`` property is a ``string`` value that is used to implement linking
+between service and SDK documentation for AWS services.
+
+This value MAY be provided, but if it's missing it will default to the ``sdkId``
+value in lower case followed by the service ``version`` property, separated by
+dashes. For the example below the value for this property would default to
+``some-value-2018-03-17``.
```
JordonPhillips, line 211:
```suggestion
This will default to the ``sdkId`` value in lower case followed by the
service ``version`` property, separated by dashes. For the example
below, the value for this property would default to
``some-value-2018-03-17``.
```

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
```diff
@@ -246,6 +298,17 @@ public ServiceTrait build(ShapeId target) {
             return new ServiceTrait(this);
         }
 
+        /**
+         * Sets the target shape to which the trait is applied.
+         *
+         * @param target the ShapeId targeted by the trait.
+         * @return Returns the builder.
+         */
+        public Builder target(ShapeId target) {
```
JordonPhillips, line 307: this will always be overwritten

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
```diff
@@ -93,6 +102,15 @@ public static Builder builder() {
         return new Builder();
     }
 
+    /**
+     * Get the ShapeId of the target of this trait.
+     *
+     * @return Returns the ShapeId which this trait targets.
+     */
+    public ShapeId getTarget() {
```
JordonPhillips, line 110:
```suggestion
    private ShapeId getTarget() {
```

We do not need to and should not expose this. Better to set the default value in the `build` method honestly.

JordonPhillips (changes requested):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
jvschneid: That's fair, I added it mostly to consistency, making sure all fields have a public getter, but I don't mind changing it at all as there's no clear need for it.

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
jvschneid: Not if I build the object without passing a `ShapeId` to the builder, such as `serviceTrait.toBuilder().build()`. I can make it private though (:

JordonPhillips (approved):

docs/source-2.0/aws/aws-core.rst
```diff
@@ -194,6 +197,32 @@ Amazon CloudWatch is ``monitoring.amazonaws.com``. Such services will
 need to explicitly configure the ``cloudTrailEventSource`` setting.
 
 
+.. _service-doc-id:
+
+``docId``
+===========================
```
kstich, line 203: Header line should match the length of the header string.

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
```diff
@@ -140,6 +148,28 @@ public String getCloudTrailEventSource() {
...
+     * Returns the documentation identifier value for the service.
+     *
+     * <p> When value on trait is not set, this method defaults to the lower
+     * cased value of the sdkId followed by the service version, separated by
+     * dashes.
+     *
+     * @return Returns the documentation identifier value for the service name.
+     */
+    public String getDocId(Model model) {
+        return getDocId().orElse(buildDefaultDocId(model));
```
kstich, line 161: minor: This should use `orElseGet` with a Supplier that returns the value currently place. Otherwise, the `buildDefaultDocId` method is called even when `getDocId()` has a value.

PR #1866, opened 2023-07-19
Slightly relax paginated pageSize types

kstich:


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PaginatedTraitValidator.java
```diff
@@ -319,6 +321,10 @@ Set<ShapeType> validTargets() {
             return PAGE_SHAPES;
         }
 
+        Set<ShapeType> dangerTargets() {
```
sugmanue, line 324: An `@Override` would be helpful to understand the flow here.

PR #1870, opened 2023-07-20
Improve segment parsing errors with offsets

### LINKED ISSUE #904: Unused offset within smithy.model.pattern.SmithyPattern.Segment.Parse
Hi, I think there's a small bug with the Segment Parser. 

It appears that the `offset` variable is currently unused within the `smithy.model.pattern.SmithyPattern.Segment.Parse` function. 
https://github.com/awslabs/smithy/blob/2e1cdc4af7e7cdab3138f5f3912a3af930b569a2/smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java#L276

All callers of this function also appear to be calling this `parse` function incorrectly and are exhibiting correct parsing behaviour because the `offset` variable is unused.

The call to `Segment.parse` at https://github.com/awslabs/smithy/blob/2e1cdc4af7e7cdab3138f5f3912a3af930b569a2/smithy-model/src/main/java/software/amazon/smithy/model/pattern/UriPattern.java#L80
Should always have an offset of 0 because the string is first split into chunks by `/`.

The call to `Segment.parse` at https://github.com/awslabs/smithy/blob/2e1cdc4af7e7cdab3138f5f3912a3af930b569a2/smithy-model/src/test/java/software/amazon/smithy/model/pattern/SmithyPatternTest.java#L39
Same as above, should always be 0 given that the string is chunked by '/' first.

The call to `Segment.parse` at https://github.com/awslabs/smithy/blob/2e1cdc4af7e7cdab3138f5f3912a3af930b569a2/smithy-model/src/main/java/software/amazon/smithy/model/traits/EndpointTrait.java#L66
This should also have the offset always set to 0 because the string is first tokenized into chunks.

A new test should be added to verify that the offset is used correctly.

Alternatively, the `offset` parameter could be removed but this would be a breaking API change so less desirable, up to you though.

I'm happy to submit a PR to correct the issue if the above is correct.

kstich:
Fixes #904 

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java
@@ -235,10 +235,14 @@ public class SmithyPattern {
         private final Type segmentType;
 
         public Segment(String content, Type segmentType) {
+            this(content, segmentType, null);
+        }
+
+        public Segment(String content, Type segmentType, Integer offset) {
             this.content = Objects.requireNonNull(content);
             this.segmentType = segmentType;
 
-            checkForInvalidContents();
+            checkForInvalidContents(offset);
 
             if (segmentType == Type.GREEDY_LABEL) {
                 asString = "{" + content + "+}";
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/pattern/SmithyPattern.java
```diff
@@ -249,27 +253,33 @@ public Segment(String content, Type segmentType) {
             }
         }
 
-        private void checkForInvalidContents() {
+        private void checkForInvalidContents(Integer offset) {
```
kstich, line 256: This is boxed because `0` is a valid offset.

AndrewFossAWS (approved):

PR #1874, opened 2023-07-21
Add workflow to handle long running issues

jvschneid:
*Description of changes:* Add workflow to handle long running issues and auto-close given labels and time windows.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -22,7 +22,7 @@ jobs:
 
           # Labels
           stale-issue-label: closing-soon
-          exempt-issue-label: bug,documentation,feature-request,guidance,protocol-test
+          exempt-issue-label: bug,documentation,feature-request,guidance,investigating,protocol-test
           response-requested-label: response-requested
 
           # Messages
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -22,7 +22,7 @@ jobs:
 
           # Labels
           stale-issue-label: closing-soon
-          exempt-issue-label: bug,documentation,feature-request,guidance,investigating,protocol-test
+          exempt-issue-label: bug,documentation,feature-request,investigating,protocol-test
           response-requested-label: response-requested
 
           # Messages
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -17,7 +17,7 @@ jobs:
 
           # Issue timing
           days-before-stale: 60
-          days-before-close: 7
+          days-before-close: 14
           days-before-ancient: 365
 
           # Labels
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,43 @@
+name: "Close stale issues"
+
+# Controls when the action will run.
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,43 @@
...
+
+          # If you don't want to mark a issue as being ancient based on a
+          # threshold of "upvotes", you can set this here. An "upvote" is
+          # the total number of +1, heart, hooray, and rocket reactions
+          # on an issue.
+          minimum-upvotes-to-exempt: 5
+
+          # Don't set closed-for-staleness label to skip closing very old issues
+          # regardless of label
+          closed-for-staleness-label: closed-for-staleness
```
kstich, line 40: This label doesn't exist.

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,43 @@
...
+    name: Stale issue job
+    steps:
+      - uses: aws-actions/stale-issue-cleanup@v3
+        with:
+          # Targeted types
+          issue-types: issues
+
+          # Issue timing
+          days-before-stale: 60
+          days-before-close: 7
```
kstich, line 20: Let's up this to 14 days.

kstich (changes requested):

.github/workflows/stale-issues.yml
jvschneid: Yes, my plan is to create it once the PR is approved and there are no concerns with naming (:

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,43 @@
...
+          issue-types: issues
+
+          # Issue timing
+          days-before-stale: 60
+          days-before-close: 14
+          days-before-ancient: 365
+
+          # Labels
+          stale-issue-label: closing-soon
+          exempt-issue-label: documentation,feature-request
```
kstich, line 25: Let's look at the labels on the repo/in-use and see if this is the right list. It probably needs at least `bug` added.

kstich:
Should also make sure all the issues in the repo are properly labeled before merging this, so it doesn't start auto-closing things we previously didn't label properly.

jvschneid:
> Should also make sure all the issues in the repo are properly labeled before merging this, so it doesn't start auto-closing things we previously didn't label properly.

I did a pass on the issues this morning, feel free to do a quick glance at the unlabeled ones and see if I missed any that should be labeled though (:

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,43 @@
...
+          issue-types: issues
+
+          # Issue timing
+          days-before-stale: 60
+          days-before-close: 14
+          days-before-ancient: 365
+
+          # Labels
+          stale-issue-label: closing-soon
+          exempt-issue-label: bug,documentation,feature-request,guidance,protocol-test
```
kstich, line 25: Shouldn't close `investigating`.

I think `guidance` will need to wait to come off this list until we've either tagged them as `documentation` or chosen that they need no further update.

kstich: I've labeled everything with a `guidance` label that needs more work with another label representing that work. The `guidance` label can be removed from this list now and we can move forward.

kstich (approved):

PR #1875, opened 2023-07-21
Fix indentation when formatting text blocks

milesziemer:
Updates the formatter to indent the node value of a node kvp when they are text blocks.

A test case was added for a text block as the node value of a node kvp, and a test case was updated for when text blocks are in traits.

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/Formatter.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/Formatter.java
@@ -17,10 +17,12 @@ package software.amazon.smithy.syntax;
 
 import com.opencastsoftware.prettier4j.Doc;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


smithy-syntax/src/main/java/software/amazon/smithy/syntax/Formatter.java
```diff
@@ -427,9 +443,18 @@ private Doc visit(TreeCursor cursor) {
                            : Doc.text(tree.concatTokens());
                 }
 
+                case TEXT_BLOCK: {
+                    // Dispersing the lines of the text block preserves any indentation applied from formatting parent
+                    // nodes.
+                    List<Doc> lines = Arrays.stream(tree.concatTokens()
+                            .split(System.lineSeparator()))
```
kstich, line 450: This belongs on the line above, as it's an operation on the result of `concatTokens()` and not on the result of `Arrays.stream()`.

kstich (approved):

Issue #1844, opened 2023-07-04
OpenAPI plugin generates incorrect schema for enums with default values


christophercurrie:
Given the following Smithy model:

```
$version: "2.0"

namespace org.fizmo.test

use aws.protocols#restJson1

@restJson1
service OpenApiDefault {
    version: "1"
    operations: [
        TestDefault
    ]
}

@http(
    method: "GET",
    uri: "/test"
)
@readonly
operation TestDefault {
    input := {}
    output := {
        test: TestEnum = "FOO"
    }
}

enum TestEnum {
    FOO
}
```

configured with the `smithy-build.json`:

```
{
  "version": "2.0",
  "plugins": {
    "openapi": {
      "service": "org.fizmo.test#OpenApiDefault",
      "protocol": "aws.protocols#restJson1"
    }
  }
}
```

The generator produces this in the "schemas" section:

```
"TestDefaultResponseContent": {
    "type": "object",
    "properties": {
        "test": {
            "$ref": "#/components/schemas/TestEnum",
            "default": "FOO"
        }
    }
}
```

According to [the OpenAPI spec](https://spec.openapis.org/oas/latest.html#referenceObject), ReferenceObjects are not allowed to take any properties other than `$ref`, `summary`, and `description`. When the generated spec is given to the [Swagger UI](https://editor.swagger.io/), it shows an error at the "default" line:

> Sibling values alongside $refs are ignored.
> To add properties to a $ref, wrap the $ref into allOf, or move the extra properties into the referenced definition (if applicable).

This suggests that the correct generated schema should be along the lines of:

```
"TestDefaultResponseContent": {
    "type": "object",
    "properties": {
        "test": {
            "allOf": [
                { "$ref": "#/components/schemas/TestEnum" },
                { "default": "FOO" }
            ]
        }
    }
}
```

PR #1887, opened 2023-07-27
Wrap refs with defaults in an allOf

srchase:
Fixes https://github.com/smithy-lang/smithy/issues/1844.

According to both the [3.1.0 spec](https://spec.openapis.org/oas/latest.html#referenceObject), and the [3.0.2 spec](https://spec.openapis.org/oas/v3.0.2#referenceObject), additional fields on `referenceObjects` are ignored.

This PR adds a JsonSchema configuration option, `enableDefaultValues` which is enabled by default. JsonSchemaShapeVisitor is updated to wrap any members with a `@default` trait applied which target a `$ref` with an `allOf` so that the default can be handled correctly by OpenAPI when this configuration option is enabled.

While JsonSchema's configuration will use this setting by default, it is disabled by default for API Gateway. API Gateway does not support default values anyways, so disabling this option will make `$ref` and top-level definitions consistent by not using defaults in either form.

### DIFF
--- a/docs/source-2.0/guides/converting-to-openapi.rst
+++ b/docs/source-2.0/guides/converting-to-openapi.rst
@@ -790,6 +790,89 @@ supportNonNumericFloats (``boolean``)
         }
 
 
+.. _generate-openapi-setting-disableDefaultValues:
+
+disableDefaultValues (``boolean``)
+    Set to true to disable adding default values.
+
+    .. code-block:: json
+
+        {
+            "version": "2.0",
+            "plugins": {
+                "openapi": {
+                    "service": "example.weather#Weather",
+                    "disableDefaultValues": true
+                }
+            }
+        }
+
+    With this disabled, default values will not appear in the output:
+
+    .. code-block:: json
+
+        {
+            "Foo": {
+                "type": "object",
+                "properties": {
+                    "bam": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    },
+                    "bar": {
+                        "type": "number"
+                    },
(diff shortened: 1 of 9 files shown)


mtdowling:
Can you test this to make sure it works with APIGW?

smithy-jsonschema/src/test/resources/software/amazon/smithy/jsonschema/default-values.jsonschema.v07.json
```diff
@@ -14,11 +14,28 @@
...
                     "default": 0
                 },
+                "bat": {
+                    "allOf": [
+                        {
+                            "$ref": "#/definitions/TestEnum"
+                        },
+                        {
+                            "default": "FOO"
+                        }
```
mtdowling, line 24: I guess this is a better way to do member documentation traits too

srchase:
> Can you test this to make sure it works with APIGW?

Updated the issue description with API Gateway testing.

docs/source-2.0/guides/converting-to-openapi.rst
```diff
@@ -790,6 +790,89 @@ supportNonNumericFloats (``boolean``)
         }
 
 
+.. _generate-openapi-setting-enableDefaultValues:
+
+enableDefaultValues (``boolean``)
```
mtdowling, line 795: I'd make this `disableDefaultValues` and default it to `false`. booleans with `true` default values are unintutitive.

mtdowling (changes requested):

mtdowling (approved):

PR #1896, opened 2023-08-01
Cache template directory in init command

hpmellema:
*Description of changes:*
Caches a copy of the smithy init template directory in the default system temp directory. This greatly speeds up subsequent executions of both the list and create template subcommands of the `smithy init`. 
Previously, creating a template could take up to 6 seconds. With this change that time is down to ~0.3sec after the initial caching of the template repo.

---

*Example Executions*
Create a Template
cli

List Templates 
list

Note: the gif recorder used here adds additional lag so these do not accurately represent the time these commands take to execute.
___

### DIFF
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/InitCommand.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/commands/InitCommand.java
@@ -45,8 +45,6 @@ import software.amazon.smithy.utils.StringUtils;
 
 
 final class InitCommand implements Command {
-    private static final Path DEFAULT_TEMP_DIR_PATH = Paths.get(System.getProperty("java.io.tmpdir"));
-    private static final Path SMITHY_INIT_CACHE_DIR = DEFAULT_TEMP_DIR_PATH.resolve("smithy-templates");
     private static final String SMITHY_TEMPLATE_JSON = "smithy-templates.json";
     private static final String DEFAULT_REPOSITORY_URL = "https://github.com/smithy-lang/smithy-examples.git";
     private static final String DEFAULT_TEMPLATE_NAME = "quickstart-cli";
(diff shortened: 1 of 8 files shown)


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/InitCommand.java
```diff
@@ -86,32 +90,100 @@ private int run(Arguments arguments, Env env) {
...
         StandardOptions standardOptions = arguments.getReceiver(StandardOptions.class);
 
-        try {
-            Path root = Paths.get(".");
-            Path temp = Files.createTempDirectory("temp");
+        boolean isLocalRepo = isLocalRepo(options.repositoryUrl);
+        Path templateRepoDirPath = getTemplateRepoDirPath(options.repositoryUrl, isLocalRepo);
+
+        // If the cache directory does not exist, create it
+        if (!Files.exists(templateRepoDirPath)) {
```
AndrewFossAWS, line 97: Does it make sense to add a param to allow user to invalidate the cache?

mtdowling: `smithy clean` does that

mtdowling:
rather than cache to the temp directory, I suggest caching to the build directory, allowing `smithy clean` to be used to clear any caches.

mtdowling (changes requested):
Is it possible to add integration tests for the added functionality of the init command?

mtdowling (approved):

hpmellema:
For some reason the file lock on the `.git` pack files on the windows runner do not seem to be released (possibly due to the windows git file watcher?). I am working on trying to find the root cause and a fix, but in the mean time this PR will have to wait for that fix.

PR #1902, opened 2023-08-02
Clarify constraint trait enforcement

### LINKED ISSUE #1039: Question concerning constraint traits when applied to output shapes
Is Smithy opinionated, or is it specified, about how constraint traits should be handled when applied to output shapes?

For example, given the (somewhat contrived) example:

```
service ExampleService {
  version: "1",
  operations: [ ExampleOp ]
}

operation ExampleOp {
  output: ExampleOpOutput
}

structure ExampleOpOutput {
  @range(min: 1, max: 10)
  exampleField: Integer
}

```

What is the expected behavior for server-side code that implements this specification?
- Are constraint traits for output shapes ignored?  Meaning the server is free to return a value of, say, 11 for `exampleField`?
- Must conforming server implementations (and code-generators that generate server code) emit code that ensures that return values  also adhere to any constraints applied to them?  

For context as to what prompted the question: as part of the implementation of our own Smithy code-generator I've been looking into the existing open source smithy-code generators (smithy-typescript and smithy-rs, which may or may not be authored by teams separate to yours), neither of which enforce constraint traits on output members for generated server code. I realize that this might be due to this not having been implemented _yet_ in either of those two libraries and might be implemented in the future.  My concern is that I've overlooked a piece of documentation that permits this behavior, which is why the functionality in those libraries does not exist.  But **as the authors/owners of the base smithy library and specification, would your general expectation be that any code-generators that emit server code can only be described as specification-conforming if constraint traits are respected for output values?**

Knowing this information, if constraint traits should be respected for output shapes, then I can follow this up with separate issues/conversations with the smithy-typescript/smithy-rs maintainers. 

Thanks!

mtdowling:
Closes #1039

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source-2.0/spec/constraint-traits.rst
+++ b/docs/source-2.0/spec/constraint-traits.rst
@@ -12,6 +12,20 @@ impact the types signatures of generated code.
        that also functions like a constraint.
 
 
+Constraint trait enforcement
+============================
+
+Constraint traits SHOULD be enforced after deserializing input. For example,
+when a server deserializes a request from a client, the server SHOULD enforce
+any defined constraint traits and reject the request if appropriate.
+
+Constraint traits SHOULD NOT be enforced when serializing shapes or when
+deserializing output. For example, when returning a response from a server
+to a client, failing to serialize a response due to a constraint trait
+violation would prevent a client from observing a state change of the server
+and would provide no real recourse for the client or server to recover.
+
+
 .. smithy-trait:: smithy.api#idRef
 .. _idref-trait:
 


docs/source-2.0/spec/constraint-traits.rst
```diff
@@ -12,6 +12,20 @@ impact the types signatures of generated code.
...
+when a server deserializes a request from a client, the server SHOULD enforce
+any defined constraint traits and reject the request if appropriate.
+
+Constraint traits SHOULD NOT be enforced when serializing shapes or when
+deserializing output. For example, when returning a response from a server
+to a client, failing to serialize a response due to a constraint trait
+violation would prevent a client from observing a state change of the server
+and would provide no real recourse for the client or server to recover.
+
+
```
sugmanue, line 28: A link or note to [this section](https://smithy.io/2.0/guides/building-codegen/mapping-shapes-to-languages.html#why-don-t-we-validate-constraint-traits-on-the-client) would be helpful here, some users get confused as of why they can build invalid structs.

sugmanue (approved):

docs/source-2.0/spec/constraint-traits.rst
mtdowling: I try not to link from the normative spec out of the spec to non-normative content

PR #1903, opened 2023-08-02
Fix NPE when docId is null when ServiceTrait.equals is called

AndrewFossAWS:


jvschneid (approved):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/ServiceTrait.java
```diff
@@ -232,9 +232,9 @@ public boolean equals(Object other) {
                     && arnNamespace.equals(os.arnNamespace)
                     && cloudFormationName.equals(os.cloudFormationName)
                     && cloudTrailEventSource.equals(os.cloudTrailEventSource)
-                    && docId == null
+                    && (docId == null
                         ? os.docId == null
-                        : docId.equals(os.docId)
+                        : docId.equals(os.docId))
```
sugmanue, line 237: `Objects.equals` is designed for this case in particular, e.g.,

```
&& Objects.equals(docId, os.docId)
```

sugmanue (approved):

Issue #1664, opened 2023-03-07
Improve OpenAPI support for enums as map keys


denisrosca:
Improve enum support as map keys by utilizing the newly added `propertyNames` JSON Schema in OpenAPI 3.1.

### Background

There's a mismatch in support of enums as map keys between Smithy and OpenAPI. Currently when converting a smithy definition containing a map shape with an enum as key, the generated Open API spec discards all enum information and treats the map shape a simple string to object mapping.

Given this simple Smithy service:
```
$version: "2"

namespace enumkeys.example

use aws.protocols#restJson1

@restJson1
service ExampleService {
  version: "1.0.0"
  operations: [MajorCitiesByCountry]
}

@http(method: "GET", uri: "/cities", code: 200)
@readonly
operation MajorCitiesByCountry {
  output : Response
}

enum Country {
  UK = "UK"
  FR = "FR"
  US = "US"
  DE = "DE"
}

string City

list Cities {
    member: City
}

map CitiesByCountry {
  key: Country
  value: Cities
}

structure Response {
  @required
  majorCities: CitiesByCountry
}
```

the output open api spec looks like this (notice how the `CitiesByCountry` object has lost all traces of the enum and it's values):

```
{
    "openapi": "3.1.0",
    "info": {
        "title": "ExampleService",
        "version": "1.0.0"
    },
    "paths": {
        "/cities": {
            "get": {
                "operationId": "MajorCitiesByCountry",
                "responses": {
                    "200": {
                        "description": "MajorCitiesByCountry 200 response",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/MajorCitiesByCountryResponseContent"
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "components": {
        "schemas": {
            "CitiesByCountry": {
                "type": "object",
                "additionalProperties": {
                    "type": "array",
                    "items": {
                        "type": "string"
                    }
                }
            },
            "MajorCitiesByCountryResponseContent": {
                "type": "object",
                "properties": {
                    "majorCities": {
                        "$ref": "#/components/schemas/CitiesByCountry"
                    }
                },
                "required": [
                    "majorCities"
                ]
            }
        }
    }
}
```

### Suggested approach
Instead of discarding all enum information, we could leverage the newly added support for the `propertyNames` JSON Schema and output the `CitiesByCountry` object like this:

```
...
    "components": {
        "schemas": {
            "CitiesByCountry": {
[...]

PR #1905, opened 2023-08-02
Support enum map keys on OpenAPI 3.1.0

srchase:
Closes: https://github.com/smithy-lang/smithy/issues/1664

This PR updates OpenAPI configuration to allow `propertyNames` to be set when using version `3.1.0`.

Previously, `propertyNames` was stripped from outputs as an unsupported keyword, since OpenAPI 3.0 did not support it.

By allowing `propertyNames`, enums can be used as map keys, without discarding all of the enum information.

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiVersion.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiVersion.java
@@ -15,25 +15,32 @@
 
 package software.amazon.smithy.openapi;
 
+import java.util.Set;
 import software.amazon.smithy.jsonschema.JsonSchemaVersion;
+import software.amazon.smithy.utils.SetUtils;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
 /**
  * OpenAPI versions supported by the converter.
  */
 public enum OpenApiVersion {
-
-    VERSION_3_0_2("3.0.2", false, JsonSchemaVersion.DRAFT07),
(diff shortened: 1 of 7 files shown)


smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/OpenApiJsonSchemaMapperTest.java
```diff
@@ -67,23 +66,6 @@ public void convertsModels() {
         assertTrue(document.toNode().expectObjectNode().getMember("components").isPresent());
     }
 
-    @Test
-    public void stripsUnsupportedKeywords() {
```
srchase, line 71: This behavior is now covered by the `convertsToOpenAPI3_0_2` and `convertsToOpenAPI3_1_0` tests in OpenApiConverterTest.

sugmanue (approved):

smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiVersion.java
```diff
@@ -46,6 +53,11 @@ public boolean supportsContentEncodingKeyword() {
         return supportsContentEncodingKeyword;
     }
 
+    @SmithyInternalApi
+    public Set<String> getUnsupportedKeywords() {
```
mtdowling, line 57: Oh interesting to make this version specific. Is this internal because you aren't sure about the design? Maybe unstable instead if that's the concern?

PR #1911, opened 2023-08-07
RpcV2 Protocol tests

adwsingh:
*Issue #, if available:*

*Description of changes:*

Draft PR, I am going to add more protocol tests. Looking for early feedback.


smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,173 @@
...
+namespace aws.protocoltests.rpcv2
+
+use smithy.protocols#rpcv2
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+
+
+@httpRequestTests([
+    {
+        id: "no_input",
```
david-perez, line 12: I think the convention for these ids is to be camelCase.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,173 @@
...
+        uri: "/service/aws.protocoltests.rpcv2.RpcV2Protocol/operation/OptionalInputOutput",
+        body: "v/8=",
+    },
+])
+@httpResponseTests([
+    {
+        id: "optional_output",
+        protocol: rpcv2,
+        documentation: "When output is empty we write CBOR equivalent of {}",
+        body: "",
```
david-perez, line 161: My implementation fails this test because it generates:

```rust
pub fn ser_optional_input_output_output(
    encoder: &mut ::aws_smithy_cbor::Encoder,
    input: &crate::output::OptionalInputOutputOutput,
) -> Result<(), ::aws_smithy_http::operation::error::SerializationError> {
    encoder.begin_map();
    if let Some(var_1) = &input.value {
        encoder.str("value").str(var_1.as_str());
    }
    encoder.end();
    Ok(())
}
```

So when no members are set, it writes an empty indefinite-length CBOR map (a body with two bytes: [0xbf, 0xff]) instead of an empty map.

I'd expect all implementations of the protocol to do what I'm doing here: it's costly to calculate the amount of set members in a structure in order to use a definite-length map or not write anything at all.

I'm also unsure about whether expecting nothing at all here is compliant, or desirable. A deserializer is going to expect a map; to accomodate for the case where nothing is on the wire it would have to (1) write the expectation fallibly, and (2) in the error case check that the decoder is at the end of the document, returning an empty structure.

---

The question about how to tell Smithy protocol tests that it should accept definite-length as well as indefinite-length encodings is still open and I don't know how to solve.

rhernandez35: > I don't see why can't both definite (0xa0) and indefinite-length map (0xbf,0xff) both represent an empty map.

That's valid and also not what your test is asserting. The CBOR equivalent of `{}` is a map type with zero elements, but we're asserting that no payload at all should be written in this case.

smithy-aws-protocol-tests/model/rpcV2/cbor-structs.smithy
```diff
@@ -0,0 +1,141 @@
...
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
+        }
+        method: "POST",
+        bodyMediaType: "application/cbor",
+        uri: "/service/aws.protocoltests.rpcv2.RpcV2Protocol/operation/SimpleScalarProperties",
+        body: "v2lieXRlVmFsdWUFa2RvdWJsZVZhbHVl+z/+OVgQYk3TcWZhbHNlQm9vbGVhblZhbHVl9GpmbG9hdFZhbHVl+kDz989saW50ZWdlclZhbHVlGQEAaWxvbmdWYWx1ZRkmkWpzaG9ydFZhbHVlGSaqa3N0cmluZ1ZhbHVlZnNpbXBsZXB0cnVlQm9vbGVhblZhbHVl9f8="
+        params: {
```
david-perez, line 25: `longValue` is not set [but we're expecting it to be set to 9873](https://cbor.nemo157.com/?type=hex&value=bf696279746556616c7565056b646f75626c6556616c7565fb3ffe395810624dd37166616c7365426f6f6c65616e56616c7565f46a666c6f617456616c7565fa40f3f7cf6c696e746567657256616c7565190100696c6f6e6756616c75651926916a73686f727456616c75651926aa6b737472696e6756616c75656673696d706c657074727565426f6f6c65616e56616c7565f5ff).

smithy-aws-protocol-tests/model/rpcV2/cbor-structs.smithy
```diff
@@ -0,0 +1,142 @@
...
+    {
+        id: "simple_scalar_structure",
+        protocol: rpcv2,
+        documentation: "Serializes simple scalar properties",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Content-Type": "application/cbor"
+        }
+        bodyMediaType: "application/cbor",
+        body: "v2lieXRlVmFsdWUFa2RvdWJsZVZhbHVl+z/+OVgQYk3TcWZhbHNlQm9vbGVhblZhbHVl9GpmbG9hdFZhbHVl+kDz989saW50ZWdlclZhbHVlGQEAaWxvbmdWYWx1ZRkmkWpzaG9ydFZhbHVlGSaqa3N0cmluZ1ZhbHVlZnNpbXBsZXB0cnVlQm9vbGVhblZhbHVl9f8=",
```
david-perez, line 82: `longValue` is set to 9873 in `body` but is not set in `params`.

smithy-aws-protocol-tests/model/rpcV2/cbor-structs.smithy
```diff
@@ -0,0 +1,142 @@
...
+        protocol: rpcv2,
+        documentation: "Serializes simple scalar properties",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
+        }
+        method: "POST",
+        bodyMediaType: "application/cbor",
+        uri: "/service/aws.protocoltests.rpcv2.RpcV2Protocol/operation/SimpleScalarProperties",
```
david-perez, line 23: `uri` must only contain the service name's shape name, a Smithy identifier.

> Every request for the `rpcv2` protocol MUST be sent to a URL with the following form: `{prefix?}/service/{serviceName}/operation/{operationName}`
> 
> * The optional `prefix` slug may span multiple path segments and is ignored by Smithy RPC v2.
> * The `serviceName` slug MUST be replaced by the [`shape name`](https://smithy.io/2.0/spec/model.html#grammar-token-smithy-Identifier) of the service's [Shape ID](https://smithy.io/2.0/spec/model.html#shape-id) in the Smithy model. The `serviceName` MUST NOT contain the namespace of the `service` shape.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,173 @@
...
+        body: ""
+    },
+    {
+        id: "no_input_server_allows_accept",
+        protocol: rpcv2,
+        documentation: "Servers should allow the Accept header to be set to the default content-type.",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
```
david-perez, line 34: `Content-Type` should be forbidden here; the operation has no input.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,173 @@
...
+        uri: "/service/aws.protocoltests.rpcv2.RpcV2Protocol/operation/NoInputOutput",
+        body: ""
+    },
+    {
+        id: "no_input_server_allows_accept",
+        protocol: rpcv2,
+        documentation: "Servers should allow the Accept header to be set to the default content-type.",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
```
david-perez, line 33: The server can't really honor sending a CBOR response here right? The operation has no output; the payload must be empty.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,173 @@
...
+        appliesTo: "server"
+    },
+    {
+        id: "no_input_server_allows_empty_cbor",
+        protocol: rpcv2,
+        documentation: "Servers should accept CBOR empty struct if no input.",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
```
david-perez, line 48: Same comments apply here.

david-perez (changes requested):
I did another pass against my implementation. I found a few bugs that I have now fixed. As of writing, against the latest pushed commit, https://github.com/smithy-lang/smithy-rs/pull/2544/commits/636bba95139e9aebb7044afcad8df884cfc53ab3 in [my PR](https://github.com/smithy-lang/smithy-rs/pull/2544), the following 11 tests fail:

```
failures:
    operation::server_greeting_with_errors_test::rpc_v2_cbor_empty_complex_error_response
    operation::server_no_input_output_test::no_input_server_allows_accept_request
    operation::server_no_input_output_test::no_input_server_allows_empty_cbor_request
    operation::server_no_input_output_test::no_input_server_ingores_unexpected_fields_request
    operation::server_no_input_output_test::no_output_response
    operation::server_rpc_v2_cbor_maps_test::rpc_v2_cbor_deserializes_dense_set_map_response
    operation::server_rpc_v2_cbor_maps_test::rpc_v2_cbor_deserializes_null_map_values_response
    operation::server_rpc_v2_cbor_maps_test::rpc_v2_cbor_serializes_dense_set_map_request
    operation::server_rpc_v2_cbor_maps_test::rpc_v2_cbor_serializes_null_map_values_request
    operation::server_simple_scalar_properties_test::rpc_v2_cbor_server_doesnt_de_serialize_null_structure_values_request
    operation::server_simple_scalar_properties_test::rpc_v2_cbor_simple_scalar_properties_response
```

Most fail due to minor mistakes in the tests' definitions; I've commented on each one indicating where I think they're wrong. However, in general, in the spec:

- we need to be clear on what to accept/reject when (1) the modeled operation has no input/output, (2) the input/output shapes are unit shapes, and (3) the input/output shapes are empty. My preference would be to be as strict as possible and only accept empty payloads for (1) and (2), and and empty map for (3).
- we need to be clear on when to set the `application/cbor` media type in `Accept` and `Content-Type`. My preference would be to be as strict as possible and only set it when we should only accept or return a non-empty HTTP body, respectively.

In terms of test suite completeness, I think we still have a long way to go. I haven't thought exhaustively about what's missing, but off the top of my head, we need tests for (in order of importance):

- [indefinite-length encodings](https://www.rfc-editor.org/rfc/rfc8949.html#name-indefinite-lengths-for-some) (as far as I can tell all string encodings are definite)
- boxed (i.e. `@box`) shapes
- event streams
- `document` shapes
- [default values](https://smithy.io/2.0/spec/type-refinement-traits.html#default-trait)

I would also like to ask the Smithy team that they consider encoding bodies in a text format that is amenable for review: https://github.com/smithy-lang/smithy/issues/2108

smithy-aws-protocol-tests/model/rpcV2/cbor-maps.smithy
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }
+    },
+    {
+        id: "RpcV2CborSerializesNullMapValues",
+        documentation: "Serializes map values in sparse maps",
+        protocol: rpcv2Cbor,
+        method: "POST",
+        uri: "/service/RpcV2Protocol/operation/RpcV2CborMaps",
+        body: "v3BzcGFyc2VCb29sZWFuTWFwv2F49v9vc3BhcnNlTnVtYmVyTWFwv2F49v9vc3BhcnNlU3RydWN0TWFwv2F49v//",
```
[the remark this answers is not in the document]
lucix-aws: Confirmed in go although I don't think the above link is right.

https://cbor.nemo157.com/?type=hex&value=bf70737061727365426f6f6c65616e4d6170bf6178f6ff6f7370617273654e756d6265724d6170bf6178f6ff6f7370617273655374727563744d6170bf6178f6ffff#type=hex&value=bf70737061727365426f6f6c65616e4d6170bf6178f6ff6f7370617273654e756d6265724d6170bf6178f6ff6f7370617273655374727563744d6170bf6178f6ffff

smithy-aws-protocol-tests/model/rpcV2/cbor-maps.smithy
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }
+    },
+    {
+        id: "RpcV2CborSerializesDenseSetMap",
+        documentation: "A request that contains a dense map of sets.",
+        protocol: rpcv2Cbor,
+        method: "POST",
+        uri: "/service/RpcV2Protocol/operation/RpcV2CborMaps",
+        body: "v2xzcGFyc2VTZXRNYXC/YXmfYWFhYv9heJ////8=",
```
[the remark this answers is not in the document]
lucix-aws: Seconded, corrected link: https://cbor.nemo157.com/#type=hex&value=bf6c7370617273655365744d6170bf61799f61616162ff61789fffffff

smithy-aws-protocol-tests/model/rpcV2/cbor-structs.smithy
```diff
@@ -0,0 +1,322 @@
...
+        protocol: rpcv2Cbor,
+        documentation: "Serializes simple scalar properties",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
+        }
+        method: "POST",
+        bodyMediaType: "application/cbor",
+        uri: "/service/aws.protocoltests.rpcv2Cbor.RpcV2Protocol/operation/SimpleScalarProperties",
```
lucix-aws, line 22: Saw this commented elsewhere and marked fixed, but we still have a handful of instances of the full namespace in the path when it shouldn't be.

smithy-aws-protocol-tests/model/rpcV2/cbor-structs.smithy
```diff
@@ -0,0 +1,322 @@
...
+        documentation: "Supports handling NaN float values.",
+        headers: {
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
+        }
+        method: "POST",
+        bodyMediaType: "application/cbor",
+        uri: "/service/RpcV2Protocol/operation/SimpleScalarProperties",
+        body: "v2tkb3VibGVWYWx1Zft/+AAAAAAAAGpmbG9hdFZhbHVl+n/AAAD/"
```
lucix-aws, line 82: Worth noting that different implementations may expose different values for general NaN. We've encoded f64 NaN here as `7ff8000000000000` but go has defined it as `0x7ff8000000000001`, both technically NaN but fails a bitwise comparison. Not sure if we care, though. We should probably be recommending that people do document-based equality checks for these comparisons.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,187 @@
...
+        },
+        forbidHeaders: [
+            "X-Amz-Target"
+        ]
+        method: "POST",
+        uri: "/service/aws.protocoltests.rpcv2Cbor.RpcV2Protocol/operation/EmptyInputOutput",
+        body: "v/8=",
```
lucix-aws, line 130: This needs `bodyMediaType` if it's expecting a CBOR body.

smithy-aws-protocol-tests/model/rpcV2/empty-input-output.smithy
```diff
@@ -0,0 +1,187 @@
...
+            "smithy-protocol": "rpc-v2-cbor",
+            "Accept": "application/cbor",
+            "Content-Type": "application/cbor"
+        },
+        forbidHeaders: [
+            "X-Amz-Target"
+        ]
+        method: "POST",
+        uri: "/service/aws.protocoltests.rpcv2Cbor.RpcV2Protocol/operation/OptionalInputOutput",
+        body: "v/8=",
```
lucix-aws, line 167: [same](https://github.com/smithy-lang/smithy/pull/1911/files#r1506777892)

kstich:
Closing in favor of #2188

PR #1914, opened 2023-08-09
Add support for suppressions to smithy-diff

### LINKED ISSUE #1861: Does the `ChangedMemberOrder` validation in `smithy diff` apply to anything other than C or C++ ?
I have created a `smithy diff` step in my release system to check for breaking changes. I started getting `ChangedMemberOrder` errors with this message

> union shape members were reordered. This can cause ABI compatibility issues in languages like C and C++ where the layout and alignment of a data structure matters.

## Questions

1. Does this validation apply to anything other than C or C++? I'm most interested in Javascript, Python, Java, and Rust. We don't use C or C++.
2. Is there a way to suppress specific errors codes from `smithy diff`? I don't see anything in `smithy diff --help` which I can use. As long as my languages aren't impacted by `ChangedMemberOrder`, I'd like to be able to do something like `smithy diff --suppress ChangedMemberOrder`

mtdowling:
Closes #1861

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/ModelDiff.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/ModelDiff.java
@@ -27,6 +27,8 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidatedResult;
 import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationEventDecorator;
+import software.amazon.smithy.model.validation.suppressions.ModelBasedEventDecorator;
 import software.amazon.smithy.utils.SmithyBuilder;
 
 /**
@@ -280,8 +282,17 @@ public final class ModelDiff {
             List<DiffEvaluator> evaluators = new ArrayList<>();
             ServiceLoader.load(DiffEvaluator.class, classLoader).forEach(evaluators::add);
             Differences differences = Differences.detect(oldModel, newModel);
+
+            // Applies suppressions and elevates event severities.
+            ValidationEventDecorator decoratorResult = new ModelBasedEventDecorator()
+                    .createDecorator(newModel)
+                    .getResult()
+                    .orElse(ValidationEventDecorator.IDENTITY);
+
             List<ValidationEvent> diffEvents = evaluators.parallelStream()
                     .flatMap(evaluator -> evaluator.evaluate(differences).stream())
+                    // No need to call canDecorate first since that method will always return true in any code path.
+                    .map(decoratorResult::decorate)
                     .collect(Collectors.toList());
 
             return new Result(differences, diffEvents, oldModelEvents, newModelEvents);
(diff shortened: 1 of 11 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/ModelDiff.java
```diff
@@ -280,8 +282,16 @@ public Result compare() {
...
+
+            // Applies suppressions and elevates event severities.
+            ValidationEventDecorator decoratorResult = new ModelBasedEventDecorator()
+                    .createDecorator(newModel)
+                    .getResult()
+                    .orElse(ValidationEventDecorator.IDENTITY);
+
             List<ValidationEvent> diffEvents = evaluators.parallelStream()
                     .flatMap(evaluator -> evaluator.evaluate(differences).stream())
+                    .map(decoratorResult::decorate)
```
sugmanue, line 294: Not calling here `canDecorate()` has an implicit assumption that the resulting decorator will return always `true`, I'd think that just calling the method as done somewhere else or adding a comment to make it explicit will avoid confusing readers.

smithy-model/src/main/java/software/amazon/smithy/model/validation/suppressions/ModelBasedEventDecorator.java
```diff
@@ -0,0 +1,198 @@
...
+    private final List<Suppression> suppressions = new ArrayList<>();
+    private final List<SeverityOverride> severityOverrides = new ArrayList<>();
+
+    /**
+     * Sets custom suppressions not found in the model.
+     *
+     * @param suppressions Suppressions to set.
+     * @return Returns the ModelBasedEventDecorator.
+     */
+    public ModelBasedEventDecorator suppressions(Collection<? extends Suppression> suppressions) {
```
sugmanue, line 52: This one is not use at the moment, is the idea to keep it around as a possible extension point in the future or can we remove it for now?

smithy-model/src/main/java/software/amazon/smithy/model/validation/suppressions/ModelBasedEventDecorator.java
```diff
@@ -0,0 +1,198 @@
...
+            } catch (SourceException e) {
+                events.add(ValidationEvent.fromSourceException(e));
+            }
+        });
+    }
+
+    private static void loadMetadataSuppressions(
+            Model model,
+            List<Suppression> suppressions,
+            List<ValidationEvent> events) {
```
sugmanue, line 130:
```suggestion
            List<ValidationEvent> events
    ) {
```

sugmanue (approved):

smithy-diff/src/main/java/software/amazon/smithy/diff/ModelDiff.java
mtdowling: Sure can add a comment. It doesn't matter here since the canDecorate method for IDENTITY and for ModelBasedEventDecorator always returns true.

sugmanue (approved):

PR #1916, opened 2023-08-10
Add versioning for API Gateway defaults

srchase:
This PR adds versioning to manage API Gateway defaults.

By default, API Gateway conversions will use defaults of version `2023-08-11` which will set the following OpenAPI config options:
* alphanumericOnlyRefs: true
* disabledFeatures: [default]
* disableDefaultValues: true
* disableIntegerFormat: true

All of these can be disabled by setting `apiGatewayDefaults` to `disabled` in smithy-build.json:
```
{
    "version": "1.0",
    "plugins": {
        "openapi": {
            "service": "example.weather#Weather",
            "protocol": "aws.protocols#restJson1",
            "apiGatewayDefaults": "DISABLED"
        }
    }
}
```


smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/ApiGatewayDefaults.java
```diff
@@ -0,0 +1,65 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.aws.apigateway.openapi;
+
+import java.util.Set;
+import software.amazon.smithy.utils.SetUtils;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+public enum ApiGatewayDefaults {
```
mtdowling, line 22: Enums are usually singular.

```suggestion
public enum ApiGatewayDefault {
```

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddDefaultConfigSettings.java
```diff
@@ -16,30 +16,41 @@
...
-        config.getDisableFeatures().add("default");
-        config.setDisableDefaultValues(true);
-        // If the `useIntegerType` config has been set, this assures that
-        // `int32` and `int64` formats are not set on those integer types.
-        // See https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html
-        config.setDisableIntegerFormat(true);
+    public void updateDefaultSettings(Model model, OpenApiConfig openApiConfig) {
+        ApiGatewayConfig config = openApiConfig.getExtensions(ApiGatewayConfig.class);
+        if (config.getApiGatewayDefaults() == null) {
+            LOGGER.warning(String.format("`apiGateWayDefaults` configuration not set for opeanapi plugin, using %s.",
```
mtdowling, line 45:
```suggestion
            LOGGER.warning(String.format("`apiGatewayDefaults` configuration not set for opeanapi plugin. Assuming %s.",
```

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddDefaultConfigSettings.java
```diff
@@ -16,30 +16,41 @@
...
+                    ApiGatewayDefaults.VERSION_2023_08_11));
+            config.setApiGatewayDefaults(ApiGatewayDefaults.VERSION_2023_08_11);
+        }
+        ApiGatewayDefaults defaults = config.getApiGatewayDefaults();
+
+        openApiConfig.setAlphanumericOnlyRefs(defaults.alphanumericOnlyRefs());
```
mtdowling, line 51: I think I'd rather polymorhpically let the chosen default implementations apply whatever configuration they want, rather than ask the implementations for their configuration options.

mtdowling (changes requested):

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/ApiGatewayDefault.java
```diff
@@ -0,0 +1,53 @@
...
+        this.version = version;
+    }
+
+    @Override
+    public String toString() {
+        return version;
+    }
+
+    @SmithyUnstableApi
+    public OpenApiConfig setDefaults(OpenApiConfig config) {
```
mtdowling, line 38: I think we should instead make this method abstract and implement in each enum value. I think a separate method for loading the right value from a string makes sense too (I think this is created using a NodeMapper already, so it just works with the toString implementation already too).

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/ApiGatewayDefault.java
```diff
@@ -0,0 +1,53 @@
...
+        if (version.equals("disabled")) {
+            config.setAlphanumericOnlyRefs(false);
```
mtdowling, line 40: I don't think this branch is right. "disabled" means don't do anything to the configs, it shouldn't mean overwrite anything that may have been given with false.

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddDefaultConfigSettings.java
```diff
@@ -16,30 +16,36 @@
...
-        config.getDisableFeatures().add("default");
-        config.setDisableDefaultValues(true);
-        // If the `useIntegerType` config has been set, this assures that
-        // `int32` and `int64` formats are not set on those integer types.
-        // See https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html
-        config.setDisableIntegerFormat(true);
+    public void updateDefaultSettings(Model model, OpenApiConfig openApiConfig) {
+        ApiGatewayConfig config = openApiConfig.getExtensions(ApiGatewayConfig.class);
+        if (config.getApiGatewayDefault() == null) {
+            LOGGER.warning(String.format("`apiGateWayDefault` configuration not set for opeanapi plugin. Assuming %s.",
```
mtdowling, line 45:
```suggestion
            LOGGER.warning(String.format("`apiGatewayDefault` configuration not set for opeanapi plugin. Assuming %s.",
```

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddDefaultConfigSettings.java
```diff
@@ -16,30 +16,36 @@
...
+                    ApiGatewayDefault.VERSION_2023_08_11));
```
mtdowling, line 46: nit - Maybe this could be a private static final constant of this class.

mtdowling (changes requested):

docs/source-2.0/guides/converting-to-openapi.rst
```diff
@@ -1176,6 +1176,33 @@ dependency on ``software.amazon.smithy:smithy-aws-apigateway-openapi``.
...
+            "version": "1.0",
+            "plugins": {
+                "openapi": {
+                    "service": "example.weather#Weather",
+                    "apiGatewayDefault": "2023-08-11"
+                }
+            }
+        }
+
+    If the version is set to ``DISABLED``, the configuration settings will not be set.
```
mtdowling, line 1204: Would be good to make this a list and document "disabled" just like "2023-08-11". Also, is it "DISABLED" or "disabled"? The toString method is "disabled". Might be good to also add an ".. important" admonition to call out setting this explicitly.

smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/AddDefaultConfigSettings.java
```diff
@@ -16,30 +16,37 @@
...
+ * <p>By default, this disables OpenAPI and JSON Schema features not
+ * supported by API Gateway.</p>
  *
  * <p>API Gateway does not allow characters like "_". API Gateway
  * doesn't support the "default" trait or `int32` or `int64` "format"
  * values.
  */
 final class AddDefaultConfigSettings implements ApiGatewayMapper {
+    private static final Logger LOGGER = Logger.getLogger(AddDefaultConfigSettings.class.getName());
+    private static final String DEFAULT_VERSION_MESSAGE = String.format("`apiGatewayDefault` configuration not set for"
```
mtdowling, line 35: I was more thinking of making something like `ApiGatewayDefault DEFAULT_VERSION = ApiGatewayDefault.VERSION_2023_08_11;` and then referencing that from the non-static message template and as the default value to apply when none is set.

mtdowling (changes requested):

docs/source-2.0/guides/converting-to-openapi.rst
srchase: Made this `DISABLED` consistency with the other settings. `apiGatewayType` uses REST|HTTP|DISABLED.

docs/source-2.0/guides/converting-to-openapi.rst
```diff
@@ -1176,6 +1176,40 @@ dependency on ``software.amazon.smithy:smithy-aws-apigateway-openapi``.
 Amazon API Gateway configuration settings
 =========================================
 
+apiGatewayDefault (``string``)
```
mtdowling, line 1179: I think it makes sense to pluralize this

mtdowling (approved):

PR #1917, opened 2023-08-11
Improve rename usage code example

virajh:
*Issue #, if available:*

*Description of changes:*
This change improves the example code block demonstrating the usage of rename property within the service types specification page.

In the original example, the disambiguated shape name is not actually used in the example code. It makes the example ambiguous as it is no longer clear that the newly assigned name can be used identically as the local shape names.

### DIFF
--- a/docs/source-2.0/spec/service-types.rst
+++ b/docs/source-2.0/spec/service-types.rst
@@ -221,7 +221,7 @@ the conflicting shapes.
     @output
     structure GetSomethingOutput {
         widget1: Widget
-        fooWidget: foo.example#Widget
+        widget2: FooWidget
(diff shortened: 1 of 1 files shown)


mtdowling:
I don't think the update is correct. The example refers to a hypothetical shape in a different namespace since it's not possible to define two shapes with the same name in the same namespace. The `rename` property of a service is more of a logical code generation construct; it doesn't allow you to refer to renamed shapes using a synthetic shape ID or anything like that.

virajh:
If what you're saying is true, then what is the _value_ "FooWidget" assigned to the _key_ "foo.example#Widget" used for?

PR #1924, opened 2023-08-15
Add ServiceIndex method to account for smithy.api#noAuth

gosar:
Add a new `getNoAuthAwareEffectiveAuthSchemes` method to ServiceIndex that wraps `getEffectiveAuthSchemes` to account for `smithy.api#noAuth` scheme.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java
@@ -29,8 +29,11 @@ import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.ToShapeId;
 import software.amazon.smithy.model.traits.AuthDefinitionTrait;
 import software.amazon.smithy.model.traits.AuthTrait;
+import software.amazon.smithy.model.traits.OptionalAuthTrait;
 import software.amazon.smithy.model.traits.ProtocolDefinitionTrait;
 import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.traits.synthetic.NoAuthTrait;
+import software.amazon.smithy.utils.MapUtils;
 
 /**
  * An index that resolves service protocols and auth schemes.
@@ -61,6 +64,25 @@ public final class ServiceIndex implements KnowledgeIndex {
 
     public static ServiceIndex of(Model model) {
         return model.getKnowledge(ServiceIndex.class, ServiceIndex::new);
+    }
+
+     /**
+     * Defines the type of auth schemes returned by {@link #getEffectiveAuthSchemes}.
+     */
+    public enum AuthSchemeMode {
+
+        /**
+         * Use only the modeled auth schemes. This is the default.
+         */
+        MODELED,
+
+        /**
+         * Use the modeled auth schemes, as well as the synthetic {@link NoAuthTrait} where applicable.
+         *
+         * <p>The Smithy Reference Architecture recommends using the {@code smithy.api#noAuth} auth scheme to represent
+         * no authentication which is available as the {@link NoAuthTrait}.
+         */
+        NO_AUTH_AWARE;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java
```diff
@@ -198,6 +201,42 @@ public Map<ShapeId, Trait> getEffectiveAuthSchemes(ToShapeId service, ToShapeId
...
+     *
+     * @param service Service the operation is within.
+     * @param operation Operation to get the effective authentication schemes of.
+     * @return Returns a map of the trait shape ID to the auth trait itself.
+     */
+    public Map<ShapeId, Trait> getNoAuthAwareEffectiveAuthSchemes(ToShapeId service, ToShapeId operation) {
+        LinkedHashMap<ShapeId, Trait> authSchemes = new LinkedHashMap<>(getEffectiveAuthSchemes(service, operation));
+        if (authSchemes.isEmpty() || isOptionalAuth(operation)) {
+            // TODO: should instantiate the trait once and reuse instance?
+            authSchemes.put(NoAuthTrait.ID, new NoAuthTrait());
```
gosar, line 227: Should instantiate the trait once and reuse instance?

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java
```diff
@@ -211,4 +250,24 @@ private Map<ShapeId, Trait> getAuthTraitValues(Shape service, Shape subject) {
...
         return result;
     }
+
+    // TODO: should this be a public class in software.amazon.smithy.model.traits.synthetic
+    /**
+     * An auth scheme trait for {@code smithy.api#noAuth} which indicates no authentication. This is not a real trait
+     * in the semantic model, but a valid auth scheme for use in
+     * {@link #getNoAuthAwareEffectiveAuthSchemes(ToShapeId, ToShapeId)}.
+     */
+    static final class NoAuthTrait extends AnnotationTrait {
```
gosar, line 258: Should this be a public class in software.amazon.smithy.model.traits.synthetic?

smithy-model/src/test/java/software/amazon/smithy/model/knowledge/ServiceIndexTest.java
```diff
@@ -66,88 +68,166 @@ public void returnsProtocolsOfService() {
...
+    public void authSchemesOfServiceWithInvalidOperation() {
+        ServiceIndex serviceIndex = ServiceIndex.of(model);
+        ShapeId service = ShapeId.from("smithy.example#ServiceWithoutAuthTrait");
+        ShapeId operation = ShapeId.from("smithy.example#InvalidOperation");
+
+        Map<ShapeId, Trait> auth = serviceIndex.getEffectiveAuthSchemes(service, operation);
+        assertAuthSchemes(auth);
+
+        auth = serviceIndex.getNoAuthAwareEffectiveAuthSchemes(service, operation);
+        assertAuthSchemes(auth, ServiceIndex.NoAuthTrait.ID);
```
gosar, line 225: This does not throw similar to getEffectiveAuthSchemes. Should it throw instead of returning map of size 1?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ServiceIndex.java
```diff
@@ -198,6 +200,41 @@ public Map<ShapeId, Trait> getEffectiveAuthSchemes(ToShapeId service, ToShapeId
...
+     * {@code smithy.api#optionalAuth} and avoids having to special case handling empty result. The returned map will
+     * always contain at least 1 entry.
+     *
+     * <p>The {@code smithy.api#noAuth} scheme, if present, is always the last scheme.
+     *
+     * @param service Service the operation is within.
+     * @param operation Operation to get the effective authentication schemes of.
+     * @return Returns a map of the trait shape ID to the auth trait itself.
+     */
+    public Map<ShapeId, Trait> getNoAuthAwareEffectiveAuthSchemes(ToShapeId service, ToShapeId operation) {
```
syall, line 223: I was wondering if overloading `getEffectiveAuthSchemes(ToShapeId, ToShapeId)` with a mode that would denote adding `@noAuth` could work. E.g. `getEffectiveAuthSchemes(ToShapeId, ToShapeId, AuthMode)`, where `AuthMode` could either be `DEFAULT` (which would be existing behavior) and `NO_AUTH_AWARE`.

This way, we can still use the reasonable existing method name `getEffectiveAuthSchemes` and be backwards-compatible with existing users.

Also, this would need an equivalent `getNoAuthAwareEffectiveAuthSchemes(ToShapeId service)` that only gets the auth schemes on a service shape.

syall (changes requested):

syall (approved):
Looks good. nits are only about possible name changes:

- `AuthMode` -> `AuthSchemeMode`
- `DEFAULT` -> `MODELED`

mtdowling (approved):

PR #1925, opened 2023-08-15
Allow creating specific tree types

milesziemer:
Adds method to TokenTree that allows creating a tree with a specific TreeType. Previously, there was no way to do this since the parse methods on TreeType are package private. This addition supports use cases where new trees need to be added to a root tree.

A test was added for this method, and other tests were updated to use it.

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TokenTree.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/TokenTree.java
@@ -45,6 +45,27 @@ public interface TokenTree extends FromSourceLocation {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


smithy-syntax/src/main/java/software/amazon/smithy/syntax/TokenTree.java
```diff
@@ -44,6 +44,20 @@ static TokenTree of(IdlTokenizer tokenizer) {
         return capturingTokenizer.getRoot();
     }
 
+    /**
+     * Create a TokenTree of the given {@link TreeType} from an {@link IdlTokenizer}.
+     *
```
mtdowling, line 49: Would be great to show a code example in the documentation for how this works.

mtdowling (approved):

PR #1929, opened 2023-08-16
Add validator for services with no auth trait

milesziemer:
Adds a validator that warns when there's a service shape that has multiple authDefinition traits applied, but no auth trait. This encourages the use of the auth trait so auth scheme priority is being modeled explicitly. It also allows for tooling to look for and use this validation event.

Documentation was also updated to explain that without the auth trait, auth scheme ordering is alphabetical.

A test for the auth trait (auth-trait-must-target-service-schemes) was updated to avoid getting the warning message from the new validator.

### DIFF
--- a/docs/source-2.0/spec/authentication-traits.rst
+++ b/docs/source-2.0/spec/authentication-traits.rst
@@ -276,6 +276,11 @@ the ``auth`` trait, then the operation is expected to support each of the
 service. Each entry in the ``auth`` trait is a shape ID that MUST refer to an
 authentication scheme trait applied to the service in which it is bound.
 
+.. note::
+    When a service has multiple authentication scheme traits applied and no
+    ``auth`` trait, the ordering of authentication schemes is alphabetical
+    based on the absolute shape ID of each authentication scheme trait.
+
 The following example defines all combinations in which ``auth`` can be applied
 to services and operations:
 
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ServiceAuthDefinitionsValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ServiceAuthDefinitionsValidator.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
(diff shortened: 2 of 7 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ServiceWithNoAuthTraitValidator.java
```diff
@@ -0,0 +1,54 @@
...
+                .filter(serviceShape -> !serviceShape.hasTrait(AuthTrait.ID))
+                .filter(serviceShape -> index.getAuthSchemes(serviceShape).size() > 1)
+                .collect(Collectors.toList());
+
+        for (ServiceShape service : services) {
+            events.add(warning(service, "This service uses multiple authentication schemes but is not annotated "
+                                        + "with the @auth trait. The @auth trait defines a priority ordering of "
+                                        + "auth schemes for a client to use. Without it, the ordering of auth "
+                                        + "schemes is alphabetical based on the absolute shape id of the auth "
+                                        + "schemes."));
```
syall, line 49: nit: small wording and stylistic changes

```suggestion
            events.add(warning(service, "This service uses multiple authentication schemes but does not have the "
                                        + "`@auth` trait applied. The `@auth` trait defines a priority ordering of "
                                        + "auth schemes for a client to use. Without it, the ordering of auth "
                                        + "schemes is alphabetical based on the absolute shape ID of the auth "
                                        + "schemes."));
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ServiceWithNoAuthTraitValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.traits.AuthTrait;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.ValidationEvent;
+
+/**
+ * Validates that services with multiple @authDefinition traits applied also
+ * have an @auth trait applied.
+ */
+public class ServiceWithNoAuthTraitValidator extends AbstractValidator {
```
gosar, line 32: Instead of being so specific, I think this could be ServiceAuthDefinitionsValidator. In the future maybe we'll want to warn if service has 0 auth schemes?

milesziemer: Updated. Also changed the javadoc for the class

gosar (approved):

syall (approved):

PR #1938, opened 2023-08-21
Bump action version to v6 and set to dryrun

jvschneid:
*Description of changes:*
Workflow wrongly tagged some of the issues due to running v3 version of the action which does not support multiple exemption labels. This change upgrades it to v4, which does have that support, and sets the workflow to dryrun to allow us to verify the run before actually turning the workflow on next time.

Closes #1937 

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Stale issue job
     steps:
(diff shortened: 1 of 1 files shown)


hpmellema (approved):

kstich:
#1937 says there's a V6, should this use that with related config updates and close that PR?

jvschneid:
That makes sense, let me update this to v6 (:

kstich (approved):

PR #1942, opened 2023-08-22
Fix NPE on trait with implicit null

srchase:
This PR fixes smithy-syntax throwing a NullPointerException when encountering a trait with an implicit null.

Following this fix, if a trait is encountered with an empty body, the parentheses will be omitted.

This:
```
structure Struct {
    @default()
    a: String

    @default( )
    b: String

    @default(null)
    c: String
}
```

Will now be formatted to this:
```
structure Struct {
    @default
    a: String

    @default
    b: String

    @default(null)
    c: String
}
```

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/FormatVisitor.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/FormatVisitor.java
@@ -328,24 +328,29 @@ final class FormatVisitor {
                             .detectHardLines(cursor)
                             .write();
                 } else {
-                    // Check the inner trait node for hard line breaks rather than the wrapper.
-                    TreeCursor traitNode = cursor
-                            .getFirstChild(TreeType.TRAIT_NODE)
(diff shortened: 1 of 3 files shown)


hpmellema (approved):

mtdowling:
Let's make sure this works the way we think it should

mtdowling:
```
$version: "2"
namespace smithy.example

@trait
structure foo {}

@foo
string MyString1

@foo()
string MyString2

@foo(null)
string MyString3
```

Results in:
```
──  ERROR  ────────────────────────────────────────────────────────── TraitValue
Shape: smithy.example#MyString3
File:  b.smithy:13:1

13| @foo(null)
  | ^
14| string MyString3

Error validating trait smithy.example#foo: Expected object value for structure
shape, smithy.example#foo; found null value
```

I think the only reliable approach is to omit `()` in the formatter:

```
@foo()
```

Becomes

```
@foo
```

mtdowling (changes requested):
See comment about this creating invalid models

mtdowling (approved):

PR #1943, opened 2023-08-22
Update release_prefix for migrated repo

srchase:
Update the update-brew workflow to use the updated repository URL.

### DIFF
--- a/.github/workflows/update-brew.yml
+++ b/.github/workflows/update-brew.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch: # on button click
 
 env:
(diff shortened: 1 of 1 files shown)


hpmellema (approved):

hpmellema (changes requested):
I think there may also be an issue with the "create PR" step not executing within the `homebrew-tap` directory. 
I think both of the last run commands should run using the `working-directory` set to `./homebrew-tap` per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

hpmellema (approved):

PR #1946, opened 2023-08-23
Http-Date Trait Value Validation Bug Fix

lewisjkl:
Found an issue where if the http-date timestamp format was used in a trait, then the validation would always fail (because it was checked twice, each time with a different format).

e.g. the following would fail:

```smithy
@trait
structure myTrait {
  @timestampFormat("http-date")
  httpDate: Timestamp
}

@myTrait({ httpDate: "Sun, 06 Nov 1994 08:49:37 GMT"  }) // Invalid string value, `Sun, 06 Nov 1994 08:49:37 GMT`, provided for timestamp, `smithy.api#Timestamp`. Expected an RFC 3339 formatted timestamp (e.g., "1985-04-12T23:20:50.52Z")
string Whatever
```

A workaround at the moment is to do:

```smithy
@timestampFormat("http-date")
timestamp HttpDate

@trait
structure myTrait {
  httpDate: HttpDate
}

@myTrait({ httpDate: "Sun, 06 Nov 1994 08:49:37 GMT"  }) // no error
string Whatever
```

This PR contains a test that reproduces the issue and one way of fixing it (no problem if you all want to do it differently or suggest a different way to me).

---

YES

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
@@ -15,12 +15,8 @@
 
 package software.amazon.smithy.model.validation;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
+
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.SourceLocation;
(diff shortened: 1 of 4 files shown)


mtdowling:
Thanks for the PR and proposed fix! I decided to fix this differently that doesn't need the creation of synthetic shapes: https://github.com/smithy-lang/smithy/pull/1948

lewisjkl:
> Thanks for the PR and proposed fix! I decided to fix this differently that doesn't need the creation of synthetic shapes: https://github.com/smithy-lang/smithy/pull/1948

Great, thank you for such a quick turn around!

PR #1949, opened 2023-08-24
Allow constraint trait errors on error example inputs

srchase:
This PR adds a `disableConstraints` boolean to the @examples trait, allowing for input constraint trait validations to be lowered from `ERROR` to `WARNING`.

By lowering the severity from an `ERROR`, Smithy model authors can include examples of invalid inputs and their corresponding errors.

### DIFF
--- a/docs/source-2.0/spec/documentation-traits.rst
+++ b/docs/source-2.0/spec/documentation-traits.rst
@@ -139,22 +139,17 @@ Each ``example`` trait value is a structure with the following members:
       - :ref:`examples-ErrorExample-structure`
       - Provides an error shape ID and example error parameters for the
         operation.
-    * - lowerInputValidationSeverity
-      - ``[string]``
-      - Lowers input validation events from ERROR to WARNING for specific
-        validation types. List can include: ``BLOB_LENGTH_WARNING``,
-        ``MAP_LENGTH_WARNING``, ``PATTERN_TRAIT_WARNING``,
-        ``RANGE_TRAIT_WARNING``, ``REQUIRED_TRAIT_WARNING``,
-        ``STRING_LENGTH_WARNING``.
-
+    * - disableConstraints
+      - ``boolean``
+      - Set to true to lower input constraint trait validations to warnings.
 
 When ``input`` and ``output`` members are present, both MUST be compatible
 with the shapes and constraints of the corresponding structure. When ``input``
 and ``error`` members are present, input validation events will be emitted as
-an ``ERROR`` by default. Specific validation events for the ``input`` can be
-lowered to a ``WARNING`` by setting the appropriate
-``lowerInputValidationSeverity`` value. ``input`` and ``output`` members use
-the same semantics and format as :ref:`custom trait values <trait-node-values>`.
+an ``ERROR`` by default. Constraint trait validation events for the ``input``
+can be lowered to a ``WARNING`` by setting ``disableConstraints`` to true.
+``input`` and ``output`` members use the same semantics and format as
+:ref:`custom trait values <trait-node-values>`.
 
 A value for ``output`` or ``error`` SHOULD be provided. However, both
 MUST NOT be defined for the same example.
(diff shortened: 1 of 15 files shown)


mtdowling (changes requested):
I don't think error definitions using values that don't match the types of for the error are possible to use. If examples let you give exact HTTP messages or something, then sure. But they're structured based on the model. So I think we would want to disable constraints, but not type checking.

srchase:
> I don't think error definitions using values that don't match the types of for the error are possible to use. If examples let you give exact HTTP messages or something, then sure. But they're structured based on the model. So I think we would want to disable constraints, but not type checking.

Update this PR to only allow lowering specific trait validations. Type checking validation issues are still emitted as an `ERROR`.

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
```diff
@@ -144,6 +167,13 @@ public Optional<ErrorExample> getError() {
             return Optional.ofNullable(error);
         }
 
+        /**
+         * @return Gets the list of lowered input validation severities.
+         */
+        public Optional<List<NodeValidationVisitor.Feature>> getLowerInputValidationSeverity() {
+            return Optional.ofNullable(lowerInputValidationSeverity);
```
hpmellema, line 174: I think we could remove the optional return here. The list could always be non null and we could check for the list having a zero length instead of optional exists

hpmellema (approved):

docs/source-2.0/spec/documentation-traits.rst
```diff
@@ -186,7 +192,8 @@ MUST NOT be defined for the same example.
                 content: {
                     message: "Invalid 'foo'. Special character not allowed."
                 }
-            }
+            },
+            lowerInputValidationSeverity: ["PATTERN_TRAIT_WARNING"]
```
hpmellema, line 196: This needs to be updated to to reflect most recent commit

hpmellema (changes requested):

docs/source-2.0/spec/documentation-traits.rst
```diff
@@ -139,10 +139,16 @@ Each ``example`` trait value is a structure with the following members:
       - :ref:`examples-ErrorExample-structure`
       - Provides an error shape ID and example error parameters for the
         operation.
-
-The values provided for the ``input`` and ``output`` members MUST be
-compatible with the shapes and constraints of the corresponding structure.
-These values use the same semantics and format as
+    * - disableConstraints
```
mtdowling, line 142: Shouldn't this only be allowed for errors?

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -104,10 +104,21 @@ public enum Feature {
...
-         * required. However, chaning the type of the value by un-boxing it or adjusting the range trait would
-         * be a lossy tranformation when migrating a model from 1.0 to 2.0.
+         * required. However, changing the type of the value by un-boxing it or adjusting the range trait would
+         * be a lossy transformation when migrating a model from 1.0 to 2.0.
          */
-        RANGE_TRAIT_ZERO_VALUE_WARNING
+        RANGE_TRAIT_ZERO_VALUE_WARNING,
+
+        // Lowers severity of constraint trait validations to WARNING.
+        DISABLE_CONSTRAINTS;
```
mtdowling, line 113: This name seems weird because it doesn't disable validation of constraints but lowers the severity. DISABLE_CONSTRAINTS -> IGNORE_CONSTRAINTS?

docs/source-2.0/spec/documentation-traits.rst
srchase: ExamplesTraitsValidator checks that this is only set when errors are also present. Updated the docs to make this clear.

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
srchase: Updated to `allowConstraintError` in [e5bcfd3](https://github.com/smithy-lang/smithy/pull/1949/commits/e5bcfd3a19ef1b1de341c87160f6ea5bcf41eab0)

mtdowling (approved):

PR #1956, opened 2023-08-25
Fix flaky init cache unit tests

hpmellema:
### Description of changes
Fixes issue where integration tests acting on the global init cache were occasionally failling. 
We correct this by ensuring that cache tests are not run in parallel with any other tests using the `@Isolated` flag. 

It is also necessary to remove the `max__ ` property and set parallelism via the JUnit test runner because gradle's forking of the test task does not respect JUnit test configurations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,10 +53,11 @@ subprojects {
     }
 
     dependencies {
-        testImplementation "org.junit.jupiter:junit-jupiter-api:5.4.0"
-        testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.4.0"
-        testImplementation "org.junit.jupiter:junit-jupiter-params:5.4.0"
+        testImplementation "org.junit.jupiter:junit-jupiter-api:5.10.0"
+        testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.10.0"
+        testImplementation "org.junit.jupiter:junit-jupiter-params:5.10.0"
         testImplementation "org.hamcrest:hamcrest:2.1"
+        testCompileOnly "org.apiguardian:apiguardian-api:1.1.2"
     }
 
     // Reusable license copySpec for building JARs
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -53,10 +53,11 @@ subprojects {
     }
 
     dependencies {
-        testImplementation "org.junit.jupiter:junit-jupiter-api:5.4.0"
-        testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.4.0"
-        testImplementation "org.junit.jupiter:junit-jupiter-params:5.4.0"
+        testImplementation "org.junit.jupiter:junit-jupiter-api:5.10.0"
```
hpmellema, line 56: `@Isolate` was added in 5.10.0 so we need to upgade to use it.

smithy-cli/src/it/java/software/amazon/smithy/cli/CleanCommandTest.java
```diff
@@ -11,13 +11,13 @@
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
```
hpmellema, line 14: Was unused

mtdowling (approved):

build.gradle
```diff
@@ -53,10 +53,11 @@ subprojects {
...
+        testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.10.0"
+        testImplementation "org.junit.jupiter:junit-jupiter-params:5.10.0"
         testImplementation "org.hamcrest:hamcrest:2.1"
+        testCompileOnly "org.apiguardian:apiguardian-api:1.1.2"
```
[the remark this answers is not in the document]
hpmellema: This is a new dependency required by the `5.10.0` version of junit. I dont really understand why they decided to split this out (it used to be bundled with the junit packages), but without it we get a ton of warnings.

PR #1958, opened 2023-08-28
Replace rubrics with linkable titles

### LINKED ISSUE #1381: Ability to link to deeper parts of the documentation
This is a feature request.

When referencing parts of the spec (with colleagues, or in smithy-rs), I'd often like to link to more specific parts of the documentation.

For example, I can link to [section 11.5, the `httpPayload trait`](https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httppayload-trait), but that section has a bunch of provisions that I can't refer to when implementing a test in code: for example, I can't link to the "Protocol-specific document payloads" section.

It'd be very nice if we could link granularly to these provisions.

I don't know how to change the documentation so that things like "Value type", "Conflicts with" are linkable, but it seems like if you changed these [`rubric` directives](https://github.com/awslabs/smithy/blob/8ac4ecdcc4553e6e8b39c9c041ba0fa9145aaf3b/docs/source-2.0/spec/http-bindings.rst#L758-L758) for regular subsections (`-`) or subsubsections (`^`), Sphinx would make them linkable in the HTML and appear in the TOC, if rendered with sufficient depth?

kstich:
Resolves #1381 

Includes minor reordering of some sections. A few rubrics were changed to note/important as they're informative content.

### DIFF
--- a/docs/source-2.0/spec/http-bindings.rst
+++ b/docs/source-2.0/spec/http-bindings.rst
@@ -497,7 +497,8 @@ The following example defines an error with an HTTP status code of ``404``.
     @httpError(404)
     structure MyError {}
 
-.. rubric:: Default HTTP status codes
+Default HTTP status codes
+-------------------------
 
 The ``httpError`` trait is used to set a *custom* HTTP response status code.
 By default, error structures with no ``httpError`` trait use the default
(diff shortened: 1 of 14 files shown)


docs/source-2.0/spec/http-bindings.rst
```diff
@@ -550,12 +552,12 @@ Conflicts with
   :rfc:`7231#section-7.1.1.1`. The :ref:`timestampFormat-trait` MAY be used
   to use a custom serialization format.
 
-.. rubric:: Do not put too much data in HTTP headers
+.. note::
```
hpmellema, line 555: Should this also be `::important`?

kstich: I think it's more consistent with other `note`s in the spec than other `important` admonitions.

hpmellema (approved):

PR #1961, opened 2023-08-30
Add sha digest to resolved artifacts

hpmellema:
## Description of change
Adds a sha digest for resolved artifacts. This will be used to validate artifacts against locked dependencies as part of implementing lock files for the smithy cli. 

A version is also added to the cache file to allow the CLI to invalidate and re-write cache file versions that are not compatible with the current implementation used by the CLI. 

Example of the new cache file: 
```
{"version":"1.0","artifacts":{"software.amazon.smithy:smithy-aws-traits:1.37.0":{"path":"/Users/hpm/.m2/repository/software/amazon/smithy/smithy-aws-traits/1.37.0/smithy-aws-traits-1.37.0.jar","shaSum":"723b00f4473c0f3f308b57b2c95b32fad669904c68d8e287606304d841cd52ff"}}}
```

### DIFF
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/FileCacheResolver.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/FileCacheResolver.java
@@ -116,7 +116,7 @@ public final class FileCacheResolver implements DependencyResolver {
         // If the version of the cache file does not match the current version or does not exist
         // invalidate it so we can replace it with a more recent version.
         if (!node.containsMember("version")
-                || !node.expectStringMember("version").getValue().equals(CURRENT_CACHE_FILE_VERSION)
+                || !CURRENT_CACHE_FILE_VERSION.equals(node.expectStringMember("version").getValue())
         ) {
             LOGGER.fine(() -> "Invalidating dependency cache: cache file uses old version");
             invalidate();
--- a/smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/ResolvedArtifact.java
+++ b/smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/ResolvedArtifact.java
@@ -129,7 +129,7 @@ public final class ResolvedArtifact implements ToNode {
 
     @Override
     public String toString() {
-        return "{path=" + path + ", coordinates='" + coordinates + " , shaSum=`" + shaSum + "'}";
+        return "{path=" + path + ", coordinates='" + coordinates + ", shaSum='" + shaSum + "'}";
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/FileCacheResolver.java
```diff
@@ -113,17 +113,28 @@ private List<ResolvedArtifact> load() {
...
         }
 
-        List<ResolvedArtifact> result = new ArrayList<>(node.getStringMap().size());
-        for (Map.Entry<String, Node> entry : node.getStringMap().entrySet()) {
-            Path artifactLocation = Paths.get(entry.getValue().expectStringNode().getValue());
-            long lastModifiedOfArtifact = artifactLocation.toFile().lastModified();
+        // If the version of the cache file does not match the current version or does not exist
+        // invalidate it so we can replace it with a more recent version.
+        if (!node.containsMember("version")
+                || !node.expectStringMember("version").getValue().equals(CURRENT_CACHE_FILE_VERSION)
```
sugmanue, line 119:
```suggestion
                || !CURRENT_CACHE_FILE_VERSION.equals(node.expectStringMember("version").getValue())
```

Rationale, we know that `CURRENT_CACHE_FILE_VERSION` will never be `null`, thus this way the code is less prone to null pointer exceptions.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/ResolvedArtifact.java
```diff
@@ -96,14 +113,28 @@ public String getVersion() {
...
+     * @return Get the last time the artifact was modified.
+     */
+    public long getLastModified() {
+        return path.toFile().lastModified();
+    }
+
     @Override
     public String toString() {
-        return "{path=" + path + ", coordinates='" + coordinates + "'}";
+        return "{path=" + path + ", coordinates='" + coordinates + " , shaSum=`" + shaSum + "'}";
```
sugmanue, line 132:
```suggestion
        return "{path=" + path + ", coordinates='" + coordinates + ", shaSum='" + shaSum + "'}";
```

smithy-utils/src/main/java/software/amazon/smithy/utils/IoUtils.java
```diff
@@ -152,6 +155,35 @@ public static String readUtf8Url(URL url) {
...
+                byte[] buf = new byte[8192];
+                int n;
+                do {
+                    n = din.read(buf);
+                } while (n > 0);
+            }
+            StringBuilder sb = new StringBuilder();
+            for (byte b: md.digest()) {
+                sb.append(String.format("%02x", b));
+            }
```
sugmanue, line 178:
```suggestion
            StringBuilder sb = new StringBuilder();
            for (byte b : md.digest()) {
                int decimal = (int) b & 0xff;
                String hex = Integer.toHexString(decimal);
                if (hex.length() == 1) {
                    sb.append('0');
                }
                sb.append(hex);
            }
```
Minor, performance: format is quite expensive.

sugmanue (approved):

sugmanue (approved):

PR #1962, opened 2023-09-01
Add warnings for ignored HTTP member bindings

### LINKED ISSUE #1379: Make clearer what HTTP bindings are ignored if not used in top-level operation input / output
The following Smithy model builds without any warnings:

```smithy
$version: "2.0"

namespace com.amazonaws.simple

use aws.protocols#restJson1

@restJson1
@title("SimpleService")
service SimpleService {
    operations: [
        AnOperation
    ]
}

@http(uri: "/operation/{label}" method: "GET")
operation AnOperation {
    input: AnOperationInput
    output: AnOperationOutput
}

@input
structure AnOperationInput {
    nested: Nested

    @required
    @httpLabel
    label: String
}

structure Nested {
    @required
    @httpHeader("header")
    header: String

    @required
    @httpLabel
    label: String

    @httpResponseCode
    responseCode: Integer

    @httpQuery("query")
    query: String

    @httpQueryParams
    queryParams: MapOfStrings

    @httpPrefixHeaders("X-Foo-")
    prefixHeaders: MapOfStrings
}

@output
structure AnOperationOutput {
    nested: Nested
}

map MapOfStrings {
    key: String
    value: String
}
```

**What HTTP bindings are ignored if not used as _top-level_ operation input or output?**

We know for sure that `httpLabel` is, because if I remove the `label` member from `AnOperationInput` in the model above, Smithy emits an error:

```
ERROR: com.amazonaws.simple#AnOperation (HttpLabelTrait)
     @ /local/home/davidpz/workplace/smithy-ws/src/SmithyRsSource/codegen-server-test/model/simple.smithy
     |
  16 | operation AnOperation {
     | ^
     = This operation uses `com.amazonaws.simple#AnOperationInput` as input, but the following URI labels found in the operation's `http` trait do not have a corresponding member marked with the `httpLabel` trait: `label`
```

However, what about the other HTTP binding traits? Many contain a phrase in the spec indicating that a trait "is only used on {input,output}". For example, take [`httpResponseCode`](https://awslabs.github.io/smithy/2.0/spec/http-bindings.html#httpresponsecode-trait):

> ### httpResponseCode is only used on output

> `httpResponseCode` is ignored when resolving the HTTP bindings of any structure except an operation's output structure. This means that if a structure that contains members marked with the `httpResponseCode` trait is not used as an output structure of an operation, then those members are sent as part of the [protocol-specific document](https://awslabs.github.io/smithy/2.0/spec/http-bindings.html#http-protocol-document-payloads) sent in the body of the request.

So it's ignored in an operation's input or error, that much is clear and reasonable. However, I don't think it's abundantly clear whether it should be ignored when used in a shape that is not _top-level_ operation output. If so, if "`httpResponseCode` is only used on output" could be rephrased as "`httpResponseCode` is only used on top-level operation output", then it'd be much clearer.

Same with all the other HTTP binding traits and especially with `httpHeader`. Since `httpHeader` can be used in operation inputs _and_ outputs, does it get ignored when used in nested shapes? The spec does not clarify.

I'd also like to propose that Smithy emit a warning or a notice everywhere where an HTTP binding gets ignored: I know shape reusability is a goal and that you can e.g. reuse a nested shape in another operation's top-level input, but reusing shapes that are bound to parts of the HTTP message across operations is probably a modeling concern.

### LINKED ISSUE #1386: Non-2xx HTTP response code emits a `DANGER` validation error
The [spec for the `@http` trait](https://awslabs.github.io/smithy/2.0/spec/http-bindings.html#http-trait) states that for the `code`:

> The provided value SHOULD be between 100 and 599, and it MUST be between 100 and 999.

However, when providing a non-2xx value for code, such as 418 (why not?), the Smithy validation emits a `DANGER` message:
```
[DANGER] aws.protocoltests.misc#ResponseCodeRequiredOperation: Expected an `http` code in the 2xx range, but found 418 | HttpResponseCodeSemantics
```

Which is correct? The spec or the error?

kstich:
This commit adds a new validator that emits a warning when a member has an HTTP member binding trait applied and its container is referenced through a relationship where it would be ignored.

Spec clarification of this behavior has been added.

*Issue #, if available:*
Resolves #1379 
Resolves #1386 


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpBindingTraitIgnoredValidator.java
```diff
@@ -0,0 +1,170 @@
...
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
+ * Emits warnings when a structure member has an HTTP binding trait that will be ignored
+ * in some contexts to which it is bound.
+ * * When httpLabel, httpQueryParams, or httpQuery is applied to a member of a shape that
```
hpmellema, line 40: Nit: could use `<ul> <li>Element</li></ul>` in javadoc here for formatting.

jvschneid (approved):

docs/source-2.0/spec/http-bindings.rst
```diff
@@ -695,15 +702,15 @@ Applying the ``httpLabel`` trait to members
...
+----------------------------------------------
 
-``httpLabel`` is ignored when resolving the HTTP bindings of an operation's
-output or an error. This means that if a structure that contains members
-marked with the ``httpLabel`` trait is used as the top-level output structure
-of an operation, then those members are sent as part of the
-:ref:`protocol-specific document <http-protocol-document-payloads>` sent in
-the body of the response.
+``httpLabel`` is ignored when resolving the HTTP bindings of any structure
+except an operation's top-level input structure. This means that if a structure
```
mtdowling, line 709: The "This means" sentence reads a little strange. Can we reword all of this?

> httpLabel is only considered when applied to top-level members of an operation's input structure. This trait has no meaning in any other context and is simply ignored.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-binding-trait-ignored.errors
```diff
@@ -0,0 +1,14 @@
...
+[WARNING] smithy.example#IgnoredBindingOperationOutput$string: This member has the `smithy.api#httpLabel` trait applied but its container has the following non-input relationships: [{OUTPUT=smithy.example#IgnoredBindingOperationOutput}] | HttpBindingTraitIgnored.Input
+[WARNING] smithy.example#IgnoredBindingOperationOutput$string2: This member has the `smithy.api#httpQuery` trait applied but its container has the following non-input relationships: [{OUTPUT=smithy.example#IgnoredBindingOperationOutput}] | HttpBindingTraitIgnored.Input
+[WARNING] smithy.example#IgnoredBindingOperationOutput$stringMap: This member has the `smithy.api#httpQueryParams` trait applied but its container has the following non-input relationships: [{OUTPUT=smithy.example#IgnoredBindingOperationOutput}] | HttpBindingTraitIgnored.Input
+[WARNING] smithy.example#NestedBindings$string: This member has the `smithy.api#httpLabel` trait applied but its container has the following non-input relationships: [{MEMBER_TARGET=smithy.example#NestedBindings}] | HttpBindingTraitIgnored.Input
```
mtdowling, line 8: Can you simplify this to:
1. "The `httpLabel` trait applied to this member is ignored in all contexts." (when it isn't ever effective, but does get referenced)
2. "The `httpLabel` trait applied to this member is ignored in some contexts: ... list here in a nicer way

mtdowling (changes requested):

mtdowling (approved):

PR #1968, opened 2023-09-05
Only serialize allowConstraintErrors when true

srchase:
Updates the Examples trait to only serialize the `allowConstraintErrors` when true. If not set, or false, it will not be serialized.


smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
```diff
@@ -178,8 +178,9 @@ public Node toNode() {
                     .withMember("title", Node.from(title))
                     .withOptionalMember("documentation", getDocumentation().map(Node::from))
                     .withOptionalMember("error", getError().map(ErrorExample::toNode))
-                    .withOptionalMember("allowConstraintErrors", BooleanNode.from(allowConstraintErrors)
-                            .asBooleanNode());
+                    .withOptionalMember("allowConstraintErrors", Optional.of(allowConstraintErrors)
```
mtdowling, line 181: You could just add an if statement which would be less yucky IMO.

mtdowling (changes requested):

mtdowling (approved):

PR #1974, opened 2023-09-08
Add Smithy Lock command

hpmellema:
### Description of changes
Adds a lock file implementation to the smithy CLI. 

Users can lock their dependencies using the `smithy lock` CLI command. This resolve all dependencies defined in the config and will generate a lock file named `smithy-lock.json` that pins the resolved dependencies. 

Future calls to the smithy cli will then use the pinned versions of dependencies defined in the lockfile rather than the dependency versions (and potentially ranges) specified in the config file whenever a lockfile is present. 

If the config file changes from what was used to generate the local lockfile an error is generated instructing users to re-lock their dependencies.

### Benchmarking
Without Cache
Before - 99.5ms +- 3.0ms (1sigma, 15 runs)
After (no-lockfile) - 101.9ms +- 3.13ms (1 sigma, 10runs)
After (lockfile) - 109ms +- 3.6ms (1 sigma, 10 runs)

With Cache
Before - 37ms +- 0.5ms (1sigma, 15 runs)
After (no lockfile) - 39ms +- 1.5ms (1 sigma, 10 runs)
After (lockfile) - 50.1ms +- 2.1ms (1 sigma, 10 runs)


smithy-cli/src/it/java/software/amazon/smithy/cli/LockCommandTest.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.stream.Collectors;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.mockserver.integration.ClientAndServer.startClientAndServer;
+import static software.amazon.smithy.cli.MavenResolverMultipleReposTest.mockArtifactAndSha;
+
+@Isolated
```
hpmellema, line 25: Necessary to prevent a socket binding issue on windows CI.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/BuildCommand.java
```diff
@@ -86,11 +86,6 @@ private static final class Options implements ArgumentReceiver {
         private String projection;
         private String plugin;
 
-        @Override
```
hpmellema, line 89: This override is removed identical to the default.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/DependencyUtils.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.logging.Logger;
+import software.amazon.smithy.build.model.MavenConfig;
+import software.amazon.smithy.build.model.MavenRepository;
+import software.amazon.smithy.build.model.SmithyBuildConfig;
+import software.amazon.smithy.cli.EnvironmentVariable;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.utils.IoUtils;
+
+
+public final class DependencyUtils {
```
mtdowling, line 26: Why public?

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/PinnedArtifact.java
```diff
@@ -0,0 +1,74 @@
...
+package software.amazon.smithy.cli.dependencies;
+
+import java.util.Objects;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+
+/**
+ * Pinned dependency definition used for locking dependencies.
+ */
+final class PinnedArtifact extends Artifact {
```
mtdowling, line 15: Why is this is a subclass? It just has custom loading from a node, and it writes out the sum that the parent defines. Seems like it could be contained in the lock file implementation.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/PinnedArtifact.java
```diff
@@ -0,0 +1,74 @@
...
+        return "{coordinates='" + coordinates + ", shaSum='" + shaSum + "'}";
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(coordinates, shaSum);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mtdowling, line 65: why is this any different than the parent?

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/LockFile.java
```diff
@@ -0,0 +1,202 @@
...
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.node.ToNode;
+import software.amazon.smithy.utils.BuilderRef;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+
+public final class LockFile implements ToSmithyBuilder<LockFile>, ToNode {
```
mtdowling, line 25: Does this need to be public?

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/Artifact.java
```diff
@@ -0,0 +1,78 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.cli.dependencies;
+
+import java.util.Objects;
+import software.amazon.smithy.model.node.ToNode;
+
+abstract class Artifact implements ToNode {
```
mtdowling, line 11: Why the need to break this out?

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ClasspathAction.java
```diff
@@ -133,13 +136,32 @@ private List<Path> resolveDependencies(
...
+        // Use the lock file dependencies and repositories if a lockfile exists
+        Optional<LockFile> lockFileOptional = DependencyUtils.loadLockfile();
+        if (lockFileOptional.isPresent()) {
+            if (lockFileOptional.get().getConfigHash() != DependencyUtils.configHash(dependencies, repositories)) {
+                throw new CliError(
+                        "Lockfile does not match configured dependencies. Re-lock dependencies or revert changes.");
+            }
+            LOGGER.fine(() -> "Lockfile found. Using pinned dependencies: "
+                    + lockFileOptional.get().getDependencyCoordinateSet());
+            dependencies = lockFileOptional.get().getDependencyCoordinateSet();
```
mtdowling, line 154: Don't you need to overwrite repositories too?

hpmellema: I dont think it is necessary. The repo URLs are checked via the config_hash so users cant actually change those without invalidating the lockfile. They could change the login info, but I think that is ok.

The dependency coordinates on the other hand could be dependency ranges so we need to replace those with the locked version for resolution.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/PinnedArtifact.java
hpmellema: I can change it to just be contained in the lockfile. I dont think the `Artifact` is necessary any more either.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/DependencyUtils.java
hpmellema: This is used in the ClasspathAction dependency resolution method.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/LockFile.java
hpmellema: If I moved it into the the Command folder then possibly not. I suppose it is only used in the ClasspathAction and LockCommand so that should be OK.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DependencyUtils.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Loads  lockfile if it exists.
+     *
+     * @param file file to load as a lockfile
+     * @return Lock file if it exists otherwise Optional.empty().
+     */
+    public static Optional<LockFile> loadLockfile(File file) {
+        if (file.exists() && file.isFile()) {
```
mtdowling, line 100: It seems like LOCK_FILE is more useful as a Path and needs to be converted to Path anyways. I think you could just give a Path, and store LOCK_FILE as a path. Then use `Files.isRegularFile()`.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DependencyUtils.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return combined hash
+     */
+    public static int configHash(Set<String> artifacts, Set<MavenRepository> repositories) {
+        Set<String> hashingSet = new LinkedHashSet<>(artifacts);
+        for (MavenRepository repo : repositories) {
+            hashingSet.add(repo.getUrl());
+        }
+        return hashingSet.hashCode();
+    }
+
```
mtdowling, line 87: nit - newline

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DependencyUtils.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.logging.Logger;
+import software.amazon.smithy.build.model.MavenConfig;
+import software.amazon.smithy.build.model.MavenRepository;
+import software.amazon.smithy.build.model.SmithyBuildConfig;
+import software.amazon.smithy.cli.EnvironmentVariable;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.utils.IoUtils;
+
```
mtdowling, line 24: nit-newline

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/DependencyUtils.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Computes the combined hash of the artifact set and repositories set.
+     *
+     * @param artifacts set of requested artifact coordinates
+     * @param repositories set of repositories used for dependency resolution
+     * @return combined hash
+     */
+    public static int configHash(Set<String> artifacts, Set<MavenRepository> repositories) {
+        Set<String> hashingSet = new LinkedHashSet<>(artifacts);
```
mtdowling, line 81: I don't think you actually want to sort these values before hashing. Changing the order of repositories can change the resolved dependencies. You could just create a String[] based on the size of both artifacts and repositories, then use Objects.hash on the array. This will maintain order which matters here.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ClasspathAction.java
```diff
@@ -133,13 +134,32 @@ private List<Path> resolveDependencies(
...
+
+        Set<MavenRepository> repositories = DependencyUtils.getConfiguredMavenRepos(smithyBuildConfig);
+        Set<String> dependencies = maven.getDependencies();
+
+        // Use the lock file dependencies and repositories if a lockfile exists
+        Optional<LockFile> lockFileOptional = DependencyUtils.loadLockfile();
+        if (lockFileOptional.isPresent()) {
+            if (lockFileOptional.get().getConfigHash() != DependencyUtils.configHash(dependencies, repositories)) {
+                throw new CliError(
+                        "Lockfile does not match configured dependencies. Re-lock dependencies or revert changes.");
```
mtdowling, line 148: Can you say smithy-lock.json or whatever it's called rather than LockFile? And also say what command to use to re-lock dependencies?

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "_AUTOGENERATED_FILE_DO_NOT_MODIFY_MANUALLY_":"",
```
mtdowling, line 2: minor, but I think we can shorten this. Maybe `"AUTOGENERATED_DO_NOT_MODIFY"`

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
```diff
@@ -0,0 +1,11 @@
...
+  "version":"1.0",
+  "config_hash":-1856284556,
```
mtdowling, line 4: We use camelCase everywhere else, so we should use it here too.
```suggestion
  "configHash":-1856284556,
```

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
```diff
@@ -0,0 +1,11 @@
...
+  "repositories":["https://repo.maven.apache.org/maven2"],
+  "artifacts":{
+    "software.amazon.smithy:smithy-aws-traits:1.37.0":{
+      "shaSum":"723b00f4473c0f3f308b57b2c95b32fad669904c68d8e287606304d841cd52ff"
```
mtdowling, line 8: Can you say what kind of sha it is? e.g. "sha256": "..."

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/LockCommand.java
```diff
@@ -0,0 +1,79 @@
...
+import software.amazon.smithy.build.model.MavenConfig;
+import software.amazon.smithy.build.model.MavenRepository;
+import software.amazon.smithy.build.model.SmithyBuildConfig;
+import software.amazon.smithy.cli.Arguments;
+import software.amazon.smithy.cli.Command;
+import software.amazon.smithy.cli.SmithyCli;
+import software.amazon.smithy.cli.dependencies.DependencyResolver;
+import software.amazon.smithy.cli.dependencies.FilterCliVersionResolver;
+import software.amazon.smithy.cli.dependencies.ResolvedArtifact;
+
```
mtdowling, line 21: nit - newline

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "_AUTOGENERATED_FILE_DO_NOT_MODIFY_MANUALLY_":"",
+  "version":"1.0",
+  "config_hash":-1856284556,
+  "repositories":["https://repo.maven.apache.org/maven2"],
+  "artifacts":{
```
mtdowling, line 6: Would this lockfile ever contain other dependency types? If so, we might want to put this in a maven section.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/LockFile.java
```diff
@@ -0,0 +1,246 @@
...
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.node.ToNode;
+import software.amazon.smithy.utils.BuilderRef;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+
```
mtdowling, line 26: nit - newline

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/LockFile.java
```diff
@@ -0,0 +1,246 @@
...
+final class LockFile implements ToSmithyBuilder<LockFile>, ToNode {
+    private static final String DEFAULT_VERSION = "1.0";
+    private static final String WARNING_STR = "_AUTOGENERATED_FILE_DO_NOT_MODIFY_MANUALLY_";
+    private static final String VERSION_NAME = "version";
+    private static final String CONFIG_HASH_NAME = "config_hash";
```
mtdowling, line 31:
```suggestion
    private static final String CONFIG_HASH_NAME = "configHash";
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/LockFile.java
```diff
@@ -0,0 +1,246 @@
...
+        builder.withMember(ARTIFACT_LIST_NAME, artifactNodeBuilder.build());
+
+        return builder.build();
+    }
+
+    public void validateArtifacts(List<ResolvedArtifact> resolvedArtifacts) {
+        for (ResolvedArtifact artifact : resolvedArtifacts) {
+            PinnedArtifact pin = artifacts.get(artifact.getCoordinates());
+            if (pin == null) {
+                throw new CliError("Resolved artifact " + artifact + "does not match pinned artifacts.");
```
mtdowling, line 90:
```suggestion
                throw new CliError("Resolved artifact `" + artifact + "` does not match pinned artifacts.");
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/LockFile.java
```diff
@@ -0,0 +1,246 @@
...
+            return this;
+        }
+
+        public LockFile.Builder repositories(Collection<String> repositories) {
+            this.repositories.clear();
+            this.repositories.get().addAll(repositories);
+            return this;
+        }
+
+        public LockFile.Builder repository(MavenRepository repository) {
```
mtdowling, line 183: If you aren't using these singular methods (repository() / artifact()), then consider removing.

mtdowling (changes requested):

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
hpmellema: Will do.

smithy-cli/src/test/resources/software/amazon/smithy/cli/commands/smithy-lock-test.json
hpmellema: I suspect we may want to do that someday however, my thinking was that we would include the type as a property of the artifact rather than as a section. I'd rather hold off on this for now and do a new version of the lockfile once we have users asking for additional dependency types.

mtdowling (changes requested):

mtdowling (approved):

smithy-cli/src/it/java/software/amazon/smithy/cli/LockCommandTest.java
```diff
@@ -0,0 +1,177 @@
...
+                assertThat(lockResult.getOutput(), containsString("software.amazon.smithy.cli.dependencies.DependencyResolver - Resolved Maven dependencies: [com.example:artifact:jar:1.0.0"));
+                assertThat(lockResult.getOutput(), containsString("Saving resolved artifacts to lockfile."));
+
+                // Confirm lockfile detected in path and used
+                RunResult validateResult = IntegUtils.run(root, ListUtils.of("validate", "--debug"), env);
+                assertThat(validateResult.getExitCode(), equalTo(0));
+                assertThat(validateResult.getOutput(), containsString("`smithy-lock.json` found. Using locked dependencies: "));
+            });
+        } finally {
+            if(mockServer!=null) {
```
sugmanue, line 75:
```suggestion
            if (mockServer!=null) {
```

smithy-cli/src/it/java/software/amazon/smithy/cli/LockCommandTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+            mockArtifactAndSha(
+                    mockServer,
+                    "/artifactory-2/com/example/artifact/maven-metadata.xml",
+                    getMetadataForVersions(ListUtils.of("1.2.0", "1.1.0", "1.0.0"))
+            );
+
+            IntegUtils.runWithEmptyCache("lockfile-used", ListUtils.of("validate", "--debug"),
+                    Collections.emptyMap(), result -> {
+                        System.out.println(result.getOutput());
```
sugmanue, line 114:
```suggestion
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ClasspathAction.java
```diff
@@ -133,13 +130,33 @@ private List<Path> resolveDependencies(
...
+        Optional<LockFile> lockFileOptional = LockFile.load();
+        if (lockFileOptional.isPresent()) {
+            LockFile lockFile = lockFileOptional.get();
+            if (lockFile.getConfigHash() != ConfigurationUtils.configHash(maven.getDependencies(), repositories)) {
+                throw new CliError(
+                        "`smithy-lock.json` does not match configured dependencies. "
+                                + "Re-lock dependencies using the `lock` command or revert changes.");
+            }
+            LOGGER.fine(() -> "`smithy-lock.json` found. Using locked dependencies: "
+                    + lockFileOptional.get().getDependencyCoordinateSet());
```
sugmanue, line 149:
```suggestion
                    + lockFile.getDependencyCoordinateSet());
```

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ConfigurationUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Computes the combined hash of the artifact set and repositories set.
+     *
+     * @param artifacts set of requested artifact coordinates
+     * @param repositories set of repositories used for dependency resolution
+     * @return combined hash
+     */
+    public static int configHash(Set<String> artifacts, Set<MavenRepository> repositories) {
```
sugmanue, line 68: I'm not sure about this implementation, I don't think we have any guaranties that java internal hashCodes are going to be stable. I'd feel much better if we were to use an stable hashing here instead such as SHA which will make the hashes consistent.

smithy-cli/src/main/java/software/amazon/smithy/cli/dependencies/DependencyUtils.java
```diff
@@ -15,7 +15,9 @@
 import java.security.NoSuchAlgorithmException;
 
 final class DependencyUtils {
-    private DependencyUtils() {}
+    private DependencyUtils() {
+        // Utility Class should not be instantiated
```
sugmanue, line 19: Nothing to do with this, and I know ppl ask for this, but I have never quite understood why "Utility Class should not be instantiated"

hpmellema: I just add the comment so intellij and checkstyle wont throw any linting warnings for an empty method. 🤷

sugmanue (approved):

PR #1975, opened 2023-09-12
Fixed duplicate NoninclusiveTerms events

rchache:
*Description of changes:*

Previously, the testcase

`"smithy.api#examples": [
                    {
                        "MasterUsername": "master"
                    }
                ]`

would produce the following ValidationEvents

``[WARNING] ns.foo#A: 'examples' trait value at path {0/MasterUsername} contains a non-inclusive term `Master`. Consider using one of the following terms instead: `Main`, `Parent`, `Primary` | NoninclusiveTerms.Trait.master.examples.0/MasterUsername``
``[WARNING] ns.foo#A: 'examples' trait value at path {0/MasterUsername} contains a non-inclusive term `master`. Consider using one of the following terms instead: `main`, `parent`, `primary` | NoninclusiveTerms.Trait.master.examples.0/MasterUsername``

Now, it will add `/key` to the TextIndex path when referencing a the key, to produce

``
[WARNING] ns.foo#A: 'examples' trait value at path {0/MasterUsername/key} contains a non-inclusive term `Master`. Consider using one of the following terms instead: `Main`, `Parent`, `Primary` | NoninclusiveTerms.Trait.master.examples.0/MasterUsername/key
``
``
[WARNING] ns.foo#A: 'examples' trait value at path {0/MasterUsername} contains a non-inclusive term `master`. Consider using one of the following terms instead: `main`, `parent`, `primary` | NoninclusiveTerms.Trait.master.examples.0/MasterUsername
``


kstich:
>which seems more dangerous in my opinion. 

What is dangerous about this approach? The portion of the id altered (`.toLowerCase()` call inside `.id()`) is done in this same case statement.

However, this approach may still be preferable since there's little difference in what to change here.

rchache:
I believe changing (not extending) event ids of validation events is dangerous, as it could break consumers who have tests on the expected output events.

kstich (approved):

PR #1977, opened 2023-09-13
Bump version to 1.38.0

kstich:


jvschneid (approved):

hpmellema (changes requested):
I'd like to revert the sha inclusion pr as there seems to be a better way to get the sha of artifacts directly from the resolver:
https://github.com/smithy-lang/smithy/pull/1978 and we shouldnt add the latency of sha computation to the CLI if it can be avoided.

kstich:
Updated to remove the line from #1961 and add #1975

PR #1990, opened 2023-09-26
Update rules engine authSchemes validation

kstich:
This commit fixes several issues with validating authSchemes properties in defined endpoints within an endpoints rule set. It now properly enforces the typing and uniqueness of names of schemes defined within an authSchemes property.

It also fixes several issues validating the presence and typing of properties configuring the sigv4, sigv4a, and "beta" schemes.

A specification has been added to clearly detail how clients should choose schemes, detail what properties of the above schemes will be validated, and explain how to add new validators.

-----

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -16,7 +16,7 @@ html1:
 html2:
 	@$(SPHINXBUILD) -M html "source-2.0" "build/2.0" $(SPHINXOPTS) -W --keep-going -n $(O)
 
-html: html1 html2 merge-versions
+html: html2 html1 merge-versions
 
 merge-versions:
 	mkdir -p build/html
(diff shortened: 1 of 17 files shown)


docs/Makefile
```diff
@@ -16,7 +16,7 @@ html1:
 html2:
 	@$(SPHINXBUILD) -M html "source-2.0" "build/2.0" $(SPHINXOPTS) -W --keep-going -n $(O)
 
-html: html1 html2 merge-versions
+html: html2 html1 merge-versions
```
jvschneid, line 19: Just out of curiosity: what's the goal of inverting these?

jvschneid (approved):

jvschneid (approved):

PR #1992, opened 2023-09-26
Add token to checkout action in brew workflow

jvschneid:

### DIFF
--- a/.github/workflows/update-brew.yml
+++ b/.github/workflows/update-brew.yml
@@ -63,7 +63,7 @@ jobs:
           repository: 'smithy-lang/homebrew-tap'
           path: 'homebrew-tap'
           ref: 'main'
-          token: ${{ GITHUB_TOKEN }}
+          token: ${{ secrets.PR_TOKEN }}
       - name: Update version
         run: |
           version="${GITHUB_REF##*/}"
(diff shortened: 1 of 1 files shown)


srchase (approved):

.github/workflows/update-brew.yml
```diff
@@ -63,6 +63,7 @@ jobs:
           repository: 'smithy-lang/homebrew-tap'
           path: 'homebrew-tap'
           ref: 'main'
+          token: ${{ GITHUB_TOKEN }}
```
hpmellema, line 66: Should this be `PR_TOKEN`?

jvschneid: I might be wrong, but from the `GITHUB_TOKEN: ${{ secrets.PR_TOKEN }}` line I thought these are interchengeable, but I don't mind being explicit here.

hpmellema: IIRC the env var used below only applies to the `Create PR` step rather than the entire workflow. You can either set the token for a single step via env var or the `token` property.

hpmellema (approved):

.github/workflows/update-brew.yml
jvschneid: That should be the step that needs this permission btw, but everything else looks right so might as well try this out.

haydenbaker: `PR_TOKEN` was originally used because the workflow needed elevated permissions (from a PAT) to push to its fork of the `aws/homebrew-tap`. I don't think we even need `PR_TOKEN` anymore (in the repo), but I may be missing something.

PR #2001, opened 2023-10-05
Set source locations on smithy-diff events

srchase:
This PR updates smithy-diff evaluators to set source locations on the emitted events where the source location was missing previously.

For events where something is removed from a model (member, trait definition, etc), the location from the old model is used. For events where something is added or modified within a model, the location from the new model is used.


smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/ChangedNullabilityTest.java
kstich, line 1: Several tests here are related to changing the nullability of specific members. For those, the member and structure should have different sources here to make sure it's pointing at the member shape.

kstich (changes requested):

kstich (approved):

PR #2005, opened 2023-10-09
Add smoke test traits

milesziemer:
This commit adds a new package containing the smithy model for smoke tests and their implementation. The model uses the `smithy.test` namespace, which it shares with other test trait models like protocol tests.


smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+/// response.
+@trait(selector: "operation")
+list smokeTests {
+    /// A list of smoke tests to perform on the operation
+    member: SmokeTestCase
+}
+
+/// A single smoke test case definition.
+@private
+structure SmokeTestCase {
```
JordonPhillips, line 16: I feel like we should share some test case definitions, maybe as mixins. A lot of this is exactly the same as protocol testsvand endpoint tests. It might make writing generators easier too if we have shared interfaces.

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+/// client can successfully connect to a service and get the right kind of
+/// response.
+@trait(selector: "operation")
+list smokeTests {
+    /// A list of smoke tests to perform on the operation
+    member: SmokeTestCase
+}
+
+/// A single smoke test case definition.
+@private
```
JordonPhillips, line 15: What is and isn't private isn't consistent

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/Expectation.java
```diff
@@ -0,0 +1,103 @@
...
+        return failure == null;
+    }
+
+    public boolean isFailure() {
+        return failure != null;
+    }
+
+    public Optional<FailureExpectation> getFailure() {
+        return Optional.ofNullable(failure);
+    }
```
JordonPhillips, line 57: public methods need docs

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestCaseValidator.java
```diff
@@ -0,0 +1,105 @@
...
+            for (int i = 0; i < testCases.size(); i++) {
+                SmokeTestCase testCase = testCases.get(i);
+
+                // Validate vendor params shape
+                Optional<ShapeId> vendorParamsShapeOptional = testCase.getVendorParamsShape();
+                Optional<ObjectNode> vendorParamsOptional = testCase.getVendorParams();
+                if (vendorParamsShapeOptional.isPresent()) {
+                    if (!vendorParamsOptional.isPresent()) {
+                        events.add(warning(shape, trait,
+                                "Smoke test case defined a `vendorParamsShape` but no `vendorParams`"));
```
JordonPhillips, line 61: We should put a warning for the case where they have vendor params but no shape too

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestCaseValidator.java
```diff
@@ -0,0 +1,105 @@
...
+                    } else {
+                        Shape vendorParamsShape = model.expectShape(vendorParamsShapeOptional.get());
```
JordonPhillips, line 63: You shouldn't use `expectShape` here since if the shape is missing then we lose out on validating all the other test cases. The `idRef` will already put an ERROR event in if the shape isn't present, so you don't need to duplicate that here, but you should tolerate the shape not being available.

(The http request test validator also has this unfortunate bit)

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/UniqueSmokeTestCaseIdValidator.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        Map<String, List<Shape>> testCaseIdsToOperations = new HashMap<>();
+        Set<Shape> shapes = model.getShapesWithTrait(SmokeTestsTrait.class);
+        for (Shape shape : shapes) {
+            SmokeTestsTrait trait = shape.expectTrait(SmokeTestsTrait.class);
+            for (SmokeTestCase testCase : trait.getTestCases()) {
+                testCaseIdsToOperations.computeIfAbsent(testCase.getId(), id -> new ArrayList<>()).add(shape);
+            }
+        }
```
JordonPhillips, line 43: You should be checking that no two cases within the scope of a service share ids (or within the trait if there's no service), but this will check every trait loaded.

JordonPhillips (changes requested):

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
milesziemer: Yea, there's a lot of crossover, but I wanted to just focus on smoke tests in this PR. If you think we should try to get it all in one though I can update it.

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+    /// share the same ID, including test cases defined for other operations
+    /// bound to the same service.
+    @required
+    @pattern("^[A-Za-z_][A-Za-z0-9_]+$")
+    id: String
+
+    /// Defines the input parameters used to generate the request. These
+    /// parameters MUST be compatible with the input of the operation.
+    ///
+    /// Parameter values that contain binary data MUST be defined using values
```
mtdowling, line 29: Is this consistent with protocol tests? Do we encode binary data that can't be represented this way with base64?

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+    /// Indicates that the call is expected to throw an error.
+    failure: FailureExpectation
+}
+
+@private
+structure FailureExpectation {
+    /// Indicates that the call is expected to throw a specific type of error
+    /// matching the targeted shape. If not specified, the error can be of
+    /// any type.
+    @idRef(
```
mtdowling, line 78: Is this how the smithy formatter would format this?

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+    errorId: String
+}
+
+@private
+list TagList {
+    member: Tag
+}
+
+@pattern("^[A-Za-z][A-Za-z0-9_\\-]+$")
+string Tag
```
mtdowling, line 91: missing `@private`

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
```diff
@@ -0,0 +1,91 @@
...
+    /// configuration, it should be tagged.
+    tags: TagList
+}
+
+/// The different kinds of expectations that can be made for a test case.
+@private
+union Expectation {
+    /// Indicates that the call is expeted to not throw an error. No other
+    /// assertions are made about the response.
+    success: Unit
```
mtdowling, line 67: I guess if we ever wanted to do assertions on the output, we could add a new Expectation variant

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestsTrait.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mtdowling, line 2: I think we remove the date now, and maybe use the short form of license headers

smithy-smoke-test-traits/src/test/resources/software/amazon/smithy/smoketests/traits/errorfiles/vendor-params-validation.errors
```diff
@@ -0,0 +1,4 @@
+[WARNING] smithy.example#SayHello: Smoke test case defined a `vendorParamsShape` but no `vendorParams` | SmokeTestCase
```
mtdowling, line 1: Would be great if this included the case ID

smithy-smoke-test-traits/src/test/resources/software/amazon/smithy/smoketests/traits/errorfiles/vendor-params-validation.errors
```diff
@@ -0,0 +1,4 @@
...
+[WARNING] smithy.example#SayHello: smithy.test#smokeTests.1.vendorParams: Invalid structure member `bar` found for `smithy.example#VendorParams` | SmokeTestCase
```
mtdowling, line 2: Is there any way to replace "1" here with the case ID?

mtdowling (changes requested):

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
JordonPhillips: This is consistent with protocol tests, yes: https://smithy.io/2.0/additional-specs/http-protocol-compliance-tests.html#smithy-test-httprequesttests-trait

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestCaseValidator.java
milesziemer: Good point. Although I think we need to update the protocol test validator then: https://github.com/smithy-lang/smithy/blob/8d2a2919c2612a21c747a6c616b62dde4a910a42/smithy-protocol-test-traits/src/main/java/software/amazon/smithy/protocoltests/traits/ProtocolTestCaseValidator.java#L127

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
milesziemer: Made `Tag` private. Is that all you meant?

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/Expectation.java
milesziemer: Added docs

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestCaseValidator.java
milesziemer: Done

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/UniqueSmokeTestCaseIdValidator.java
milesziemer: Fixed and updated test cases.

smithy-smoke-test-traits/src/main/resources/META-INF/smithy/smithy.test.smithy
milesziemer: Ran smithy format on the file and pushed the update.

smithy-smoke-test-traits/src/main/java/software/amazon/smithy/smoketests/traits/SmokeTestsTrait.java
milesziemer: Removed the date, I'm not sure what the short form of license headers is though. Do you have an example?

smithy-smoke-test-traits/src/test/resources/software/amazon/smithy/smoketests/traits/errorfiles/vendor-params-validation.errors
milesziemer: Good call, done

smithy-smoke-test-traits/src/test/resources/software/amazon/smithy/smoketests/traits/errorfiles/vendor-params-validation.errors
milesziemer: Yes, done. There's probably other places we can do the same

JordonPhillips (approved):

PR #2015, opened 2023-10-21
Mitigate possible NSE exception in PrivateAccessValidator.java

jyoo980:
*Issue #, if available:* n/a

*Description of changes:*

Within `PrivateAccessValidator#getPrivateAccessValiationEvent`, an unchecked call to `Optional#get` occurs. This might throw a `NoSuchElementException` if the optional is indeed empty.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PrivateAccessValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/PrivateAccessValidator.java
@@ -18,7 +18,6 @@ package software.amazon.smithy.model.validation.validators;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.Optional;
 
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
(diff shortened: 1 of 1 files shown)


mtdowling (changes requested):
Thanks for submitting this. Did you run into this issue in practice? By the time this validator is invoked, all model references have been validated. Given this, I think a more tactical fix is to change the call to .get() the neighbor to use expectNeighborShape instead. This way it throws a more useful error message and doesn’t need the extra complexity. See https://github.com/smithy-lang/smithy/blob/32ff2c35075aa0f549c31cd5e4987c3f6c2d7320/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/Relationship.java#L117

jyoo980:
Thanks for the review, @mtdowling; I didn't mean for you to look at this over the weekend, apologies.

I haven't seen the error in-practice, but it was reported by a static analysis tool after I ran a compilation locally. I'm happy to help improve the codebase if this is something you would find useful. Are there any places in the code in particular that actually _might_ be at risk of `NoSuchElementException`?

mtdowling:
Thanks for updating. Generally any place that we use `.get()` right now is probably fine and rather than convert to returning optionals or additional conditionals, we could switch to using these kinds of expect* style methods when available or using something like orElseThrow.

jyoo980:
Got it, thanks! I'll fix the checkstyle errors shortly.

mtdowling (approved):

mtdowling:
Thanks!

PR #2033, opened 2023-11-07
Add convenience method to get IntEnumShape from GenerateIntEnumDirective

hpmellema:
*Description of changes:*
Adds convenience method to get an IntEnumShape from the `GenerateIntEnumDirective`.

This provides parity with the `GenerateEnumDirective` that has an `expectEnumShape` method [see here](https://github.com/smithy-lang/smithy/blob/3ea52156715f7e30de026763021ef78447bf965d/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/GenerateEnumDirective.java#L79)


JordonPhillips (changes requested):
I hate to say it but you should add a test

JordonPhillips (approved):

PR #2037, opened 2023-11-09
Scalar RemovedShape events from ERROR -> NOTE

rchache:
*Description of changes:*
For scalar shapes (and non-enum strings), removing them is safe and therefore can be lowered from an ERROR to a NOTE

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
@@ -19,13 +19,15 @@ import java.util.List;
 import java.util.stream.Collectors;
 import software.amazon.smithy.diff.Differences;
 import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.PrivateTrait;
+import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
 
 /**
  * Creates an ERROR event when a non-private non-scalar shape is removed.
- * Creates a NOTE event when a non-private scalar shape is removed.
+ * Creates a WARNING event when a non-private scalar shape is removed.
  */
 public final class RemovedShape extends AbstractDiffEvaluator {
     @Override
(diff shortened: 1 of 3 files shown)


smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
```diff
@@ -19,19 +19,23 @@
...
  */
 public final class RemovedShape extends AbstractDiffEvaluator {
     @Override
     public List<ValidationEvent> evaluate(Differences differences) {
         return differences.removedShapes()
                 .filter(shape -> !shape.hasTrait(PrivateTrait.class))
                 .filter(shape -> !isMemberOfRemovedShape(shape, differences))
-                .map(shape -> error(shape, String.format("Removed %s `%s`", shape.getType(), shape.getId())))
+                .map(shape -> isScalarType(shape)
+                        ? note(shape, String.format("Removed %s `%s`", shape.getType(), shape.getId()))
```
kstich, line 37: Wonder if this should be a `warning` instead.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
```diff
@@ -40,4 +44,19 @@ private boolean isMemberOfRemovedShape(Shape shape, Differences differences) {
                 .filter(member -> !differences.getNewModel().getShapeIds().contains(member.getContainer()))
                 .isPresent();
     }
+
+    private boolean isScalarType(Shape shape) {
+        return shape.isIntegerShape()
```
kstich, line 49: Should exclude `intEnum` like is done for `enum` below.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
```diff
@@ -40,4 +44,19 @@ private boolean isMemberOfRemovedShape(Shape shape, Differences differences) {
...
+            || shape.isBigIntegerShape()
+            || shape.isBlobShape()
+            || shape.isBooleanShape()
+            || shape.isByteShape()
+            || shape.isDoubleShape()
+            || shape.isFloatShape()
+            || shape.isShortShape()
+            || shape.isTimestampShape()
+            || shape.isLongShape()
+            || (shape.isStringShape() && !shape.hasTrait(EnumTrait.class));
```
kstich, line 60: Should exclude the `enum` shape type as well.

kstich (changes requested):
Some typing updates in `isScalarType` as well as resolving the `warning` vs `note` conversation.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedShape.java
mtdowling: This really should be a warning. We need to bring this to people's attention but not fail the build, hence a warning.

mtdowling (approved):

kstich (approved):

PR #2046, opened 2023-11-16
Allow the ModifiedTrait diff evaluator to take account traits added to

teali:
newly added shapes: none

*Issue #, if available:*

*Description of changes:*
This allows the ModifiedTrait diff evaluator to consider traits that are added to newly added shapes. For our specific use case, we need to know about added waiters even when it's added as a part of a new operation.


milesziemer:
ModifiedTrait is more for traits that have been added, modified, or removed from existing shapes. Spoke offline and the recommended solution to provide diff events specifically for added waiters is a custom smithy diff evaluator, so closing this for now.

PR #2009, opened 2023-10-12
fix for ExampleTrait.Example equality

lewisjkl:
*Description of changes:*

Found this issue through:

```scala
val modelString =
  """|$version: "2"
     |
     |namespace foo
     |
     |@examples([
     |  {
     |    title: "Test This"
     |    input: {
     |      test: "something"
     |    }
     |  }
     |])
     |operation Foo {
     |  input := {
     |    test: String
     |  }
     |}
     |""".stripMargin

val one = Model
  .assembler()
  .addUnparsedModel("foo.smithy", modelString)
  .assemble()
  .unwrap()
val two = Model
  .assembler()
  .addUnparsedModel("foo.smithy", modelString)
  .assemble()
  .unwrap()

one == two // false
```

I've allowed edits from maintainers, feel free to change as you see fit or tell me what to change.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

YES

PR #2052, opened 2023-11-24
Add `equals` and `hashCode` to `ErrorExample`

miguel-vila:
*Issue #, if available:*

We looked into this because we were having some false negatives while comparing values of the `@examples` trait. Part of it was using an old version of smithy which didn't have an `equals` for `Example`, but this could happen if the trait also has an `ErrorExample`.

*Description of changes:*

Adds a simple implementation and a test.


kubukoz:
Can we have this reviewed and released?

Related: https://github.com/smithy-lang/smithy/pull/2009

smithy-model/src/main/java/software/amazon/smithy/model/traits/ExamplesTrait.java
```diff
@@ -337,6 +337,23 @@ public Builder content(ObjectNode content) {
...
+        @Override
+        public boolean equals(Object other) {
+            if (other == this) {
+                return true;
+            }
+            if (other == null || other.getClass() != this.getClass()) {
+                return false;
+            }
+            ErrorExample otherExample = (ErrorExample) other;
+            return shapeId.equals(otherExample.shapeId) && content.equals(otherExample.content);
```
gosar, line 355: just for extra safety: can we switch this to `Objects.equals`?

```suggestion
            return Objects.equals(shapeId, otherExample.shapeId) && Objects.equals(content, otherExample.content);
```

gosar (approved):

gosar:
> This branch cannot be rebased due to conflicts

Hi @miguel-vila, can you rebase? I can merge after that. Thanks.

PR #2056, opened 2023-11-30
Add configurable diff evaluator

milesziemer:
Adds a smithy-diff evaluator which can be configured via metadata property `diffEvaluators` within the model. Unlike the similar EmitEach/EmitNoneSelectorValidators, this diff evaluator is not loaded using SPI, and the ability to create your own custom configurable diff evaluators has not been exposed. Instead, smithy-diff just looks at the new model's metadata and loads any diff evaluators directly. We can expose creating custom configurable diff evaluators later if there is a need, but this diff evaluator should be enough for almost every use case, and we can also extend it easily with more configuration options.

This evaluator works as follows:
1. Get a subset of shapes to match based on the `appliesTo` property. Currently either added shapes or removed shapes.
2. Optionally filter this subset of shapes further with a selector configured in the `filter` property, which runs on either the new or old model depending on `appliesTo`.
3. Run the configured `selector` on either the new model or old model depending on `appliesTo`.
4. Match shapes returned by `selector` to the set of shapes from 2, and emit events based on `emitCondition`.

This functionality can be extended later by adding more options for `emitCondition` and `appliesTo`. For example, there is currently no configuration option for looking at changed shapes, but you could imaging wanting to see which shapes don't match in the old model, but do in the new model, which would be a new `appliesTo`. Another `emitCondition` could be `IfAnyDontMatch` or something.

Also adds a validator to smithy-diff for this metadata property so you'll know when the model is being build if evaluators have been configured incorrectly. This requires having a dependency on smithy-diff in the model package where you're configuring the diff evaluators, but the alternative is stick the validation in smithy-model, creating an implicit dependency on smithy-diff.

See the added test case for an example. 

**Next Steps**: 
- Should also add to our docs for this once we know it's concrete. 
- Another thing to look at would be moving some of this functionality into smithy-model, since its similar to how the current configurable validators work and we could provide more configuration options for those.

### DIFF
--- a/smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
+++ b/smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
@@ -1,0 +1,54 @@
+$version: "2.0"
+
+metadata diffEvaluators = [
+    {
+        id: "AddedInternalOperation"
+        message: "Added operation with `@internal` trait."
+        emitCondition: "ForEachMatch"
+        appliesTo: "AddedShapes"
+        severity: "NOTE"
+        selector: "operation [trait|internal]"
+    }
+    {
+        id: "AddedOnlyPrimitiveNumbersAndBools"
+        message: "All added numbers and booleans were primitives."
+        emitCondition: "IfAllMatch"
+        appliesTo: "AddedShapes"
+        severity: "NOTE"
+        filter: ":is(member :test(> number), number)"
+        selector: "[trait|default = 0]"
+    }
+    {
+        id: "AddedMemberWithoutClientOptional"
+        message: "One of the added members does not have `@clientOptional` trait."
+        emitCondition: "IfAnyMatch"
+        appliesTo: "AddedShapes"
+        severity: "WARNING"
+        selector: "member :not([trait|clientOptional])"
+    }
+    {
+        id: "RemovedRootShape"
+        message: "Removed root shape."
+        emitCondition: "ForEachMatch"
+        appliesTo: "RemovedShapes"
+        severity: "NOTE"
+        selector: "simpleType"
+    }
+]
+
+namespace smithy.example
+
+@internal
+operation InternalOperation {}
+
+structure Foo {
+    bar: Integer = 0
+    baz: String
+}
+
+structure Bar {
+    baz: Integer = 0
+}
+
+@default(0)
+integer PrimitiveInt
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/DiffEvaluatorsMetadataValidator.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/DiffEvaluatorsMetadataValidator.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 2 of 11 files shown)


smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
```diff
@@ -0,0 +1,54 @@
...
+        id: "AddedOnlyPrimitiveNumbersAndBools"
+        message: "All added numbers and booleans were primitives."
+        emitCondition: "IfAllMatch"
+        appliesTo: "AddedShapes"
+        severity: "NOTE"
+        filter: ":is(member :test(> number), number)"
+        selector: "[trait|default = 0]"
+    }
+    {
+        id: "AddedMemberWithoutClientOptional"
```
mtdowling, line 22: Can you provide an example of IfAnyMatch that is more realistic?

smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
```diff
@@ -0,0 +1,54 @@
...
+        emitCondition: "IfAnyMatch"
+        appliesTo: "AddedShapes"
+        severity: "WARNING"
+        selector: "member :not([trait|clientOptional])"
+    }
+    {
+        id: "RemovedRootShape"
+        message: "Removed root shape."
+        emitCondition: "ForEachMatch"
+        appliesTo: "RemovedShapes"
```
mtdowling, line 33: Can we get more removed examples?

smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
```diff
@@ -0,0 +1,54 @@
...
+        emitCondition: "IfAllMatch"
+        appliesTo: "AddedShapes"
+        severity: "NOTE"
+        filter: ":is(member :test(> number), number)"
+        selector: "[trait|default = 0]"
+    }
+    {
+        id: "AddedMemberWithoutClientOptional"
+        message: "One of the added members does not have `@clientOptional` trait."
+        emitCondition: "IfAnyMatch"
```
mtdowling, line 24: Is there any use case for IfNoneMatch?

smithy-diff/src/test/resources/software/amazon/smithy/diff/evaluators/configurable/configurable-b.smithy
```diff
@@ -0,0 +1,54 @@
+$version: "2.0"
+
+metadata diffEvaluators = [
```
mtdowling, line 3: Maybe this should emulate validators: https://smithy.io/2.0/spec/model-validation.html#validators. So more like:

```
{
    id: "AddedInternalOperation"
    name: "EmitEachMatch|EmitIfNoneMatch|EmitIfAnyMatch|..."
    severity: "NOTE"
    configuration: {
        appliesTo: "AddedShapes|RemovedShapes|ChangedShapes"
        selector: "operation [trait|internal]"
    }
}
```

This keeps the door open for future configurable diff evaluators, and maybe even the ability to configure built-in evaluators. This could also, if we really look into the future, potentially cover metadata diffs and trait diffs too.

mtdowling (changes requested):

milesziemer:
Going back to the drawing board with this, based on some discussion with the team this needs more design work. Closing for now.

PR #2059, opened 2023-12-04
Add JReleaser config

milesziemer:
Adds the jreleaser plugin and configuration for publishing to Maven Central. Using the plugin is configurable with a Gradle property, so we can continue to use the old release method if necessary. The `closeRepository` and `releaseRepository` props in the jreleaser config are set to false right now so we can check the staged artifacts before actually releasing, but can be changed later to automatically release.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,8 @@ plugins {
     id "com.github.spotbugs" version "4.7.1"
     id "io.codearte.nexus-staging" version "0.30.0"
     id "me.champeau.jmh" version "0.7.1"
-    id 'com.github.johnrengelman.shadow' version "7.1.2"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -24,7 +24,8 @@ plugins {
     id "com.github.spotbugs" version "4.7.1"
     id "io.codearte.nexus-staging" version "0.30.0"
```
JordonPhillips, line 25: We can get rid of this too right?

JordonPhillips (approved):

build.gradle
milesziemer: Yea, I left it in so we could fall back to it if necessary

PR #2063, opened 2023-12-06
Split input and output shapes for rest-xml tests

srchase:
This PR splits InputOutput shapes into separate request and response shapes for rest-xml protocol tests.


JordonPhillips (changes requested):
Go ahead and add the input and output traits while you're at it. Heck even just inlining them might be easier

JordonPhillips (approved):

PR #2072, opened 2023-12-12
Change line break formatting in brackets

milesziemer:
This commit updates how the formatter decides whether or not to put a line break after an opening bracket and before the closing bracket (bracket meaning `()`, `{}`, and `[]`). Previously, all bracket types followed the same logic - which included breaking when there are any array or object descendants. This lead to trait bodies being formatted like:
```
@someTrait(
    [
        {
	    key: "value"
	}
    ]
)
structure SomeStruct {}
```
With this change, it will be formatted like:
```
@someTrait([
    {
        key: "value"
    }
])
structure SomeStruct {}
```
the same also applies if the trait has a single object member:
```
@someTrait({
    foo: "bar"
    // ... more members to cause break onto separate lines
})
structure SomeStruct {}
```
Object nodes now also have a single space of padding inside the brackets when on a single line. Previously, it was like:
```
metadata foo = {bar: "baz"}
```
now:
```
metadata foo = { bar: "baz" }
```
Finally, when object nodes are elements of an array, they now always line break if not empty. Previously:
```
metadata foo = [
    {bar: "baz"}
]
```
now (with above padding changes):
```
metadata foo = [
    {
        bar: "baz"
    }
]
```
A few test cases were updated for this behavior, and a few more added to test the nuances of this formatting method.

### DIFF
--- a/smithy-syntax/src/main/java/software/amazon/smithy/syntax/BracketFormatter.java
+++ b/smithy-syntax/src/main/java/software/amazon/smithy/syntax/BracketFormatter.java
@@ -32,6 +32,7 @@ final class BracketFormatter {
     private Doc close = Formatter.RBRACE;
     private Collection<Doc> children;
     private boolean forceLineBreaks;
+    private boolean forceInline = false;
 
     static Function<TreeCursor, Stream<Doc>> extractor(
             Function<TreeCursor, Doc> visitor,
@@ -91,8 +92,8 @@ final class BracketFormatter {
         return false;
     }
 
-    BracketFormatter forceLineBreaks(boolean forceLineBreaks) {
-        this.forceLineBreaks = forceLineBreaks;
+    BracketFormatter forceLineBreaks() {
+        forceLineBreaks = true;
         return this;
     }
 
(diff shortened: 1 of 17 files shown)


smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/brackets.formatted.smithy
```diff
@@ -0,0 +1,192 @@
...
+
+// Line breaking values
+metadata f = [
+    """
+    foo
+    """
+]
+
+metadata g = [
+    "abc
```
mtdowling, line 42: note: we should in the future reformat this to maybe use a text block

smithy-syntax/src/main/java/software/amazon/smithy/syntax/BracketFormatter.java
```diff
@@ -103,15 +104,38 @@ BracketFormatter forceLineBreaksIfNotEmpty() {
         return this;
     }
 
+    // Don't force line breaks and don't indent inside brackets
+    BracketFormatter inline(boolean inline) {
```
mtdowling, line 108: Not sure I understand what this setter does. Is this really "forceInline"? If so, maybe it should set forceLineBreaks to false, and forceLineBreaks when set to true sets forceInline to false. That way you can remove the other conditionals around it, and the last one set wins.

mtdowling (approved):

PR #2073, opened 2023-12-13
Add Annotation processor base class for executing Smithy-Build plugins

hpmellema:
### Description of changes:
Adds a base class for annotation processors for Smithy build plugins. This annotation processor executes smithy build for a single plugin and adds the generated artifacts (java code or `META-INF/` metadata files) to the generated sources for the project. Annotation processors are easily supported in all java build systems and this processor will be used for the trait codegen plugin to easily add generated traits.


smithy-annotation-processor/src/main/java/software/amazon/smithy/processor/SmithyAnnotationProcessor.java
```diff
@@ -0,0 +1,221 @@
...
+                        "Only one package can have the " + getAnnotationClass() + " annotation.");
+            }
+        } else {
+            messager.printMessage(Diagnostic.Kind.NOTE,
+                    "Executing processor: " + this.getClass().getSimpleName() + "...");
+            SmithyBuildConfig config = createBuildConfig(getAnnotation(elements));
+            executeSmithyBuild(config).allArtifacts()
+                    .filter(path -> path.toString().contains(getPluginName()) && path.toString().contains("source"))
+                    .forEach(this::writeArtifact);
+        }
```
milesziemer, line 117:
```suggestion
        if (elements.size() > 1) {
            messager.printMessage(Diagnostic.Kind.ERROR,
                    "Only one package can have the " + getAnnotationClass() + " annotation.");
        } else if (elements.size() == 1) {
            messager.printMessage(Diagnostic.Kind.NOTE,
                    "Executing processor: " + this.getClass().getSimpleName() + "...");
            SmithyBuildConfig config = createBuildConfig(getAnnotation(elements));
            executeSmithyBuild(config).allArtifacts()
                    .filter(path -> path.toString().contains(getPluginName()) && path.toString().contains("source"))
                    .forEach(this::writeArtifact);
        }
```
Nit

milesziemer (approved):

smithy-annotation-processor/src/main/java/software/amazon/smithy/processor/SmithyAnnotationProcessor.java
```diff
@@ -0,0 +1,219 @@
...
+        messager = processingEnv.getMessager();
+        filer = processingEnv.getFiler();
+    }
+
+    @Override
+    public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+        Set<? extends Element> elements = roundEnv.getElementsAnnotatedWith(getAnnotationClass());
+        if (elements.size() > 1) {
+            messager.printMessage(Diagnostic.Kind.ERROR,
+                    "Only one package can have the " + getAnnotationClass() + " annotation.");
```
JordonPhillips, line 107: Why?

smithy-annotation-processor/src/main/java/software/amazon/smithy/processor/SmithyAnnotationProcessor.java
```diff
@@ -0,0 +1,219 @@
...
+     *
+     * @param annotation instance of generator annotation to use to create the build config.
+     * @return ObjectNode to use as plugin configuration node.
+     */
+    protected abstract ObjectNode createPluginNode(A annotation);
+
+    private SmithyBuildConfig createBuildConfig(A annotation) {
+        Map<String, ObjectNode> pluginMap = new HashMap<>();
+        pluginMap.put(getPluginName(), createPluginNode(annotation));
+        return SmithyBuildConfig.builder().version("1.0").plugins(pluginMap).build();
```
JordonPhillips, line 150: I think you're gonna need to inherit existing configuration. `imports`, `sources`, and `maven` in particular are important since the processor is gonna want full context.

JordonPhillips (changes requested):

smithy-annotation-processor/src/main/java/software/amazon/smithy/processor/SmithyAnnotationProcessor.java
hpmellema: I was thinking that we dont want to have two of the same plugins applied twice as would likely lead to duplicate/conflicting output files which would almost certainly fail because these annotation processors write to shared standard locations. 

For example we wouldn't want to allow: 
```
@MySmithyPlugin
@MySmithyPlugin
package com.example.smithy;
```
But the following might be OK, but would already be allowed by the check performed here:
```
@MySmithyPlugin
@MyOtherSmithyPlugin
package com.example.smithy;
```

smithy-annotation-processor/src/main/java/software/amazon/smithy/processor/SmithyAnnotationProcessor.java
hpmellema: The model files used by this annotation processor are expected to be already resolved/built and available (along with deps) on the compile classpath. I.e. the processor is not handling dep resolution or full smithy build execution.

I think this is easiest to see how it currently works is with a gradle example for the trait codegen processor:
```
description = "Custom Smithy structure trait with multiple inputs"

plugins {
    `java-library`
    id("software.amazon.smithy.gradle.smithy-jar").version("0.9.0")
}

dependencies {
    val smithyVersion: String by project

    // Trait codegen deps
    annotationProcessor("software.amazon.smithy:smithy-trait-codegen-processor-processor:$smithyVersion")
    compileOnly("software.amazon.smithy:smithy-trait-codegen-processor-annotation:$smithyVersion")

    // base library dependencies
    implementation("software.amazon.smithy:smithy-model:$smithyVersion")
}

// Add resources to the compile classpath so they are available to the annotation processors
sourceSets {
    main {
        java {
            compileClasspath += sourceSets["main"].resources.sourceDirectories
        }
    }
}

repositories {
    mavenLocal()
    mavenCentral()
}
```
Here we add the resources which are staged and packaged correctly by the jar plugin (i.e. selected projection output is added to a `resources/META-INF/smithy` directory or smithy models already exist there) to the compile classpath for discovery by the processor. The processor then picks up these files and any deps already on the classpath when executing. If we wanted to inherit the imports and sources directly from a smithy-build.json and use those within the processor, both the smithy build file and the imports and sources would need to be available on the compile classpath.

JordonPhillips: This kinda forces you to separate your traits from your actual model right? Because requiring the source model to have been built means any other plugins won't have access to these generated classes.

hpmellema: It doesnt necessarily force you to separate traits from your model (you can still use the smithy-only definitions to resolve the model), but no, it does not allow other plugins access to the generated classes. This isnt really any different than generating the traits with just the trait-codegen plugin because plugins already do not have access to the artifacts of other smithy-build plugins. The idea here is to mimic executing smithy build for a single plugin then copying/adding the generated java artifacts from that plugin to the correct locations in the java compilation sources.

JordonPhillips (approved):

PR #2075, opened 2023-12-15
Add backticks to diff messages for trait changes

rchache:
*Description of changes:*
Adds backticks (or triple backticks) around trait values within diff event messages. This better supports markdown rendering of the messages.

Also changes pretty-printing to use platform-specific line separator through "%n"

### DIFF
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ModifiedTrait.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ModifiedTrait.java
@@ -227,7 +227,7 @@ public final class ModifiedTrait extends AbstractDiffEvaluator {
                 }
 
                 String message;
-                String pretty = Node.prettyPrintJson(right.toNode());
+                String pretty = Node.prettyPrintJsonWithBackticks(right.toNode());
                 if (path.isEmpty()) {
                     message = String.format("Added trait `%s` with value %s", trait, pretty);
                 } else {
@@ -260,7 +260,7 @@ public final class ModifiedTrait extends AbstractDiffEvaluator {
                     return Collections.emptyList();
                 }
 
-                String pretty = Node.prettyPrintJson(left.toNode());
+                String pretty = Node.prettyPrintJsonWithBackticks(left.toNode());
                 String message;
                 if (path.isEmpty()) {
                     message = String.format("Removed trait `%s`. Previous trait value: %s", trait, pretty);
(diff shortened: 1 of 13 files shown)


rchache:
Tests fail since I used System.lineSeparator() instead of "\n", but the testcases use "\n". Should I change my code to use "\n" or change the testcases to check for System.lineSeparator()?

smithy-model/src/main/java/software/amazon/smithy/model/node/Node.java
```diff
@@ -155,6 +155,27 @@ public static String prettyPrintJson(Node node, String indentString) {
...
     }
 
+    /**
+     * Writes the contents of a Node to a pretty-printed JSON string,
+     * and surrounds the result in backticks.
+     *
+     * @param node Node to write.
+     * @return Returns a string suitable for Markdown rendering.
+     */
+    public static String prettyPrintJsonWithBackticks(Node node) {
```
mtdowling, line 165: I don't think this should be something on Node itself. I think this could go in ValidationUtils.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/TraitBreakingChange.java
```diff
@@ -214,15 +214,15 @@ private TraitDefinition.ChangeType isChangeBreaking(TraitDefinition.ChangeType t
...
+            String leftPretty = Node.prettyPrintJsonWithBackticks(left.toNode());
+            String rightPretty = Node.prettyPrintJsonWithBackticks(right.toNode());
 
             switch (type) {
                 case ADD:
                     if (!path.isEmpty()) {
                         return String.format("Added trait contents to `%s` at path `%s` with value %s",
                                              trait.getId(), path, rightPretty);
-                    } else if (rightPretty.equals("{}")) {
+                    } else if (Node.objectNode().equals(right.toNode())) {
```
mtdowling, line 225: I don't think you need .toNode on right here, or below on line 234

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ModifiedTrait.java
```diff
@@ -294,8 +294,8 @@ List<ValidationEvent> validate(
                     return Collections.emptyList();
                 }
 
-                String leftPretty = Node.prettyPrintJson(left.toNode());
-                String rightPretty = Node.prettyPrintJson(right.toNode());
+                String leftPretty = Node.prettyPrintJsonWithBackticks(left.toNode());
```
mtdowling, line 297: don't need .toNode on left or right. I know it was there before, but this is a good time to fix it.

mtdowling (changes requested):
For the failing tests, we usually fix the test cases to normalize line endings when we run into these issues. We want to use the right line separator when we write output, but assertions need to be normalized (e.g., find and replace "\r\n" with "\n" on the input and output when doing the assertions).

mtdowling (approved):

PR #2094, opened 2024-01-08
Add s3 invalid expires protocol test

zoewangg:
*Issue #, if available:*
N/A
*Description of changes:*
Add S3 protocol test to validate the request doesn't fail with invalid expires header

### DIFF
--- a/smithy-aws-protocol-tests/model/restXml/services/s3.smithy
+++ b/smithy-aws-protocol-tests/model/restXml/services/s3.smithy
@@ -43,7 +43,8 @@ service AmazonS3 {
     operations: [
         ListObjectsV2,
         GetBucketLocation,
-        DeleteObjectTagging
+        DeleteObjectTagging,
+        GetObject
     ],
 }
 
@@ -364,7 +365,6 @@ operation GetBucketLocation {
     output: GetBucketLocationOutput,
 }
 
-
 structure CommonPrefix {
     Prefix: Prefix,
 }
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restXml/services/s3.smithy
```diff
@@ -380,6 +380,42 @@ structure GetBucketLocationOutput {
...
+            """,
+            code: 200,
+            headers: {
+                "Expires": "foobar1234"
+            },
+            body: "",
+            protocol: restXml
+            }])
+@http(uri: "/{Bucket}/{Key+}", method: "GET")
+@s3UnwrappedXmlOutput
```
milesziemer, line 398: I don't think the actual S3 model has `s3UnwrappedXmlOutput` on `GetObject`

smithy-aws-protocol-tests/model/restXml/services/s3.smithy
```diff
@@ -515,10 +551,14 @@ string AccountId
 
 string BucketName
 
+string KeyName
+
 string Delimiter
 
 string DisplayName
 
+string Expires
```
milesziemer, line 560: Should this be a timestamp? From what I understand, clients should try deserializing as a timestamp, but not fail Expires is invalid, right?

smithy-aws-protocol-tests/model/restXml/services/s3.smithy
```diff
@@ -380,6 +380,42 @@ structure GetBucketLocationOutput {
     LocationConstraint: BucketLocationConstraint,
 }
 
+@httpResponseTests([{
+            id: "GetObjectInvalidExpires",
+            documentation: """
+                S3 clients should not fail the request with invalid expires.
+                GA SDKs should verify the value in ExpiresString param and
+                new SDKs should verifiy the value in Expires param
```
milesziemer, line 388: Reword suggestion, if accurate:
```suggestion
                S3 clients should not fail deserialization of the `Expires` header its
                value is not a valid timestamp. Instead, the `Expires` parameter of
                the response should be left empty, and the `ExpiresString` parameter
                should be populated with the raw value of the header.
```

smithy-aws-protocol-tests/model/restXml/services/s3.smithy
```diff
@@ -380,6 +380,42 @@ structure GetBucketLocationOutput {
...
+            """,
+            code: 200,
+            headers: {
+                "Expires": "foobar1234"
+            },
```
milesziemer, line 393: Just an idea, but we could add an `ExpiresString` member to `GetObjectOutput`, and update this test to have the following:
```suggestion
            },
            params: {
                ExpiresString: "foobar1234"
            },
```

milesziemer (changes requested):

smithy-aws-protocol-tests/model/restXml/services/s3.smithy
milesziemer: Nevermind, this is wrong

kstich:
Closing as we went another direction for testing this behavior.

PR #2098, opened 2024-01-13
Add "critical" validation phase to validation

mtdowling:
Runs "critical" validators before other validators so that they can be written without needing to account for things like whether traits are applied in valid places or not or if a trait's value matches its definition and constraints.

This can help to simplify validators as they can now call methods like .expectShape on a Model to throw if a shape is missing rather than having to defensively code around potentially incorrect models.

To ensure this works backward compatibly with existing errorfile based test runners, we will now detect whether to use a "legacy" validation mode that continues to validate after a critical validator emitted an event because an errorfiles mixes critical and non-critical events. This ensures backward compatibility and that any new addition of critical validators in the future will not break existing test cases.


smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -103,34 +108,43 @@ static ValidatorFactory defaultValidationFactory() {
 
     static final class Builder implements SmithyBuilder<ModelValidator> {
 
-        private final List<Validator> validators = new ArrayList<>();
-        private final List<ValidationEvent> includeEvents = new ArrayList<>();
+        private final BuilderRef<List<Validator>> validators = BuilderRef.forList();
```
mtdowling, line 111: I made minor tweaks to update this builder to use the builder ref API so that we don't have to make unnecessary copies.

smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelValidator.java
```diff
@@ -68,24 +70,27 @@ private static final class LazyValidatorFactoryHolder {
...
-            ResourceCycleValidator.class
+    private static final Map<Class<?>, Validator> CORRECTNESS_VALIDATORS = MapUtils.of(
+            TargetValidator.class, new TargetValidator(),
+            ResourceCycleValidator.class, new ResourceCycleValidator()
     );
 
     private final ValidatorFactory validatorFactory;
-    private final List<ValidationEvent> events = new ArrayList<>();
-    private final List<Validator> validators = new ArrayList<>();
-    private final List<SeverityOverride> severityOverrides = new ArrayList<>();
```
mtdowling, line 79: This wasn't actually used, so until we use it, I removed it. It was confusing me as to how it worked until I realized it's unused.

smithy-model/src/main/java/software/amazon/smithy/model/validation/ValidationUtils.java
```diff
@@ -26,21 +26,89 @@
...
+            ExclusiveStructureMemberTraitValidator.class,
+            TraitTargetValidator.class,
+            TraitValueValidator.class,
+            TraitConflictValidator.class,
+
+            // Ensure The semantic model adheres to the requirements of shape IDs and service shape uniqueness.
+            ShapeIdConflictValidator.class,
+            SingleOperationBindingValidator.class,
+            SingleResourceBindingValidator.class
+    );
```
mtdowling, line 73: This is just the set I came up with and I'm happy to add more or remove some if needed.

kstich: I think this is the right list. Here are a couple others I looked at for inclusion but don't think they should be added.

* `PrivateAccessValidator`
  * Akin to `TargetValidator`, the shape is invalid if it has a relationship to something in a private namespace. Counterpoint: this isn't as bad since `TargetValidator` would catch the thing that breaks downstream validation.
* `ShapeRecursionValidator`
  * Improperly recursive shapes being walked during validation can cause infinite recursions. Counterpoint: this is still the case without invalid recursion, as validators that need to walk the graph have to track valid shape recursion.

kstich (changes requested):

kstich (approved):

sugmanue (approved):

hpmellema (approved):

PR #2101, opened 2024-01-17
Add back vendorParamsShape if removed by transformers

milesziemer:
Adds a ModelTransformerPlugin to smithy-smoke-test-traits that will add back shapes referenced by a smoke test case vendorParamsShape that were removed by model transforms. Transforms like removeUnusedShapes will remove these shapes since they aren't connected to the model graph as @idRef's don't create edges.

To accomplish this, ModelTransformerPlugin was given an `order` used to sort the plugins. This was required to make the new plugin run after all other plugins, which may remove more shapes that need to be added back.


milesziemer:
Closing in favor of https://github.com/smithy-lang/smithy/pull/2105

PR #2105, opened 2024-01-18
Check references via idRef in UnreferencedShapes

milesziemer:
Applications of traits with `@idRef` members don't create relationships between shapes, so removing unreferenced shapes from the following:
```
service FooService {
    ...
    operations: [GetFoo]
}

operation GetFoo {
    output := {
        foo: Foo
    }
}

@trait
@idRef(failWhenMissing: true, selector: "*")
string myIdRef

@myIdRef(Referenced)
structure Foo {}

stucture Referenced {}
```
will remove `Referenced` from the model. This commit changes UnreferencedShapes so that `Referenced` is considered connected to `FooService` because it is referenced by an `@idRef` on a shape connected to the service.

Determining which shapes are connected via id ref is done by the new IdRefShapeReferences class, which uses the following process:
1. Get all trait definition shapes.
2. Find all paths from those trait definition shapes to shapes with the idRef trait. Each path found says that applications of the trait definition may have a shape reference.
3. For each path, get all applications of the trait on connected shapes and search the node value using the path to the shape that had idRef.
4. For each search result, exclude all connected shapes from being unreferenced.

Other changes:
- A few test cases that expected to receive an UnreferencedShape event were updated
- An `asShapeId` method was added to StringNode


haydenbaker (approved):

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/IdRefShapeReferences.java
```diff
@@ -0,0 +1,105 @@
...
+                    NodeQuery query = buildNodeQuery(path);
+                    addReferences(traitDef, query, connected);
+                }
+            }
+        }
+        return found;
+    }
+
+    private void addReferences(Shape traitDef, NodeQuery query, Set<Shape> connected) {
+        model.getShapesWithTrait(traitDef.getId()).stream()
```
mtdowling, line 63: you really don't need Streams here for the imperative work you're doing

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NeighborProviderIndex.java
```diff
@@ -76,6 +77,28 @@ public NeighborProvider getProviderWithTraitRelationships() {
...
 
+    /**
+     * Gets the neighbor provider that includes idRef relationships.
+     *
+     * @return Returns the provider.
+     */
+    public NeighborProvider getProviderWithIdRefRelationships() {
+        NeighborProvider result = providerWithIdRefs;
+
+        if (result == null) {
```
mtdowling, line 88: I think this is one I wouldn't cache since this isn't used in selectors. I'd either just create it on demand, or entirely remove it from this index.

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/NodeQuery.java
```diff
@@ -0,0 +1,87 @@
...
+        queries.add(node -> {
+            if (node == null || !node.isObjectNode()) {
+                return Stream.empty();
+            }
+            return node.expectObjectNode().getMember(name).map(Stream::of).orElse(Stream.empty());
+        });
+        return this;
+    }
+
+    NodeQuery anyMember() {
```
mtdowling, line 41: seems like several of these could be private static final instances of a query rather than recreated each time.

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/IdRefShapeRelationships.java
```diff
@@ -0,0 +1,107 @@
...
+                case MAP_VALUE:
+                    query.anyMember();
+                    break;
+                case LIST_MEMBER:
+                case SET_MEMBER:
+                    query.anyElement();
+                    break;
+                case UNION_MEMBER:
+                case STRUCTURE_MEMBER:
+                    MemberShape member = (MemberShape) relationship.getNeighborShape().get();
```
mtdowling, line 97: Do we need to cover enum and int enum members?

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/neighbor/IdRefShapeRelationships.java
milesziemer: I don't think so, since an enum/intEnum within a node would just be a literal string or number.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NeighborProviderIndex.java
milesziemer: Removed

mtdowling (approved):

PR #2106, opened 2024-01-19
Support internal trait when building synthetic enum trait

JordonPhillips:
When converting the normal enum trait to an enum shape, we convert the `"internal"` tag into the `@internal` trait. This updates the opposite direction conversion used to generate the synthetic enum trait to add the `"internal"` tag if the member had the `@internal` trait if it isn't already in the list of tags.

I also added a few other test cases for the member -> enum trait definition test that didn't seem to be covered, at least from the perspective of an enum shape that wasn't itself generated by converting a string shape.

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
@@ -290,9 +290,12 @@ public final class EnumShape extends StringShape {
         member.getTrait(DocumentationTrait.class).ifPresent(docTrait -> builder.documentation(docTrait.getValue()));
         member.getTrait(DeprecatedTrait.class).ifPresent(deprecatedTrait -> builder.deprecated(true));
 
-        Optional<List<String>> tags = member.getTrait(TagsTrait.class).map(StringListTrait::getValues);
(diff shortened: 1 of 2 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/shapes/EnumShape.java
```diff
@@ -286,8 +288,13 @@ static EnumDefinition enumDefinitionFromMember(MemberShape member) {
                 .orElseThrow(() -> new IllegalStateException("Enum definitions can only be made for string enums."));
         builder.value(traitValue);
         member.getTrait(DocumentationTrait.class).ifPresent(docTrait -> builder.documentation(docTrait.getValue()));
-        member.getTrait(TagsTrait.class).ifPresent(tagsTrait -> builder.tags(tagsTrait.getValues()));
         member.getTrait(DeprecatedTrait.class).ifPresent(deprecatedTrait -> builder.deprecated(true));
+
+        Optional<List<String>> tags = member.getTrait(TagsTrait.class).map(StringListTrait::getValues);
```
mtdowling, line 293: Could simplify by using an orElse(Collections.emptyList)

mtdowling (approved):

PR #2119, opened 2024-01-30
Make UnreferencedShape an opt-in linter instead

### LINKED ISSUE #2093: Allow opting out of `UnreferencedShape`?
Hi!

In our internal Smithy codebase (or should I say, specbase?), we have a bunch of entities that aren't connected to services - e.g. event or Databricks table definitions.

Currently, users have to annotate them with suppressions, which is far from an ideal experience.

Instead, it would be great if we could opt out of the default behavior (in both the LSP and our model assembler) and replace it with custom validation, such as something that'd check each shape is connected to one of these top-level entities.

For example:

```smithy
metadata disableUnreferencedShapeValidation = true
```

Alternatively, an ability to expand the list of "top-level things" from "services" to other things (e.g. using a selector):

```smithy
metadata unreferencedShapeValidation = {
  validationRootSelector: ":is(service, [trait|myEventTrait])"
}
```

Please consider this enhancement and let me know what you think!

mtdowling:
UnreferencedShape is no longer an on-by-default validator that ensures shapes must be connected to a service shape. There are plenty of cases where teams might create models that define shared shapes and don't necessarily connect those shapes to services. Emitting an UnreferencedShape validation event for these cases is harmless, but also annoying and desensitizing developers for when a shape is actually unreferenced.

This commit instead adds a new smithy-linters validator called UnreferencedShape that allows an optional "rootShapeSelector" to be provided that defines the "root" shapes that all other shape must be connected to. It defaults to "service". You might want to customize this to indicate that other shapes are your roots (e.g., maybe there is a specific trait that makes a shape a root).

Closes #2093


smithy-linters/src/main/java/software/amazon/smithy/linters/UnreferencedShapeValidator.java
```diff
@@ -0,0 +1,82 @@
...
+import software.amazon.smithy.model.node.ExpectationNotMetException;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.selector.Selector;
+import software.amazon.smithy.model.selector.SelectorSyntaxException;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidatorService;
+
+public final class UnreferencedShapeValidator extends AbstractValidator {
```
JordonPhillips, line 23: public classes need docs

JordonPhillips (changes requested):
Also moving this back to smithy-model would be nice

smithy-model/src/test/java/software/amazon/smithy/model/loader/ValidationEventDecoratorTest.java
```diff
@@ -96,11 +87,11 @@ public Optional<Validator> createValidator(String name, ObjectNode configuration
         };
     }
 
-    static class DummyHintValidationEventDecorator implements ValidationEventDecorator {
-
+    static class TestValidationEventDecorator implements ValidationEventDecorator {
         @Override
         public boolean canDecorate(ValidationEvent ev) {
-            return ev.containsId(UNREFERENCED_SHAPE_EVENT_ID) && ev.getMessage().contains("The structure ");
+            System.out.println(ev);
```
kstich, line 93:
```suggestion

```

kstich (approved):

PR #2130, opened 2024-02-07
Fix regression that allowed suppressing ERROR

### LINKED ISSUE #2128: Validations with `ERROR` severity are suppressible when documented as insuppressible
The documentation states that validations with a severity level of `ERROR` are unable to be suppressed: [docs/source-2.0/spec/model-validation.rst#severity](https://github.com/smithy-lang/smithy/blob/8977f9309c7c5a6554d52fed47d07c3298574c6b/docs/source-2.0/spec/model-validation.rst#severity). 

However in version `1.36.0` with PR #1890 the call to [`Severity.canSuppress()`](https://github.com/smithy-lang/smithy/blob/8977f9309c7c5a6554d52fed47d07c3298574c6b/smithy-model/src/main/java/software/amazon/smithy/model/validation/Severity.java#L42-L44) was removed: https://github.com/smithy-lang/smithy/pull/1890/commits/d8ee25e2673bee0fee2ddb5129eb0a98f79ff90e#diff-5088b9a4c34e5de8b7a61ddd131c7279111079cfd92d464b02f97bb794f738a5L328-L331.

It would be great if we could get this new behaviour documented and `Severity.canSuppress` fully removed if it's intentional, or if the old behaviour is to be reinstated have a validation or warning when an `ERROR` severity is attempted to be suppressed. 

I personally lean towards keeping the current behaviour as I have a use-case for suppressing `HttpUriConflict` as my team and I are writing a specification for an existing API/server and are generating OpenAPI from Smithy which allows these apparent conflicts. But if this behaviour is to be reverted we do not want to rely on the current behaviour.

mtdowling:
There was a regression in #1890 that allowed ERROR events to be suppressed. This was not intentional and we do not want to remove this restriction. ERROR events indicate that the model is in an unusable state because it violates a MUST in the spec or some other necessary requirement to make it usable (like referring only to shapes that exist).

New test cases have been added to guard against this regression in the future.

Closes #2128


sugmanue (approved):

smithy-model/src/test/resources/software/amazon/smithy/model/knowledge/member-property-index.smithy
```diff
@@ -1,12 +1,5 @@
 $version: "2.0"
 
-metadata suppressions = [
```
mtdowling, line 3: This shouldn't have been there and is one of the reasons I missed this in the first PR :(. Removing now to avoid any confusion in the future.

syall (approved):

PR #2133, opened 2024-02-08
Add csv output format to validate command

mtdowling:
You can now pass `--validation-format text|csv` to the validate command to get CSV or text output. Text continues to be the default when not specified.


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
```diff
@@ -0,0 +1,98 @@
...
+
+        void beginPrinting(CliPrinter printer) {}
+
+        abstract void print(CliPrinter printer, ValidationEventFormatter formatter, ValidationEvent event);
+
+        void endPrinting(CliPrinter printer) {}
+
+        private static String formatCsv(String value) {
+            // Replace DQUOTE with DQUOTEDQUOTE and escape newlines.
+            return value.replace("\"", "\"\"").replace("\n", "\\n");
```
sugmanue, line 60:
```suggestion
            return value.replace("\"", "\"\"").replace("\n", "\\n").replace("\r", "\\r");
```

Rationale, if someone uses `String.format` with `%n` that will will add carriage return for some OS's.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
```diff
@@ -0,0 +1,98 @@
...
+            @Override
+            void print(CliPrinter printer, ValidationEventFormatter formatter, ValidationEvent event) {
+                printer.println(formatter.format(event));
+            }
+        },
+
+        CSV {
+            @Override
+            void beginPrinting(CliPrinter printer) {
+                printer.println("severity,id,shape,file,message,hint,suppressionReason");
```
hpmellema, line 32: The `file` field also includes the line number so maybe `location` would be clearer

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
```diff
@@ -0,0 +1,98 @@
...
+import software.amazon.smithy.cli.HelpPrinter;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationEventFormatter;
+
+final class ValidationEventFormatOptions implements ArgumentReceiver {
+
+    private static final String VALIDATION_FORMAT = "--validation-format";
+
+    enum ValidationFormat {
```
sugmanue, line 21: Nit, the name can be more explicit
```suggestion
    enum OutputFormat {
```

sugmanue (approved):

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
hpmellema, line 60: Do we also need to escape carriage returns for windows systems?

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
mtdowling: Hmmm... since I don't fully agree here, I think I'll just go ahead and split this into file, line, column. Checkstyle does this in their reports too.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
```diff
@@ -0,0 +1,98 @@
...
+import software.amazon.smithy.cli.CliError;
+import software.amazon.smithy.cli.CliPrinter;
+import software.amazon.smithy.cli.HelpPrinter;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.model.validation.ValidationEventFormatter;
+
+final class ValidationEventFormatOptions implements ArgumentReceiver {
+
+    private static final String VALIDATION_FORMAT = "--validation-format";
```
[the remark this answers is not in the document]
mtdowling: That's true. I should apply this to diff.  --format sounds fine to me.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
mtdowling: I'll call it Format.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/ValidationEventFormatOptions.java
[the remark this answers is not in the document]
mtdowling: Updated

hpmellema (approved):

PR #2151, opened 2024-02-20
Add docs-publishing workflow

haydenbaker:
#### Background
- To reduce effort in releasing new versions of smithy, we can automate doc-publishing
- The `mkdocs` script is mostly 1-1 with this workflow, but this workflow will also allow us to do off-cycle docs releases without releasing javadocs (which may contain docs for new code not officially release yet)
- There are two flags that can be toggled: `javadocs` and `release`L
  - `javadoc`: whether to build javadocs as part of the workflow
  - `release`: whether to release the artifacts as part of the workflow
- One potential use case of toggling release off is to generate docs for PRs, so that we can see the rendered docs instead of a diff view

#### Testing
- Tested in personal fork of smithy
- Ran workflow by hand with each flag combo (javadoc, release, javadoc + release)
- Triggered workflow by creating a new release

---

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -22,7 +22,7 @@ merge-versions:
 	mkdir -p build/html
 	cp -R "build/1.0/html" "build/html/1.0"
 	cp -R "build/2.0/html" "build/html/2.0"
-	cp -R "root/" "build/html"
+	cp -R root/* "build/html"
 
 openhtml:
 	open "build/html/index.html"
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -22,7 +22,7 @@ merge-versions:
 	mkdir -p build/html
 	cp -R "build/1.0/html" "build/html/1.0"
 	cp -R "build/2.0/html" "build/html/2.0"
-	cp -R "root/" "build/html"
+	cp -R root/* "build/html"
```
haydenbaker, line 25: This change was needed because the `root` directory was being copied over as a directory instead of its contents.

hpmellema (approved):

Issue #2111, opened 2024-01-24
Relax `map` shape validation to allow for other key types than "string" 


Baccata:
![Screenshot 2024-01-24 at 18 18 26](https://github.com/smithy-lang/smithy/assets/4439335/84dc4603-afe7-4157-b6df-827a74d72f6b)

Currently smithy maps can only take string shapes as keys. Considering the protocol-agnostic nature of smithy, this is somewhat arbitrary and prevents its usage in some legitimate usecases. 

Intuitively, I guess this limitation is driven by the fact that it makes sense in the context of AWS protocols and tools. In smithy itself, having arbitrary shapes as map keys may not play well with the current canonical "node" representation of maps (representation used by traits in particular, but also other things). 

I personally think the limitation solvable : 

1. protocol specific validators could be created to enforce the same limitation in the context of AWS protocols 
2. the canonical object-node representation of maps could be used for a lot more keys than just strings (numerics in particular) 
3. for more complex shapes, a canonical representation of `[{"key" : ???, "value: ???}]` could be used. 
4. Alternatively to 3, the constraint on map keys could be relaxed to incorporate strings/numerics/booleans/timestamps/bytes/enums/intEnums but exclude structures/unions/documents/blobs.

mtdowling:
I can expand on some of the reasons behind this limitation.

Maps right now support string and enum keys. I wouldn't characterize this limitation as arbitrary, but rather a deliberate simplification in the model that in turn simplifies code generators (e.g., no need to worry about making every generated type hashable) and protocols (like you mentioned).

There hasn't been a lot of demand or use cases over the years (inside or outside of Amazon) to expand maps to support other key types. I'd expect that if such a problem occurred, the model could adopt a list of structs approach to avoid any protocol or programming language pitfalls. If a generator really needed to make this convert to a map in a target programming language, I suppose a generator could introduce their own custom traits to facilitate this as well.

Supporting all types as map keys (e.g., structures, unions, etc) isn't something we'd do since that brings in new requirements to code generators that every generated type can be used in their programming language's map type (this is a protocol agnostic concern and unrelated to what protocols AWS supports).

Types that aren't easily comparable or hashable like floats, doubles, documents (since it's sort of arbitrary), and probably timestamps won't work across languages for map keys either (e.g., Rust doesn't let you use floats as a map key).

So that leaves potentially adding support for non-float numbers (byte, short, integer, long, bigInteger, bigDecimal), intEnum, blob, and boolean.

[...]

PR #2156, opened 2024-02-21
Add trait that can constrain shape closures

mtdowling:
This commit adds a new meta-trait called `constrainShapes` that is used to constrain the closure of shapes that can be used when a trait is applied to a shape.

This can be used on a protocol definition trait to disallow the use of certain kinds of shapes that are incompatible with the protocol (e.g., some AWS protocols that use XML don't support document types). This trait is a replacement for the `noInlineDocument` property of the `protocolDefinition` trait, but generalizes the concept to work with any kind of shape or trait.

While this could have been acheived with metadata based validators before, introducing a trait makes it easier and more performant to constrain shapes. For example:

```
metadata validators = [
    {
        name: "EmitEachSelector"
        id: "MyProtocol.QueryBoolean"
        configuration: {
            selector: "service [trait|smithy.example#xProto] $service(*) ~> member :test(> document)"
            messageTemplate: "xProto found on @{var|service} does not support document types"
        }
    }
]
```

The advantage of a trait is that you get good validation messages for free, more performant scans of models when multiple constraints are needed (no need to crawl shapes multiple times), and the validation is reusable since it can be applied and shared via mixins.

Partially addresses #2111

#### Background
* What do these changes do? See message.
* Why are they important? It helps open the door to things like protocol changes that aren't compatible with all protocols. It also helps to reduce the need to write code and we can better leverage meta-modeling validation.

#### Testing
* How did you test these changes? Unit tests and errorfiles, checking jacoco output.

#### Links
* Issue #, if applicable (see [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for a list of keywords to use for linking issues) #2111 

---


haydenbaker (approved):

Baccata:
This is great !

kubukoz:
This also allows actually emitting errors, right? `EmitEachSelector` didn't allow ERROR severity.

mtdowling:
Yes it allows ERROR

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -218,6 +218,7 @@ structure protocolDefinition {
     traits: TraitShapeIdList
 
     /// Set to true if inline documents are not supported by this protocol.
+    @deprecated(message: "Use the `@constrainShapes` trait instead")
```
kubukoz, line 221: oops 😅

PR #2162, opened 2024-02-26
Update cloudformation guide to use updated plugins and cli

hpmellema:
#### Background
* Update the `Generating CloudFormation Resource Schemas from Smithy`: 
1.  Update Gradle examples to use new plugin version
2. Add kotlin and groovy versions of gradle examples.
3. Add CLI examples

#### Testing
* [Generated docs](https://github.com/smithy-lang/smithy/actions/runs/8053548585/artifacts/1276161112)

---

### DIFF
--- a/docs/source-2.0/guides/model-translations/generating-cloudformation-resources.rst
+++ b/docs/source-2.0/guides/model-translations/generating-cloudformation-resources.rst
@@ -44,48 +44,105 @@ Generating Schemas with smithy-build
 ------------------------------------
 
 The ``cloudformation`` plugin contained in the ``software.amazon.smithy:smithy-aws-cloudformation``
-package can be used with smithy-build and the `Smithy Gradle plugin`_ to
+package can be used with either the `Smithy Gradle plugin`_ or :ref:`Smithy CLI <smithy-cli>` to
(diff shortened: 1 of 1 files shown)


docs/source-2.0/guides/model-translations/generating-cloudformation-resources.rst
```diff
@@ -44,48 +44,105 @@ Generating Schemas with smithy-build
...
+    <artifacts-from-smithy-models>`:
+
+    .. tab:: Kotlin
+
+        .. code-block:: kotlin
+            :caption: build.gradle.kts
+
+            plugins {
+                java
+                id("software.amazon.smithy.gradle.smithy-jar").version("0.10.0")
```
haydenbaker, line 94: It might be nice to have a substitution for the gradle plugin versions that we can change when needed, instead of having to do a find-and-replace-all. But, if it makes the doc-build longer, maybe not worth.

haydenbaker (approved):

docs/source-2.0/guides/model-translations/generating-cloudformation-resources.rst
hpmellema: I will add this as a separate PR.

PR #2171, opened 2024-03-06
Enable building on JDK 21

JordonPhillips:
When setting up a new laptop I noticed that Smithy was failing to build with JDK 21. This PR does a few things to address the issue:

* Adds JDK21 to the CI matrices (and switches to corretto)
* Updates Gradle to 8.6
* Updates spotbugs to 6.0.8 (with some additional filters and fixes)
* Updates other build dependencies

---


config/spotbugs/filter.xml
```diff
@@ -194,4 +194,15 @@
         <Bug pattern="DE_MIGHT_IGNORE"/>
     </Match>
 
+    <!-- This checks for mutability, but it's full of false positives. Among other things, it doesn't
+         recognize SetUtils.copyOf and similar as providing immutable collections. -->
+    <Match>
+        <Bug code="EI,EI2"/>
+    </Match>
+
+    <!-- This class is effectively finalized when the exception is thrown. -->
```
kstich, line 203: This reads like it should be scoped to a class but isn't.

.github/workflows/checkout-and-build-smithy/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
     - uses: actions/setup-java@v2
       with:
         java-version: 17
```
kstich, line 12: Should this be 21 now?

kstich (changes requested):

.github/workflows/checkout-and-build-smithy/action.yml
JordonPhillips: I don't think it really matters, but I can update it

kstich (approved):

PR #2180, opened 2024-03-12
Protocol test for `: null` values in unions

rcoh:
#### Background
Servers are allowed (but not recommended) to send `"key": null` for unset fields in unions. However, this behavior was not covered by protocol tests.

#### Testing
These tests were dry-run on the Rust SDK.

#### Links
* https://github.com/awslabs/aws-sdk-rust/issues/1095

---

### DIFF
--- a/smithy-aws-protocol-tests/model/awsJson1_0/unions.smithy
+++ b/smithy-aws-protocol-tests/model/awsJson1_0/unions.smithy
@@ -556,8 +556,7 @@ apply JsonUnions @httpResponseTests([
                   "mapValue": null,
                   "structureValue": {
                       "hi": "hello"
-                  },
-                  "unitValue": null
+                  }
                 }
             }"""
         bodyMediaType: "application/json"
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/awsJson1_0/unions.smithy
```diff
@@ -43,6 +43,7 @@ union MyUnion {
     listValue: StringList,
     mapValue: StringMap,
     structureValue: GreetingStruct,
+    unitValue: Unit,
```
kstich, line 46: Can the `Unit` target entry be separated out to its own test case?

rcoh: removed—I thought it was just an oversight that it was missing. Turns out this actually causes compilation failure in the Rust server codegen: https://github.com/smithy-lang/smithy-rs/issues/2546

So we should definitely get a test with this at some point but it will probably break a few folks

kstich (approved):

PR #2210, opened 2024-03-22
Create smithy-protocol-tests for rpcv2Cbor tests

kstich:
#### Background
* What do these changes do? 
Move the RPCv2 CBOR tests to a new package.
* Why are they important?
The protocol is not AWS specific, so these tests belong in their own package for clean dependency separation.

#### Testing
Build + check jar files.

---


smithy-aws-protocol-tests/model/shared-types.smithy
```diff
@@ -121,10 +121,6 @@ list LongSet {
     member: Long,
 }
 
-list TimestampList {
```
kstich, line 124: This was a duplicate definition that was also cleaned up in the `smithy-protocol-tests` copy.

smithy-protocol-tests/build.gradle
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+plugins {
+    id "software.amazon.smithy.gradle.smithy-jar" version "$smithyGradleVersion"
+}
+
+description = "Defines protocol tests for Smithy HTTP protocols."
```
hpmellema, line 10: RPCv2?

smithy-protocol-tests/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+
+ext {
+    displayName = "Smithy :: Protocol Tests"
```
hpmellema, line 13: Should this be updated to be specific to cbor? Same with the modulename.

smithy-protocol-tests/build.gradle
kstich: Package is not specific to RPCv2, even if it only contains that for now.

smithy-protocol-tests/build.gradle
kstich: These aren't specific to RPCv2 CBOR.

hpmellema (approved):

PR #2220, opened 2024-03-28
Change the HTTP bindings validation to support any non-conflicting URI

sugmanue:
#### Background

Changes the Smithy validation of the HTTP bindings to allow URI patterns that before were considered in conflict. In particular the following constraint was removed

> A label and a literal SHOULD NOT both occupy the same segment in patterns which are equivalent to that point if they share a host.

And the docs changed as

> A label and a literal MAY both occupy the same segment in patterns that are equivalent to that point if they share a host. Server implementations MUST route ambiguous requests to the operation with the most specific URI pattern

This feature will allow customers the use of URI path patterns that are not ambiguous but that are not currently supported, such as:

* `/abc/{xyz}` and
* `/abc/{xyz}/cde`

and also those that can be ambiguous such as

* `/abc/bcd/{xyz}`
* `/abc/{xyz}/cde`

Servers are expected to use Specificity Routing to route request that might be ambiguous at runtime as explained in the design doc ["URI Pattern Validation and Conflict Resolution"](https://github.com/smithy-lang/smithy/blob/main/designs/uri-conflict-resolution.md).

#### Testing
* Unit tests 

#### Links

---


docs/source-2.0/spec/http-bindings.rst
```diff
@@ -435,41 +444,138 @@ following rules are in place:
...
+position. Similarly a segment with a non-greedy label is considered
+more specific than a segment with a greedy label segment in the same
+position.”
+
+The following algorithm is used to compare two paths
+
+Given two ambiguous URI patterns ``A`` and ``B`` with segments ``[A0,
+…, An]`` and ``[B0, …, Bm]`` with query string literals ``[AQ0, …,
+AQp]`` and ``[BQ0, …, BQq]`` (with both ``p`` and ``q`` possibly zero,
+i.e., without query string literals), the following steps are taken to
```
syall, line 485:
```suggestion
i.e. without query string literals), the following steps are taken to
```

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-greedy-lables-validator1.smithy
```diff
@@ -0,0 +1,28 @@
+$version: "2"
+namespace smithy.example
+
+service SmithyExample {
+    operations: [
+        OperationOne
+    ]
+}
+
+@http(code: 200, method: "GET", uri: "/example/{greedyOne+}/bar/{greedyTwo+}")
```
syall, line 10: Should `greedyOne+` emit a "greedy label should be the last label" event?

syall (changes requested):
Only had a question about the greedy label validator.

Also for the greedy label test files, "labels" is spelled as "lables" in the file names.

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-greedy-lables-validator1.smithy
sugmanue: It doesn't as the logic is written, 

```java
if (segments.get(j).isGreedyLabel()) {
    events.add(danger(shape, trait,
                      "At most one greedy label segment may exist in a pattern: " + pattern));
} else if (segments.get(j).isLabel()) {
    events.add(danger(shape, trait,
                      "A greedy label must be the last label in its pattern: " + pattern));
}
```
It could if we convert the `else if` into an `if` by itself, but that will emit two events for this case, I feel like one would be better as the second is *implied* by the first.

docs/source-2.0/spec/http-bindings.rst
sugmanue: > In modern American English, a comma should follow both e.g. and i.e. And because they have
both become so commonplace, it is unnecessary to place the abbreviations in italics, even though
they are abbreviated Latin phrases

See this [doc](https://warnell.uga.edu/sites/default/files/inline-files/ieVeg.pdf) and other references in the web.

sugmanue:
> Also for the greedy label test files, "labels" is spelled as "lables" in the file names.

🤦 good catch, I will fix it in the next revision.

syall (approved):

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/http-uri-greedy-lables-validator1.smithy
syall: IMO I think it’ll help to over-emit events since these are events that could be suppressed.

The best case is to force users to explicitly suppress events per extra greedy label and per greedy label that’s not last, but here it’s “blanketed” into one event.

syall (approved):

sugmanue:
> Hi folks, have there been any additions of tests that would validate this behavior? I'm looking for some samples beyond what's in the docs, for smithy4s.

Hi Jakub, no, we don't have any in the form of protocol tests, but we should. Feel free to open an issue, and I will ask for it to be prioritized.

PR #2222, opened 2024-04-02
Use `@deprecated` info when building OpenAPI description

miguel-vila:
#### Background
The objective is to add `@deprecated` to the description of the definition in OpenAPI.

Let me know if this makes sense or if the way the description is built makes sense.

#### Testing

Added a couple of tests.

### DIFF
--- a/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaShapeVisitor.java
+++ b/smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaShapeVisitor.java
@@ -333,11 +333,10 @@ final class JsonSchemaShapeVisitor extends ShapeVisitor.Default<Schema> {
     }
 
     private Optional<String> descriptionMessage(Shape shape) {
-        String deprecatedMessageSuffix = shape.getTrait(DeprecatedTrait.class)
-                .map(deprecatedTrait -> deprecatedTrait.getDeprecatedDescription(shape.getType()))
-                .map(message -> "\n" + message).orElse("");
-        return shape.getTrait(DocumentationTrait.class).map(DocumentationTrait::getValue)
-                .map(documentation -> documentation + deprecatedMessageSuffix);
+        StringBuilder builder = new StringBuilder();
+        shape.getTrait(DocumentationTrait.class).ifPresent(trait -> builder.append(trait.getValue());
+        shape.getTrait(DeprecatedTrait.class).ifPresent(trait -> builder.append("\n").append(trait.getValue());
+        return builder.toString().trim();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/traits/DeprecatedTrait.java
```diff
@@ -72,6 +74,17 @@ public Optional<String> getMessage() {
...
 
+    public String getDeprecatedDescription(ShapeType shapeType) {
+        String shapeTypeString = shapeType == ShapeType.OPERATION ? "operation" : "shape";
+        if (Objects.isNull(this.since)) {
+            return this.message != null ? this.message : "This " + shapeTypeString + " is deprecated.";
+        } else if (Objects.isNull(this.message)) {
+            return "This " + shapeTypeString + " is deprecated since " + this.since;
+        } else {
+            return "This " + shapeTypeString + " is deprecated since " + this.since + ": " + this.message;
+        }
```
kstich, line 85: This would be less duplicate using a `StringBuilder`:

```suggestion
        StringBuilder builder = new StringBuilder("This ");
        builder.append(shapeType == ShapeType.OPERATION ? "operation" : "shape").append(" is deprecated");
        if (since != null) {
            builder.append(" since ").append(since);
        }
        if (message != null) {
            builder.append(": ").append(message);
        } else {
            builder.append(".");
        }
        return builder.toString();
```

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/JsonSchemaShapeVisitor.java
```diff
@@ -332,6 +332,14 @@ private Schema.Builder updateBuilder(Shape shape, Schema.Builder builder) {
         return builder;
     }
 
+    private Optional<String> descriptionMessage(Shape shape) {
+        String deprecatedMessageSuffix = shape.getTrait(DeprecatedTrait.class)
+                .map(deprecatedTrait -> deprecatedTrait.getDeprecatedDescription(shape.getType()))
+                .map(message -> "\n" + message).orElse("");
+        return shape.getTrait(DocumentationTrait.class).map(DocumentationTrait::getValue)
+                .map(documentation -> documentation + deprecatedMessageSuffix);
```
kstich, line 340:
```suggestion
        StringBuilder builder = new StringBuilder();
        shape.getTrait(DocumentationTrait.class).ifPresent(trait -> builder.append(trait.getValue());
        shape.getTrait(DeprecatedTrait.class).ifPresent(trait -> builder.append("\n").append(trait.getValue());
        return builder.toString().trim();
```

Also simpler with a `StringBuilder`.

kstich (changes requested):

kstich (approved):

PR #2248, opened 2024-04-18
Add typed properties for TypedPropertiesBag

hpmellema:
#### Background
Adds a strongly typed Key to the typed properties bag used by Symbols in codegen. These changes are intended to improve the experience of using Symbol properties in code generators. Usage of these new typed properties would look like: 

```java
Property<String> myStringProperty = Property.named("string property"); 

Symbol symbol = Symbol.builder()
                .name("foo")
                .putProperty(myStringProperty, "foo")
                .build()
                
String result = symbol.expectProperty(myStringProperty);
```

#### Testing
Added new unit tests

---

### DIFF
--- a/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TypedPropertiesBag.java
+++ b/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TypedPropertiesBag.java
@@ -23,9 +23,11 @@ import software.amazon.smithy.utils.BuilderRef;
 class TypedPropertiesBag {
 
     private final Map<String, Object> properties;
+    private final Map<Property<?>, Object> typedProperties;
 
     TypedPropertiesBag(Builder<?> bagBuilder) {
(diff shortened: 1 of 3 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/TypedPropertiesBag.java
```diff
@@ -37,6 +39,15 @@ public Map<String, Object> getProperties() {
         return properties;
     }
 
+    /**
+     * Gets the additional typed properties of the object.
+     *
+     * @return Returns a map of additional typed properties.
+     */
+    public Map<Property<?>, Object> getTypedProperties() {
```
mtdowling, line 47: Rather than return the whole map, I suggest just returning the keys.

mtdowling (changes requested):

mtdowling (approved):

PR #2245, opened 2024-04-15
Add SigV4 and SigV4A migration diff validation

syall:
## Overview

~~***Dependent on #2250, which is the first commit of this PR***~~

These changes add SigV4 (`aws.auth#sigv4`, `sigv4`) and SigV4A (`aws.auth#sigv4a`, `sigv4a`) migration diff validation.

**This is important since not all SigV4 credentials work with SigV4A, so migration from SigV4 to SigV4A is not backward compatible.**

At the time of writing, it is uncertain whether migrating from SigV4A to SigV4 is backward compatible.

Simplified, the SigV4 migration validation events are emitted when:

- `aws.auth#sigv4` is replaced by `aws.auth#sigv4a`, or vice versa
- the order of `aws.auth#sigv4` and `aws.auth#sigv4a` is changed in the effective auth schemes
- `aws.auth#sigv4a` is added before an existing `aws.auth#sigv4`, or vice versa

## Validation Tables

Keywords:

- `sigv4` = `aws.auth#sigv4`
- `sigv4a` = `aws.auth#sigv4a`
- `+` = set addition
- `,` = list addition (order matters)
- `...` = set, e.g. `any`, `no sigv4`, `with sigv4`

For both `@auth` and `@smithy.rules#endpointRuleSet` auth schemes, these are the following validations:

- `SigV4Migration` checks for service and operation `@auth` diffs.
- `EndpointSigV4Migration` checks for service `@auth` and `@smithy.rules#endpointRuleSet` diffs.
  - At the time of writing, `sigv4` also includes `sigv4-` sub-schemes except `sigv4-s3express`.
  - At the time of writing, `sigv4a` also includes `sigv4-s3express`.
  - `beta` auth schemes are not considered during SigV4 migration, and will have to manually be added.

Old Model Auth | New Model Auth | Expectation
--- | --- | ---
[`no sigv4` + `no sigv4a`] | [`any`] | Skips validation since no migration
[`with sigv4` + `no sigv4a`] | [`no sigv4a`] | Skips validation since no migration
[`with sigv4a` + `no sigv4`] | [`no sigv4`] | Skips validation since no migration
[`with sigv4`, `with sigv4a`] | [`with sigv4`, `with sigv4a`] | Skips validation since no migration
[`with sigv4` + `no sigv4a`] | [`with sigv4a` + `no sigv4`] | Danger: `sigv4a` replaced `sigv4`, but not all `sigv4` credentials are compatible with `sigv4a`
[...]

PR #2252, opened 2024-04-20
Add `@auth` trait breaking change rules

syall:
#### Background
* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
+++ b/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
@@ -191,7 +191,42 @@ map externalDocumentation {
 }
 
 /// Defines the ordered list of supported authentication schemes.
-@trait(selector: ":is(service, operation)")
+@trait(
+    selector: ":is(service, operation)"
+    breakingChanges: [
+        {
+            change: "add"
+            severity: "DANGER"
+            message: """
+                Adding the `@auth` trait will explicitly define the order of the \
+                authentication scheme traits for the target shape. \
+                If the target shape is a service, all contained operations will \
+                inherit the same `@auth` trait unless a contained operation has \
+                its own `@auth` trait applied."""
+        }
+        {
+            change: "remove"
+            severity: "DANGER"
+            message: """
+                Removing the `@auth` trait will default to using the alphabetical \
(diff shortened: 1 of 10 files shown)


smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -191,7 +191,42 @@ map externalDocumentation {
...
-@trait(selector: ":is(service, operation)")
+@trait(
+    selector: ":is(service, operation)"
+    breakingChanges: [
+        {
+            change: "add"
+            severity: "DANGER"
+            message: """
+                Adding the `@auth` trait will explicitly define the order of the \
+                authentication scheme traits for the target shape. \
```
syall, line 202: > From this message it is unclear why explicitly defining the order might be a danger. Probably talk about the default like in the remove case below.
> These messages seem to be geared towards the `service` shape. Seems like they'd need to be slightly different for the `operation` shape. You only have tests for the service shape cases. Maybe having operation shape cases would allow us to see if the error messages are helpful enough for the case.
> I also wonder if we should instead do this with the getEffectiveAuthSchemes approach to eliminate noise where there is no effective change. Like you have in the Migration classes, but generically, with sigv4 cases being special cases?
> I agree, the validation for `@auth` should be more comprehensive with `ServiceIndex::getEffectiveAuthSchemes` for both service and operation levels.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -191,7 +191,42 @@ map externalDocumentation {
...
+                If the target shape is a service, all contained operations will \
+                inherit the same `@auth` trait unless a contained operation has \
+                its own `@auth` trait applied."""
+        }
+        {
+            change: "remove"
+            severity: "DANGER"
+            message: """
+                Removing the `@auth` trait will default to using the alphabetical \
```
syall, line 211: Previous comment thread: https://github.com/smithy-lang/smithy/pull/2245#discussion_r1567891793

> If removed from operation, will default to using the service level `@auth` if present. I think the emphasis again is on the risk that the effective auth schemes may change.

PR #2263, opened 2024-04-29
Fix @externalDocumentation rendering

msosnicki:
#### Background

Currently the `key` and `value` from the `@externalDocumentation` trait is taken in the wrong order, leading to rendering of the following:

```
<a href="External docs">https://external.docs/structureTrait</a>
```

* What do these changes do? 
* Why are they important?

#### Testing
* How did you test these changes? Added unit test

#### Links

---

### DIFF
--- a/smithy-trait-codegen/src/test/java/software/amazon/smithy/traitcodegen/TraitCodegenPluginTest.java
+++ b/smithy-trait-codegen/src/test/java/software/amazon/smithy/traitcodegen/TraitCodegenPluginTest.java
@@ -14,8 +14,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
 import software.amazon.smithy.build.MockManifest;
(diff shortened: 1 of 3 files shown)


smithy-trait-codegen/src/test/java/software/amazon/smithy/traitcodegen/TraitCodegenPluginTest.java
```diff
@@ -163,4 +93,43 @@ public void doesNotFormatContentInsideHtmlTags() {
...
+
+        assertFalse(manifest.getFiles().isEmpty());
+        assertEquals(EXPECTED_NUMBER_OF_FILES, manifest.getFiles().size());
+        Optional<String> fileStringOptional = manifest.getFileString(
+                Paths.get("com/example/traits/structures/StructureTrait.java").toString());
+        assertTrue(fileStringOptional.isPresent());
+        String expectedStructureExternalDocs = "<a href=\"https://external.docs/structureTrait\">External docs</a>";
+        assertTrue(fileStringOptional.get().contains(expectedStructureExternalDocs));
+//        String expectedFieldExternalDocs = "<a href=\"https://external.docs/structureTrait#fieldA\">External docs</a>";
+//        assertTrue(fileStringOptional.get().contains(expectedFieldExternalDocs));
```
msosnicki, line 109: Made these just for demonstration purposes, can revert as this is not related to this PR. Uncommenting these lines will make the test fail, even though there is 

```
    @externalDocumentation(
        "External docs": "https://external.docs/structureTrait#fieldA"
    )
    fieldA: String
```.

To make it work, it has to contain a documentation section:
```

```
    /// some other docs
    @externalDocumentation(
        "External docs": "https://external.docs/structureTrait#fieldA"
    )
    fieldA: String
```

I believe this is how the interceptors are implemented, the `ExternalDocsInterceptor` has  
```
    @Override
    public Class<JavaDocSection> sectionType() {
        return JavaDocSection.class;
    }
```
but this condition does not hold in the failing case.

I can raise an issue if you also feel that this is a problem

PR #2264, opened 2024-04-29
Add OperationContextParams + validator

alextwoods:
#### Background
* Adds new OperationContextParams trait to support binding of array parameters to nested operation inputs and an associated validator.

#### Testing
* New validator tests (valid/invalid)

---


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/OperationContextParamsTraitValidator.java
```diff
@@ -0,0 +1,455 @@
...
+            return LiteralExpression.ANY;
+        }
+
+        @Override
+        public Object stringShape(StringShape shape) {
+            // Create a random string that does not exceed or go under the length trait.
+            int chars = computeLength(shape);
+
+            // Fill a string with "a"'s up to chars.
+            return new String(new char[chars]).replace("\0", "a");
```
sugmanue, line 215: This expression is equivalent and more idiomatic
```suggestion
            return "a".repeat(chars);
```

smithy-rules-engine/src/test/resources/software/amazon/smithy/rulesengine/language/errorfiles/valid/operation-context-params.smithy
```diff
@@ -0,0 +1,42 @@
+$version: "1.0"
```
sugmanue, line 1:
```suggestion
$version: "2.0"
```

smithy-rules-engine/src/test/resources/software/amazon/smithy/rulesengine/language/errorfiles/invalid/invalid-operation-context-params.smithy
```diff
@@ -0,0 +1,35 @@
+$version: "1.0"
```
sugmanue, line 1:
```suggestion
$version: "2.0"
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/OperationContextParamsTraitValidator.java
```diff
@@ -0,0 +1,455 @@
...
+                    .map(OperationContextParamsTrait::getParameters)
+                    .orElse(Collections.emptyMap());
+            for (Map.Entry<String, OperationContextParamDefinition> entry : definitionMap.entrySet()) {
+
+                try {
+                    JmespathExpression path = JmespathExpression.parse(entry.getValue().getPath());
+                    StructureShape input = OperationIndex.of(model).expectInputShape(operationShape);
+                    LinterResult linterResult = path.lint(createCurrentNodeFromShape(input, model));
+
+                    if (!linterResult.getProblems().isEmpty()) {
```
sugmanue, line 92: Can we add some tests for the linter result?

E.g.,

* Using an array as object, `someArray.key`
* Using an object as array, `someObject[*]`
* Field not present in the structure `someObject.fieldThatDoesNotExist`
* Array/object operations on simpleTypes
* A test for recursive shapes

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/OperationContextParamsTraitValidator.java
```diff
@@ -0,0 +1,455 @@
...
+        // shapes rather than a shared set. This allows a shape to be used
+        // multiple times in the closure of a single shape without causing the
+        // reuse of the shape to always be assumed to be a recursive type.
+        private Object withCopiedVisitors(Supplier<Object> supplier) {
+            // Account for recursive shapes at the current
+            Set<MemberShape> visitedCopy = new HashSet<>(visited);
+            Object result = supplier.get();
+            visited = visitedCopy;
+            return result;
+        }
```
sugmanue, line 243: This logic feels incorrect, in particular, it assumes a single level and there might be a chain of things (more than one) that are recursive which will break that assumption, I don't have a concrete example right now but I'd like to know if you think otherwise and why.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/OperationContextParamsTraitValidator.java
alextwoods: I believe `repeat` is only supported in java 11+ :-(   Not sure theres a nicer way in Java 8 that I could find.

sugmanue: You're right, I missed that bit. I'm not sure if it's better but we can use `Arrays.fill`, up to you, e.g.,

```
        @Override
        public Object stringShape(StringShape shape) {
            // Create a random string that does not exceed or go under the length trait.
            int chars = computeLength(shape);
            // Fill a string with "a"'s up to chars.
            char[] tmp = new char[chars];
            Arrays.fill(tmp, 'a');
            return new String(tmp);
        }
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/OperationContextParamsTraitValidator.java
alextwoods: Added more linter tests

sugmanue (approved):

PR #2277, opened 2024-05-07
Fix formatter to correctly convert invalid doc comments

haydenbaker:
#### Background
- Fixes an issue where a doc-comment above a trait-statement and member may get converted to a line-comment
- Updates the expected formatting of test case file (`.formatted.smithy`)
- Updates tests to use `assertEquals`, since it provides diff functionality in IDEs
- Caveat:
  - A trailing doc-comment won't get converted to a line-comment
    - TODO was added to handle this case, if it is possible 

#### Testing
- Ran the tests, no failures

---

### DIFF
--- a/smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/fixes-invalid-doc-comments.formatted.smithy
+++ b/smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/fixes-invalid-doc-comments.formatted.smithy
@@ -23,19 +23,19 @@ use smithy.api#String
 
 /// 12 (keep)
 @deprecated
-/// 13 (change)
+// 13 (change)
 structure Foo {
     /// 14 (keep)
     @length(
         // 15 (change)
         min: 1
     )
-    /// 16 (change)
+    // 16 (change)
     @since("1.x")
-    /// 17 (TODO: change)
+    // 17 (change)
     bar: String
 
-    // 18 (change)
+    /// 18 (TODO: handle this case somehow)
 }
 
 // 19 (change)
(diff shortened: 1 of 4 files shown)


smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/fixes-invalid-doc-comments.formatted.smithy
```diff
@@ -23,19 +23,19 @@ use smithy.api#String
...
     )
-    /// 16 (change)
+    // 16 (change)
     @since("1.x")
-    /// 17 (TODO: change)
+    // 17 (change)
     bar: String
 
-    // 18 (change)
+    /// 18 (TODO: handle this case somehow)
```
kstich, line 38: This looks like a regression.

haydenbaker: Unfortunately, it is. In my limited knowledge and time, I couldn't figure a way of accomplishing both. The compromise is retaining the right comments, but not converting trailing doc comments attached to nothing.

milesziemer:
I pushed a commit that fixes the regression, which is captured by what I added to the test case [here](https://github.com/smithy-lang/smithy/pull/2277/files#diff-5f0819eaa722d6e5033442a41baf6e785999ef798507bc0b2308207c814f970eR53-R57). The TODO [here](https://github.com/smithy-lang/smithy/pull/2277/files#diff-5f0819eaa722d6e5033442a41baf6e785999ef798507bc0b2308207c814f970eR38-R42) is a separate issue that occurs when the comment is after a default value assignment. We will need to follow up to address that.

smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/fixes-invalid-doc-comments.formatted.smithy
milesziemer: See my comment here: https://github.com/smithy-lang/smithy/pull/2277#issuecomment-2101091519 - this specific regression should be fixed now

milesziemer:
Followup work: https://github.com/smithy-lang/smithy/issues/2279

kstich (approved):

PR #2298, opened 2024-05-23
Replace outdated RFC references

JordonPhillips:
The HTTP RFCs have received some facelifts recently. The vast majority of the HTTP spec is now collected in [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110) with version-specific information broken out into their own RFCs (HTTP 1.1 is in [RFC 9112](https://datatracker.ietf.org/doc/html/rfc9112) for example).

This PR updates any references we have to RFCs obsoleted by RFC 9110 and RFC 9112 to those newer RFCs. The semantic content of these sections has not changed, only the presentation.

---


nateprewitt (approved):
:shipit:

PR #2313, opened 2024-06-07
Add missing node mapper for document types

### LINKED ISSUE #2312: Trait codegen doesn't like Document in collections
Tested against 1.49.0.

Example 1:

```smithy
@trait()
list documents {
    member: Document
}
```

```java
public static DocumentsTrait fromNode(Node node) {
    Builder builder = builder();
    node.expectArrayNode()
        .getElements().stream()
        .map(n -> ) //error here
        .forEach(builder::addValues);
    return builder.build();
}
```

Example 2:

```smithy
@trait()
map documents {
    key: String
    value: Document
}
```

```java
public static DocumentsTrait fromNode(Node node) {
    Builder builder = builder();
    node.expectObjectNode().getMembers().forEach((k, v) -> {
        builder.putValues(k.expectStringNode().getValue(), ); //error here
    });
    return builder.build();
}
```

hpmellema:
fixes #2312 

#### Background
Adds missing node mapper for documents in collection traits.

#### Testing
Added new integration tests for collections traits with document members 

---

### DIFF
--- a/smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
+++ b/smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
@@ -4,14 +4,14 @@ namespace test.smithy.traitcodegen
 
 use test.smithy.traitcodegen.lists#DocumentListTrait
 
-
 @DocumentListTrait([
-    { a : "a"}
     {
-        b : "b"
-        c : 1
+        a: "a"
+    }
+    {
+        b: "b"
+        c: 1
     }
     "string"
 ])
-structure myStruct {
-}
+structure myStruct {}
--- a/smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
+++ b/smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
@@ -1,0 +1,16 @@
+$version: "2.0"
+
+namespace test.smithy.traitcodegen
+
+use test.smithy.traitcodegen.lists#DocumentListTrait
+
+
+@DocumentListTrait([
+    { a : "a"},
+    {
+        b : "b"
+        c : "c"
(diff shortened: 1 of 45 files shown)


smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/FromNodeMapperVisitor.java
```diff
@@ -112,6 +112,7 @@ public Void floatShape(FloatShape shape) {
 
     @Override
     public Void documentShape(DocumentShape shape) {
+        writer.write("$L.expectObjectNode()", varName);
```
mtdowling, line 115: A document can also be a string or any other value. Would this still work for those cases?

smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
```diff
@@ -0,0 +1,16 @@
+$version: "2.0"
+
+namespace test.smithy.traitcodegen
+
+use test.smithy.traitcodegen.lists#DocumentListTrait
+
+
```
mtdowling, line 7: It's probably a good idea to format these Smithy files. I think we did a bulk formatting update recently.

smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/FromNodeMapperVisitor.java
hpmellema: Good point. The document needs to be a `Node` not `ObjectNode`. Then this can just be changed to transparently pass the node here.

smithy-trait-codegen/src/it/resources/software/amazon/smithy/traitcodegen/test/lists/document-list-trait.smithy
hpmellema: Ran on all the trait codegen models.

mtdowling (approved):

PR #2316, opened 2024-06-11
Add IDL serializer option to coerce inline IO

kstich:
This commit adds a new IDL serializer configuration, coerceInlineIo, that will use inline IO to write input and output shapes that are only bound to one operation - even if they don't have the input or output trait applied.

It also fixes a bug where an empty newline was written before empty inline IO shape definitions.

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
@@ -34,6 +34,7 @@ import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.OperationIndex;
 import software.amazon.smithy.model.loader.Prelude;
 import software.amazon.smithy.model.node.ArrayNode;
 import software.amazon.smithy.model.node.BooleanNode;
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -833,7 +874,7 @@ private Collection<MemberShape> writeInlineableProperty(String key, ShapeId shap
         }
 
         private boolean hasOnlyDefaultTrait(Shape shape, ShapeId defaultTrait) {
-            return shape.getAllTraits().size() == 1 && shape.hasTrait(defaultTrait);
+            return shape.getAllTraits().isEmpty() || (shape.getAllTraits().size() == 1 && shape.hasTrait(defaultTrait));
```
kstich, line 877: This is the aforementioned fix.

sugmanue (approved):

smithy-model/src/test/resources/software/amazon/smithy/model/shapes/idl-serialization/coerced-io/before.smithy
```diff
@@ -0,0 +1,46 @@
...
+
+structure MultipleBindResponse {}
+
+structure WithoutTraitsRequest {}
+
+structure WithoutTraitsResponse {}
+
+structure WithTraitsRequest {
+    a: String
+}
```
syall, line 44: nit: this doesn't have any traits?

milesziemer (approved):

PR #2322, opened 2024-06-12
Add restJson1 protocol test for content-type and operation with no input

haydenbaker:
#### Background
- Requests should not have the content-type header added when there is no input shape (and therefore no payload)
  - This doesn't appear to be a case in the aws sdk (operations with no input), but could be in generic clients 

#### Testing
- Tested by running protocol tests in the typescript sdk

#### Links
- Related to https://github.com/smithy-lang/smithy-typescript/pull/1304

---

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
@@ -284,3 +284,31 @@ structure TestNoPayloadInputOutput {
     @httpHeader("X-Amz-Test-Id")
     testId: String,
 }
+
+/// This example operation has no input and serializes a request without an HTTP body.
+///
+/// These tests are to ensure we do not attach a body or related headers
+/// (Content-Length, Content-Type) to operations that semantically
+/// cannot produce an HTTP body.
+///
+@readonly
+@http(uri: "/no_input_no_payload", method: "GET")
+operation TestNoInputNoPayload {
+    output: TestNoPayloadInputOutput
+}
+
+apply TestNoInputNoPayload @httpRequestTests([
+    {
+        id: "RestJsonHttpWithNoInput",
+        documentation: "Serializes a GET request for an operation with no input, and therefore no modeled body",
+        protocol: restJson1,
+        method: "GET",
+        uri: "/no_input_no_payload",
+        body: "",
+        forbidHeaders: [
+            "Content-Length",
+            "Content-Type"
+        ],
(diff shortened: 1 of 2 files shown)


hpmellema (approved):

smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
```diff
@@ -284,3 +284,31 @@ structure TestNoPayloadInputOutput {
...
+/// (Content-Length, Content-Type) to operations that semantically
+/// cannot produce an HTTP body.
+///
+@readonly
+@http(uri: "/no_input_no_payload", method: "GET")
+operation TestNoInputNoPayload {
+    output: TestNoPayloadInputOutput
+}
+
+apply TestNoInputNoPayload @httpRequestTests([
```
david-perez, line 300: There's already the [RestJsonNoInputAndOutput](https://github.com/smithy-lang/smithy/blob/4af16861a7c90a4fcf05a5bead08a54811e2d3f9/smithy-aws-protocol-tests/model/restJson1/empty-input-output.smithy#L120) test, to which we can add the `forbidHeaders`.

david-perez: Interestingly, that operation is `POST`, when it defines no input shape. That probably shouldn't build, or at least a warning should be emitted.

haydenbaker: IMO, it's better to add a new test here, as the goal of this test is to specifically verify that the content-type header doesn't get set.

nateprewitt: The key distinction between `POST` and `GET` is idempotency rather than the presence of a body. I'm pretty sure we have APIs today that are `POST` but bind their parameters to the URI/Headers. I'd agree with Hayden this is reasonable to add as a new test.

nateprewitt (approved):

smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy
david-perez: > I'm pretty sure we have APIs today that are POST but bind their parameters to the URI/Headers

Yeah that's fine because you're sending over data. But when the definition of the operation is such that no data could ever be sent, POST usage is probably a modeling mistake on behalf of the user.

PR #2324, opened 2024-06-13
Remove debugging comments from the v2Cbor protocol tests

drganjoo:
A few debugging comments were left out in the last commit for rpcv2Cbor protocol tests, which this PR removes.


drganjoo:
Didn't see the patch earlier.

PR #2333, opened 2024-06-19
Add protocol tests for sparse list primitives

adamthom-amzn:
#### Background
* What do these changes do? 
Adds a list of sparse shorts to existing sparse list tests

* Why are they important?
In some languages, like Java, Strings are naturally nullable and thus handling sparse collections of String values is straightforward, but sparse collection of values with primitives (such as shorts) may introduce complexities.

#### Testing
Ran these tests to verify that serde in a language implementation was broken, and then fixed the implementation and verified that tests pass.

---


JordonPhillips (approved):
Doing the same for maps would be great, though I'm not going to hold back this PR on that

adamthom-amzn:
> Doing the same for maps would be great, though I'm not going to hold back this PR on that

We [already have them](https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/json-maps.smithy#L294); that's why I knew to look again at list protocol tests, because the map ones failed and the list ones did not.

PR #2339, opened 2024-06-26
Add flexible timestamp validation option for nodes

JordonPhillips:
This updates the node validator to add a timestamp validation strategy that will accept any one of the valid timestamp formats if there is no timestamp format trait. This can be useful when writing protocol tests or example traits for protocols where the default format is dependent on the binding location.

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/TimestampFormatPlugin.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/TimestampFormatPlugin.java
@@ -36,6 +36,12 @@ final class TimestampFormatPlugin implements NodeValidatorPlugin {
     private static final DateTimeFormatter DATE_TIME_Z = DateTimeFormatter.ISO_INSTANT;
     private static final Logger LOGGER = Logger.getLogger(TimestampFormatPlugin.class.getName());
 
+    private final boolean flexibleDefault;
+
+    TimestampFormatPlugin(boolean flexibleDefault) {
+        this.flexibleDefault = flexibleDefault;
+    }
+
     @Override
     public void apply(Shape shape, Node value, Context context, Emitter emitter) {
         if (shape instanceof TimestampShape) {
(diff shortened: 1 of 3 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/node/TimestampFormatPlugin.java
```diff
@@ -81,17 +87,48 @@ private void validate(Shape shape, TimestampFormatTrait trait, Node value, Emitt
...
-        // that the value is either a number or a string that matches the date-time format.
-        if (!value.isNumberNode()) {
-            if (value.isStringNode()) {
-                validateDatetime(shape, value, emitter);
-            } else {
-                emitter.accept(value, "Invalid " + value.getType() + " value provided for timestamp, `"
-                                      + shape.getId() + "`. Expected a number that contains epoch seconds with "
-                                      + "optional millisecond precision, or a string that contains an RFC 3339 "
-                                      + "formatted timestamp (e.g., \"1985-04-12T23:20:50.52Z\")");
+        // that the value is matches one of the supported timestamp formats according to
```
kstich, line 90:
```suggestion
        // that the value matches one of the supported timestamp formats according to
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/TimestampFormatPlugin.java
```diff
@@ -102,27 +139,35 @@ private void validateDatetime(Shape shape, Node value, Emitter emitter) {
...
                                   + shape.getId() + "`. Expected an RFC 3339 formatted timestamp (e.g., "
                                   + "\"1985-04-12T23:20:50.52Z\")");
         }
     }
 
+    private boolean isValidDatetime(String timestamp) {
+        // Newer versions of Java support parsing instants that have an offset.
+        // See: https://bugs.openjdk.java.net/browse/JDK-8166138
+        // However, Smithy doesn't allow offsets for timestamp shapes.
+        return timestamp.endsWith("Z") && isValidFormat(timestamp, DATE_TIME_Z);
```
kstich, line 153: nit: sync order with `isValidHttpDate` on doing `isValidFormat` then `endsWith`

kstich (approved):

mtdowling:
Even if we have a mode for allowing this, it clashes with trait node values in the specification: https://smithy.io/2.0/spec/model.html#trait-node-values. To prevent our protocol test definitions from becoming a specialized and non-compatible trait definition, this PR would inherently mean that all timestamp validation must support this modality and we'd need to update the specification to reflect it. I'm not sure it's worth it.

> If a number is provided, it represents Unix epoch seconds with optional millisecond precision. If a string is provided, it MUST be a valid [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339.html) string with optional fractional precision but no UTC offset (for example, 1985-04-12T23:20:50.52Z).

JordonPhillips:
> Even if we have a mode for allowing this, it clashes with trait node values in the specification: https://smithy.io/2.0/spec/model.html#trait-node-values. To prevent our protocol test definitions from becoming a specialized and non-compatible trait definition, this PR would inherently mean that all timestamp validation must support this modality and we'd need to update the specification to reflect it. I'm not sure it's worth it.
> 
> > If a number is provided, it represents Unix epoch seconds with optional millisecond precision. If a string is provided, it MUST be a valid [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339.html) string with optional fractional precision but no UTC offset (for example, 1985-04-12T23:20:50.52Z).

I see what you're saying, but it kinda doesn't make sense to not support all of our formats. I'd be in favor of changing the default to be more lenient since I do agree that it might be confusing if we only do this in special cases. Anything that implements handling timestamps from nodes already has to handle all three formats anyway.

mtdowling:
Yeah, we don't want to special case any traits.

You can add a `timestampFormat` trait to a node value, and it'll work just fine. The built-in handling of this should be strict IMO and limited to what it is now. Adding support for more formats by default means attempting to parse the format with date-time, then if it fails, try with http-date. That's more work to load a model, more compliance tests, and more specs. It also doesn't scale in the long term if we add more formats (e.g., if we ever added epoch-millis, then we don't know if 100 is 100 seconds or milliseconds).

PR #2352, opened 2024-07-19
Use venv when building docs

mtdowling:
MacOS and homebrew don't like installing global Python packages, so this solves that automatically.


sugmanue (approved):

mtdowling:
A future followup to this could be making install a dependency of the other tasks, while ensuring it only runs install if dependencies have changed or the installation is missing.

Issue #2347, opened 2024-07-08
Parsing for `ParameterTypes` has unintuitive casing rules


landonxjames:
The `string` and `boolean` types are `equalsIgnoreCase`, but the `stringArray` type is just compared with `equals`. This leads to unintuitive behavior like `String` and `Boolean` being valid types, but `StringArray` failing.

https://github.com/smithy-lang/smithy/blob/2cd3579c7fad1322798ebb62dd9f47b13ca4beb3/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/ParameterType.java#L46-L60

sugmanue:
Hi Landon, 

Thank you for reporting this issue, I've created an PR to also use `equalsIgnoreCase` for the `stringArray` type.

sugmanue:
I discussed this with the team, this inconsistency is there by design (see [thread here](https://github.com/smithy-lang/smithy/pull/2266#discussion_r1586579002)). We wanted all the type names to be not case sensitive but the first two when in and we cannot change those to avoid breaking compatibility. Going forward all new type names will be not case sensitive.

sugmanue:
Let us know if you have any further questions or concerns.

PR #2355, opened 2024-07-23
Check the stringArray type ignoring case

sugmanue:
As reported in #2347 the stringArray parameter type is tested using equals whereas the other, `boolean` and `string` are tested ignoring case. This change aligns `stringArray` to also be tested ignoring case.

#### Background
* What do these changes do? 
* Why are they important?

#### Links
* Issue #2347

---

### DIFF
--- a/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/ParameterType.java
+++ b/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/ParameterType.java
@@ -51,7 +51,7 @@ public enum ParameterType {
         if (value.equalsIgnoreCase("boolean")) {
             return BOOLEAN;
         }
-        if (value.equals("stringArray")) {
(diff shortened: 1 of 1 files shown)


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/ParameterType.java
```diff
@@ -51,7 +51,7 @@ public static ParameterType fromNode(StringNode node) throws RuleError {
         if (value.equalsIgnoreCase("boolean")) {
             return BOOLEAN;
         }
-        if (value.equals("stringArray")) {
+        if (value.equalsIgnoreCase("stringArray")) {
```
gosar, line 54: I think we wanted to make it all `.equals()` but kept string/boolean as `equalsIgnoreCase` for backwards compatibility reason.

sugmanue:
This was done by design, I will add a note to the issue.

PR #2357, opened 2024-07-25
Add intEnum validation to NodeValidationVisitor

maximk000:
#### Background

Adds a new plugin to validate intEnums in NodeValidationVisitor. Previously you could provide values outside of the allowed set without failing validation. IntEnum Shapes also qualify as Integer Shapes, so they will pass through the same validation as Integer Shapes in [`validateNaturalNumber()`](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java#L228-L248) to make sure floats aren't provided.

#### Testing

Added test cases:
- two for valid values defined by the intEnum
- one for wrong shape
- one for a number with a floating point, similar to Integer tests

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/IntEnumPlugin.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/IntEnumPlugin.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/node/IntEnumPlugin.java
```diff
@@ -0,0 +1,42 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
sugmanue, line 14: Let's use the new shorter copyright notice.

```suggestion
/*
 * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 * SPDX-License-Identifier: Apache-2.0
 */
```

sugmanue (approved):

PR #2358, opened 2024-07-26
Add validator to warn on ignored idempotency token trait

sugmanue:
#### Background
* What do these changes do? 

Added a validator to warn on ignored use of @idempotencyToken trait. As per [spec](https://smithy.io/2.0/spec/behavior-traits.html#idempotencytoken-trait), 

> Only a single member of the input of an operation can be targeted by the idempotencyToken trait; only top-level structure members of the input of an operation are considered.

This validator warns when we find a member using the trait in a context where it will be ignored. We cannot ERROR here as the structures might be reused in valid and invalid context, thus we only warning about the context where it will be ignored.

* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
@@ -5,8 +5,6 @@
 
 package software.amazon.smithy.model.validation.validators;
 
-import static java.lang.String.format;
-
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -15,6 +13,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.neighbor.NeighborProvider;
@@ -29,7 +28,6 @@ import software.amazon.smithy.model.traits.MixinTrait;
 import software.amazon.smithy.model.traits.Trait;
 import software.amazon.smithy.model.validation.AbstractValidator;
 import software.amazon.smithy.model.validation.ValidationEvent;
-import software.amazon.smithy.utils.ListUtils;
 
 /**
  * Emits warnings when a structure member has an idempotency token trait that will be ignored.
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        List<Relationship> relationships = reverse.getNeighbors(containerShape);
+        int checkedRelationshipCount = relationships.size();
+        for (Relationship relationship : relationships) {
+            // Skip members of the container.
+            if (relationship.getRelationshipType() == RelationshipType.MEMBER_CONTAINER) {
+                checkedRelationshipCount--;
+                continue;
+            }
+            if (relationship.getRelationshipType() != RelationshipType.INPUT) {
+                ignoredRelationships.merge(relationship.getRelationshipType(),
```
mtdowling, line 80: Can this be simplified to just use ignoredRelationships.computeIfAbsent? Might not need mergeShapeIdsList either at that point.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private ValidationEvent emit(
+        MemberShape memberShape,
+        Trait trait,
+        int checkedRelationshipCount,
+        Map<RelationshipType, List<ShapeId>> ignoredRelationships
+    ) {
+        String mixedIn = memberShape.getMixins().isEmpty() ? "" : " mixed in";
+        String message = "The `%s` trait applied to this%s member is ";
```
mtdowling, line 107: Can we say something more clear like "The `idempotencyToken` trait only has an effect when applied to a top-level operation input member, but it was applied and ignored in the following contexts: " ... etc

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        if (checkedRelationshipCount == ignoredRelationships.size()) {
```
mtdowling, line 108: Why omit the context here? It looks like we emit the event on the member that has the trait, but don't say how the containing shape was referenced in a way that caused the trait to be ignored.

mtdowling (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
sugmanue: I will remove.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
sugmanue: I will change as described.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
```diff
@@ -0,0 +1,112 @@
...
+        MemberShape memberShape,
+        Trait trait,
+        NeighborProvider reverse,
+        List<ValidationEvent> events
+    ) {
+
+        // Store relationships so we can emit one event per ignored binding.
+        Map<RelationshipType, List<ShapeId>> ignoredRelationships = new TreeMap<>();
+        List<Relationship> relationships = reverse.getNeighbors(containerShape);
+        int checkedRelationshipCount = relationships.size();
```
mtdowling, line 67: looks like this doesn't do anything now

mtdowling (approved):
Ship it but remove what I think is some dead code

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotencyTokenIgnoredValidator.java
sugmanue: 🤦 yeah, let me remove that.

PR #2364, opened 2024-08-05
Make the `NonEmptyString` type public

jivanic-demystdata:
Any reason to keep it private?

It's a quite useful type.
Making it private forces everyone to redefine it

### DIFF
--- a/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
+++ b/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
@@ -552,7 +552,6 @@ structure xmlNamespace {
     prefix: NonEmptyString
 }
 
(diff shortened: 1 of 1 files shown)


JordonPhillips:
Typically when people put traits directly on a shape like this, it's not just one trait. There's other constraints, documentation, etc. Shapes defined in the prelude can't have traits applied to them, so there's limited use to providing a shape like this with just one trait. The base shapes are different in that everybody uses them, and being bare they're a useful base case applicable almost anywhere. The primitive shapes are similar in that regard.

Number-wise, string shapes with just this trait only appear 86 times in the AWS models, out of a total of 20,548 string shapes. Of those, only 9 appear to be intended to be used generically. I don't think that's enough usage to demonstrate a need to include it in the prelude.

PR #2367, opened 2024-08-11
Improve BottomUpIndex performance

milesziemer:
When doing some ad-hoc profiling on the language server using aws service models I noticed that BottomUpIndex was taking up a lot of CPU time. This was because it used selectors and PathFinder to determine the paths from resources/operations up to their enclosing service. Selectors are pretty slow in comparison to regular java code, and when the model is large with many services/operations, this index can take a while to compute. BottomUpIndex is used for validating `@cfnResource` (through CfnResourcePropertyValidator -> CfnResourceIndex), so anyone using this trait is paying this performance cost (in particular, SDK code generators).

To get a rough idea of how much faster this commit makes BottomUpIndex, I ran model validation using the cli on all aws service models with and without this change a few times each:
```
smithy validate --quiet # 14 - 16 seconds

/path/to/smithy/smithy-cli/build/image/smithy-cli-darwin-aarch64/bin/smithy
validate --quiet # 8 - 9 seconds
```

I also added a jmh benchmark for BottomUpIndex (named it KnowledgeIndicies in case we want to add more, but we could keep them separate too and I can change the name). The benchmark results:
Before this change:
```
KnowledgeIndicies.createsBottomUpIndex  avgt    3  6.300 ± 0.245  us/op
```
After this change:
```
KnowledgeIndicies.createsBottomUpIndex  avgt    3  0.467 ± 0.042  us/op
```

### DIFF
--- a/smithy-model/src/jmh/java/software/amazon/smithy/model/jmh/KnowledgeIndicies.java
+++ b/smithy-model/src/jmh/java/software/amazon/smithy/model/jmh/KnowledgeIndicies.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.model.jmh;
+
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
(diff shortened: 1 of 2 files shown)


JordonPhillips (commented):
This looks good, but can you add a jmh benchmark?

JordonPhillips (changes requested):
Whoops, set the review to comment instead of request changes

JordonPhillips (approved):

PR #2393, opened 2024-09-04
Add validation for null members in structures, lists and maps

maximk000:
#### Background
Previously, a NullNode could be provided as a required structure member without emitting a ValidationEvent. So an example test below would fail. 

```    @Test
    public void nullRequiredStructureMember() {
        ObjectNode structure = ObjectNode.builder()
                .withMember("foo", Node.nullNode())
                .build();
        NodeValidationVisitor visitor = NodeValidationVisitor.builder()
                .value(structure)
                .model(MODEL)
                .allowOptionalNull(true)
                .build();
        List<ValidationEvent> events = MODEL
                .expectShape(ShapeId.from("ns.foo#Structure"))
                .accept(visitor);

        assertThat(events, hasSize(1));
        assertThat(events.get(0).getMessage(), containsString(
                "foo: Required member `foo` for structure ns.foo#Structure for cannot be null"));
    }

ns.foo#Structure schema:
"ns.foo#Structure": {
    "type": "structure",
    "members": {
        "foo": {
            "target": "ns.foo#String",
            "traits": {
                "smithy.api#required": {}
            }
        },
...
```

This happens because the memberShape function would check the member's [target](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java#L385) instead of itself, which always satisfies the !shape.isMemberShape() [condition](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java#L410) in the invalidShape function.

To fix this I added a new function to check null member nodes for structures, lists and maps. It relies on the `isMemberNullable` function, which checks that:
- If a structure member, member cannot be required
- If a list member, the list must have the sparse trait
- If a map value, the map must have the sparse trait (this case is not explicit but happens due to a fall through to the [default case](https://github.com/smithy-lang/smithy/blob/1e4458ee825b24490c03fbf22f4b2f79111c7ae5/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java#L191))

#### Testing
Added testcases:
- One that emits ValidationEvent when NullNode is provided as a required structure member
- One that emits ValidationEvent when NullNode is provided as member for a non-sparse list
- One that doesn't emit ValidationEvent when NullNode is provided as member for a sparse list
- One that emits ValidationEvent when NullNode is provided as a value for a non-sparse map
- One that doesn't emit ValidationEvent when NullNode is provided as a value for a sparse map

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
@@ -392,29 +392,28 @@ public final class NodeValidationVisitor implements ShapeVisitor<List<Validation
     }
 
     public List<ValidationEvent> checkNullMember(MemberShape shape) {
-        List<ValidationEvent> events = new ArrayList<>();
-        if (!shape.asMemberShape().filter(nullableIndex::isMemberNullable).isPresent()) {
+        if (!nullableIndex.isMemberNullable(shape)) {
             switch (model.expectShape(shape.getContainer()).getType()) {
                 case LIST:
-                    events.add(event(
+                    return ListUtils.of(event(
                             String.format(
-                                    "Non-sparse list shape %s cannot contain null values", shape.getContainer())));
+                                    "Non-sparse list shape `%s` cannot contain null values", shape.getContainer())));
                     break;
                 case MAP:
-                    events.add(event(
+                    return ListUtils.of(event(
                             String.format(
-                                    "Non-sparse map shape %s cannot contain null values", shape.getContainer())));
+                                    "Non-sparse map shape `%s` cannot contain null values", shape.getContainer())));
                     break;
                 case STRUCTURE:
-                    events.add(event(
-                            String.format("Required member `%s` for structure %s for cannot be null",
+                    return ListUtils.of(event(
+                            String.format("Required structure member `%s` for `%s` cannot be null",
                                     shape.getMemberName(), shape.getContainer())));
                     break;
                 default:
                     break;
             }
         }
-        return events;
+        return ListUtils.of();
     }
 
     @Override
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
@@ -398,17 +398,14 @@ public final class NodeValidationVisitor implements ShapeVisitor<List<Validation
                     return ListUtils.of(event(
                             String.format(
                                     "Non-sparse list shape `%s` cannot contain null values", shape.getContainer())));
-                    break;
                 case MAP:
                     return ListUtils.of(event(
                             String.format(
                                     "Non-sparse map shape `%s` cannot contain null values", shape.getContainer())));
-                    break;
                 case STRUCTURE:
                     return ListUtils.of(event(
                             String.format("Required structure member `%s` for `%s` cannot be null",
                                     shape.getMemberName(), shape.getContainer())));
-                    break;
                 default:
                     break;
             }
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
@@ -378,6 +378,9 @@ public final class NodeValidationVisitor implements ShapeVisitor<List<Validation
     @Override
     public List<ValidationEvent> memberShape(MemberShape shape) {
         List<ValidationEvent> events = applyPlugins(shape);
+        if (value.isNullNode()) {
+            events.addAll(checkNullMember(shape));
+        }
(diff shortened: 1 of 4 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
         return events;
     }
 
+    public List<ValidationEvent> checkNullMember(MemberShape shape) {
+        List<ValidationEvent> events = new ArrayList<>();
```
rationull, line 395: Not sure if this is a concern for NodeValidationVisitor performance, but we could return a static empty list (`Collections.emptyList()`) unless at least one error is actually found.

kstich (changes requested):
Some comments on event message formatting (that will also require test updates) and a simplification of the conditional check.

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
kstich: True, this could be rewritten to return `ListUtils.of` based collections instead of creating resizable lists.

kstich, line 395:
```suggestion
```

Note that the switch can have its `break;` lines removed as well, since each case will have a `return`.

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
...
+        if (!shape.asMemberShape().filter(nullableIndex::isMemberNullable).isPresent()) {
+            switch (model.expectShape(shape.getContainer()).getType()) {
+                case LIST:
+                    events.add(event(
```
kstich, line 399:
```suggestion
                    return ListUtils.of(event(
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
...
+                            String.format(
+                                    "Non-sparse list shape %s cannot contain null values", shape.getContainer())));
+                    break;
+                case MAP:
+                    events.add(event(
```
kstich, line 404:
```suggestion
                    return ListUtils.of(event(
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
...
+                            String.format(
+                                    "Non-sparse map shape %s cannot contain null values", shape.getContainer())));
+                    break;
+                case STRUCTURE:
+                    events.add(event(
```
kstich, line 409:
```suggestion
                    return ListUtils.of(event(
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
...
+                            String.format("Required member `%s` for structure %s for cannot be null",
+                                    shape.getMemberName(), shape.getContainer())));
+                    break;
+                default:
+                    break;
+            }
+        }
+        return events;
+    }
```
kstich, line 418:
```suggestion
            }
        }
        return ListUtils.of();
    }
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -388,6 +391,32 @@ public List<ValidationEvent> memberShape(MemberShape shape) {
         return events;
     }
 
+    public List<ValidationEvent> checkNullMember(MemberShape shape) {
+        List<ValidationEvent> events = new ArrayList<>();
+        if (!shape.asMemberShape().filter(nullableIndex::isMemberNullable).isPresent()) {
```
rationull, line 396: Nitpick: `shape` is passed in as `MemberShape` so we don't need the `asMemberShape().filter()` do we? Switching to just `shape` removes any apparent (but not practical) ambiguity between the "member shape is not nullable" and "not a member shape" cases that are logically possible.

rationull: Ninja'd by Kevin with a better version of this comment!

kstich (approved):

PR #2398, opened 2024-09-17
Add breakingChanges for httpChecksum trait

gosar:
#### Background
The `aws.protocols#httpChecksum` trait was missing any rules for detecting backwards incompatible changes to the applied trait. This PR adds `breakingChanges` to the trait definition.

#### Testing
Added tests to assert the events for diff.

#### Links

---


smithy-aws-traits/src/main/resources/META-INF/smithy/aws.protocols.smithy
```diff
@@ -123,7 +123,42 @@ structure ec2Query {}
 string ec2QueryName
 
 /// Indicates that an operation supports checksum validation.
-@trait(selector: "operation")
+@trait(
+    selector: "operation",
+    breakingChanges: [
+        {
+            change: "remove",
+            severity: "ERROR",
```
hpmellema, line 131: Error is the default severity for breaking changes, so we can probably remove that here.

smithy-aws-traits/src/main/resources/META-INF/smithy/aws.protocols.smithy
```diff
@@ -123,7 +123,42 @@ structure ec2Query {}
...
+                `requestAlgorithmMember` was removed so newly generated clients will no longer be able to pick the \
+                request checksum algorithms The service MUST continue to support old clients by supporting \
+                `requestAlgorithmMember`."""
+        },
+        {
+            change: "remove",
+            path: "/requestValidationModeMember",
+            severity: "ERROR",
+            message: """
+                `requestValidationModeMember` was removed so newly generated clients will no longer validate response \
```
hpmellema, line 147:
```suggestion
                `requestValidationModeMember` was removed, so newly generated clients will no longer validate response \
```

hpmellema: Same above.

smithy-aws-traits/src/main/resources/META-INF/smithy/aws.protocols.smithy
yefrig, line 131: Should we include a message when the entire trait is removed?

yefrig (approved):

smithy-aws-traits/src/main/resources/META-INF/smithy/aws.protocols.smithy
hpmellema: I noticed that the request compression rules use `DANGER` [see](https://github.com/smithy-lang/smithy/blob/e3348620935216be2943f92c5288ec0a6762353e/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy#L1253). These seem like they would have the same severity.

gosar: I debated it and felt removal of compression being less severe than removal of checksums was reasonable. But don't feel strongly about it, so changing to DANGER.

gosar: oh I wanted to add message later after adding test, but forgot. thanks for catching.

hpmellema (approved):

Issue #2400, opened 2024-09-19
open-api plugin does not obey "effective documentation" precedence for member structures


nated0g:
There appears to be a bug in the `open-api` plugin that leads to incorrectly applied documentation.  This is causing some friction for us, requiring refactoring of our model to display documentation correctly.  If the plugin behaved according to the documentation here [Effective documentation](https://smithy.io/2.0/spec/documentation-traits.html#effective-documentation), we would not have any issue.

To repro:

**model.smithy**
```smithy
$version: "2"

namespace example

use aws.api#service
use aws.protocols#restJson1

/// Cutting edge foo barring service, now with AI
@service(sdkId: "Foo")
@restJson1
service FooService {
    version: "2006-03-01"
    operations: [
        BarOperation
    ]
}

/// Performs a bar operation.
@http(uri: "/bar", method: "POST")
operation BarOperation {
    input: BarInput
    output: BarOutput
}

@documentation("It's a Quux")
structure Quux {
    quux: String
}

/// Input for the BarOperation.
structure BarInput {
    @required
    @documentation("It's a Bar")
    bar: Quux
}

/// Output for the BarOperation.
structure BarOutput {
    /// The result of the bar operation.
    result: String
}
```

**smithy-build.json**
```json
{
  "version": "1.0",
  "maven": {
    "dependencies": [
      "software.amazon.smithy:smithy-aws-traits:1.51.0",
      "software.amazon.smithy:smithy-openapi:1.51.0"
    ]
  },
  "sources": ["model.smithy"],
  "plugins": {
    "openapi": {
      "service": "example#FooService",
      "protocol": "aws.protocols#restJson1"
    }
  }
}
```

This produces a `model.json` that **correctly** contains `"It's a Bar"` for the member's documentation:
```json
 "example#BarInput": {
            "type": "structure",
            "members": {
                "bar": {
                    "target": "example#Quux",
                    "traits": {
                        "smithy.api#documentation": "It's a Bar",
                        "smithy.api#required": {}
                    }
                }
            },
            "traits": {
                "smithy.api#documentation": "Input for the BarOperation."
            }
        },
```

however, the generated `FooService.openapi.json` contains a schema that looks like this:

```json
{
        "schemas": {
            "BarOperationRequestContent": {
                "type": "object",
                "description": "Input for the BarOperation.",
                "properties": {
                    "bar": {
                        "$ref": "#/components/schemas/Quux"
                    }
                },
                "required": [
                    "bar"
                ]
[...]

PR #2402, opened 2024-09-23
Allow adding member docs to $ref

milesziemer:
Addresses https://github.com/smithy-lang/smithy/issues/2400.

When a member targets a structure, it becomes a schema reference when converted to JSON Schema. Previously, we didn't add member docs to the converted object, possibly because earlier versions of open api or JSON Schema did not support it. Reading through [this issue](https://github.com/OAI/OpenAPI-Specification/issues/1514) the [OAI spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#reference-object), and the [JSON Schema Spec](https://json-schema.org/draft/2020-12/json-schema-core#section-8.2.3), it seems that OpenAPI 3.1 and JSON Schema 2020-12 support the `description` property alongside `$ref`. I wasn't able to find anything about whether it is supported in [JSON Schema 07](https://json-schema.org/draft-07/json-schema-release-notes). This commit adds a new config option, `addReferenceDescriptions` that will add the `description` property alongside `$ref` when the member has documentation. I made it opt-in through the config option so we don't cause any existing documentation to be changed unexpectedly.

---

### DIFF
--- a/docs/source-2.0/guides/model-translations/converting-to-openapi.rst
+++ b/docs/source-2.0/guides/model-translations/converting-to-openapi.rst
@@ -1053,6 +1053,58 @@ disableIntEnums (``boolean``)
         }
 
 
+.. _generate-openapi-setting-addReferenceDescriptions:
+
+addReferenceDescriptions (``boolean``)
+    Set to ``true`` to add the ``description`` property to ``$ref`` members
+    with the value of the member's :ref:`documentation-trait` trait.
+
+    .. important::
+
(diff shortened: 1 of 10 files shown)


JordonPhillips (approved):
Looks good! We should add this to the examples repo [here](https://github.com/smithy-lang/smithy-examples/blob/main/conversion-examples/smithy-to-openapi/smithy-build.json)

JordonPhillips (approved):

PR #2412, opened 2024-10-01
Fix JSON Schema pointer conflicts with trait definitions

kstich:
This commit updates the DeconflictingStrategy to scrub trait definitions from the model used to pre-compute JSON Schema pointers. These shapes wouldn't have made their way into converted schemas anyways, and existing behavior for non-computed pointers is maintained.

---


yasmewad (approved):
Looks good to me.

hpmellema (approved):

PR #2403, opened 2024-09-23
serialize empty strings and lists to headers in restXml and restJson1

kuhe:
#### Background
* What do these changes do? 
  * changes header serialization test assertions to include empty strings and lists, but still exclude null
* Why are they important?
  * absence and emptiness of headers differ in meaning

#### Testing
* How did you test these changes?

Tested protocol-test codegen and test execution in smithy-typescript + aws-sdk-js-v3

#### Links
https://github.com/smithy-lang/smithy-typescript/pull/1412

PR #2415, opened 2024-10-03
fix: update prefix header tests to allow for empty headers

Madrigal:
#### Background
* What do these changes do? 
Update protocol tests to catch an issue where SDKs were not serializing empty headers 

* Why are they important?
Had an issue where calling S3 with metadata with an empty value caused SDKs to not serialize the value and not send the header. While #2403 addresses the general case of headers present but with no values, it wouldn't cover this scenario since S3 metadata headers are not modeled as [httpHeader](https://smithy.io/2.0/spec/http-bindings.html#httpheader-trait) but as [httpPrefixHeader](https://smithy.io/2.0/spec/http-bindings.html#httpprefixheaders-trait)

#### Testing
Generated tests on [smithy-go](https://github.com/aws/smithy-go) and ran the new tests on [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2/)

#### Links

---

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-prefix-headers.smithy
@@ -42,7 +42,7 @@ apply HttpPrefixHeaders @httpRequestTests([
     },
     {
         id: "RestJsonHttpPrefixHeadersAreNotPresent",
-        documentation: "No prefix headers are serialized because the value is empty",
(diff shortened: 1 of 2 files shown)


kuhe (approved):

sugmanue (approved):

sugmanue:
Just a side comment, please notice that Smithy does not use conventional commit messages.

PR #2433, opened 2024-10-24
Update `NullAndEmptyHeadersServer` tests for `restJson1` and `restXml`

landonxjames:
#### Background
* What do these changes do? 
* Why are they important?

`NullAndEmptyHeaderClient` tests were [recently updated](https://github.com/smithy-lang/smithy/pull/2403) to expect empty headers to be serialized. Similar changes were not made for the server variants of those tests. This leads to an unexpected difference in serialization between servers and clients and prevents clients from receiving semantically meaningful headers from the server.

The test now align with the similar tests for clients. They expect empty headers to be serialized as "" and null headers to not be serialized

#### Testing
Confirmed that the same changes made for client header serialization in https://github.com/smithy-lang/smithy-rs/pull/3887 worked for the server tests as well.

#### Links

---

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
@@ -412,7 +412,7 @@ structure InputAndOutputWithHeadersIO {
     headerIntegerEnumList: IntegerEnumList,
 }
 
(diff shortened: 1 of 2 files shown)


Madrigal:
Can confirm that this also impacts Go SDK

milesziemer (approved):

PR #2438, opened 2024-10-31
Removing `appliesTo` from `NullAndEmptyHeaders` tests 

landonxjames:
#### Background
* What do these changes do? 
* Why are they important?
Removing `appliesTo` from `NullAndEmptyHeaders` tests per conversation with @kstich 

#### Testing

#### Links

---


smithy-aws-protocol-tests/model/restJson1/http-headers.smithy
milesziemer, line 1: Since these two operations and their tests (NullAndEmptyHeaders(Client/Server)) are now identical, can we just combine them into one and get rid of the client-only/server-only tags? For restXml too

milesziemer (changes requested):

kstich:
This older PR needs consideration across implementations on top of its requested updates, closing for now.

PR #2439, opened 2024-11-01
Add validator for xmlFlattened + xmlName

milesziemer:
Adds a validator for the xmlFlattened trait that checks if the member's target is a list that has a member with xmlName, and that xmlName doesn't match the name of the xmlFlattened member. xmlFlattened causes the xmlName of the list member to be ignored. This is called out in our docs:
https://smithy.io/2.0/spec/protocol-traits.html#flattened-list-serialization, but we were missing a validator for it. It is a warning because you might mean to do this, i.e. if the list is used in multiple places.

I also added some suppressions to our protocol tests that are testing the behavior for models that would trip up this validator.

---

### DIFF
--- a/smithy-aws-protocol-tests/model/awsQuery/main.smithy
+++ b/smithy-aws-protocol-tests/model/awsQuery/main.smithy
@@ -1,5 +1,15 @@
 $version: "2.0"
 
+metadata suppressions = [
+    {
+        id: "XmlFlattenedTrait"
+        namespace: "aws.protocoltests.query"
+        reason: """
+            Some tests are for asserting the correct behavior in the case that
+            xmlFlattened is wrong and trips this validator."""
+    }
+]
+
 namespace aws.protocoltests.query
 
 use aws.api#service
--- a/smithy-aws-protocol-tests/model/awsQuery/main.smithy
+++ b/smithy-aws-protocol-tests/model/awsQuery/main.smithy
@@ -1,15 +1,5 @@
 $version: "2.0"
 
-metadata suppressions = [
-    {
-        id: "XmlFlattenedTrait"
-        namespace: "aws.protocoltests.query"
-        reason: """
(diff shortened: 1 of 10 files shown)


smithy-aws-protocol-tests/model/awsQuery/main.smithy
```diff
@@ -1,5 +1,15 @@
 $version: "2.0"
 
+metadata suppressions = [
```
kstich, line 3: Is it prohibitive to use the `@suppress` trait instead of a metadata suppression? It's a nice bit of documentation on that specific member as well that it's intending to test the behavior.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/XmlFlattenedTraitValidator.java
```diff
@@ -0,0 +1,51 @@
...
+        for (MemberShape member : model.getMemberShapesWithTrait(XmlFlattenedTrait.class)) {
+            // Don't emit the event if they're being explicit about the xmlName on this member
+            if (member.hasTrait(XmlNameTrait.class)) {
+                continue;
+            }
+
+            model.getShape(member.getTarget())
+                    .flatMap(Shape::asListShape)
+                    .flatMap(listShape -> validateMemberTargetingList(member, listShape))
+                    .ifPresent(events::add);
```
kstich, line 37: IMO reading this as a stream is harder than just reading `if`s since we know the member target should exist.

```
Shape target = model.expectShape(member.getTarget());
if (target.isListShape()) {
    validateMemberTargetListing(member, targetList).ifPresent(events::add);
}
```

milesziemer: True I forgot we run that type of validation first, so we can expect the member target to exist.

PR #2445, opened 2024-11-07
Add round-trip convertibility of resource schemas

haydenbaker:
#### Background
- Before these changes, converting a node to resource-schema is lossy
- These changes include the necessary changes to keep round-trips between node and resource-schema lossless
- Property was adapted to incorporate its separate "properties" as extensions to the schema**
  - **only when these "properties" (`insertionOrder`, `dependencies`) is set to a non-default
- Unknown properties are added as extensions to the Schema

#### Testing
- Unit testing

---

### DIFF
--- a/smithy-aws-cloudformation/src/main/java/software/amazon/smithy/aws/cloudformation/schema/model/Property.java
+++ b/smithy-aws-cloudformation/src/main/java/software/amazon/smithy/aws/cloudformation/schema/model/Property.java
@@ -16,7 +16,9 @@
 package software.amazon.smithy.aws.cloudformation.schema.model;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import software.amazon.smithy.jsonschema.Schema;
 import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.node.ObjectNode;
@@ -32,8 +34,6 @@ import software.amazon.smithy.utils.ToSmithyBuilder;
  * @see <a href="https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/data/schema/provider.definition.schema.v1.jsonL74">Resource Type Properties JSON Schema</a>
  */
 public final class Property implements ToNode, ToSmithyBuilder<Property> {
-    private final boolean insertionOrder;
-    private final List<String> dependencies;
     private final Schema schema;
     // Other reserved property names in definition but not in the validation
     // JSON Schema, so not defined in code:
@@ -49,14 +49,12 @@ public final class Property implements ToNode, ToSmithyBuilder<Property> {
             schemaBuilder = builder.schema.toBuilder();
         }
 
-        this.insertionOrder = builder.insertionOrder;
-        if (this.insertionOrder) {
+        if (builder.insertionOrder) {
             schemaBuilder.putExtension("insertionOrder", Node.from(true));
         }
 
-        this.dependencies = builder.dependencies;
-        if (!this.dependencies.isEmpty()) {
-            schemaBuilder.putExtension("dependencies", Node.fromStrings(this.dependencies));
+        if (!builder.dependencies.isEmpty()) {
+            schemaBuilder.putExtension("dependencies", Node.fromStrings(builder.dependencies));
         }
 
         this.schema = schemaBuilder.build();
@@ -69,19 +67,12 @@ public final class Property implements ToNode, ToSmithyBuilder<Property> {
 
     @Override
     public Builder toBuilder() {
-        return builder()
-                .insertionOrder(insertionOrder)
-                .dependencies(dependencies)
-                .schema(schema);
+        return builder().schema(schema);
     }
 
     public static Property fromNode(Node node) {
         ObjectNode objectNode = node.expectObjectNode();
         Builder builder = builder();
-
-        objectNode.getBooleanMember("insertionOrder", builder::insertionOrder);
-        objectNode.getArrayMember("dependencies", StringNode::getValue, builder::dependencies);
-
         builder.schema(Schema.fromNode(objectNode));
 
         return builder.build();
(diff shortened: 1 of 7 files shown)


smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/Schema.java
```diff
@@ -970,5 +1000,138 @@ public Builder disableProperty(String propertyName) {
                     return this;
             }
         }
+
+        Builder applyNode(String key, Node node) {
```
JordonPhillips, line 1004: This is a lot - can NodeMapper handle it?

JordonPhillips (dismissed):

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/Schema.java
haydenbaker: No, there's no way of adding custom behavior with unknown fields -`setWhenMissingSetter` takes an enum, `WhenMissing`. I guess we could add overload it, but adding a `fromNode` is nice and simple.

smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/ResourceSchemaTest.java
```diff
@@ -0,0 +1,61 @@
...
+        Node node = Node.parse(json);
+        ResourceSchema schemaFromNode = mapper.deserialize(node, ResourceSchema.class);
+        Node nodeFromSchema = schemaFromNode.toNode();
+
+        Node.assertEquals(nodeFromSchema.withDeepSortedKeys(), node.withDeepSortedKeys());
+    }
+
+    public static List<String> resourceSchemaFiles() {
+        try {
+            Path definitionPath = Paths.get(ResourceSchemaTest.class.getResource("aws-sagemaker-domain.json").toURI());
```
kstich, line 50: Should this be `.cfn.` file to be included in the test suite?

smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/ResourceSchemaTest.java
```diff
@@ -0,0 +1,61 @@
...
+public class ResourceSchemaTest {
+
+    @ParameterizedTest
+    @MethodSource("resourceSchemaFiles")
+    public void validateResourceSchemaFromNodeToNode(String resourceSchemaFile) {
+        NodeMapper mapper = new NodeMapper();
+        String json = IoUtils.readUtf8File(resourceSchemaFile);
+
+        Node node = Node.parse(json);
+        ResourceSchema schemaFromNode = mapper.deserialize(node, ResourceSchema.class);
```
kstich, line 42: This should be a static `fromNode` method so using it doesn't require configuring the `NodeMapper`.

kstich (changes requested):

smithy-aws-cloudformation/src/main/java/software/amazon/smithy/aws/cloudformation/schema/model/Property.java
```diff
@@ -41,24 +41,30 @@ public final class Property implements ToNode, ToSmithyBuilder<Property> {
...
+        if (builder.schema == null) {
+            schemaBuilder = Schema.builder();
+        } else {
+            schemaBuilder = builder.schema.toBuilder();
+        }
 
-        // Only serialize these properties if set to non-defaults.
-        if (insertionOrder) {
-            builder.withMember("insertionOrder", Node.from(insertionOrder));
+        this.insertionOrder = builder.insertionOrder;
```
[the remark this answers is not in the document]
haydenbaker: I initially did that, but kept the existing fields so that we don't need to run the `toNode()` methods each time we access them:
```
        this.insertionOrder = schema.getExtension("insertionOrder")
                .map(n -> n.toNode().expectBooleanNode().getValue()).orElse(false);

        this.dependencies = schema.getExtension("dependencies")
                .map(n -> n.toNode().expectArrayNode().getElementsAs(StringNode::getValue))
                .orElseGet(ArrayList::new);
```
This logic would essentially move into the getters if we got rid of the fields.

smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/ResourceSchemaTest.java
haydenbaker: Interestingly, adding that to the filename causes the node comparison to break - the order of array items in the node change... 

I think the way we deep-sort is incorrect.

smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/ResourceSchemaTest.java
kstich: I think adding that method is preferable, basically just containing this code, so creating/configuring the `NodeMapper` isn't required.

smithy-aws-cloudformation/src/test/java/software/amazon/smithy/aws/cloudformation/schema/fromsmithy/ResourceSchemaTest.java
kstich: It could be that or something isn't using an ordered collection.

haydenbaker: Ah, yep.

haydenbaker: Took a crack at it and defined a custom sorter, but decided against it. I just updated the cfn file to have correctly ordered items.

smithy-aws-cloudformation/src/main/java/software/amazon/smithy/aws/cloudformation/schema/model/Property.java
```diff
@@ -87,7 +106,7 @@ public Schema getSchema() {
 
     public static final class Builder implements SmithyBuilder<Property> {
         private boolean insertionOrder = false;
-        private final List<String> dependencies = new ArrayList<>();
+        private final List<String> dependencies = new LinkedList<>();
```
kstich, line 109: An `ArrayList` is insertion ordered, why a `LinkedList` here?

haydenbaker: Artifact from a previous change - reverted.

kstich (approved):

PR #2452, opened 2024-11-12
Add transforms to remove deprecated shapes

hpmellema:
### Description 
Adds two transforms to `smithy-model` and `directed codegen` that allow users to filter out deprecated shapes by version or date. 

These transforms allow a user of directed codegen to quickly implement the recommended [`relativeVersion`](https://smithy.io/2.0/guides/building-codegen/configuring-the-generator.html#relativeversion-client-and-type-codegen-only) and [`relativeDate`](https://smithy.io/2.0/guides/building-codegen/configuring-the-generator.html#relativedate-client-and-type-codegen-only) settings for client and type code generators.

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.model.transform;
+
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.DeprecatedTrait;
+
+final class FilterDeprecatedRelativeDate {
+    private static final Logger LOGGER = Logger.getLogger(FilterDeprecatedRelativeDate.class.getName());
+    // YYYY-MM-DD calendar date with optional hyphens.
+    private static final Pattern ISO_8601_DATE_REGEX = Pattern.compile(
+            "^([0-9]{4})-?(1[0-2]|0[1-9])-?(3[01]|0[1-9]|[12][0-9])$"
+    );
+
+    public final String relativeDate;
+
+    FilterDeprecatedRelativeDate(String relativeDate) {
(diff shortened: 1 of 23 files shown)


mtdowling:
Can we get a smithy-build.json transform for this too?

smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public Model transform(ModelTransformer transformer, Model model) {
+        // If there is no filter. Exit without traversing shapes
+        if (relativeDate == null) {
+            return model;
+        }
+
+        Set<Shape> shapesToRemove = new HashSet<>();
+        for (Shape shape : model.getShapesWithTrait(DeprecatedTrait.class)) {
+            if (shape.hasTrait(DeprecatedTrait.class)) {
```
milesziemer, line 44: Is this if necessary considering what we're looping over?

smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
```diff
@@ -0,0 +1,70 @@
...
+import software.amazon.smithy.model.traits.DeprecatedTrait;
+
+final class FilterDeprecatedRelativeDate {
+    private static final Logger LOGGER = Logger.getLogger(FilterDeprecatedRelativeDate.class.getName());
+    // YYYY-MM-DD calendar date with optional hyphens.
+    private static final Pattern ISO_8601_DATE_REGEX = Pattern.compile(
+            "^([0-9]{4})-?(1[0-2]|0[1-9])-?(3[01]|0[1-9]|[12][0-9])$"
+    );
+
+    public final String relativeDate;
```
milesziemer, line 24: Does this need to be public? Same with the other transformer

milesziemer: Some of these methods could be private/package-private (in the other transformer too) I think

smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
hpmellema: Definitely not. Will remove.

smithy-model/src/main/java/software/amazon/smithy/model/transform/FilterDeprecatedRelativeDate.java
hpmellema: Yeah. That is an error. Will update.

milesziemer (approved):

PR #2453, opened 2024-11-12
Converting existing gradle DSL from groovy to kotlin.

yasmewad:
**Why is this needed ?**
This is the first step in moving to a more consistent and composable experience for gradle configuration in the smithy repository. Writing it in kotlin to make it more expressive as compared to groovy. This commit just converts the existing gradle build from groovy to kotlin.

**Next steps:**
Once this change is merged, few other PRs will follow to introduce nice features such as `buildSrc` plugins to enhance configurability of how each sub project builds in this package. It will introduce flexibility in things such as allowing each subproject to compile with different JVM toolchain, and much more.

**Testing**
- `gradle clean && gradle build`.
- Any more testing? 

---


settings.gradle.kts
```diff
@@ -0,0 +1,51 @@
...
+    repositories {
+        mavenLocal()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+
+    plugins {
+        val smithyGradleVersion : String by settings
+        id("software.amazon.smithy.gradle.smithy-jar") version smithyGradleVersion
+    }
```
yasmewad, line 11: Previously we would pin the version in specific subprojects' `build.gradle.kts` like this:
```
plugins {
     id("software.amazon.smithy.gradle.smithy-jar") version smithyGradleVersion
}
		
```

However I found that the correct way to do version pinning for dependencies from `gradle.properties` is this.

This is also temporary while we move to the `versions.toml` approach like other projects and `buildSrc` based plugins in future PRs.

References:
 - https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_management 
- Github issue on gradle side: https://github.com/gradle/gradle/issues/1697

smithy-aws-apigateway-traits/build.gradle.kts
```diff
@@ -16,10 +16,10 @@
 description = "This module provides Smithy traits and validators for API Gateway."
 
 ext {
-    displayName = "Smithy :: AWS :: API Gateway Traits"
-    moduleName = "software.amazon.smithy.aws.apigateway.traits"
+    set("displayName", "Smithy :: AWS :: API Gateway Traits")
```
hpmellema, line 19: I think these should likely be `extra` properties rather than extensions. Interested if you found any guidance in the gradle guides that may suggest otherwise.

smithy-model/build.gradle.kts
```diff
@@ -17,21 +17,23 @@ description = "This module provides the core implementation of loading, validati
...
+    api(project(":smithy-utils"))
+    jmh(project(":smithy-utils"))
 }
 
 jmh {
     timeUnit = "us"
 }
 
-processJmhResources {
+tasks.processJmhResources {
```
hpmellema, line 37:
```suggestion
tasks { 
    processJmhResources {
```

smithy-protocol-tests/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+dependencies {
+    implementation(project(path= ":smithy-cli", configuration= "shadow"))
+    implementation(project(":smithy-protocol-test-traits"))
+    implementation(project(":smithy-protocol-traits"))
+    api(project(":smithy-validation-model"))
+}
+
+tasks.sourcesJar {
```
hpmellema, line 26:
```suggestion
tasks {
   sourcesJar {
```

smithy-syntax/build.gradle.kts
```diff
@@ -16,35 +16,37 @@
...
     dependencies {
-        exclude(project(':smithy-utils'))
-        exclude(project(':smithy-model'))
+        exclude(project(":smithy-utils"))
+        exclude(project(":smithy-model"))
     }
 }
 
-tasks['jar'].finalizedBy(tasks['shadowJar'])
+tasks.jar {
```
hpmellema, line 50: same as above. Just makes it faster to add more task configurations in the future. Could group with shadowjar configuration above in one `tasks` block.

smithy-trait-codegen/build.gradle.kts
```diff
@@ -0,0 +1,78 @@
...
+// Set up Integration testing source sets
+sourceSets {
+    create("it") {
+        compileClasspath += sourceSets.main.get().output +
+                configurations["testRuntimeClasspath"] +
+                configurations["testCompileClasspath"]
+
+        runtimeClasspath += output +
+                compileClasspath +
+                sourceSets.test.get().runtimeClasspath +
```
hpmellema, line 26: I would suggest just adding a block for: 
```
val main by getting
val test by getting
```
Then you can change ` sourceSets.test.get().runtimeClasspath` and such to `test.runtimeClasspath`

smithy-trait-codegen/build.gradle.kts
```diff
@@ -0,0 +1,78 @@
...
+                sourceSets.test.get().output
+
+        java {
+            srcDir("${layout.buildDirectory.get()}/integ/")
+        }
+
+        resources {
+            srcDirs(layout.buildDirectory.get().dir("generated-resources"))
```
hpmellema, line 34: nit: get can be after `dir()` call.

smithy-trait-codegen/build.gradle.kts
```diff
@@ -0,0 +1,78 @@
...
+    useJUnitPlatform()
+    testClassesDirs = sourceSets["it"].output.classesDirs
+    classpath = sourceSets["it"].runtimeClasspath
+}
+
+// Do not run checkstyle on generated trait classes
+tasks["checkstyleIt"].enabled = false
+
+// Force correct ordering so generated sources are available
+tasks["compileItJava"].dependsOn("generateTraits")
```
hpmellema, line 68: could probably move all of these task configs into a `tasks {}` block

smithy-validation-model/build.gradle.kts
```diff
@@ -14,22 +14,24 @@
...
 
-tasks["sourcesJar"].dependsOn("smithyJarStaging")
+tasks.sourcesJar {
+    dependsOn("smithyJarStaging")
+}
 
 smithy {
     smithyBuildConfigs.set(project.files())
-}
+}
```
hpmellema, line 37: nit: newlines.

hpmellema (changes requested):

smithy-aws-apigateway-traits/build.gradle.kts
yasmewad: I see, I found an interesting explanation here: 
https://stackoverflow.com/a/56306736

Seeing this it makes sense for us to use `extra`. 

Previously I used :

```kotlin
extra.apply{
   set("displayName", " some display name")
   set("moduleName", "some module name")
}

```

and it generated some weird directory named `property(fixed` in all subprojects.


I will use :

```kotlin
extra["displayName"] = "Smithy :: Waiters"
extra["moduleName"] = "software.amazon.smithy.waiters"

```

smithy-model/build.gradle.kts
yasmewad: That makes sense.

smithy-protocol-tests/build.gradle.kts
yasmewad: Will update.

smithy-syntax/build.gradle.kts
yasmewad: Ack.

smithy-trait-codegen/build.gradle.kts
yasmewad: What is the difference? Filtering first and then getting?

smithy-trait-codegen/build.gradle.kts
yasmewad: Ack.

smithy-validation-model/build.gradle.kts
yasmewad: ack.

build.gradle.kts
```diff
@@ -106,54 +101,57 @@ subprojects {
...
+                url = uri(stagingDirectory)
             }
         }
 
         publications {
-            mavenJava(MavenPublication) { publication ->
-                if (tasks.shadowJar.enabled) {
-                    project.shadow.component(publication)
+            create<MavenPublication>("mavenJava") {
+                if (tasks.findByName("shadowJar")?.enabled == true) {
```
hpmellema, line 143:
```suggestion
                if (tasks.findByName("shadowJar")?.enabled) {
```

settings.gradle.kts
```diff
@@ -0,0 +1,51 @@
...
+    "smithy-smoke-test-traits",
+    "smithy-syntax",
+    "smithy-aws-endpoints",
+    "smithy-aws-smoke-test-model",
+    "smithy-protocol-traits",
+    "smithy-protocol-tests",
+    "smithy-trait-codegen"
+)
+
+subProjects.forEach {
```
hpmellema, line 49: Just use `include(":subproject") for each subproject rather than this loop. This makes it a bit easier to add renames and nested stuff in the future. Plus the loop isnt exactly saving a lot of typing.

hpmellema (changes requested):

JordonPhillips:
We deliberately switched to groovy from kotlin, though I don't remember the context on why. @kstich did it iirc and he might remember.

build.gradle.kts
yasmewad: Hah Classic mistake!! :D I never thought I would make this but in these many files in hindsight I did. Will update.

settings.gradle.kts
yasmewad: This is anyway gonna go away in favor of buildSrc config for subprojects. So I am not in favor of changing this now. Next PR will have the changes for buildSrc plugins.

yasmewad: I will update the forEach

hpmellema (approved):

PR #2462, opened 2024-11-13
Add a tool to create and render stageable changelog entries

JordonPhillips:
Writing changelog entries is not super fun. Writing them after the fact, especially when you didn't write the original feature, is even less fun. And merge conflicts because of synchronizing a changelog file are SUPER not-fun.

This PR solves those problems. It introduces a changelog tool similar to what many of the SDK teams use (and indeed cribs some code from them).

## Workflow

* When creating a pull request, you create a changelog entry by running `.changes/new-change`. This creates a JSON file in `.changes/next-release/` containing the changelog entry and its metadata (e.g. whether it's a documentation change, feature change, etc.)
* When the pull request is opened a GitHub action will automatically add a PR comment with a suggestion that you can click on to add the PR number.
* During a release, the changelog is updated by running `.changes/render --release-version 1.2.3`. This compiles all pending changes in `.changes/next-release/` into a single file in the `.changes/releases/` directory. It then renders all of the releases in `.changes/releases/` into a markdown file that can be written to `CHANGELOG.md`. If for some reason the changelog needs to be re-rendered out of band, `./change/render` can be called without a version to just do the rendering.

In effect, the canonical source for the changelog becomes the changelog JSON files, and the actual markdown representation is just a derived artifact.

## Why?

Using distinct files for changelog entries solves a major problem: merge conflicts. These are horribly annoying to deal with during development, especially on a file that is essentially documentation and semantically has no reason to ever conflict. This problem is less bad for Smithy itself than it is for SDKs that have daily releases, but it's a problem for us too with our relatively large team.

## What's different?

This differs from what other SDKs do in a few ways. Firstly, and most notably, we aren't deriving a version number from these. We could! It would not be hard! But I elected not to do that in this PR since it's not as big a problem for us.

We also have a habit of linking the PRs for all of our changes in the changelog. This is actually kinda annoying because it creates a bit of a chicken-egg scenario. This is why I added the github action to automatically add it. If the action is annoying we can always get rid of it and just stop relying on the links. But that would be a shame.

Mechanically there's also a change in that I used modern python to write this, and updated any borrowed code to also be modern python.

## Why not use jmeslog?

jmeslog is a standalone tool derived from the python team's changelog tool (which started it all). It is essentially exactly what they do and want, but there's a few things that prevent us from using it. Notably, we need the current date and pr links. jmeslog doesn't have that and doesn't currently allow for easy extension in that way.

It also has things we don't necessarily want, like a category. That's really mostly used in the sdks to indicate changes made for a particular service. With hundreds of services, that's an issue worthy of calling out. But smithy itself doesn't need it.

But also, it's a dependency, and it has dependencies. I'd rather not have dependencies if I can avoid it. And what we need is ultimately not that complex. And doing it ourselves gives us some flexibility.

All that said, there's definitely a world where jmeslog is updated to a place where we can more comfortably use it.

## Why not derive changelogs from commits?

The audience for changelogs and the audience for commit messages are two separate (though sometimes intersecting) audiences. Commits may contain conext about technical junk that only mattters to people working on the code itself.

There is also not necessarily a 1-1 relationship between a changelog entry and a commit. Large features are often broken up into logical commits that make reviewing and code diving easier, but aren't important to the changelog. You might also have commits that make changes unimportant to the changelog audience, such as formatting changes or changes to CI configurations.

Commits are also, critically, <ins>***immutable***</ins>. Did you make a typo in your commit? Now it's preserved forever in your changelog. With JSON files, you can just change them. Those changes are now part of the commit record!

Also, on a more subjective level, have you seen repos that have semantic commits? They're ugly. Having more than half your commits starting with `chore` just makes it look like your project is in maintenance mode and/or like you hate your job. And if you're following best practices for commit messages, that's eating into your 50 character limit for the title.

## Any unfortunate bits?

There's no line wrapping, so you better believe the changelog is gonna have some long lines. It doesn't really matter though, because it'll all be natural when you're seeing it fully rendered. The line boundary thing these days mostly helps with readability when you're editing a file, but you will never manually edit the file.

Also the automated commit will hide the status checks in GitHub's UI for some reason. That's why it's only added *after* all the other checks and isn't run if said checks fail. Also why you can do it manually.

## What's next?

Provided this goes through, the next step will be to backfill all the current changelog entries. That'll certainly be a task.

A github action to create the version bump pr would be sweet. We'd need to add the ability to derive a version number from the feature list, which isn't hard. Then the script could just aslo perform the version bump by writing that one file. Then it could be just a button press away in github. Maybe as a fully automated release?

---

### DIFF
--- a/.changes/amend
+++ b/.changes/amend
@@ -1,0 +1,61 @@
+#! /usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from argparse import ArgumentParser
+
+from tool.amend import amend
+
+
+def main() -> None:
+    parser = ArgumentParser(
+        description="""\
+            Amend recently-introduced changelog entries to include a pull request \
+            link. This is intended to be run in GitHub as an action, but can be run \
+            manually by specifying parameters GitHub would otherwise provide in \
+            environment variables.
+            
+            This only checks entries that have been staged in the current branch, \
+            using git to get a list of newly introduced files. If the entry already \
+            has one or more associated pull requests, it is not amended.""",
+    )
+    parser.add_argument(
+        "-n",
+        "--pull-request-number",
+        required=True,
+        help="The numeric identifier for the pull request.",
+    )
+    parser.add_argument(
+        "-r",
+        "--repository",
+        help="""\
+            The name of the repository, defaulting to 'smithy-lang/smithy'. This is \
+            provided by GitHub in the GITHUB_REPOSITORY environment variable.""",
+    )
+    parser.add_argument(
+        "-b",
+        "--base",
+        help="""\
+            The name of the base branch to diff against, defaulting to 'main'. This \
+            is provided by GitHub in the GITHUB_BASE_REF environment variable.""",
+    )
+    parser.add_argument(
+        "-c",
+        "--review-comment",
+        action="store_true",
+        default=False,
+        help="""\
+            Instead of amending the local files on disk, post a review comment on the \
+            PR. This will also post a normal comment if no changelog entry is found.""",
+    )
+
+    args = parser.parse_args()
+    amend(
+        base=args.base,
+        repository=args.repository,
+        pr_number=args.pull_request_number,
+        review_comment=args.review_comment,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/.changes/new-change
+++ b/.changes/new-change
@@ -1,0 +1,47 @@
+#! /usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from argparse import ArgumentParser
+
+from tool import ChangeType
+from tool.new import new_change
+
+
+def main() -> None:
+    parser = ArgumentParser(
+        description="""\
+            Create a new changelog entry to be staged for the next release. Required \
+            values not provided on the command line will be prompted for.""",
+    )
+    parser.add_argument(
+        "-t",
+        "--type",
+        choices=[t.name.lower() for t in ChangeType],
+        help="The type of change being logged.",
+    )
+    parser.add_argument(
+        "-d", "--description", type=str, help="A concise description of the change."
+    )
+    parser.add_argument(
+        "-p",
+        "--pull-requests",
+        nargs="+",
+        help="The pull request that implements the change.",
+    )
+    parser.add_argument(
+        "-r",
+        "--repository",
+        help="The name of the repository, defaulting to 'smithy-lang/smithy'.",
+    )
+
+    args = parser.parse_args()
+    new_change(
+        change_type=ChangeType[args.type.upper()],
+        description=args.description,
+        pull_requests=args.pull_requests,
+        repository=args.repository,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/.changes/tool/render.py
+++ b/.changes/tool/render.py
@@ -1,0 +1,35 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from . import RELEASES_DIR, Change, Release
+
+TITLE = "Changelog"
+
+
+def render() -> None:
+    rendered = f"# {TITLE}\n\n"
+    for release in get_releases():
+        rendered += f"## {release.version} ({release.date})\n\n"
+
+        for change_type, changes in release.change_map().items():
+            rendered += f"### {change_type.section_title}\n\n"
+
+            for change in changes:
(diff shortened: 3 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -63,3 +67,25 @@ jobs:
...
+              ref: ${{ github.head_ref }}
+
+          - uses: actions/setup-python@v5
+            with:
+              python-version: "3.13"
+
+          - name: "Amend and push any staged changelogs"
+            run: |
+              git config user.name "github-actions[bot]"
+              git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
JordonPhillips, line 89: This is just the standard github actions bot. We could use something else if we want

.changes/amend
```diff
@@ -0,0 +1,67 @@
+#! /usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from argparse import ArgumentParser
+
+from tool.amend import amend
+
+
+def main() -> None:
```
milesziemer, line 9: Looking through our changelog, we have some instances where we've grouped multiple different PRs under the same changelog entry (like, "Fixed various issues with protocol tests"). We could do that because the changelog came only after all changes for the release were known. With this, changelog entries either come with each individual PR, or not at all (if you omit one for a PR). Given how it will now be certainly easier to just add a changelog entry for each PR and not worry about trying to combine them at the end, I don't mind this, but just something to note.

.changes/new-change
```diff
@@ -0,0 +1,49 @@
+#! /usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from argparse import ArgumentParser
+
+from tool import ChangeType
+from tool.new import new_change
+
+
+def main() -> None:
```
milesziemer, line 10: Just a suggestion - after trying this tool out a bit locally it does seem really simple to use, but a small readme in this .changes directory showing how to use this in a common workflow would be helpful when down the road the commit message in this PR isn't as readily available

.changes/tool/render.py
```diff
@@ -0,0 +1,33 @@
...
+def render() -> None:
+    rendered = f"# {TITLE}\n\n"
+    for release in get_releases():
+        rendered += f"## {release.version} ({release.date})\n\n"
+
+        for change_type, changes in release.change_map().items():
+            rendered += f"### {change_type.value}\n\n"
+
+            for change in changes:
+                rendered += render_change(change)
```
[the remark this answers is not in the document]
milesziemer: Oh, also, I think we want stable ordering of change types (I'm not sure how to do that in python)

milesziemer (changes requested):

.changes/next-release/documentation-b1f9dba71fb0e7538b4b526ba9a872489ce79be9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "documentation",
+  "description": "Added a tool to generate stageable, non-conflicting changelog entries and render a changelog based on staged entries.",
+  "pull_requests": [
+    "[#7](https://github.com/JordonPhillips/smithy/pull/7)"
```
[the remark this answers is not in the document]
JordonPhillips: I was testing the whole process out on my fork and forgot to take down that pr. I guess it finished first!

.changes/amend
JordonPhillips: You can still add them later, or amend them. This is why the associated pr list is a list

.changes/new-change
JordonPhillips: Ah yeah I should add this to the contributing file

JordonPhillips:
Since initially opening this PR I had to change the way the GitHub action worked. Previously when it detected an introduced changelog without an associated PR it would just amend it and push a commit. It turns out that this is a permissions nightmare on a PR from a fork that isn't easily resolveable as it would need permissions from each of the forks. A bit annoying, to say the least.

So I changed it to instead post a *review comment*. The genius of this being that  whoever opens the PR is doing the committing, so no permission issues! Except there were permission issues. Solveable this time though. The new problem was related to the old - when handling a PR from a fork, posting comments would fail. In this case, I lacked permissions for the main repo. I lacked those permissions because the maximum permissions level for a `pull_request` GitHub action triggered by a fork is `read`.

The workaround is to instead use `pull_request_target`, which runs the action from the context of the target (i.e. `main`) instead of the PR. With this, we can have those elevated permissions. The downside is that the action won't run on *this* PR because it doesn't exist on main yet. You can however see how it works in [this closed testing pr](https://github.com/smithy-lang/smithy/pull/2471).

I split the PR into two commits. One of those commits is just the action, so that it can be reverted if it ends up being a menace without having to revert the entire tool.

PR #2480, opened 2024-11-27
Reduce false positives of ShouldHaveUsedTimestampValidator

rchache:
#### Background
* Reduce false positives of ShouldHaveUsedTimestampValidator, when a member shape targets a list of Timestamps or structure of solely Timestamps
* Also allows timey members to target enums without emitting events
* Improves the CX by reducing unhelpful noise in smithy validator

#### Testing
* added new unit tests

#### Links

---

### DIFF
--- a/smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
+++ b/smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
@@ -31,7 +31,6 @@ import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.shapes.ShapeVisitor;
 import software.amazon.smithy.model.shapes.StructureShape;
-import software.amazon.smithy.model.shapes.TimestampShape;
 import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.validation.AbstractValidator;
 import software.amazon.smithy.model.validation.ValidationEvent;
@@ -114,7 +113,7 @@ public final class ShouldHaveUsedTimestampValidator extends AbstractValidator {
             @Override
             protected List<ValidationEvent> getDefault(Shape shape) {
                 if (shape.isStringShape() || shape instanceof NumberShape) {
-                    return validateSimpleShape(shape, patterns);
+                    return validateSimpleShape(shape);
                 } else {
                     return Collections.emptyList();
                 }
@@ -122,17 +121,12 @@ public final class ShouldHaveUsedTimestampValidator extends AbstractValidator {
 
             @Override
             public List<ValidationEvent> structureShape(StructureShape shape) {
-                return validateStructure(shape, model, patterns);
+                return validateStructure(shape, model);
             }
 
             @Override
             public List<ValidationEvent> unionShape(UnionShape shape) {
-                return validateUnion(shape, model, patterns);
-            }
-
-            @Override
-            public List<ValidationEvent> timestampShape(TimestampShape shape) {
-                return Collections.emptyList();
+                return validateUnion(shape, model);
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
```diff
@@ -127,6 +129,16 @@ public List<ValidationEvent> structureShape(StructureShape shape) {
             public List<ValidationEvent> unionShape(UnionShape shape) {
                 return validateUnion(shape, model, patterns);
             }
+
+            @Override
+            public List<ValidationEvent> timestampShape(TimestampShape shape) {
+                return Collections.emptyList();
```
hpmellema, line 135: Timestamp shapes are not instances of string or number so they would be covered by existing default.

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
```diff
@@ -160,31 +172,43 @@ private List<ValidationEvent> validateUnion(
...
             String name,
-            MemberShape target,
+            MemberShape memberShape,
             Model model,
             List<Pattern> patterns
     ) {
-        return OptionalUtils.stream(model.getShape(target.getTarget())
-                .flatMap(shape -> validateName(name, shape.getType(), target, patterns)));
+        return OptionalUtils.stream(model.getShape(memberShape.getTarget())
+                .flatMap(targetShape -> validateName(name, targetShape, memberShape, patterns, model)));
```
hpmellema, line 180: I dont think the `patterns` value needs to be passed in as the validateName method is an instance method and patterns is an instance field.

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
```diff
@@ -160,31 +172,43 @@ private List<ValidationEvent> validateUnion(
...
+            Shape targetShape,
             Shape context,
-            List<Pattern> patterns
+            List<Pattern> patterns,
+            Model model
     ) {
+        ShapeType type = targetShape.getType();
         if (type == ShapeType.TIMESTAMP) {
             return Optional.empty();
+        } else if (type == ShapeType.ENUM) {
```
hpmellema, line 206: nit: can be combine with previous

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
```diff
@@ -160,31 +172,43 @@ private List<ValidationEvent> validateUnion(
...
+            return Optional.empty();
+        } else if (type == ShapeType.STRUCTURE || type == ShapeType.LIST) {
```
hpmellema, line 208: Do we also need to check maps?

hpmellema (changes requested):

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
rchache: Do you know if the compiler optimizes it better this way vs combining them?

hpmellema: I'm not concerned with micro-optimizations here, more just conciseness.

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
hpmellema: Could we add that in this PR or do you feel that is too out of scope?

rchache: Well at least from what we've seen in AWS services, maps are not a cause of false positives the same way list members and structure members have been. I can add it if you want, but I don't think it would add much value tbh

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
rchache: ok

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
rchache: Ah ok, I'll remove it

smithy-linters/src/main/java/software/amazon/smithy/linters/ShouldHaveUsedTimestampValidator.java
hpmellema: Fair enough. Let's consider it out of scope for this change.

hpmellema (approved):

PR #2483, opened 2024-12-06
Add service id to the error message to aid debugging

sugmanue:
#### Background

When a resource is in the closure of more than one service the `TaggableResource` error won't say for which one it failed, e.g.,

```
[ERROR] example.weather#City: Resource does not have tagging CRUD operations and is not compatible with service-wide tagging operations. | TaggableResource 
```

This change add the service to better understand where the issue is.

```
[ERROR] example.weather#City: Resource does not have tagging CRUD operations and is not compatible with service-wide tagging operations for service `example.weather#Weather`.
```

While there I also made minor changes to improve the code around.

#### Testing

* Added tests for this case

#### Links

---

### DIFF
--- a/smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
+++ b/smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
@@ -16,8 +16,8 @@
 package software.amazon.smithy.aws.traits.tagging;
 
 import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -40,7 +40,7 @@ import software.amazon.smithy.model.validation.ValidationEvent;
 public final class TaggableResourceValidator extends AbstractValidator {
     @Override
     public List<ValidationEvent> validate(Model model) {
-        List<ValidationEvent> events = new LinkedList<>();
+        List<ValidationEvent> events = new ArrayList<>();
         TopDownIndex topDownIndex = TopDownIndex.of(model);
         AwsTagIndex tagIndex = AwsTagIndex.of(model);
         for (ServiceShape service : model.getServiceShapes()) {
@@ -72,7 +72,7 @@ public final class TaggableResourceValidator extends AbstractValidator {
             ServiceShape service,
             AwsTagIndex awsTagIndex
     ) {
-        List<ValidationEvent> events = new LinkedList<>();
+        List<ValidationEvent> events = new ArrayList<>();
         // Generate danger if resource has tag property in update API.
         if (awsTagIndex.isResourceTagOnUpdate(resource.getId())) {
             Shape operation = resource.getUpdate().isPresent()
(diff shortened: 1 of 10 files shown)


smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
```diff
@@ -40,13 +40,13 @@
...
     public List<ValidationEvent> validate(Model model) {
-        List<ValidationEvent> events = new LinkedList<>();
+        List<ValidationEvent> events = new ArrayList<>();
         TopDownIndex topDownIndex = TopDownIndex.of(model);
         AwsTagIndex tagIndex = AwsTagIndex.of(model);
         for (ServiceShape service : model.getServiceShapes()) {
             for (ResourceShape resource : topDownIndex.getContainedResources(service)) {
                 boolean resourceLikelyTaggable = false;
-                if (resource.hasTrait(TaggableTrait.class)) {
+                if (resource.hasTrait(TaggableTrait.class) && service.getResources().contains(resource.getId())) {
```
kstich, line 49: This isn't right. It's only checking `resource` shapes bound directly to the service, meaning this won't run on nested resources (service -> parent -> child - child won't be checked.)

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
```diff
@@ -40,13 +40,13 @@
...
                     events.addAll(validateResource(model, resource, service, tagIndex));
```
kstich, line 50: The error messages in `validateResource` that are related to the service's tags should include the ID of the service shape to aid in debugging.

kstich (changes requested):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
```diff
@@ -87,7 +87,6 @@ private List<ValidationEvent> validateResource(
         //    list or tag keys input or output member
         // 2. Tagging via APIs specified in the @taggable trait which are validated
         //    through the tag property, and must be resource instance operations
-        //Caution: avoid short circuiting behavior.
```
sugmanue, line 90: This comment is misleading, you only want to avoid short circuiting if you are expecting a side effect, that's not the case here.

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggableResourceValidator.java
```diff
@@ -96,9 +95,11 @@ private List<ValidationEvent> validateResource(
                     + " It must use the `aws.api@arn` trait."));
         }
 
-        if (!(isServiceWideTaggable || isInstanceOpTaggable)) {
-            events.add(error(resource, "Resource does not have tagging CRUD operations and is not compatible"
-                    + " with service-wide tagging operations."));
+        if (!isServiceWideTaggable && !isInstanceOpTaggable) {
```
sugmanue, line 98: The condition written this way expresses more clearly what we want.

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/tagging/TaggingShapeUtils.java
```diff
@@ -190,10 +190,10 @@ static boolean isTagPropertyInInput(
         Model model,
         ResourceShape resource
     ) {
-        PropertyBindingIndex propertyBindingIndex = PropertyBindingIndex.of(model);
-        Optional<String> property = resource.expectTrait(TaggableTrait.class).getProperty();
-        if (property.isPresent()) {
-            if (operationId.isPresent()) {
+        if (operationId.isPresent()) {
```
sugmanue, line 193: The only way for this code to branch to `true` is for when the optional `operationId` is present. I moved it up to avoid unneeded work when that's not true.

kstich (approved):

PR #2484, opened 2024-12-09
Migrate build logic to conventions plugins

haydenbaker:
#### Background
* Overhauls our builds to use convention plugins, which is a far cleaner approach
* Spotless and formatting will be added/applied in a follow-up PR (separate to keep this PR from being several pages)

#### Testing
* `./gradlew clean build publish -c -Plog-tests`
* tested jreleaser configs (for main and cli release)
* checked shadowed publications, verified shadowed deps are not in the pom

---


buildSrc/src/main/kotlin/Extensions.kt
```diff
@@ -0,0 +1,11 @@
+import org.gradle.api.Project
+import org.gradle.api.file.Directory
+import org.gradle.api.provider.Provider
+import org.gradle.kotlin.dsl.*
```
yasmewad, line 4: nit: add single line imports?

build.gradle.kts
```diff
@@ -13,296 +16,50 @@
...
-    `maven-publish`
-    signing
-    checkstyle
-    jacoco
-    id("com.github.spotbugs") version "6.0.8"
-    id("io.codearte.nexus-staging") version "0.30.0"
-    id("me.champeau.jmh") version "0.7.2"
-    id("com.github.johnrengelman.shadow") version "7.1.2"
-    id("org.jreleaser") version "1.12.0" apply false
+    id("org.jreleaser") version "1.15.0" apply false
```
hpmellema, line 20: could we add an alias here and move this into version library? Having everything in one place makes the dependabot changes tiny.

build.gradle.kts
```diff
@@ -13,296 +16,50 @@
...
-    }
-}
-
-// The root project doesn't produce a JAR.
-tasks.named("jar") {
-    enabled = false
-}
-
-// ==== Javadoc ====
-afterEvaluate {
```
hpmellema, line 253: I dont see any aggregate javadoc task anymore? Maybe I missed it, but we still need an aggregate javadoc task.

buildSrc/src/main/kotlin/smithy.module-conventions.gradle.kts
```diff
@@ -0,0 +1,5 @@
+plugins {
+    id("smithy.java-conventions")
+    id("smithy.integ-test-conventions")
```
hpmellema, line 3: I dont think all modules need integ test conventions. I would prefer to isolate this to the 2 instances that need this.

smithy-cli/build.gradle.kts
```diff
@@ -0,0 +1,367 @@
...
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+import org.jreleaser.gradle.plugin.JReleaserExtension
+import org.jreleaser.model.Active
+import org.jreleaser.model.Distribution
+import org.jreleaser.model.Stereotype
+
+plugins {
+    application
+    id("org.beryx.runtime") version "1.13.1"
```
hpmellema, line 16: as with jreleaser, can we move these into version lib?

hpmellema (changes requested):

hpmellema (approved):

PR #2488, opened 2024-12-16
Add smithy-docgen to the repository

haydenbaker:
#### Background
- Adds [smithy-docegen](https://github.com/smithy-lang/smithy-docgen) to the repository
- Previously, this was not easily possible with how the repository was structured and built
  - With #2484 and #2487, this is now (easily and cleanly) possible to do
- The core module was renamed from `smithy-docgen-core` to `smithy-docgen` to fit our conventional naming scheme
  - The actual module _structure_ was not changed, everything is still under `core` (up for debate whether to change)
- **Reviewers**: Please double-check that surface area is as small as possible (everything that can be package-protected or marked as unstable is)
  

#### Testing
- Migrated the `-tests` package to a couple of simple integration (sanity-check) tests
  - The actual example usage (as a project) should be moved to [smithy-examples](https://github.com/smithy-lang/smithy-examples)

---


smithy-docgen/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+plugins {
```
hpmellema, line 5: Could we also add a `README.md` matching the existing one ([here](https://github.com/smithy-lang/smithy-docgen))

smithy-docgen/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+description = "This module contains support for generating API documentation based on Smithy models."
+
+extra["displayName"] = "Smithy :: DocGen"
+extra["moduleName"] = "software.amazon.smithy.docgen"
+
+dependencies {
+    implementation(project(":smithy-build"))
+    implementation(project(":smithy-codegen-core"))
+    implementation(project(":smithy-linters"))
+    implementation(project(":smithy-model"))
```
hpmellema, line 19: Model and utils should be unnecessary here b/c they are already api deps of smithy-build
https://github.com/smithy-lang/smithy/blob/main/smithy-build/build.gradle.kts#L16C1-L17C34

Similarly, codegen-core defines `model` and `build` as api deps, so this can be simplified to just `smithy-codegen-core` and `smithy-linters`

smithy-docgen/src/main/java/software/amazon/smithy/docgen/core/DirectedDocGen.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.docgen.core;
```
hpmellema, line 6: The `core` suffix doesnt seem necessary to me. Could we just have this be `software.amazon.smithy.docgen`?

smithy-docgen/build.gradle.kts
haydenbaker: Yeah, good point, I meant to add that.

smithy-docgen/README.md
```diff
@@ -0,0 +1,197 @@
...
+First, create a gradle-based Smithy model project. This can be done easily with
+the Smithy CLI: `smithy init -o /tmp/smithy-docgen -t quickstart-gradle`.
+
+Note: the generator currently cannot be run with non-gradle-based projects.
+
+Next, publish the generator to Maven local by running
+`./gradlew :smithy-docgen:publishToMavenLocal`.
+
+In `build.gradle.kts`, add
+`implementation("software.amazon.smithy.docgen:smithy-docgen:+")` under
```
hpmellema, line 20: Is this still the right artifact id?

smithy-docgen/README.md
```diff
@@ -0,0 +1,197 @@
...
+`implementation`.
```
hpmellema, line 21:
```suggestion
`dependencies`.
```

hpmellema (approved):

PR #2526, opened 2025-02-18
Fix resource identifier binding validation

kstich:
This commit fixes a bug where only the names of identifiers were validated, not the ShapeId that represent them, when testing the binding of operations to resources.

---


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceIdentifierBindingValidator.java
```diff
@@ -227,13 +228,24 @@ private Stream<ValidationEvent> validateInstanceBindings(
...
             OperationShape operation,
             IdentifierBindingIndex bindingIndex
     ) {
-        return bindingIndex.getOperationInputBindings(resource, operation)
-                .keySet()
-                .containsAll(resource.getIdentifiers().keySet());
+        StructureShape inputShape = model.expectShape(operation.getInputShape(), StructureShape.class);
+        Map<String, String> bindings = bindingIndex.getOperationInputBindings(resource, operation);
+        for (Map.Entry<String, ShapeId> identifier : resource.getIdentifiers().entrySet()) {
+            if (!bindings.containsKey(identifier.getKey())) {
```
yefrig, line 239: I think in-lining these if's statements (into return) might be a bit more readable

yefrig (approved):

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceIdentifierBindingValidator.java
kstich:
```
if (!bindings.containsKey(identifier.getKey())
                    || !inputShape.getMember(bindings.get(identifier.getKey()))
                            .get()
                            .getTarget()
                            .equals(identifier.getValue())) {
                return false;
            }
```

This didn't look as readable to me as the split.

mcmasn-amzn:
I think this broke one of my models. Is there a way to model an API which accept a single string field which can be one of multiple ways to identify a resource? Example -- I have an API that is similar to [`lambda:GetFunction`](https://docs.aws.amazon.com/lambda/latest/api/API_GetFunction.html#API_GetFunction_RequestSyntax), which accepts have multiple name formats (full arn or name or partial arn), but only one is considered the true, full identifier in the resource model.

PR #2530, opened 2025-02-19
Return an empty collection instead of null

sugmanue:
#### Background

The method `get<MemberName>OrEmpty()` for collections, either lists or maps, was supposed to return an empty collection when the collection was not set, instead of null, but somehow we missed to add this expected behavior. This change fixes this, if the collection was not set we return either `Collections.emptyList()` or
`Collections.emptyMap()` depending on the collection type instead of `null`. The user can still find whether the value was set by using the `Optional<...> get<MemberName>()` variant that returns en empty optional when the collection was not set.

#### Testing

Validated the generated code. We don't currently have asserts for the generated code but only tests that asserts that everything compiles as expected. Asserting on the expected generated code would be good but can be added in a different change.

For the `structure` trait in the tests now the code generated look like this for the members `fieldD` and `fieldE`.

```java
    public Optional<List<String>> getFieldD() {
        return Optional.ofNullable(fieldD);
    }

    public List<String> getFieldDOrEmpty() {
        return fieldD == null ? Collections.emptyList() : fieldD;
    }

    public Optional<Map<String, String>> getFieldE() {
        return Optional.ofNullable(fieldE);
    }

    public Map<String, String> getFieldEOrEmpty() {
        return fieldE == null ? Collections.emptyMap() : fieldE;
    }
```

#### Links

---

### DIFF
--- a/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
+++ b/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
@@ -143,10 +143,10 @@ final class GetterGenerator implements Runnable {
                                 "}",
                                 symbolProvider.toSymbol(member),
                                 symbolProvider.toMemberName(member),
-                                () -> writer.write("return $1L == null ? $2T.empty$3L() : $1L;",
+                                () -> writer.write("return $1L == null ? $2T.$3L() : $1L;",
                                         symbolProvider.toMemberName(member),
                                         Collections.class,
-                                        isListShape ? "List" : "Map"));
+                                        isListShape ? "emptyList" : "emptyMap"));
                     }
                 } else {
                     writer.openBlock("public $T get$U() {",
--- a/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
+++ b/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
@@ -4,6 +4,7 @@
  */
 package software.amazon.smithy.traitcodegen.generators;
 
+import java.util.Collections;
 import java.util.Optional;
 import software.amazon.smithy.codegen.core.Symbol;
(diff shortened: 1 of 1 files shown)


smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
```diff
@@ -136,12 +137,16 @@ public Void structureShape(StructureShape shape) {
...
+                    if (isListShape || target.isMapShape()) {
                         writer.openBlock("public $T get$UOrEmpty() {",
                                 "}",
                                 symbolProvider.toSymbol(member),
                                 symbolProvider.toMemberName(member),
-                                () -> writer.write("return $L;", symbolProvider.toMemberName(member)));
+                                () -> writer.write("return $1L == null ? $2T.empty$3L() : $1L;",
+                                        symbolProvider.toMemberName(member),
+                                        Collections.class,
+                                        isListShape ? "List" : "Map"));
```
haydenbaker, line 149: Could you do just `Collections.emptyList()`/`Collections.emptyMap()` instead of having to interpolate? It would be a tad clearer, but no big deal.

haydenbaker (approved):

smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/generators/GetterGenerator.java
sugmanue: I guess we can, but we need to branch somewhere. What about spelling `emptyList` and `emptyMap` instead of just `List` and `Map`. That might be clearer.

sugmanue, line 149: @haydenbaker What about something like this?
```suggestion
                                () -> writer.write("return $1L == null ? $2T.$3L() : $1L;",
                                        symbolProvider.toMemberName(member),
                                        Collections.class,
                                        isListShape ? "emptyList" : "emptyMap"));
```

PR #2532, opened 2025-02-20
Don't make oneOf errors tagged unions

### LINKED ISSUE #2512: OpenApi onErrorStatusConflict: OneOf transforms errors into a tagged union
Setting openApi configuration parameter `onErrorStatusConflict` to `oneOf` transforms the error shapes under the same http status code to a tagged union set. However, server SDKs do not generate tagged unions for errors, leading to a mismatch between client and server. Additionally, Smithy Errors are not tagged unions. They are designed as an array of possible outputs, with the shape name providing the discriminating tag. I believe the problem occurs because the code generation relies on the UnionShape, which are a distinct concept from the errors array.

```smithy
namespace example

 use aws.protocols#restJson1
 use smithy.framework#ValidationException

 @restJson1
 service Example {
     version: "2006-03-01",
     operations: [Op],
     errors: [
         Error1,
         Error2
     ]
 }

@http(uri: "/op", method: "GET")
 operation Op {
     input: String,
     output: String
 }

 @httpError(418)
 structure Error1 {
     TeapotName: String
 }

 @httpError(418)
 structure Error2 {
     IsHot: Boolean
 }
```

```json
{
   "openapi": "3.0.2",
   "info": {
     "title": "Example",
     "version": "2006-03-01"
   },
   "paths": {
     "/op": {
       "get": {
         "operationId": "Op",
         "responses": {
           "200": {
             "description": "GetOp 200 response",
             "content": {
               "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/GetOpResponseContent"
                 }
               }
             }
           },
           "418": {
             "description": "GetOp418Error 418 response",
             "content": {
               "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/GetOp418ErrorResponseContent"
                 }
               }
             }
           }
         }
       }
     }
   },
   "components": {
     "schemas": {
       "Error1": {
         "type": "object",
         "properties": {
           "TeapotName": {
             "type": "string"
           }
         }
       },
       "Error2": {
         "type": "object",
         "properties": {
           "IsHot": {
             "type": "boolean"
           }
       },
       "GetOp418ErrorResponseContent": {
         "oneOf": [
           {
             "type": "object",
             "title": "Error1",
             "properties": {
               "Error1": {
                 "$ref": "#/components/schemas/Error1"
               }
             },
             "required": [
               "Error1"
             ]
           },
           {
             "type": "object",
             "title": "Error2",
             "properties": {
               "Error2": {
                 "$ref": "#/components/schemas/Error2"
               }
             },
             "required": [
               "Error2"
             ]
           }
         ]
       },
       "GetOpResponseContent": {
         "type": "string"
       }
     }
   }
 }
```

Issue introduced in #1995 . Creating a tagged union of errors is not necessary for OpenApi per [this comment](https://github.com/smithy-lang/smithy/issues/1649#issuecomment-1687014958).

The correct output shape for the 418 component is:

```json
"GetOp418ErrorResponseContent": {
   "oneOf": [
     {
         "$ref": "#/components/schemas/Error1"
     },
     {
         "$ref": "#/components/schemas/Error2"
     }
   ]
 }

```

JordonPhillips:
OpenAPI only lets you have one error per HTTP response code. This is a problem when converting things because Smithy has no such limitation. To address that, we added a toggle that lets you disambiguate by making the body contents a `oneOf`, which is an untagged union. The problem is that we made this a tagged union by leaning on how Smithy converts Smithy's union shape, which is tagged.

To illustrate the problem, the OpenAPI modeled body technically now declares the following body valid:

```json
{
    "Error1" {
        "foo": "bar"
    }
}
```

But the model for the structure is:

```
@error("client")
@httpError(400)
structure Error1 {
    foo: String
}
```

And an actual valid body would be:

```json
{
    "foo": "bar"
}
```

This commit turns the OpenAPI body into an untagged union instead, which is how it should be.

The question is: is this a backwards compatibility concern somehow? I think not, but maybe it should be a new flag?

Resolves #2512

---

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AwsRestJson1Protocol.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AwsRestJson1Protocol.java
@@ -6,6 +6,7 @@ package software.amazon.smithy.openapi.fromsmithy.protocols;
 
 import static software.amazon.smithy.openapi.OpenApiConfig.ErrorStatusConflictHandlingStrategy.ONE_OF;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
@@ -130,16 +131,15 @@ public final class AwsRestJson1Protocol extends AbstractRestProtocol<RestJson1Tr
         if (context.getConfig().getOnErrorStatusConflict() != null
                 && context.getConfig().getOnErrorStatusConflict().equals(ONE_OF)
                 && targetsSyntheticError(cleanedShape, context)) {
-            UnionShape.Builder asUnion = UnionShape.builder().id(cleanedShape.getId());
             UnionShape targetUnion = context.getModel()
                     .expectShape(
                             cleanedShape.getAllMembers().values().stream().findFirst().get().getTarget(),
                             UnionShape.class);
+            List<Schema> oneOfMembers = new ArrayList<>();
             for (MemberShape member : targetUnion.getAllMembers().values()) {
-                String name = member.getMemberName();
-                asUnion.addMember(member.toBuilder().id(cleanedShape.getId().withMember(name)).build());
+                oneOfMembers.add(context.createRef(member));
             }
-            return context.getJsonSchemaConverter().convertShape(asUnion.build()).getRootSchema();
+            return Schema.builder().type(null).oneOf(oneOfMembers).build();
         }
         return context.getJsonSchemaConverter().convertShape(cleanedShape).getRootSchema();
     }
--- a/smithy-openapi/src/test/resources/software/amazon/smithy/openapi/fromsmithy/protocols/error-code-collision-test-use-oneof.openapi.json
+++ b/smithy-openapi/src/test/resources/software/amazon/smithy/openapi/fromsmithy/protocols/error-code-collision-test-use-oneof.openapi.json
@@ -154,104 +154,32 @@
       "GetCurrentTime404ErrorResponseContent": {
         "oneOf": [
           {
-            "type": "object",
-            "title": "Error7",
-            "properties": {
-              "Error7": {
-                "$ref": "#/components/schemas/Error7"
-              }
-            },
-            "required": [
-              "Error7"
-            ]
+            "$ref": "#/components/schemas/Error7"
           },
           {
-            "type": "object",
-            "title": "Error8",
-            "properties": {
-              "Error8": {
-                "$ref": "#/components/schemas/Error8"
-              }
-            },
-            "required": [
-              "Error8"
-            ]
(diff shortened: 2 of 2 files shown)


yasmewad (commented):
Question: Could there be a case, where we can have one and the other i.e. a ref, and an object defined at the same time? 

Or this is the expected output to OpenAPI ? This is referring to the JSON that you've added for test.

JordonPhillips:
> Question: Could there be a case, where we can have one and the other i.e. a ref, and an object defined at the same time?

There's nothing stopping you from intermixing them in OpenAPI if that's what you're asking. You could have this for example:

```json
{
    "oneOf": [
        { "$ref": "#/components/schemas/Foo" },
        { "type": "string" }
    ]
}
```

In most cases we won't do that in our conversions though, because you effectively have to re-define shapes if you do.

In this case in particular we're referencing HTTP payload definitions. These aren't necessarily whole shapes though, like here:

```smithy
structure HttpRequest {
    @httpHeader("x-foo")
    foo: String
    
    bar: String
    baz: String
}
```

In this case the `bar` and `baz` members form the HTTP payload. In OpenAPI you can't intermix headers and payload members like this though. We *could* resolve that problem by inlining an anonymous shape. But that's not ideal because OpenAPI tooling can make a mess of those, so we instead create a synthetic shape that only has payload members that gets converted to a named OpenAPI schema. Thus we'll always have refs in cases like this, and never anonymous schemas.

yasmewad:
> > Question: Could there be a case, where we can have one and the other i.e. a ref, and an object defined at the same time?
> 
> There's nothing stopping you from intermixing them in OpenAPI if that's what you're asking. You could have this for example:
> 
> ```json
> {
>     "oneOf": [
>         { "$ref": "#/components/schemas/Foo" },
>         { "type": "string" }
>     ]
> }
> ```
> 
> In most cases we won't do that in our conversions though, because you effectively have to re-define shapes if you do.
> 
> In this case in particular we're referencing HTTP payload definitions. These aren't necessarily whole shapes though, like here:
> 
> ```smithy
> structure HttpRequest {
>     @httpHeader("x-foo")
>     foo: String
>     
>     bar: String
>     baz: String
> }
> ```
> 
> In this case the `bar` and `baz` members form the HTTP payload. In OpenAPI you can't intermix headers and payload members like this though. We _could_ resolve that problem by inlining an anonymous shape. But that's not ideal because OpenAPI tooling can make a mess of those, so we instead create a synthetic shape that only has payload members that gets converted to a named OpenAPI schema. Thus we'll always have refs in cases like this, and never anonymous schemas.

I see, thanks for this context. I was asking with the perspective that we needed that intermixed test-case or not. 

If this is just our conversion output, then I agree we won't need to have intermixed type output in OpenAPI.

yasmewad (approved):

PR #2539, opened 2025-02-26
Add primaryIdentifier to cfnResource trait

kstich:
This commit adds the `primaryIdentifier` field to the `@cfnResource` trait. This supports service teams that have CFN support that deviates from their APIs in their primary identifier, normally from a human readable value to an ARN. This field is `@deprecated` from the start, as it is not the preferred path for supporting resource identifiers.

The value of the `primaryIdentifier` field, when set, MUST be the name of a property defined on the resource shape and that property MUST be defined as a `string` or `enum` shape.

---

### DIFF
--- a/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourcePropertyValidator.java
+++ b/smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourcePropertyValidator.java
@@ -4,6 +4,8 @@
  */
 package software.amazon.smithy.aws.cloudformation.traits;
 
+import static java.lang.String.format;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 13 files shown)


smithy-aws-cloudformation-traits/src/main/java/software/amazon/smithy/aws/cloudformation/traits/CfnResourcePropertyValidator.java
```diff
@@ -44,17 +46,48 @@ private List<ValidationEvent> validateResource(
...
-                        validateResourceProperty(model, resource, resourceName, property).ifPresent(events::add);
-                    }
-                });
+        if (trait.getPrimaryIdentifier().isPresent()) {
+            validateResourcePrimaryIdentifier(model, resource, trait.getPrimaryIdentifier().get())
+                    .ifPresent(events::add);
+        }
+
+        Optional<CfnResource> cfnResourceOptional = cfnResourceIndex.getResource(resource);
+        if (cfnResourceOptional.isPresent()) {
```
JordonPhillips, line 55: Man I wish we could use something like Python's walrus assignment expression

JordonPhillips (changes requested):
Needs updates to the docs

Issue #2556, opened 2025-03-06
Lots of time of the OpenAPI conversion spent in `removeShapes`


kubukoz:
Hi! I had a brief look at profiling our code to see what takes the most time in the build process. One of the things we do in that process is building OpenAPI for a bunch of services. That's actually the longest part of the process, at the moment.

One thing stood out to me - there are two `removeShapes` calls in the call tree, which together take ~54% of the OpenAPI conversion - 324 seconds for the whole thing (we parallelize this by service, but that only gets us so far).

<img width="1418" alt="Image" src="https://github.com/user-attachments/assets/420555f3-62d1-4580-89fa-6a2540a1f9f8" />

I wonder if there are some optimization opportunities in there. Here's a breakdown of`removeShapes`:

<img width="1436" alt="Image" src="https://github.com/user-attachments/assets/74d08fc0-e0eb-40d3-8907-4058b986a786" />

I wonder if there's something obvious we could do about `CleanStructureAndUnionMembers`.

kubukoz:
Now that I think of it... as your model grows and you have more services, computing scoped models by removing shapes is a hopeless cause - it'll just get consistently slower as more shapes are added.

Shouldn't the `scopeModelToService` be changed to use a builder that only adds the shapes the service still refers to? That seems much easier.

PR #2561, opened 2025-03-11
Optimize transforms and validators

kstich:
This commit optimizes two ModelTransformPlugin implementations and one Validator plugin to use fewer intermediate objects and streams.

Helps with #2556 

---


yasmewad (approved):
LGTM reads simple too.

PR #2562, opened 2025-03-11
Prefer ID for hasTrait lookups

kstich:
This commit updates all calls to `hasTrait` from using Class based lookups to use ID based lookups. Traits are stored on a `Shape` in a `ShapeId`-keyed map. Using this will be faster than testing values via `isInstance`.

The lookup for `hasTrait(ShapeId)` is also updated to do a simple `containsKey` check on the map. This eliminates using `findTrait` which creates an intermediate `Optional` only to throw it away.

```
TraitLookups.hasTraitByClass       avgt    3  37.269 ± 2.546  us/op
TraitLookups.hasTraitByString      avgt    3  21.824 ± 1.066  us/op
TraitLookups.hasTraitByShapeIdOld  avgt    3  15.233 ± 5.002  us/op
TraitLookups.hasTraitByShapeId     avgt    3  11.611 ± 2.044  us/op
```

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpMethodSemanticsValidator.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpMethodSemanticsValidator.java
@@ -118,7 +118,7 @@ public final class HttpMethodSemanticsValidator extends AbstractValidator {
             events.add(warning(shape, trait, semantics.warningWhenModeled, method));
         }
 
-        boolean isReadonly = shape.getTrait(ReadonlyTrait.class).isPresent();
+        boolean isReadonly = shape.hasTrait(ReadonlyTrait.ID);
         if (semantics.isReadonly != null && semantics.isReadonly != isReadonly) {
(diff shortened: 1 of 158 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/HttpMethodSemanticsValidator.java
```diff
@@ -129,7 +129,7 @@ private List<ValidationEvent> validateOperation(
                     isReadonly ? UNNECESSARY_READONLY_TRAIT : MISSING_READONLY_TRAIT));
         }
 
-        boolean isIdempotent = shape.getTrait(IdempotentTrait.class).isPresent();
+        boolean isIdempotent = shape.hasTrait(IdempotentTrait.class);
```
sugmanue, line 132:
```suggestion
        boolean isIdempotent = shape.hasTrait(IdempotentTrait.ID);
```

sugmanue (approved):

PR #2573, opened 2025-03-30
Updating build script to remove some previous workarounds.

yasmewad:
#### Background
* What do these changes do? 
This pull request updates the build script for Smithy repository to remove a few previous workarounds set in place due to issues in Jreleaser. It also removes redundant tagging of our repository when releasing a new version. These changes were co-ordinated and discussed with @haydenbaker during 1.56.0 release process.

* Why are they important?
To ensure that future releases occur without the issues.

#### Testing
These changes were pending to be pushed and were required to successfully release version 1.56.0 of Smithy.

---

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,45 +53,43 @@ afterEvaluate {
     }
 }
 
-    apply(plugin = "org.jreleaser")
+configure<JReleaserExtension> {
+    dryrun = false
 
-    configure<JReleaserExtension> {
-        dryrun = false
-
-        release {
-            generic {
-                enabled = false
-            }
+    release {
+        generic {
+            skipTag = true
         }
+    }
 
-        // Used to announce a release to configured announcers.
-        // https://jreleaser.org/guide/latest/reference/announce/index.html
-        announce {
-            active = Active.NEVER
-        }
+    // Used to announce a release to configured announcers.
+    // https://jreleaser.org/guide/latest/reference/announce/index.html
+    announce {
+        active = Active.NEVER
+    }
 
-        // Signing configuration.
-        // https://jreleaser.org/guide/latest/reference/signing.html
-        signing {
-            active = Active.ALWAYS
-            armored = true
-        }
+    // Signing configuration.
+    // https://jreleaser.org/guide/latest/reference/signing.html
+    signing {
+        active = Active.ALWAYS
+        armored = true
+    }
 
-        // Configuration for deploying to Maven Central.
-        // https://jreleaser.org/guide/latest/examples/maven/maven-central.html#_gradle
-        deploy {
-            maven {
-                nexus2 {
-                    create("maven-central") {
-                        active = Active.ALWAYS
-                        url = "https://aws.oss.sonatype.org/service/local"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -58,21 +58,9 @@ afterEvaluate {
...
     configure<JReleaserExtension> {
         dryrun = false
 
-        // Used for creating and pushing the version tag, but this configuration ensures that
-        // an actual GitHub release isn't created (since the CLI release does that)
-        release {
-            github {
-                skipRelease = true
-                tagName = "{{projectVersion}}"
-            }
```
haydenbaker, line 73: We need to set `skipTag`. If you recall, it still pushed the tag even though we didn't configure an actual release.
```
            generic {
                skipTag = true
            }
```
or maybe we can even disable the generic (git) release entirely:
```
            generic {
                enabled = false
            }
```

yasmewad: Yes. Let me update that.

yasmewad: can you provide the snippet where this should be added? I looked at https://jreleaser.org/guide/latest/reference/release/generic.html but it keeps failing the build since I might be adding it at a wrong place.

haydenbaker:
```
       release {
            generic {
                enabled = false
            }
```

yasmewad: Thanks. Hayden.

build.gradle.kts
```diff
@@ -58,18 +58,12 @@ afterEvaluate {
 if (project.hasProperty("release.main")) {
```
haydenbaker, line 58: Can you actually remove this property check, and apply the plugin at the top with other plugins? This used to be a problem, but shouldn't be anymore since we removed the jreleaser config in `:smithy-cli`.

build.gradle.kts
```diff
@@ -53,23 +53,14 @@ afterEvaluate {
     }
 }
 
-// We're using JReleaser in the smithy-cli subproject, so we want to have a flag to control
-// which JReleaser configuration to use to prevent conflicts
-if (project.hasProperty("release.main")) {
     apply(plugin = "org.jreleaser")
```
haydenbaker, line 56: Don't need this. It's applied at the top.

haydenbaker: Fix indentation please.

build.gradle.kts
```diff
@@ -53,23 +53,14 @@ afterEvaluate {
...
         dryrun = false
 
-        // Used for creating and pushing the version tag, but this configuration ensures that
-        // an actual GitHub release isn't created (since the CLI release does that)
         release {
-            github {
-                skipRelease = true
-                tagName = "{{projectVersion}}"
+            generic {
+                enabled = false
```
haydenbaker, line 63:
```suggestion
                skipTag = true
```

If we disable it, we actually get:
```
* What went wrong:
Execution failed for task ':jreleaserConfig'.
> JReleaser has not been properly configured.
  No release provider has been configured
```

haydenbaker (approved):

PR #2575, opened 2025-04-01
add AWS_RECOMMENDED as partitional endpoint pattern

rchache:
#### Background
* The two existing patterns do not match what the AWS DNS team recommends to be used for partitional service endpoints
* This PR adds a new third type which will the be recommended flag going forward, and will match the prescribed patterns 1:1 for all partitions

#### Testing
* Added a sanity test

---

### DIFF
--- a/docs/source-2.0/aws/aws-endpoints-region.rst
+++ b/docs/source-2.0/aws/aws-endpoints-region.rst
@@ -269,9 +269,7 @@ Trait value
       - Description
     * - endpointPatternType
       - ``string``
-      - **Required** The pattern type to use for the partition endpoint.  This value can be set to ``service_dnsSuffix`` to
-        use the ``https://{service}.{dnsSuffix}`` pattern or ``service_region_dnsSuffix`` to use
-        ``https://{service}.{region}.{dnsSuffix}``.
+      - **Required** The pattern type to use for the partition endpoint.  This value should be set to ``aws_recommended`` in almost all cases.
     * - partitionEndpointSpecialCases
       - ``map`` of partition to `PartitionEndpointSpecialCase object`_
       - A map of partition to partition endpoint special cases - partitions that do not follow the
@@ -282,13 +280,9 @@ Conflicts with
     :ref:`aws.endpoints#standardRegionalEndpoints-trait`
 
 Partitional services (also known as "global" services) resolve a single endpoint per partition.
-That single endpoint is located in the partition's ``defaultGlobalRegion``. Partitional
-services should follow one of two standard patterns:
+That single endpoint is located in the partition's ``defaultGlobalRegion``.
 
-- ``service_dnsSuffix``: ``https://{service}.{dnsSuffix}``
-- ``service_region_dnsSuffix``: ``https://{service}.{region}.{dnsSuffix}``
-
-The following example defines a partitional service that uses ``{service}.{dnsSuffix}``:
+The following example defines a partitional service that uses AWS recommended patterns for each partition:
 
 .. code-block:: smithy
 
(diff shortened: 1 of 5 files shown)


smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/traits/EndpointPatternType.java
```diff
@@ -12,10 +12,15 @@
  */
 public enum EndpointPatternType {
     /** An endpoint with pattern `{service}.{dnsSuffix}`.*/
+    @Deprecated
```
alextwoods, line 15: We should maybe add a message + since here

docs/source-2.0/aws/aws-endpoints-region.rst
```diff
@@ -269,9 +269,7 @@ Trait value
       - Description
     * - endpointPatternType
       - ``string``
-      - **Required** The pattern type to use for the partition endpoint.  This value can be set to ``service_dnsSuffix`` to
-        use the ``https://{service}.{dnsSuffix}`` pattern or ``service_region_dnsSuffix`` to use
-        ``https://{service}.{region}.{dnsSuffix}``.
+      - **Required** The pattern type to use for the partition endpoint.  This value should be set to ``aws_recommended`` in almost all cases.
```
alextwoods, line 272: Should we still require this to be set?  We could default to using recommended

docs/source-2.0/aws/aws-endpoints-region.rst
```diff
@@ -282,13 +280,9 @@ Conflicts with
...
 Partitional services (also known as "global" services) resolve a single endpoint per partition.
-That single endpoint is located in the partition's ``defaultGlobalRegion``. Partitional
-services should follow one of two standard patterns:
+That single endpoint is located in the partition's ``defaultGlobalRegion``.
 
-- ``service_dnsSuffix``: ``https://{service}.{dnsSuffix}``
-- ``service_region_dnsSuffix``: ``https://{service}.{region}.{dnsSuffix}``
-
-The following example defines a partitional service that uses ``{service}.{dnsSuffix}``:
+The following example defines a partitional service that uses AWS recommended patterns for each partition:
```
alextwoods, line 285: Should we provide documentation on what those are?  At least for AWS partition.  We provide a list of the "standard patterns" eg here: https://smithy.io/2.0/aws/aws-endpoints-region.html#aws-endpoints-standardregionalendpoints-trait

rchache: It's fairly complex though, and currently only captured on an internal w.amazon page, which I don't think is fit for the public docs. If you think we should go the full mile and put the detailed pattern cases in the docs, we can do that though

docs/source-2.0/aws/aws-endpoints-region.rst
rchache: Yeah I don't have a strong preference. The only thing I had in mind was that it might lead to a small bug downstream in some code that expects it to always be nonnull if we miss changing it

docs/source-2.0/aws/aws-endpoints-region.rst
rchache: And just adding aws partition is a little misleading since aws partition is itself nonstandard

alextwoods: Yeah thats fair - I think its reasonable to leave it unspecified here.

docs/source-2.0/aws/aws-endpoints-region.rst
alextwoods: Yeah - I dont have a strong preference either.  Its more explicit to require it and the documentation here and on the trait itself makes it clear what to use.

alextwoods (approved):

docs/source-2.0/aws/aws-endpoints-region.rst
rchache: If anyone asks, we can presumably link them to the authoritative doc for clarity :)

PR #2577, opened 2025-04-07
Use tab sets to avoid synchronization of input/ouput tabs in Sphinx

yefrig:
#### Background
The current implementation in Smithy-docgen synchronizes all "Input" and "Output" tabs across operation examples, which creates a poor user experience when dealing with multiple examples in Sphinx.

It is not yet [possible](https://github.com/pradyunsg/sphinx-inline-tabs/issues/18) to avoid tab synchronization without changing the tab labels which in this case are fixed to "Input" and "Output" for all examples.

We can instead use [tab sets](https://sphinx-design.readthedocs.io/en/latest/tabs.html) supported in sphinx-design.

#### Testing
- I did an end-to-end doc generation run for a quick-start example and verified the tabs are no longer synchronized. 
- I also did an end-to-end doc generation run without the inline_tab dependency to verify it is no longer needed.

---

### DIFF
--- a/smithy-docgen/src/main/java/software/amazon/smithy/docgen/integrations/SphinxIntegration.java
+++ b/smithy-docgen/src/main/java/software/amazon/smithy/docgen/integrations/SphinxIntegration.java
@@ -98,7 +98,6 @@ public final class SphinxIntegration implements DocIntegration {
     private static final List<String> MARKDOWN_REQUIREMENTS = parseRequirements("requirements-markdown.txt");
 
     private static final List<String> BASE_EXTENSIONS = List.of(
-            "sphinx_inline_tabs",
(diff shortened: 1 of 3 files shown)


JordonPhillips (changes requested):
The motivating example isn't actually a problem. There's only an `Input` and `Output` tab right now to showcase that examples are tabbed. Both should be removed entirely when we have actual examples in languages being generated. Each tab then will be a language, and we definitely want those synchronized.

yefrig:
I think the use case for the language tabs makes sense. In the meantime, I would not be opposed to merging the change given that tab synchronization [was added](https://sphinx-design.readthedocs.io/en/latest/tabs.html#synchronised-tabs) to sphinx-design as well.

PR #2606, opened 2025-04-25
Add CI to build against api-models-aws

milesziemer:
Adds a job to the CI workflow that builds https://github.com/aws/api-models-aws using new Smithy changes.

api-models-aws is a pretty large test suite, so this should give us much more confidence that we aren't making breaking changes to a validator or something.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,3 +63,42 @@ jobs:
       with:
         name: built-docs
         path: docs/build/html
+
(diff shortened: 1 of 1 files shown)


haydenbaker (approved):

haydenbaker:
I wonder if we could pTML from the 17 build, and then cache that so build doesn't have to be done twice. Nevertheless, an optimization we can look at later.

PR #2618, opened 2025-05-08
Add static methods for more endpoint rules functions

mtdowling:
Adds static methods I can use in smithy-java for substring and isValidHostLabel. Improves isValidHostLabel performance.

---


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/Substring.java
```diff
@@ -97,30 +97,42 @@ public Value evaluate(List<Value> arguments) {
...
 
-            if (startIndex >= stopIndex || str.length() < stopIndex) {
-                return Value.emptyValue();
+    /**
+     * Get the substring of {@code value} using the same logic as the substring function.
+     *
+     * @param value Value to get the substring of.
+     * @param startIndex Start index.
+     * @param stopIndex Stop index.
+     * @param reverse True if the slice if from the end.
```
kstich, line 116:
```suggestion
     * @param reverse True if the slice is from the end.
```

kstich (approved):

PR #2622, opened 2025-05-09
Improve EndpointTests built-in validation

kstich:
This commit adds new validation for built-in parameters when used with the `endpointTests` trait.

1. Validates that built-in values with defaults are specified to the correct value if they are mismatched in `operationInput` cases.
2. Validates that built-in values without defaults are specified in `operationInput` cases.

Also fixes skipping DANGER/ERROR in endpoint test validation and adds a convenience method to the `Parameter` class.

---

### DIFF
--- a/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
+++ b/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.StructureShape;
@@ -17,6 +18,8 @@ import software.amazon.smithy.model.validation.AbstractValidator;
 import software.amazon.smithy.model.validation.NodeValidationVisitor;
 import software.amazon.smithy.model.validation.Severity;
 import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.rulesengine.language.EndpointRuleSet;
+import software.amazon.smithy.rulesengine.language.syntax.parameters.Parameter;
 import software.amazon.smithy.utils.SmithyUnstableApi;
 
 /**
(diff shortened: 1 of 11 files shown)


sugmanue (approved):

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -36,7 +39,28 @@ public List<ValidationEvent> validate(Model model) {
...
+                    }
+                }
+            }
+
             for (EndpointTestCase testCase : serviceShape.expectTrait(EndpointTestsTrait.class).getTestCases()) {
+                // If values for built-in parameters don't match the default, they MUST
+                // be specified in the operation inputs. Precompute the ones that don't match
+                // and capture their value.
+                Map<Parameter, Node> mismatchedBuiltInParams =
+                        getMismatchedBuiltInParams(builtInParamsWithDefaults, testCase);
```
alextwoods, line 62: "Mismatched" doesn't quite feel like the right description - to be it sounds like these are a invalid (esp in the context of a validator) - maybe something like `builtInsWithNonDefaultValues`

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -47,42 +71,102 @@ public List<ValidationEvent> validate(Model model) {
...
+            // value in this test case aren't the same.
+            if (testCase.getParams().containsMember(parameterName)
+                    && !testCase.getParams().expectMember(parameterName).equals(defaultValue)) {
+                mismatchedBuiltInParams.put(parameter, testCase.getParams().expectMember(parameterName));
+            }
+        }
+        return mismatchedBuiltInParams;
+    }
+
+    private void validateBuiltInValuesAreValid(
```
alextwoods, line 118: minor naming nit: `validateBuiltInValues` (validate values are valid seems a little redundant)

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -47,42 +71,102 @@ public List<ValidationEvent> validate(Model model) {
...
+                        String.format("Test case does not supply the `%s` value for the `%s` parameter's "
+                                + "`%s` built-in.",
+                                Node.printJson(mismatchedBuiltInParam.getValue()),
+                                mismatchedBuiltInParam.getKey().getNameString(),
+                                mismatchedBuiltInParam.getKey().getBuiltIn().get())));
+            }
+        }
+    }
+
+    private void validateNoDefaultBuiltInParamsHaveValues(
```
alextwoods, line 143: Sorry, another slight naming nit - I read this name as validating that built ins with defaults do not have values.  Maybe something like `validateBuiltInsWithoutDefaultsHaveValues`.  I think the logic below is correct, I just had to read it a few times to figure out what the method was supposed to be validating.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTraitValidator.java
```diff
@@ -92,6 +176,18 @@ private List<ValidationEvent> validateOperationInput(
                 .startingContext("The operationInput value for an endpoint test "
                         + "does not match the operation's input shape")
                 .build();
-        return inputShape.accept(validator);
+
+        // Error test cases may use invalid inputs as the mechanism to trigger their error,
```
alextwoods, line 180: I'm still not entirely convinced that this is valid for SDKs that implement operation input tests. If a test's inputs are invalid, it seems likely to trigger an error in the SDK outside of the endpoint resolver and would therefore not match the error specified by the test?

kstich: This isn't changing from before, it's just moving to clean up the `validate` method above.

alextwoods (approved):

PR #2629, opened 2025-05-12
Fix declarative endpoint traits toNode

alextwoods:
#### Background
Calling `toNode` on standardRegionalEndpoints or standardPartitionalEndpoints can raise NPE when special cases are missing non-required members.

This PR fixes that and adds equals/toHash methods as well.

#### Testing
* Modified existing tests to test that traits converted to node and back are equal.

---

### DIFF
--- a/smithy-aws-endpoints/src/test/java/software/amazon/smithy/rulesengine/aws/traits/StandardPartitionalEndpointsTraitTest.java
+++ b/smithy-aws-endpoints/src/test/java/software/amazon/smithy/rulesengine/aws/traits/StandardPartitionalEndpointsTraitTest.java
@@ -30,7 +30,7 @@ class StandardPartitionalEndpointsTraitTest {
         assertEquals(trait,
                 new StandardPartitionalEndpointsTrait.Provider()
                         .createTrait(StandardPartitionalEndpointsTrait.ID,
-                                trait.toBuilder().sourceLocation(trait.getSourceLocation()).build().toNode()));
+                                trait.toBuilder().build().toNode()));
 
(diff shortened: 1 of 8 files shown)


smithy-aws-endpoints/src/test/java/software/amazon/smithy/rulesengine/aws/traits/StandardPartitionalEndpointsTraitTest.java
```diff
@@ -27,6 +27,10 @@ public void loadsFromModel() {
 
         assertEquals(trait.getEndpointPatternType(), EndpointPatternType.SERVICE_DNSSUFFIX);
         assertEquals(trait.getPartitionEndpointSpecialCases().size(), 0);
+        assertEquals(trait,
+                new StandardPartitionalEndpointsTrait.Provider()
+                        .createTrait(StandardPartitionalEndpointsTrait.ID,
+                                trait.toBuilder().sourceLocation(trait.getSourceLocation()).build().toNode()));
```
kstich, line 33: Instead of all these `.sourceLocation` sets for both traits, the traits should probably set it in the `toBuilder` call themselves.

kstich (changes requested):

smithy-aws-endpoints/src/test/java/software/amazon/smithy/rulesengine/aws/traits/StandardPartitionalEndpointsTraitTest.java
alextwoods: done!

kstich (approved):

PR #2633, opened 2025-05-13
Improve isVirtualHostableBucket function

mtdowling:
This commit removes the use of three regular expressions and string splitting when checking if a value is a virtual hostable bucket. This should help with both startup and runtime performance. It also exposes a static method for performing this check that I'll use in smithy-java.

#### Background
* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/ParseUrl.java
+++ b/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/ParseUrl.java
@@ -174,10 +174,7 @@ public final class ParseUrl extends LibraryFunction {
                         || (ch2 < '0' || ch2 > '9')) {
                     return false;
                 }
-                int value = ((ch0 - '0') * 100) + ((ch1 - '0') * 10) + (ch2 - '0');
-                if (value > 255) {
-                    return false;
-                }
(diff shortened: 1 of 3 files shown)


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/ParseUrl.java
```diff
@@ -174,10 +174,6 @@ public static boolean isIpAddr(String host) {
                         || (ch2 < '0' || ch2 > '9')) {
                     return false;
                 }
-                int value = ((ch0 - '0') * 100) + ((ch1 - '0') * 10) + (ch2 - '0');
```
sugmanue, line 177: What about adding a comment here about the heuristic? Otherwise someone is going to think that we can do better and attempt to change it.

sugmanue (approved):

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/ParseUrl.java
mtdowling: Sure

sugmanue (approved):

PR #2641, opened 2025-05-15
Add a test for explicit payload member and empty body responses

sugmanue:
#### Background

Add a new protocol tests for the JSON REST protocol. When we have a response object with a member that's bound to the HTTP payload, such as

```smithy
@output
structure HttpPayloadWithStructureInputOutput {
    @httpPayload
    nested: NestedPayload,
}
```

And the response body is empty, then the member MUST be set to `null`.

Regression for this case was reported to the Java SDK V2 and fixed on [#6111](https://github.com/aws/aws-sdk-java-v2/pull/6111).

#### Testing

#### Links

---

### DIFF
--- a/smithy-aws-protocol-tests/model/restJson1/http-payload.smithy
+++ b/smithy-aws-protocol-tests/model/restJson1/http-payload.smithy
@@ -281,6 +281,20 @@ apply HttpPayloadWithStructure @httpResponseTests([
     }
 ])
 
+apply HttpPayloadWithStructure @httpResponseTests([
+    {
+        id: "RestJsonHttpPayloadWithStructureAndEmptyResponseBody",
+        documentation: "Serializes a structure in the payload",
+        protocol: restJson1,
+        code: 200,
+        body: "",
+        bodyMediaType: "application/json",
+        params: {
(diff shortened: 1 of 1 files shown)


smithy-aws-protocol-tests/model/restJson1/http-payload.smithy
```diff
@@ -281,6 +281,20 @@ apply HttpPayloadWithStructure @httpResponseTests([
...
 ])
 
+apply HttpPayloadWithStructure @httpResponseTests([
+    {
+        id: "RestJsonHttpPayloadWithStructureAndEmptyResponseBody",
+        documentation: "Serializes a structure in the payload",
+        protocol: restJson1,
+        code: 200,
+        body: "",
+        bodyMediaType: "application/json",
```
mtdowling, line 291: Not sure this would be set if there's no payload, or at least, you don't want to enforce it for servers do you?

sugmanue: Yeah, probably not, let me get rid of it.

mtdowling (approved):

PR #2656, opened 2025-06-06
Improve Smithy landing page and documentation page

joewyz:
Improvement list:
- Update "Ecosystem" button to "Awesome Smithy" and its link
- Add hover tooltip for client and server circles
- Add max width for heading and sub-heading blocks
- Move "Quickstart" button from top navigation bar to heading block and rename "Get Started"
- Update "Get Started" button's redirect  URL
- Set Smithy documentation sidebar logo to redirect to current Smithy landing page
- Modify correct URL for TypeScript section

### DIFF
--- a/docs/_templates/sidebar/brand.html
+++ b/docs/_templates/sidebar/brand.html
@@ -1,0 +1,8 @@
+<a class="sidebar-brand" href="/index.html">
+  {% if theme_light_logo %}
+  <div class="sidebar-brand-icon">
+    <img class="sidebar-logo only-light" src="{{ pathto('_static/' + theme_light_logo, 1) }}" alt="Logo">
+    <img class="sidebar-logo only-dark" src="{{ pathto('_static/' + theme_dark_logo, 1) }}" alt="Logo">
+  </div>
+  {% endif %}
+</a>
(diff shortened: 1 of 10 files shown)


docs/_templates/sidebar/brand.html
```diff
@@ -0,0 +1,8 @@
+<a class="sidebar-brand" href="/index.html">
+  {% if theme_light_logo %}
+  <div class="sidebar-brand-icon">
+    <img class="sidebar-logo only-light" src="{{ pathto('_static/' + theme_light_logo, 1) }}" alt="Logo">
+    <img class="sidebar-logo only-dark" src="{{ pathto('_static/' + theme_dark_logo, 1) }}" alt="Logo">
+  </div>
+  {% endif %}
```
sugmanue, line 7: I don't fully understand this, the test is for `theme_light_logo` but both logos, dark and light are included, and there's no else, so if `theme_light_logo` evals to false there's nothing includes. Shouldn't this be, **if** `theme_light_logo` include `theme_light_logo` **else** include `theme_dark_logo`?

joewyz: I think it is testing wether the choice of dark or light theme is allowed. If allowed, it will provide both dark and light image for the html so that it can choose which to use based on the current theme. Hope this makes sense.

sugmanue: Joe explained offline that this is just to know if changing the theme is supported. Which image is going to be used is toggled using css.

sugmanue (approved):

PR #2658, opened 2025-06-09
Fix false positive for mixed in member removal warnings

joewyz:
#### Background
Add additional check to prevent false positive warning messages being produced when the mixin shape itself is also to be removed. eg., the execution of `getModelWithoutTraitShapes`

---


sugmanue (approved):

JordonPhillips:
Could you add a test that covers this case?

PR #2666, opened 2025-06-13
Fix jsonName conversion for OpenAPI examples

### LINKED ISSUE #2659: When generating openapi spec, examples trait does not respect jsonName trait
When using the [converting Smithy to OpenAPI](https://smithy.io/2.0/guides/model-translations/converting-to-openapi.html#converting-smithy-to-openapi), generation of [examples](https://swagger.io/docs/specification/v3_0/adding-examples/) from the [`@examples`](https://smithy.io/2.0/spec/documentation-traits.html#examples-trait) documentation trait the [`@jsonName`](https://smithy.io/2.0/spec/protocol-traits.html#jsonname-trait) trait is not respected, leading to generation of examples which don't comply with the generated spec.

I adapted the `quickstart-cli` template project to showcase the problem: https://github.com/ingemaradahl/smithy-openapi-example/commit/dfada352c19bdcf40e96207985cb08f7639792ab. The [generated OpenAPI spec](https://gist.github.com/ingemaradahl/a9a16d978952205016806c1bbf81c035#file-weather-openapi-json-L212l) correctly uses the `jsonName` when generating the component schemas, but in the [example](https://gist.github.com/ingemaradahl/a9a16d978952205016806c1bbf81c035#file-weather-openapi-json-L95) uses the original smithy structure member name.

It is possible to use the same value as the `jsonName` trait argument in the example, but this leads to warnings issued when running `smithy build`. When incorporating in build tools, these warnings are sometimes treated as errors:
```
──  WARNING  ── ExamplesTrait.UnknownMember.example.weather#CityCoordinates.coordinate_system
Shape: example.weather#CreateCity
File:  models/weather.smithy:99:22

99|         coordinates: {
  |                      ^

Example input of Create a city.coordinates: Member coordinate_system does
not exist in example.weather#CityCoordinates
```

JordonPhillips:
When converting to OpenAPI for the restJson1 protocol, we convert member names if they have the jsonName trait. However, we were *not* converting those member names in examples. This updates example conversion to respect member name, if present and enabled.

Fixes #2659

---

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
@@ -109,11 +109,16 @@ abstract class AbstractRestProtocol<T extends Trait> implements OpenApiProtocol<
             MessageType messageType
     );
 
+    @Deprecated
+    Node transformSmithyValueToProtocolValue(Node value) {
+        return value;
+    };
+
     /**
      * Converts Smithy values in Node form to a data exchange format used by a protocol (e.g., XML).
      * Then returns the converted value as a long string (escaping where necessary).
      * If data exchange format is JSON (e.g., as in restJson1 protocol),
-     * method should return values without any modification.
+     * method should respect the jsonName trait, but otherwise not modify the node.
      *
      * <p> Used for the value property of OpenAPI example objects.
      * For protocols that do not use JSON as data-exchange format,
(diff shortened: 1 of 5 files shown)


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -109,11 +109,16 @@ abstract Schema createDocumentSchema(
             MessageType messageType
     );
 
+    @Deprecated
+    Node transformSmithyValueToProtocolValue(Node value) {
+        return value;
+    };
```
sugmanue: Extra semicolon 
```suggestion
    }
```

PR #2669, opened 2025-06-17
Node null check fix

lewisjkl:
#### Background
* What do these changes do? 
Updates such that the node validator will not raise a false-positive related to a Document value being null.
* Why are they important?
Currently causing an error to flag in one of our projects where a trait has a `Map` shape with a target type of `Document`.

#### Testing
See unit test included.

#### Links

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/knowledge/NullableIndex.java
@@ -179,8 +179,9 @@ public class NullableIndex implements KnowledgeIndex {
                 }
                 // fall-through.
             case LIST:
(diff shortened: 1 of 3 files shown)


kubukoz (approved):

lewisjkl:
I can understand that position, I think maybe the documentation for Document could be updated to remove mention that "null" is a valid document value. As far as I can tell, "null" for Documents is not treated any differently than it is for a String/Int or any other primitive. 

"Document types are represented by a JSON-like data model and can contain UTF-8 strings, arbitrary precision numbers, booleans, nulls, a list of these values, and a map of UTF-8 strings to these values."

That's the passage that made me infer document had a special handling of Null compared to other primitives (and the fact that this used to work, we've had this same test case for years now)

PR #2671, opened 2025-06-19
Trait codegen: less restrictive namespace check

kubukoz:
Allow namespaces like `smithy4s`, but not `smithy` or `smithy.api` or other sub-`smithy` namespaces.

#### Background
* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
+++ b/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
@@ -55,8 +55,8 @@ public final class TraitCodegenSettings {
     ) {
         this.packageName = Objects.requireNonNull(packageName);
         this.smithyNamespace = Objects.requireNonNull(smithyNamespace);
-        if (smithyNamespace.startsWith(SMITHY_API_NAMESPACE)) {
-            throw new IllegalArgumentException("The `smithy` namespace is reserved.");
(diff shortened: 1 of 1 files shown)


smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
```diff
@@ -55,8 +55,8 @@ public final class TraitCodegenSettings {
     ) {
         this.packageName = Objects.requireNonNull(packageName);
         this.smithyNamespace = Objects.requireNonNull(smithyNamespace);
-        if (smithyNamespace.startsWith(SMITHY_API_NAMESPACE)) {
-            throw new IllegalArgumentException("The `smithy` namespace is reserved.");
+        if (smithyNamespace.equals(SMITHY_API_NAMESPACE) || smithyNamespace.startsWith(SMITHY_API_NAMESPACE + ".")) {
```
kubukoz, line 58: actually, `SMITHY_API_NAMESPACE` is set to `smithy`. Shouldn't it just be `smithy.api`?

kubukoz:
Hi team, can we have some eyes on this?

sugmanue:
Yes, apologies for the delay. Let me circle back with the team about it.

smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
sugmanue: Let's keep it this way. The variable needs to be adjusted but we want this value.

sugmanue (approved):

PR #2674, opened 2025-06-23
Validate that query errors are unique

JordonPhillips:
This adds validation that ensures that no two errors in a service that have the awsQuery or awsQueryCompatible trait have the same code. This is necessary in these cases because query services may change the error code with a trait, and so do not benefit from Smithy's general restriction that every shape in the closure of a service must have a unique name.

There is at least one instance of this in the wild, they'll have to suppress it for this case.

---


smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/QueryErrorCodeValidator.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+            errorCodeBindings.computeIfAbsent(errorCode, k -> new TreeSet<>()).add(errorId);
+        }
+
+        for (Map.Entry<String, Set<ShapeId>> entry : errorCodeBindings.entrySet()) {
+            if (entry.getValue().size() == 1) {
+                continue;
+            }
+            String errorCode = entry.getKey();
+            String shapes = entry.getValue().stream().map(Objects::toString).collect(Collectors.joining(", "));
```
sugmanue, line 63: Nit, no need to create a stream here.
```suggestion
            String shapes = String.join(", ", entry.getValue());
```

SamRemis (approved):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/QueryErrorCodeValidator.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.stream.Collectors;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.ValidationEvent;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
```
sugmanue, line 23: Nit, javadoc is missing.
```suggestion

/**
 * Emits a warning if two or more error shapes share the same error code.
 */
```

sugmanue (approved):

smithy-aws-traits/src/main/java/software/amazon/smithy/aws/traits/protocols/QueryErrorCodeValidator.java
sugmanue: Oh, never mind, the values are shapes ids.

sugmanue (approved):

PR #2690, opened 2025-07-01
MQTT Topic Filter Support

bretambrose:
#### Background
Adds support (parsing, validation, conflict-checking) for traits to reference MQTT topic filters.  Topic filter support is required for the new request-response MQTT service clients being released by the IoT SDKs.

Fixes a bug where the topic parser incorrectly skips a trailing or preceding empty topic level.

#### Testing
* Tests have been added for topic filter parsing and conflict checking.
* Tests have been added to cover initial and trailing empty topic levels.

#### Links
* Further information about topics and topic filters can be found in MQTT specification docs: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901241

---

### DIFF
--- a/smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/Topic.java
+++ b/smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/Topic.java
@@ -21,12 +21,12 @@ import java.util.stream.Collectors;
 public final class Topic {
     private static final Pattern LABEL_PATTERN = Pattern.compile("^[a-zA-Z0-9_]+$");
 
-    private final TopicRules type;
+    private final TopicRule rule;
     private final String topic;
     private final List<Level> levels;
 
-    private Topic(TopicRules type, String topic, List<Level> levels) {
-        this.type = type;
+    private Topic(TopicRule rule, String topic, List<Level> levels) {
+        this.rule = rule;
         this.topic = topic;
         this.levels = Collections.unmodifiableList(levels);
     }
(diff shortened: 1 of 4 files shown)


smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/Topic.java
```diff
@@ -119,13 +160,38 @@ public boolean conflictsWith(Topic other) {
...
                 // conflict. One is more specific than the other.
+
+                // Note: I disagree with the above assertion and what it implies for the definition (which is not
+                // given anywhere) of "topic conflict."  My definition of "topic conflict" is "could potentially have
+                // a non-empty routing intersection."
+                //
+                // In particular, the above check can lead to a non-empty intersection if the label substitution
+                // yields a value that matches the non-label level's value.
+                //
+                // Despite this, I don't want to change the behavior of what currently exists.
```
sugmanue: Can you please move this note to an issue instead? Having it here is confusing and difficult to find to improve whatever the current situation is.

smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/Topic.java
```diff
@@ -21,32 +21,73 @@
 public final class Topic {
     private static final Pattern LABEL_PATTERN = Pattern.compile("^[a-zA-Z0-9_]+$");
 
+    private final TopicRule rule;
```
sugmanue: The rule is only used for parsing. This field is never used, let's remove it.

smithy-mqtt-traits/src/main/java/software/amazon/smithy/mqtt/traits/Topic.java
bretambrose: I removed the comment, but I'm not sure an issue is necessary.  I'm not advocating for anything here.

PR #2703, opened 2025-07-15
Add BDD-based rules engine trait

mtdowling:
This commit updates the smithy-rules-engine package to support binary decision diagrams (BDD) to more efficiently resolve endpoints.

We create the BDD by converting the decision tree into a control flow graph (CFG), then compile the CFG to a BDD. The CFG canonicalizes conditions for better sharing (e.g., sorts commutative functions, expands simple string templates, etc), and strips all conditions from results and hash-conses them as well. Later, we'll migrate to emitting the BDD directly in order to shave off many conditions and results that can be simplified.

Our decision-tree based rules engine requires deep branching logic to find results. When evaluating the path to a result based on given input, decision trees require descending into a branch, and if at any point a condition in the branch fails, you bail out and go back up to the next branch. This can cause pathological searches of a tree (e.g., 60+ repeated checks on things like isset and booleanEquals to resolve S3 endpoints). In fact, there are currently ~73,000 unique paths through the current decision tree for S3 rules.

Using a BDD (a fully reduced one at least) guarantees that we only evaluate any given condition at most once, and only when that condition actually discriminates the result. This is achieved by recursively converting the CFG into BDD nodes using ITE (if-then-else) operations, choosing a variable ordering that honors dependencies between conditions and variable bindings. The BDD builder applies Shannon expansion during ITE operations and uses hash-consing to share common subgraphs.

The "bdd" trait has most of the same information as the endpointRuleset trait, but doesn't include "rules". Instead it contains a base64 encoded "nodes" value that contains the zig-zag variable-length encoded node triples, one after the other (this is much more compact and efficient to decode than 1000+ JSON array nodes).

The BDD implementation uses CUDD-style complement edges where negative node references represent logical NOT, further reducing BDD size.

## BDD output examples

<details>
<summary>AWS Connect BDD output</summary>

```
Bdd{
  conditions (8):
    C0: isSet(Endpoint)
    C1: isSet(Region)
    C2: PartitionResult = aws.partition(Region)
    C3: booleanEquals(UseFIPS, true)
    C4: booleanEquals(UseDualStack, true)
    C5: booleanEquals(PartitionResult#supportsDualStack, true)
    C6: booleanEquals(PartitionResult#supportsFIPS, true)
    C7: stringEquals("aws-us-gov", PartitionResult#name)
  results (13):
    R0: NoMatchRule
    R1: Error: "Invalid Configuration: FIPS and custom endpoint are not supported"
    R2: Error: "Invalid Configuration: Dualstack and custom endpoint are not supported"
    R3: Endpoint: Endpoint
    R4: Endpoint: "https://connect-fips.{Region}.{PartitionResult#dualStackDnsSuffix}"
    R5: Error: "FIPS and DualStack are enabled, but this partition does not support one or both"
    R6: Endpoint: "https://connect.{Region}.amazonaws.com"
    R7: Endpoint: "https://connect-fips.{Region}.{PartitionResult#dnsSuffix}"
    R8: Error: "FIPS is enabled but this partition does not support FIPS"
    R9: Endpoint: "https://connect.{Region}.{PartitionResult#dualStackDnsSuffix}"
    R10: Error: "DualStack is enabled but this partition does not support DualStack"
    R11: Endpoint: "https://connect.{Region}.{PartitionResult#dnsSuffix}"
    R12: Error: "Invalid Configuration: Missing Region"
  root: 1
  nodes (26):
    0: terminal
    1: [C0, 21, 2]
    2: [C1, 3, 4]
    3: [C2, 5, 4]
    4: [R12, TRUE, FALSE]
    5: [C3, 11, 6]
    6: [C4, 8, 7]
    7: [R11, TRUE, FALSE]
    8: [C5, 10, 9]
    9: [R10, TRUE, FALSE]
    10: [R9, TRUE, FALSE]
    11: [C4, 17, 12]
    12: [C6, 14, 13]
    13: [R8, TRUE, FALSE]
    14: [C7, 16, 15]
    15: [R7, TRUE, FALSE]
    16: [R6, TRUE, FALSE]
    17: [C5, 18, 19]
    18: [C6, 20, 19]
    19: [R5, TRUE, FALSE]
    20: [R4, TRUE, FALSE]
    21: [C3, 25, 22]
    22: [C4, 24, 23]
    23: [R3, TRUE, FALSE]
    24: [R2, TRUE, FALSE]
    25: [R1, TRUE, FALSE]
}
```
</details>

<details>
<summary>bdd trait</summary>

```json
{
    "version": "1.3",
    "parameters": {
        "Region": {
            "builtIn": "AWS::Region",
            "required": false,
            "documentation": "The AWS region used to dispatch the request.",
            "type": "String"
        },
        "UseDualStack": {
            "builtIn": "AWS::UseDualStack",
            "required": true,
            "default": false,
            "documentation": "When true, use the dual-stack endpoint. If the configured endpoint does not support dual-stack, dispatching the request MAY return an error.",
            "type": "Boolean"
        },
        "UseFIPS": {
            "builtIn": "AWS::UseFIPS",
            "required": true,
            "default": false,
            "documentation": "When true, send this request to the FIPS-compliant regional endpoint. If the configured endpoint does not have a FIPS compliant endpoint, dispatching the request will return an error.",
            "type": "Boolean"
        },
        "Endpoint": {
            "builtIn": "SDK::Endpoint",
            "required": false,
            "documentation": "Override the endpoint used to send this request",
            "type": "String"
        }
    },
    "conditions": [
        {
            "fn": "isSet",
            "argv": [
                {
                    "ref": "Endpoint"
                }
            ]
        },
        {
            "fn": "isSet",
            "argv": [
                {
                    "ref": "Region"
                }
            ]
        },
        {
            "fn": "aws.partition",
            "argv": [
                {
                    "ref": "Region"
                }
            ],
            "assign": "PartitionResult"
        },
        {
            "fn": "booleanEquals",
            "argv": [
                {
                    "ref": "UseFIPS"
                },
                true
            ]
        },
        {
            "fn": "booleanEquals",
            "argv": [
                {
                    "ref": "UseDualStack"
                },
                true
            ]
        },
        {
            "fn": "booleanEquals",
            "argv": [
                {
                    "fn": "getAttr",
                    "argv": [
                        {
                            "ref": "PartitionResult"
                        },
                        "supportsDualStack"
                    ]
                },
                true
            ]
        },
        {
            "fn": "booleanEquals",
            "argv": [
                {
                    "fn": "getAttr",
                    "argv": [
                        {
                            "ref": "PartitionResult"
                        },
                        "supportsFIPS"
                    ]
                },
                true
            ]
        },
        {
            "fn": "stringEquals",
            "argv": [
                "aws-us-gov",
                {
                    "fn": "getAttr",
                    "argv": [
                        {
                            "ref": "PartitionResult"
                        },
                        "name"
                    ]
                }
            ]
        }
    ],
    "results": [
        {},
        {
            "error": "Invalid Configuration: FIPS and custom endpoint are not supported",
            "type": "error"
        },
        {
            "error": "Invalid Configuration: Dualstack and custom endpoint are not supported",
            "type": "error"
        },
        {
            "endpoint": {
                "url": {
                    "ref": "Endpoint"
                },
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "endpoint": {
                "url": "https://connect-fips.{Region}.{PartitionResult#dualStackDnsSuffix}",
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "error": "FIPS and DualStack are enabled, but this partition does not support one or both",
            "type": "error"
        },
        {
            "endpoint": {
                "url": "https://connect.{Region}.amazonaws.com",
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "endpoint": {
                "url": "https://connect-fips.{Region}.{PartitionResult#dnsSuffix}",
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "error": "FIPS is enabled but this partition does not support FIPS",
            "type": "error"
        },
        {
            "endpoint": {
                "url": "https://connect.{Region}.{PartitionResult#dualStackDnsSuffix}",
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "error": "DualStack is enabled but this partition does not support DualStack",
            "type": "error"
        },
        {
            "endpoint": {
                "url": "https://connect.{Region}.{PartitionResult#dnsSuffix}",
                "properties": {},
                "headers": {}
            },
            "type": "endpoint"
        },
        {
            "error": "Invalid Configuration: Missing Region",
            "type": "error"
        }
    ],
    "root": 2,
    "nodes": "AQIBACwGAggKBAwKKAIBBhgOCBIQJgIBChYUJAIBIgIBCCQaDB4cIAIBDiIgHgIBHAIBCiYoDCooGgIBGAIBBjQuCDIwFgIBFAIBEgIB"
}
```

</details>

## Endpoint rules: BDD vs Decision tree size comparison

**Regional service**

* BDD: Pretty=4.4 KB; Minified=2.8 KB
* Decision tree: Pretty=9.7 KB; Minified=3.7 KB

**S3**

* BDD: Pretty=67 KB; Minified=42 KB
* Decision tree: Pretty=427 KB; Minified=96 KB

---


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/FunctionDefinition.java
```diff
@@ -45,4 +45,16 @@ public interface FunctionDefinition {
...
+
+    /**
+     * Get the relative "cost" of the function as compared to the baseline of "isset" which equals 1.
+     *
+     * <p>If this function is considered more computationally expensive, then it has a value higher than 1. Otherwise,
+     * it has a value equal to 1. Defaults to "4" for unknown functions.
+     *
+     * @return the relative cost.
+     */
+    default int getCostHeuristic() {
```
JordonPhillips: nit: does this need to be an int? Looks like you don't define anything more than 10 or so.

edit: This intersects with other cost calculations that tragically *could* run into int range. Probably fine to keep it all int.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
```diff
@@ -0,0 +1,296 @@
...
+     * Appends a string representation to the given StringBuilder.
+     *
+     * @param sb the StringBuilder to append to
+     * @return the given string builder.
+     */
+    public StringBuilder toString(StringBuilder sb) {
+        sb.append("Bdd{\n");
+
+        // Conditions
+        sb.append("  conditions (").append(getConditionCount()).append("):\n");
```
JordonPhillips: nit: trailing colon when there are zero conditions. Similar applies to other sections if their lists can be empty.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
```diff
@@ -0,0 +1,296 @@
...
+        for (int i = 0; i < conditions.size(); i++) {
+            sb.append("    C").append(i).append(": ").append(conditions.get(i)).append("\n");
```
JordonPhillips: super nit: maybe align these in some way by padding before, after, or padding the index

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/cfg/Cfg.java
```diff
@@ -0,0 +1,261 @@
...
+        // Cache the result for this (rule, fallthrough) combination
+        context.putProcessed(rule, fallthrough, current);
+
+        return current;
+    }
+
+    // Context for tracking rule processing to handle shared rules.
+    private static final class RuleContext {
+        // Map from (Rule, fallthrough) -> CfgNode to handle shared rules
+        private final Map<RuleKey, CfgNode> processedRules = new HashMap<>();
```
JordonPhillips: Does this need to be here? It looks like just a thin wrapper over map

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/cfg/Cfg.java
```diff
@@ -0,0 +1,261 @@
...
+     *
+     * @return the lazily created and cached prepared condition data.
+     */
+    public ConditionData getConditionData() {
+        ConditionData result = data;
+        if (result == null) {
+            result = ConditionData.from(this);
+            data = result;
+        }
+        return result;
```
JordonPhillips:
```suggestion
        if (data == null) {
            data = ConditionData.from(this);
        }
        return data;
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
```diff
@@ -0,0 +1,296 @@
...
+        } else if (ref < 0) {
+            return "!" + (Math.abs(ref) - 1);
+        } else {
+            return String.valueOf(ref - 1);
+        }
+    }
+
+    public static Bdd fromNode(Node node) {
+        return BddNodeHelpers.fromNode(node);
+    }
```
JordonPhillips: Why not just have this be another overload for `from`?

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
```diff
@@ -7,6 +7,11 @@ namespace smithy.rules
 @trait(selector: "service")
 document endpointRuleSet
 
+/// Defines an endpoint rule-set using a binary decision diagram (BDD).
+@unstable
+@trait(selector: "service")
+document bdd
```
JordonPhillips: I would prefer that this be a full structure. I know that we used a document for endpoint ruleset, but I think that was a mistake. This is one of the more complicated traits, and I think the user experience will be much better if there are safety guards as early on in the process as possible. For example, if somebody mistypes `conditions` as `condtions` or similar then they'll get no error at build time and things will just silently not work. When parsing does fail, it'll be with unhelpful java `Node` errors instead of Smithy errors referencing shapes.

I know there was some desire to have the endpoint ruleset versioned separately from smithy, which this seems to inherit. While that use-case is one of the purposes of documents existing, I don't think it applies to us. We are the central source of information, the implementation lives here and cannot change without us updating.

I also notice that while there are tests for the BDD implementation, there aren't any tests for the trait.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddBuilder.java
```diff
@@ -0,0 +1,599 @@
...
+     * <p>This is the fundamental BDD operation from which all others are derived.
+     * Includes optimizations for special cases and complement edges.
+     *
+     * @param f the condition (must be boolean)
+     * @param g the "then" branch
+     * @param h the "else" branch
+     * @return the resulting BDD
+     * @throws IllegalArgumentException if f is a result terminal
+     */
+    public int ite(int f, int g, int h) {
```
JordonPhillips, line 326: Why `f, g, h` instead of `i, t, e` or something longer?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/DefaultOrderingStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+        List<Condition> queue = new ArrayList<>();
+        Collections.addAll(queue, conditions);
+
+        queue.sort(Comparator
+                // fewer deps first
+                .comparingInt((Condition c) -> deps.getDependencies(c).size())
+                // isSet() before everything else
+                .thenComparing(c -> !"isSet".equals(c.getFunction().getName()))
+                // variable-defining conditions first
+                .thenComparing(c -> infos.get(c).getReturnVariable() == null)
```
JordonPhillips: Shouldn't these be reversed?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddEquivalenceChecker.java
```diff
@@ -0,0 +1,345 @@
...
+import software.amazon.smithy.rulesengine.logic.cfg.Cfg;
+import software.amazon.smithy.rulesengine.logic.cfg.CfgNode;
+import software.amazon.smithy.rulesengine.logic.cfg.ConditionNode;
+import software.amazon.smithy.rulesengine.logic.cfg.ResultNode;
+
+/**
+ * Verifies functional equivalence between a CFG and its BDD representation.
+ *
+ * <p>This verifier uses structural equivalence checking to ensure that for every possible
+ * combination of condition evaluations, both representations produce the same result.
```
JordonPhillips: This makes it seem like it's always exhaustive, even though exhaustiveness is conditional upon the number of conditions and how long evaluation takes.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddNodeHelpers.java
```diff
@@ -0,0 +1,128 @@
...
+        return builder
+                .withMember("parameters", bdd.getParameters().toNode())
+                .withMember("conditions", Node.fromNodes(conditions))
+                .withMember("results", Node.fromNodes(results))
+                .withMember("root", bdd.getRootRef())
+                .withMember("nodes", encodeNodes(bdd))
+                .build();
+    }
+
+    static Bdd fromNode(Node node) {
```
JordonPhillips: This should assert that there's no unexpected members, perhaps after validating the version.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/OrderConstraints.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param conditions the conditions in their specific order
+     */
+    public OrderConstraints(ConditionDependencyGraph graph, List<Condition> conditions) {
+        int n = conditions.size();
+        if (n != graph.size()) {
+            throw new IllegalArgumentException(
+                    "Condition count (" + n + ") doesn't match dependency graph size (" + graph.size() + ")");
+        }
+
+        this.conditions = conditions.toArray(new Condition[0]);
```
JordonPhillips: It's a bit strange to me that they haven't optimized this so you don't need to take the un-intuitive step of passing in a zero sized array

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/OrderConstraints.java
```diff
@@ -0,0 +1,97 @@
...
+        for (int i = 0; i < n; i++) {
+            conditionToIndex.put(this.conditions[i], i);
+        }
+
+        // Build dependencies and compute valid positions in one pass
+        for (int i = 0; i < n; i++) {
+            maxValidPosition[i] = n - 1; // Initialize max position
+
+            Set<Condition> deps = graph.getDependencies(this.conditions[i]);
+            if (!deps.isEmpty()) {
```
JordonPhillips: This is unnecessary

JordonPhillips (changes requested):
When are we going to be running the BDD optimizations? I think it would make sense to do either prior to code generation, or better as a sort of pre-compile/formatting step. The latter would make sure it's only done once, but maybe a generator wouldn't want to trust that

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/cfg/Cfg.java
mtdowling: What I have is a lazy initialization idiom in Java that is slightly more performant and slight more thread safe (though tbf, I didn't use a volatile, but that's ok because we don't need to use identity comparisons on these and they're cheap enough to compute).

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
mtdowling: Yeah, I did this because I was being lazy and just copied the other trait :) I'll address it.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
mtdowling: fromNode makes it work with the NodeMapper stuff

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddBuilder.java
mtdowling: f, g, h, for whatever reason, is the typical way this is written. For example, check out page 12 of the CUDD guide: https://my.ece.utah.edu/~kalla/ECE6745/cudd.pdf#page19

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
mtdowling: eh, I think it's fine. It'll just be an empty section.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/OrderConstraints.java
mtdowling: agreed!

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddNodeHelpers.java
mtdowling: I'll do a warning instead to allow for new stuff to be added in the future

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/cfg/Cfg.java
mtdowling: no, it did more in previous code that got refactored. Will remove.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/DefaultOrderingStrategy.java
mtdowling: `!"isSet".equals(c.getFunction().getName())` -> returns false for isset functions, which makes them come earlier.
`getReturnVariable() == null` returns false for variable defining functions, which makes them come first.
I'll update the code to be more clear though.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/Bdd.java
mtdowling: okidae

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
mtdowling: Done. that was a doozy.

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
```diff
@@ -7,9 +7,203 @@ namespace smithy.rules
...
+    @required
+    @range(min: -1)
+    root: Integer
+
+    /// The number of nodes contained in the BDD.
+    @required
+    @range(min: 0)
+    nodeCount: Integer
+
+    /// Base64-encoded array of BDD nodes representing the decision graph structure.
```
JordonPhillips, line 40: I think the base64 encoding will make this trait difficult to write and review. How do you envision the development process for these?

mtdowling: The alternative is to embed thousands of numbers in arrays of arrays, which is just as unreadable and significantly more JSON to parse. The zig-zag encoding of the binary of the numbers gives a much more compact representation and lets consumers of the trait parse it directly into whatever data structure they want (e.g., in Java, we'd use int[][] instead of List for performance).

I don't envision people authoring BDDs by hand. They're going to typically generate them from something else. I will probably add some code in future PRs to make that easier too.

JordonPhillips: So the expectation is that people write an `endpointRuleset` and then use some transformer like `convertRulesetToBdd` that also filters the `endpointRuleset` trait? That should be fine.

We'd also want some other tooling to make it easy to work with, like being able to compile/optimize from the command line, e.g. `smithy rules optimize --timeout X --exhaustiveness X ...`. Back-porting the optimizations to the `endpointRuleset` trait would also be cool. We do have one in the CFG that we use while optimizing. And something to pretty-print the BDD.

All that can be done later though.

JordonPhillips (approved):

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
mtdowling: Yep. I plan to add an API that contributes paths to results to a CFG, and then combines them all into one big ITE chain that the BDD can turn into a compressed DAG representation.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/Parameter.java
```diff
@@ -272,7 +273,7 @@ public Node toNode() {
         if (documentation != null) {
             node.withMember(DOCUMENTATION, documentation);
         }
-        node.withMember(TYPE, type.toString());
+        node.withMember(TYPE, StringUtils.uncapitalize(type.toString()));
```
kstich, line 276: Is this value read elsewhere in a case sensitive manner where a value could mismatch?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/rule/Rule.java
```diff
@@ -150,11 +171,13 @@ public Type typeCheck(Scope<Type> scope) {
...
 
-        ArrayNode.Builder conditionsBuilder = ArrayNode.builder();
-        for (Condition condition : conditions) {
-            conditionsBuilder.withValue(condition.toNode());
+        if (!conditions.isEmpty()) {
+            ArrayNode.Builder conditionsBuilder = ArrayNode.builder();
+            for (Condition condition : conditions) {
+                conditionsBuilder.withValue(condition.toNode());
+            }
+            builder.withMember(CONDITIONS, conditionsBuilder.build());
```
kstich: Each of the three Rule implementations require serializing this value - stated as "Zero or more" in [the spec.](https://smithy.io/2.0/additional-specs/rules-engine/specification.html#smithy-rules-endpointruleset-trait)

mtdowling: I added it back, but we really should remove this requirement at some point. It's useless for BDD.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/parameters/Parameter.java
mtdowling: The current approach is unable to read its own serialized JSON back. This fixes it.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddBuilder.java
```diff
@@ -0,0 +1,663 @@
...
+
+    // Terminal constants
+    private static final int TRUE_REF = 1;
+    private static final int FALSE_REF = -1;
+
+    // ITE operation cache for memoization
+    private final Map<TripleKey, Integer> iteCache;
+
+    // Node storage: three separate arrays
+    private int[] variables = new int[1024];
```
kstich: Is 1024 based on existing rulesets? I imagine we have a significant amount of smaller rulesets and a few very large ones that need a lot of growth.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddCompiler.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        int result;
+        if (cfgNode == null) {
+            result = bddBuilder.makeResult(noMatchIndex);
+
+        } else if (cfgNode instanceof ResultNode) {
+            Rule rule = ((ResultNode) cfgNode).getResult();
+            result = bddBuilder.makeResult(getOrCreateResultIndex(rule));
+
```
kstich:
```suggestion
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddCompiler.java
```diff
@@ -0,0 +1,140 @@
...
+    private int convertCfgToBdd(CfgNode cfgNode) {
+        Integer cached = nodeCache.get(cfgNode);
+        if (cached != null) {
+            return cached;
+        }
+
+        int result;
+        if (cfgNode == null) {
+            result = bddBuilder.makeResult(noMatchIndex);
+
```
kstich:
```suggestion
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddEquivalenceChecker.java
```diff
@@ -0,0 +1,372 @@
...
+import software.amazon.smithy.rulesengine.logic.cfg.CfgNode;
+import software.amazon.smithy.rulesengine.logic.cfg.ConditionNode;
+import software.amazon.smithy.rulesengine.logic.cfg.ResultNode;
+
+/**
+ * Verifies functional equivalence between a CFG and its BDD representation.
+ *
+ * <p>This verifier uses structural equivalence checking to ensure that both representations produce the same result.
+ * When the BDD has fewer than 20 conditions, the checking is exhaustive. When there are more, random samples are
+ * checked up the earlier of max samples being reached, or the max duration being reached.
```
kstich:
```suggestion
 * checked up to the earlier of max samples being reached or the max duration being reached.
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddEquivalenceChecker.java
```diff
@@ -0,0 +1,372 @@
...
+            verifyCase(allTrue ^ (1L << i));
+        }
+
+        // Alternating patterns
+        if (!hasEitherLimitBeenExceeded()) {
+            verifyCase(0x5555555555555555L & ((1L << bdd.getConditionCount()) - 1));
+        }
+
+        if (!hasEitherLimitBeenExceeded()) {
+            verifyCase(0xAAAAAAAAAAAAAAAAL & ((1L << bdd.getConditionCount()) - 1));
```
kstich: What are these magic bitwise values?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddTrait.java
```diff
@@ -0,0 +1,357 @@
...
+
+    @Override
+    protected Node createNode() {
+        ObjectNode.Builder builder = ObjectNode.builder();
+        builder.withMember("parameters", parameters.toNode());
+
+        List<Node> conditionNodes = new ArrayList<>();
+        for (Condition c : conditions) {
+            conditionNodes.add(c.toNode());
+        }
```
kstich: Why not build an `ArrayNode` directly? Goes for `results` as well.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddBuilder.java
mtdowling: I'll do 256. This grows to like 30K for S3.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddEquivalenceChecker.java
mtdowling: Hah, I should have a better comment. These are just alternating 0101, and 1010. Added a comment.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/BddTrait.java
mtdowling: fixed, good call

kstich (approved):

docs/source-2.0/additional-specs/rules-engine/specification.rst
```diff
@@ -74,6 +74,184 @@ or :ref:`error rules, <rules-engine-endpoint-rule-set-error-rule>` with an
...
+``smithy.rules#endpointBdd`` trait
+----------------------------------
+
+.. warning:: Experimental
+
+    This trait is experimental and subject to change.
+
+Summary
+    Defines a `Binary Decision Diagram (BDD) <https://en.wikipedia.org/wiki/Binary_decision_diagram>`_ representation
+    of endpoint rules for efficient runtime evaluation.
```
alextwoods: May want to add something about compact storage/representation as well here (I know its mentioned later, but I think worth highlighting both benefits up front).

docs/source-2.0/additional-specs/rules-engine/specification.rst
```diff
@@ -45,8 +46,7 @@ The content of the ``endpointRuleSet`` document has the following properties:
       - Description
     * - version
       - ``string``
-      - **Required**. The rule set schema version. This specification covers
-        version 1.0 of the endpoint rule set.
+      - **Required**. The rules engine version (e.g., 1.0).
```
alextwoods, line 57: Do we have documentation anywhere of what specifically is supported in each version and what backwards/forwards compatibility considerations are?

docs/source-2.0/additional-specs/rules-engine/specification.rst
```diff
@@ -74,6 +74,184 @@ or :ref:`error rules, <rules-engine-endpoint-rule-set-error-rule>` with an
...
+    ``structure``
+
+The ``endpointBdd`` trait provides a BDD representation of endpoint rules, optimizing runtime evaluation by
+eliminating redundant condition evaluations and reducing the decision tree to a minimal directed acyclic graph.
+This trait is an alternative to ``endpointRuleSet`` that trades compile-time complexity for significantly improved
+runtime performance and reduced artifact sizes.
+
+.. note::
+
+   The ``endpointBdd`` trait can be generated from an ``endpointRuleSet`` trait through compilation. Services may
```
alextwoods, line 111: Since we can go from rules -> bdd, but (as far as I know) not the other way from bdd -> rules, I might make an argument that we (for now at least) prefer endpointRules trait since consumers can use a simple projection to remove endpoint rules or compile them to bdd.

docs/source-2.0/additional-specs/rules-engine/standard-library.rst
```diff
@@ -38,6 +38,53 @@ parameter is equal to the value ``false``:
     }
 
 
+.. _rules-engine-standard-library-coalesce:
+
+``coalesce`` function
+=====================
+
+Summary
+    Evaluates arguments in order and returns the first non-empty result, otherwise returns the result of the last
```
alextwoods, line 47: We should make sure that we have test cases for implementors that cover coalesce with boolean types to ensure that handling of False values is consistent.

docs/source-2.0/additional-specs/rules-engine/standard-library.rst
```diff
@@ -38,6 +38,53 @@ parameter is equal to the value ``false``:
...
+Return type
+    * ``coalesce(T, T, ...)`` → ``T``
+    * ``coalesce(option<T>, T, ...)`` → ``T`` (if any argument is non-optional)
+    * ``coalesce(T, option<T>, ...)`` → ``T`` (if any argument is non-optional)
+    * ``coalesce(option<T>, option<T>, ...)`` → ``option<T>`` (if all arguments are optional)
+Since
+    1.1
+
+The ``coalesce`` function provides null-safe chaining by evaluating arguments in order and returning the first
+non-empty result. If all arguments evaluate to empty, it returns the result of the last argument. This is
```
alextwoods: > If all arguments evaluate to empty, it returns the result of the last argument
I found this a little confusing - the last argument here would be empty right?

docs/source-2.0/additional-specs/rules-engine/specification.rst
alextwoods: Another thought on this - we have been enforcing 100% test coverage (on conditions, endpoints and errors) - how does this translate to BDD - I assume we could still measure the coverage of test cases against the BDD (as edges traversed?) and my intuition is that a ruleset with 100% coverage would compile to a BDD with 100% coverage.

mtdowling: We should eventually stop shipping the tree in the models because they cause cold start issues with SDKs that use models dynamically (e.g., botocore). That's what the guidance is steering us toward here.

As for coverage, we run the same modeled test cases against the BDD too. We don't do coverage analysis there, but I can see if that's possible.

docs/source-2.0/additional-specs/rules-engine/standard-library.rst
mtdowling: Where would such a test be added?

docs/source-2.0/additional-specs/rules-engine/standard-library.rst
mtdowling: Yeah. If the last argument is empty it returns empty, if it's not empty, it returns a not-empty value. When evaluating this condition, the implementation doesn't need to inspect the result of the last condition and can just return it (empty or not). I made a small clarification.

docs/source-2.0/additional-specs/rules-engine/specification.rst
mtdowling: I added a simple section about the version number.

docs/source-2.0/additional-specs/rules-engine/standard-library.rst
alextwoods: I think we've generally added them under the test rule files, eg: [parse-url.smithy](https://github.com/smithy-lang/smithy/blob/main/smithy-rules-engine/src/test/resources/software/amazon/smithy/rulesengine/language/errorfiles/valid/parse-url.smithy) as well as in internal documentation.

docs/source-2.0/additional-specs/rules-engine/specification.rst
mtdowling: I just pushed up a BDD coverage checker, similar to the ruleset checker. When run with S3 and its test cases, we get 100% coverage (yay)

alextwoods (approved):

docs/source-2.0/additional-specs/rules-engine/specification.rst
```diff
@@ -74,6 +82,184 @@ or :ref:`error rules, <rules-engine-endpoint-rule-set-error-rule>` with an
...
+----------------------------------
+``smithy.rules#endpointBdd`` trait
+----------------------------------
+
+.. warning:: Experimental
+
+    This trait is experimental and subject to change.
+
+Summary
+    A Binary Decision Diagram (BDD) <https://en.wikipedia.org/wiki/Binary_decision_diagram>_ representation of endpoint
```
JordonPhillips:
```suggestion
    A `Binary Decision Diagram (BDD) <https://en.wikipedia.org/wiki/Binary_decision_diagram>`_ representation of endpoint
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/Coalesce.java
```diff
@@ -0,0 +1,154 @@
...
+import software.amazon.smithy.rulesengine.language.evaluation.value.Value;
+import software.amazon.smithy.rulesengine.language.syntax.ToExpression;
+import software.amazon.smithy.rulesengine.language.syntax.expressions.Expression;
+import software.amazon.smithy.rulesengine.language.syntax.expressions.ExpressionVisitor;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+/**
+ * A coalesce function that returns the first non-empty value.
+ *
+ * <p>This variadic function requires two or more arguments. At runtime, returns the first arguments that returns a
```
kstich:
```suggestion
 * <p>This variadic function requires two or more arguments. At runtime, returns the first argument that contains a
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/Coalesce.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Creates a {@link Coalesce} function from a list of expressions.
+     *
+     * @param args the expressions to coalesce
+     * @return The resulting {@link Coalesce} function.
+     */
+    public static Coalesce ofExpressions(List<? extends ToExpression> args) {
+        return ofExpressions(args.toArray(new ToExpression[0]));
```
kstich, line 70:
```suggestion
        return ofExpressions(args.toArray(new ToExpression[args.size()]));
```

From the `List` JavaDocs:

>If the list fits in the specified array, it is returned therein. Otherwise, a new array is allocated with the runtime type of the specified array and the size of this list.

This reads as if this array is going to be thrown away if the list has any elements, but used if we allocate the right size to start.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/EndpointBddTrait.java
```diff
@@ -0,0 +1,367 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.rulesengine.logic.bdd;
```
kstich:
```suggestion
package software.amazon.smithy.rulesengine.traits;
```

This probably necessitates a lot of visibility changes, which means `@SmithyInternalApi` annotations, too.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/SiftingOptimization.java
```diff
@@ -0,0 +1,592 @@
...
+    private final int mediumMinNodes;
+    private final int mediumMaxPasses;
+    private final int granularMaxNodes;
+    private final int granularMaxPasses;
+
+    // Internal effort levels for the tiered optimization stages.
+    private enum OptimizationEffort {
+        COARSE(11, 4, 0, 20, 4_000, 6),
+        MEDIUM(2, 20, 6, 20, 1_000, 6),
+        GRANULAR(1, 50, 12, 20, 8_000, 12);
```
kstich, line 58: The `exhaustiveThreshold` value doesn't change, should it?

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/SiftingOptimization.java
```diff
@@ -0,0 +1,592 @@
...
+                (1.0 - (double) state.bestSize / state.initialSize) * 100,
+                totalTimeInSeconds));
+
+        return trait.toBuilder().conditions(state.orderView).results(state.results).bdd(state.bestBdd).build();
+    }
+
+    private OptimizationState initializeOptimization(EndpointBddTrait trait) {
+        // Use the trait's existing ordering as the starting point
+        List<Condition> initialOrder = new ArrayList<>(trait.getConditions());
+        Condition[] order = initialOrder.toArray(new Condition[0]);
```
kstich, line 142: Similar comment about the array sizing as before. Though this also seems to be creating an extra `List` that may not be necessary.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/cfg/SsaTransform.java
```diff
@@ -0,0 +1,277 @@
...
+    private SsaTransform(VariableAnalysis variableAnalysis) {
+        scopeStack.push(new HashMap<>());
+        this.variableAnalysis = variableAnalysis;
+
+        this.referenceRewriter = new TreeRewriter(
+                ref -> {
+                    String originalName = ref.getName().toString();
+                    String uniqueName = resolveReference(originalName);
+                    return Expression.getReference(Identifier.of(uniqueName));
+                },
```
kstich: Make this non-anonymous.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/BddTraitValidator.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        // Validate nodes
+        int nodeCount = bdd.getNodeCount();
+
+        for (int i = 0; i < nodeCount; i++) {
+            // Skip terminal node at index 0
+            if (i == 0) {
+                continue;
+            }
```
kstich:
```suggestion
        // Skip terminal node at index 0
        for (int i = 1; i < nodeCount; i++) {
```

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
```diff
@@ -7,9 +7,196 @@ namespace smithy.rules
...
+    /// - high: Reference to follow when the condition evaluates to true
+    /// - low: Reference to follow when the condition evaluates to false
+    ///
+    /// Reference Encoding:
+    /// - 0: Invalid/unused reference (never appears in valid BDDs)
+    /// - 1: TRUE terminal (treated as "no match" in endpoint resolution)
+    /// - -1: FALSE terminal (treated as "no match" in endpoint resolution)
+    /// - 2, 3, 4, ...: Node references pointing to nodes[ref-1]
+    /// - -2, -3, -4, ...: Complement node references (logical NOT of nodes[abs(ref)-1])
+    /// - 2000000+: Result terminals (2000000 + resultIndex)
```
kstich:
```suggestion
    /// - 100000000+: Result terminals (100000000 + resultIndex)
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/Coalesce.java
JordonPhillips: This is one of those weird unintuitive optimizations. IIRC it's because constructing an array with known elements is faster than constructing a sized array and then filling it. The JVM can even optimize away the extra allocation so there's no downside except readability. JDK 11+ has a [more readable version](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#toArray(java.util.function.IntFunction)), but tragically we're still stuck on 8.

JordonPhillips (changes requested):

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/SiftingOptimization.java
mtdowling: It's all sort of heuristics and mushy as to what is optimal for any given BDD. These settings all worked well for S3 and simple regional services, so I'll keep it as-is. The other settings changed though, which did influence the output in my testing.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/BddTraitValidator.java
mtdowling: doh

smithy-rules-engine/src/main/resources/META-INF/smithy/smithy.rules.smithy
mtdowling: Very nice catch

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/logic/bdd/EndpointBddTrait.java
mtdowling: Moved the trait

PR #2705, opened 2025-07-17
Add additional properties to OpenApiConfig

msosnicki:
#### Background
* The `JsonSchemaMapper` and `OpenApiMapper` have access to `OpenApiConfig`. In some cases (see related [smithy4s PR](github.com/disneystreaming/alloy/pull/224)), there is a need to pass additional information to such mappers. This PR adds `Map<String, String> additionalProperties` field to enable that

---

### DIFF
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
@@ -98,6 +98,7 @@ public class OpenApiConfig extends JsonSchemaConfig {
     private boolean syncCorsPreflightIntegration = false;
     private ErrorStatusConflictHandlingStrategy onErrorStatusConflict;
     private OpenApiVersion version = OpenApiVersion.VERSION_3_0_2;
+    private Map<String, String> additionalProperties = Collections.emptyMap();
 
     public OpenApiConfig() {
         super();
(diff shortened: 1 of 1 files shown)


smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
```diff
@@ -375,6 +376,18 @@ public void setOnErrorStatusConflict(ErrorStatusConflictHandlingStrategy onError
...
 
+    public Map<String, String> getAdditionalProperties() {
+        return additionalProperties;
+    }
+
+    /**
+     * Sets additional non-standard config properties that might be used by other components
+     * @param additionalProperties Map of additional properties
+     */
+    public void setAdditionalProperties(Map<String, String> additionalProperties) {
```
kubukoz: **suggestion:** shall we have a method to just `put` into the properties, without having to pass the entire map?

i.e. `void putAdditionalProperty(String key, String value)`

kubukoz: also, looks like `ModelAssembler` allows `Object` values so maybe we should be as flexible here as well. `String` is fairly limiting, after all.

msosnicki: Yeah, makes sense, changed so it resembles how it's done in `ModelAssembler`

kubukoz (approved):

kstich:
Is the behavior you're looking for already filled by the `JsonSchemaConfig.getExtensions(YourCustomConfig.class)` method? That will populate additional fields into the `YourCustomConfig` class from on content that was unknown to the underlying `JsonSchemaConfig` when loaded.

PR #2709, opened 2025-07-22
Add a reserved namespaces deny list to restrict trait-codegen from using smithy specific namespaces.

yasmewad:
#### Background
• **What do these changes do?** 
 Adds validation to prevent trait code generation on  `smithy` keyword itself or reserved Smithy namespaces  ( `smithy.api`, `smithy.waiters`, `smithy.test`) and their sub-namespaces but we still want to allow namespaces like `smithy.foo` or `smithy.rules` . 

This change adds a RESERVED_NAMESPACES set to act as a deny list for future extensibility.

For instance: 

`smithy` -> not allowed
`smithy.api` , `smithy.waiters` , and `smithy.test` -> not allowed
`smithy.foo`, `smithyfoo.bar`, -> allowed

• **Why are they important?**
 Prevents namespace conflicts with core Smithy functionality and protects framework integrity by blocking user code generation on reserved namespaces.

#### Testing
• **How did you test these changes?**
 Added comprehensive unit tests covering reserved namespace validation, case-insensitive matching, edge cases, and legitimate namespace acceptance.

#### Links
• No related issues

---

### DIFF
--- a/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
+++ b/smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
@@ -6,8 +6,11 @@ package software.amazon.smithy.traitcodegen;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
+import java.util.Set;
 import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.utils.SetUtils;
 import software.amazon.smithy.utils.SmithyUnstableApi;
 
 /**
@@ -29,6 +32,15 @@ import software.amazon.smithy.utils.SmithyUnstableApi;
  */
 @SmithyUnstableApi
 public final class TraitCodegenSettings {
+    private static final String SMITHY_KEYWORD = "smithy";
+    private static final Set<String> RESERVED_NAMESPACES = SetUtils.of("smithy.api",
+            "smithy.framework",
+            "smithy.mqtt",
+            "smithy.openapi",
+            "smithy.protocols",
+            "smithy.rules",
+            "smithy.test",
+            "smithy.waiters");
 
     private final String packageName;
     private final String smithyNamespace;
@@ -53,6 +65,9 @@ public final class TraitCodegenSettings {
             List<String> excludeTags
     ) {
         this.packageName = Objects.requireNonNull(packageName);
+        if (isReservedNamespace(smithyNamespace)) {
+            throw new IllegalArgumentException("The `smithy` namespace and its sub-namespaces are reserved.");
+        }
         this.smithyNamespace = Objects.requireNonNull(smithyNamespace);
         this.headerLines = Objects.requireNonNull(headerLines);
         this.excludeTags = Objects.requireNonNull(excludeTags);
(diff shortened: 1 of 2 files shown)


smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
```diff
@@ -29,6 +32,8 @@
  */
 @SmithyUnstableApi
 public final class TraitCodegenSettings {
+    private static final String SMITHY_KEYWORD = "smithy";
+    private static final Set<String> RESERVED_NAMESPACES = SetUtils.of("smithy.api", "smithy.waiters", "smithy.test");
```
sugmanue: You are missing some, the full list as of now is:

```
smithy.api
smithy.framework
smithy.mqtt
smithy.openapi
smithy.protocols
smithy.rules
smithy.test
smithy.waiters
```

kstich: This needs several other sub-namespaces: `.mqtt`, `.openapi`, `.protocols`, `.rules`, `.framework`

yasmewad: Ack. Are these from just the github search? Or some other mechanism you used?

yasmewad: Will update.

sugmanue: A shell script:

```
find . -name \*.smithy | xargs egrep '^namespace ' | grep ' smithy.' | cut -d: -f2  | grep -v ' test' | grep -v example | sort | uniq | cut -d ' ' -f2 | cut -d '.' -f1,2 | uniq
```

yasmewad: Nice Thanks!

yasmewad: I have updated the namespace list.

yasmewad: I have updated the namespace list.

smithy-trait-codegen/src/main/java/software/amazon/smithy/traitcodegen/TraitCodegenSettings.java
```diff
@@ -110,4 +125,26 @@ public List<String> headerLines() {
...
+
+    /**
+     * Checks if a namespace is reserved. A namespace is considered reserved if it is
+     * "smithy", one of the predefined reserved namespaces or if it starts with one of the reserved namespaces followed by a dot.
+     *
+     * @param namespace the namespace to check
+     * @return true if the namespace is reserved, false otherwise
+     */
+    private static boolean isReservedNamespace(String namespace) {
+        String namespaceLowerCase = namespace.toLowerCase(Locale.ROOT);
```
sugmanue, line 137: Let's remove this conversion, namespaces are case sensitive.

sugmanue: Confirmed offline that this needs to be case sensitive.

yasmewad: Yeah, while namespaces are case-sensitive, the rationale of this was anyone should be forbidden from creating a trait with `SMITHY` or `SMITHY.API` or other versions of it as it conflicts with our reserved namespaces. 

I am leaning towards blocking those case-insensitively.  @mtdowling and I discussed, the reasoning for case-insensitive checking is that shapes are case-insensitive.

FWIW, for future record putting this here from our [docs](https://smithy.io/2.0/spec/model.html) :

> While shape ID references within the semantic model are case-sensitive, no two shapes in the semantic model can have the same case-insensitive shape ID. This restriction makes it easier to use Smithy models for code generation in programming languages that do not support case-sensitive identifiers or that perform some kind of normalization on generated identifiers (for example, a Python code generator might convert all member names to lower snake case). To illustrate, com.Foo#baz and com.foo#BAZ are not allowed in the same semantic model. This restriction also extends to member names: com.foo#Baz$bar and com.foo#Baz$BAR are in conflict.

sugmanue (approved):

yefrig (approved):

PR #2742, opened 2025-08-14
Use ordered sets for shape id sequences

### LINKED ISSUE #2735: Unintuitive behavior of ModelSerializer
## Description
`ModelSerializer` currently sorts map key–value pairs during serialization. This makes round-tripping fail: 

```
deserialize(serialize(model)) != model
``` 

While producing stable output can be useful, it is surprising for a serializer to mutate the logical structure of a model. Model sanitization/normalization should be an explicit, opt-in step, not part of serialization.

## Why this is a problem?
- breaks the usual serializer invariant: `deserialize(serialize(x)) == x`
- makes it impossible to directly compare:
   a) A model loaded from disk (`Model.assembler(...)`)
   b) A model still in memory from generation
    ..without serializing & deserializing the latter just to normalize it, which is redundant and costly.

## Reproduction

In our particular case, we observe `ModelSerializer` changing the order of operation errors:

```scala
//> using scala 3.7.1
//> using dep software.amazon.smithy:smithy-model:1.60.3
//> using option -Wunused:all
import software.amazon.smithy.model.shapes.ModelSerializer
import software.amazon.smithy.model.knowledge.TopDownIndex
import software.amazon.smithy.model.shapes.ServiceShape

val m = Model
  .assembler()
  .addUnparsedModel(
    "test.smithy",
    """$version: "2"
      |namespace demo
      |
      |service DummyService {
      |    version: "0.0"
      |    operations: [
      |        Op1
      |    ]
      |}
      |
      |
      |operation Op1 {
      | errors: [Error1, Error3, Error2]
      |}
      |@error("server")
      |@httpError(500)
      |structure Error1 {}
      |
      |@error("server")
      |@httpError(500)
      |structure Error2 {}
      |
      |@error("server")
      |@httpError(500)
      |structure Error3 {}
      |""".stripMargin,
  )
  .assemble()
  .unwrap()


val serialized = ModelSerializer.builder().build().serialize(m)

val m2 = Model.assembler().addDocumentNode(serialized).assemble().unwrap()

println(m == m2)
println(m.expectShape(ShapeId.from("demo#Op1")).asOperationShape().get().getErrors)
println(m2.expectShape(ShapeId.from("demo#Op1")).asOperationShape().get().getErrors)
```

gives:
```
false
[demo#Error1, demo#Error3, demo#Error2]
[demo#Error1, demo#Error2, demo#Error3]
```

## Suggested fix

Make the sorting behavior opt-in:
```java
ModelSerializer.serialize(ModelSorter.sort(model))
```

JordonPhillips:
Several shapes have properties that are sequences of shape ids, such as an OperationShape's errors. Previously these shapes were using the List structure to store these in Java, but since the order of these sequences isn't actually semantically important, it was causing equality checks to fail where the should not.

This replaces those usages of `List` to `LinkedHashSet`. That way the order in the model may still be preserved, but the equality is set equality. Additionally, duplicates in these sequences have no meaning, so the deduplication of sets is a benefit in that sense too.

Fixes #2735

Note that we already use ordered sets for the lists of operations and resources on service and resource shapes

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
@@ -141,7 +141,7 @@ public final class OperationShape extends Shape implements ToSmithyBuilder<Opera
     }
 
     /**
-     * <p>Gets a list of the error shape IDs bound directly to the operation
+     * <p>Gets a set of the error shape IDs bound directly to the operation
      * that can be encountered.
      *
      * <p>This DOES NOT include errors that are common to a service. Operations


smithy-model/src/main/java/software/amazon/smithy/model/shapes/OperationShape.java
```diff
@@ -134,6 +135,11 @@ public ShapeId getOutputShape() {
         return output;
     }
 
+    @Deprecated
+    public List<ShapeId> getErrors() {
+        return ListUtils.copyOf(errors);
+    }
+
     /**
      * <p>Gets a list of the error shape IDs bound directly to the operation
```
yefrig:
```suggestion
     * <p>Gets a set of the error shape IDs bound directly to the operation
```

yefrig (approved):

PR #2743, opened 2025-08-15
Consistently use BuilderRefs for collections in builders

### LINKED ISSUE #2734: @cors trait causes non-deterministic OpenAPI output key ordering
### Description

When using the [openapi](https://smithy.io/2.0/guides/model-translations/converting-to-openapi.html) plugin, we’ve noticed that applying the [@cors trait](https://smithy.io/2.0/spec/http-bindings.html#cors-trait) results in non-deterministic ordering of keys in the generated OpenAPI JSON output.

Specifically, the order of keys under `x-amazon-apigateway-integration/responseParameters/responses/default` appears to vary between runs. Here’s a sample diff that illustrates the inconsistency:

```
"responseParameters": {
-  "method.response.header.Access-Control-Allow-Origin": "'*'",
-  "method.response.header.Access-Control-Expose-Headers": "'Content-Length,Content-Type,Date,X-Amzn-ErrorType,X-Amzn-Requestid'"
+  "method.response.header.Access-Control-Expose-Headers": "'Content-Length,Content-Type,Date,X-Amzn-ErrorType,X-Amzn-Requestid'",
+  "method.response.header.Access-Control-Allow-Origin": "'*'"
}
```

This behavior appears to be specific to the @cors trait. While we couldn’t find documentation specifying a guaranteed key order in the generated OpenAPI, this inconsistency makes diffs noisy and complicates CI processes that depend on deterministic output so wanted to raise as a ticket. Question: Is this expected behavior or should we expect the output order to be deterministic?

JordonPhillips:
This updates every map, set, or list in a builder to use BuilderRef. BuilderRef exists to minimize the number of copies we have to do as well as to make it harder to forget to make immutable copies of collections.

In a few cases the underlying collection was changed to preserve or enforce order. This fixes #2734 because the source of that issue was not having a consistent order for responseParameters. That's now kept sorted, like basically every other part of the openapi document.

This resulted in fixing a few other unknown bugs. In some instances we had been directly referencing collections from the builder instead of making any kind of copies. There was also at least one test that asserted non-existent behavior that only passed because of the accident of ordering.

Note that mutability was preserved where it existed already. I had initially removed all of that, but there were some areas relying on it (notably openapi plugins).

---

### DIFF
--- a/smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/IntegrationResponse.java
+++ b/smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/IntegrationResponse.java
@@ -4,14 +4,13 @@
  */
 package software.amazon.smithy.aws.apigateway.traits;
 
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import software.amazon.smithy.model.FromSourceLocation;
 import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.node.NodeMapper;
 import software.amazon.smithy.model.node.ToNode;
-import software.amazon.smithy.utils.MapUtils;
+import software.amazon.smithy.utils.BuilderRef;
 import software.amazon.smithy.utils.SmithyBuilder;
 import software.amazon.smithy.utils.ToSmithyBuilder;
 
@@ -31,8 +30,8 @@ public final class IntegrationResponse implements ToNode, ToSmithyBuilder<Integr
     private IntegrationResponse(Builder builder) {
         statusCode = SmithyBuilder.requiredState("statusCode", builder.statusCode);
         contentHandling = builder.contentHandling;
-        responseTemplates = MapUtils.copyOf(builder.responseTemplates);
-        responseParameters = MapUtils.copyOf(builder.responseParameters);
+        responseTemplates = builder.responseTemplates.copy();
+        responseParameters = builder.responseParameters.copy();
         sourceLocation = builder.sourceLocation;
     }
 
@@ -151,8 +150,8 @@ public final class IntegrationResponse implements ToNode, ToSmithyBuilder<Integr
     public static final class Builder implements SmithyBuilder<IntegrationResponse> {
         private String statusCode;
         private String contentHandling;
-        private Map<String, String> responseTemplates = new HashMap<>();
-        private Map<String, String> responseParameters = new HashMap<>();
+        private final BuilderRef<Map<String, String>> responseTemplates = BuilderRef.forSortedMap();
+        private final BuilderRef<Map<String, String>> responseParameters = BuilderRef.forSortedMap();
         private FromSourceLocation sourceLocation;
 
         @Override
@@ -193,7 +192,7 @@ public final class IntegrationResponse implements ToNode, ToSmithyBuilder<Integr
          */
         public Builder responseTemplates(Map<String, String> responseTemplates) {
             this.responseTemplates.clear();
-            this.responseTemplates.putAll(responseTemplates);
+            responseTemplates.forEach(this::putResponseTemplate);
             return this;
         }
 
@@ -206,7 +205,7 @@ public final class IntegrationResponse implements ToNode, ToSmithyBuilder<Integr
          * @see IntegrationResponse#getResponseTemplates()
          */
         public Builder putResponseTemplate(String mimeType, String template) {
-            responseTemplates.put(mimeType, template);
+            responseTemplates.get().put(mimeType, template);
             return this;
         }
 
@@ -217,7 +216,7 @@ public final class IntegrationResponse implements ToNode, ToSmithyBuilder<Integr
          * @return Returns the builder.
          */
         public Builder removeResponseTemplate(String mimeType) {
-            responseTemplates.remove(mimeType);
+            responseTemplates.get().remove(mimeType);
             return this;
         }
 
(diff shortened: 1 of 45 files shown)


smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/IntegrationResponse.java
```diff
@@ -151,8 +150,8 @@ public int hashCode() {
     public static final class Builder implements SmithyBuilder<IntegrationResponse> {
         private String statusCode;
         private String contentHandling;
-        private Map<String, String> responseTemplates = new HashMap<>();
-        private Map<String, String> responseParameters = new HashMap<>();
+        private BuilderRef<Map<String, String>> responseTemplates = BuilderRef.forSortedMap();
```
milesziemer: There's a bunch of these that can now be final

smithy-aws-cloudformation/src/main/java/software/amazon/smithy/aws/cloudformation/schema/model/ResourceSchema.java
```diff
@@ -36,43 +35,42 @@ public final class ResourceSchema implements ToNode, ToSmithyBuilder<ResourceSch
...
-        if (builder.properties.isEmpty()) {
+        if (builder.properties.peek().isEmpty()) {
             throw new CfnException(format("Expected CloudFormation resource %s to have properties, "
                     + "found none", typeName));
         }
-        properties.putAll(builder.properties);
+        properties = builder.properties.copy();
 
-        required.addAll(builder.required);
+        required = builder.required.copy();
```
milesziemer: These don't preserve existing mutability

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/Schema.java
```diff
@@ -761,10 +759,9 @@ public Builder minProperties(Integer minProperties) {
         }
 
         public Builder required(Collection<String> required) {
-            if (required == null) {
-                this.required.clear();
-            } else {
-                this.required = new ArrayList<>(required);
+            this.required.clear();;
```
milesziemer: Extra semi

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/Schema.java
```diff
@@ -776,17 +773,23 @@ public Builder properties(Map<String, Schema> properties) {
                 properties.forEach(this::putProperty);
             }
 
+            for (String property : this.required.copy()) {
```
milesziemer, line 776: Nice find

smithy-jsonschema/src/main/java/software/amazon/smithy/jsonschema/SchemaDocument.java
```diff
@@ -35,7 +35,7 @@ private SchemaDocument(Builder builder) {
         idKeyword = builder.idKeyword;
         schemaKeyword = builder.schemaKeyword;
         rootSchema = builder.rootSchema != null ? builder.rootSchema : Schema.builder().build();
-        definitions = new LinkedHashMap<>(builder.definitions);
+        definitions = builder.definitions.copy();
```
milesziemer: Doesn't preserve existing mutability

smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
```diff
@@ -285,27 +282,29 @@ public Builder deprecated(boolean deprecated) {
         }
 
         public Builder security(Collection<Map<String, List<String>>> security) {
-            this.security = new LinkedHashSet<>();
-            this.security.addAll(security);
+            if (this.security == null) {
```
milesziemer: Might be able to use `BuilderRef::hasValue` in the ctor, then you can just make the builder member final and skip these checks

smithy-protocol-test-traits/src/main/java/software/amazon/smithy/protocoltests/traits/HttpMessageTestCase.java
```diff
@@ -243,10 +240,10 @@ abstract static class Builder<B extends Builder<?, ?>, T extends HttpMessageTest
         private ShapeId vendorParamsShape;
         private ObjectNode vendorParams = Node.objectNode();
         private AppliesTo appliesTo;
-        private final Map<String, String> headers = new TreeMap<>();
-        private final List<String> forbidHeaders = new ArrayList<>();
-        private final List<String> requireHeaders = new ArrayList<>();
-        private final List<String> tags = new ArrayList<>();
+        private final BuilderRef<Map<String, String>> headers = BuilderRef.forOrderedMap();
```
milesziemer: This was previously a sorted map (TreeMap)

milesziemer (changes requested):

smithy-openapi/src/main/java/software/amazon/smithy/openapi/model/OperationObject.java
```diff
@@ -37,22 +34,22 @@ public final class OperationObject extends Component implements ToSmithyBuilder<
...
         requestBody = builder.requestBody;
-        responses = Collections.unmodifiableMap(new TreeMap<>(builder.responses));
+        responses = builder.responses.copy();
         deprecated = builder.deprecated;
-        callbacks = Collections.unmodifiableMap(new TreeMap<>(builder.callbacks));
-        if (builder.security != null) {
-            security = ListUtils.copyOf(builder.security);
+        callbacks = builder.callbacks.copy();
+        if (builder.security.hasValue()) {
+            security = ListUtils.copyOf(builder.security.peek());
```
milesziemer, line 48: Can just do `copy`

JordonPhillips: Nah, it goes from a set to a list

milesziemer: Oh that's right

PR #2746, opened 2025-08-18
Return contained operations/resources in order

### LINKED ISSUE #2699: TopDownIndex doesn't retain order of operations
Related: #591 / #602 - the `getOperations()` method has been made to retain the order of operation in the service's original definition, but `TopDownindex.getContainedOperations()` didn't get the same treatment.

This results in unpredictable ordering, which has the same flaws as the previous behavior of `getOperations`.

For example, you can see the ordering change in the output of the following Scala script:

```scala
//> using scala 3.7.1
//> using dep software.amazon.smithy:smithy-model:1.60.3
//> using option -Wunused:all
import software.amazon.smithy.model.knowledge.TopDownIndex
import software.amazon.smithy.model.shapes.ServiceShape

val m = Model
  .assembler()
  .addUnparsedModel(
    "test.smithy",
    """$version: "2"
      |namespace demo
      |
      |service DummyService {
      |    version: "0.0"
      |    operations: [
      |        DummyHostPrefix
      |        Dummy
      |        DummyPath
      |    ]
      |}
      |
      |
      |operation DummyHostPrefix {}
      |operation Dummy {}
      |operation DummyPath {}
      |""".stripMargin,
  )
  .assemble()
  .unwrap()
// m: Model = software.amazon.smithy.model.Model@90925b8

TopDownIndex
  .of(m)
  .getContainedOperations(ShapeId.from("demo#DummyService"))
// res0: Set[OperationShape] = [(operation: `demo#Dummy`), (operation: `demo#DummyHostPrefix`), (operation: `demo#DummyPath`)]

m.expectShape(ShapeId.from("demo#DummyService"), classOf[ServiceShape]).getOperations()
// res1: Set[ShapeId] = [demo#DummyHostPrefix, demo#Dummy, demo#DummyPath]
```

JordonPhillips:
This updates the results of TopDownIndex to be in modeled order. Previously the resuls were being sorted. This problem extended to the Walker, which was returning shapes in reverse order.

This fixes #2699
---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/Walker.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/neighbor/Walker.java
@@ -158,7 +158,7 @@ public final class Walker {
                 // Use removeLast to retrieve relationships in their defined order rather
                 // than the reverse of the defined order. Note that this only preserves
                 // the order of a particular relationship type, not the order of all
-                // relationship types. So a resources `operation` relationships will be
+                // relationship types. So a resource's `operation` relationships will be
                 // resolved in the order of that list, but the resources's `resource`
                 // relationships will nevertheless always appear first because that is
                 // simply the order that the NeighborVisitor checks them in.


smithy-model/src/main/java/software/amazon/smithy/model/neighbor/Walker.java
```diff
@@ -154,7 +154,15 @@ public boolean hasNext() {
             while (!stack.isEmpty()) {
                 // Every relationship is returned, even if the same shape is pointed
                 // to multiple times from a single shape.
-                Relationship relationship = stack.pop();
+
+                // Use removeLast to retrieve relationships in their defined order rather
+                // than the reverse of the defined order. Note that this only preserves
+                // the order of a particular relationship type, not the order of all
+                // relationship types. So a resources `operation` relationships will be
```
kstich:
```suggestion
                // relationship types. So a resource's `operation` relationships will be
```

PR #2754, opened 2025-08-25
Fix OpenAPI error examples missing when service-level and operation-level errors share HTTP status codes

justincange:
We noticed that when a Smithy service defines both service-level errors and operation-specific errors that map to the same HTTP status code, the generated OpenAPI spec includes the correct schema structure but completely omits error response examples.

Expected: 
ErrorExamples from operation @examples should appear in HTTP error response sections.
Actual: Error response examples are missing while request and success examples work correctly.

This affects API documentation quality for services using both service-level and operation-specific errors.

This PR has the test case and a potential fix. Thanks!


smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
```diff
@@ -350,14 +354,17 @@ private Map<String, Node> createErrorBodyExamples(
         // unique numbering for unique example names in OpenAPI.
         int uniqueNum = 1;
         Optional<ExamplesTrait> examplesTrait = operation.getTrait(ExamplesTrait.class);
+        
+        Set<ShapeId> errorShapeIds = getErrorShapeIdsForMatching(context, error, operation);
```
haydenbaker, line 358: When the deconflict logic is run, the error shape that is passed here is actually a (synthesized) error structure that targets a (synthesized) union of the conflicting error shapes. 

So do something like (but you should add some edge case handling, and better names)
```suggestion
        // handle synthesized error unions when deconflicting is enabled
        Set<ShapeId> errorShapeIds = new HashSet<>();
        if (error.getMember("errorUnion").isPresent()) {
            ShapeId t = error.getMember("errorUnion").get().getTarget();
            Shape tt = context.getModel().getShape(t).get();
            for (MemberShape memberShape : tt.getAllMembers().values()) {
                errorShapeIds.add(memberShape.getTarget());
            }
        } else {
            // if not deconflicted, then the error passed is an actual error (not synthesized)
            errorShapeIds.add(error.toShapeId());
        }
```
And then you can get rid of the helper functions that aren't needed.

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ServiceLevelErrorOverrideTest.java
```diff
@@ -0,0 +1,141 @@
...
+import software.amazon.smithy.openapi.OpenApiConfig;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests that service-level errors don't override operation-specific error examples.
+ * This reproduces the healthcare service issue where ValidationException (service-level)
+ * overrides EligibilityCheckOutputValidationErrors (operation-level) examples.
+ */
+public class ServiceLevelErrorOverrideTest {
```
haydenbaker: Rename this to something like `ErrorDeconflictingExamplesTest`, and get rid of Stedi-specific jargon, like `EligibilityCheckOutputValidationErrors`, `healthcare service`, etc.

haydenbaker (changes requested):

smithy-openapi/src/main/java/software/amazon/smithy/openapi/fromsmithy/protocols/AbstractRestProtocol.java
justincange: just fixed and took this approach instead, thanks!

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests that service-level errors don't override operation-specific error examples.
+ */
+public class ErrorDeconflictingExamplesTest {
+
+    @Test
+    public void serviceLevelErrorsShouldNotOverrideOperationSpecificExamples() {
+        String modelText = "$version: \"2.0\"\n" +
```
haydenbaker: Please make this a test resource (file) instead of a string.

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,134 @@
...
+            "namespace test.service.error\n" +
+            "use aws.protocols#restJson1\n" +
+            "\n" +
+            "@restJson1\n" +
+            "service TestService {\n" +
+            "    version: \"1.0.0\",\n" +
+            "    operations: [TestOp]\n" +
+            "    errors: [ServiceError]  // Service-level 400 error\n" +
```
haydenbaker: Let's call this ClientError instead, because it's a little confusing to call it a "Service" error but it's a 400.

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,134 @@
...
+            .filter(example -> example.getMember("value").isPresent())
+            .map(example -> example.expectMember("value").expectObjectNode())
+            .anyMatch(value ->
+                value.getMember("code").filter(node -> "CUSTOM_ERROR".equals(node.expectStringNode().getValue())).isPresent() &&
+                value.getMember("message").filter(node -> "Custom error occurred".equals(node.expectStringNode().getValue())).isPresent()
+            );
+
+        assertTrue(foundCustomErrorExample,
+            "Expected to find CustomError example with code 'CUSTOM_ERROR' and message 'Custom error occurred' " +
+            "but service-level ValidationException appears to be overriding it");
```
haydenbaker: Instead of doing all of this manual assertion, let's just have the expected generated openapi file as a test resource and assert that it matches. Check `AwsRestJson1ProtocolTest` to see how that's done.

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,134 @@
...
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.openapi.OpenApiConfig;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests that service-level errors don't override operation-specific error examples.
```
haydenbaker:
```suggestion
 * Tests that service-level error and operation-level error examples are both included in the openapi when error deconflicting (`onErrorStatusConflict`) is set to `oneOf`.
```

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        OpenApiConfig config = new OpenApiConfig();
+        config.setService(ShapeId.from("test.service.error#TestService"));
+        config.setOnErrorStatusConflict(OpenApiConfig.ErrorStatusConflictHandlingStrategy.ONE_OF);
+
+        Node result = OpenApiConverter.create().config(config).convertToNode(model);
+        ObjectNode openApi = result.expectObjectNode();
+
+        System.out.println("Generated OpenAPI:");
+        System.out.println(Node.prettyPrintJson(result));
```
haydenbaker: Remove println statements from this file

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
justincange: Done, tried to do it in a more focused, concise way just for this use case

smithy-openapi/src/test/java/software/amazon/smithy/openapi/fromsmithy/ErrorDeconflictingExamplesTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Tests that all error examples are included in the openapi when error deconflicting (`onErrorStatusConflict`) is set to `oneOf`.
+ */
+public class ErrorDeconflictingExamplesTest {
+
+    @Test
+    public void serviceLevelErrorsShouldNotOverrideOperationSpecificExamples() {
+        Model model = Model.assembler()
+                .addImport(getClass().getResource("/META-INF/smithy/aws.protocols.smithy"))
```
milesziemer:
```suggestion
                .discoverModels()
```
This will pick up aws traits since they're on the classpath. Unless there's some reason you don't want this.

PR #2759, opened 2025-08-28
Fix loading jars with encoded paths

### LINKED ISSUE #2103: `createSmithyJarManifestUrl`: incorrect handling of complex paths
Hi! I first bumped into this over a year ago, but so far had a workaround that seemed decent, and the usecase seemed pretty niche. Now me and my team maintain about 4 copies of the same workaround, so it's reaching a point where it'd be best to resolve upstream - right here.

## What I'm trying to do

- grab a bunch of jars, find Smithy model files in them, and add them to a model assembler
- ...allowing some of these jars to be missing a manifest altogether (simply ignoring them).

Originally, here's how we did it in smithy4s:

```scala
val modelsInJars = deps.flatMap { file => // java.io.File
   val manifestUrl = 
     ModelDiscovery.createSmithyJarManifestUrl(file.getAbsolutePath()) 
   try { ModelDiscovery.findModels(manifestUrl).asScala } 
   catch { 
     case _: ModelManifestException => Seq.empty //fallback to nothing
   } 
 }
```

So far so good.

## The problem

Trouble begins when certain artifacts are involved, namely:

- artifacts fetched with Coursier
- ...which have urlencoded characters in their path. Which is typical for snapshots named after their corresponding commit and <commit count since last tag>.

For example, `org.polyvariant:test-library-core_2.13:0.0.1+123-SNAPSHOT` from Sonatype Snapshots would be placed in:

```
$COURSIER_CACHE_LOCATION/v1/https/s01.oss.sonatype.org/content/repositories/snapshots/org/polyvariant/test-library-core_2.13/0.0.1%2B123-SNAPSHOT/test-library-core_2.13-0.0.1%2B123-SNAPSHOT.jar
```

Here's what happens when you use that file with `ModelDiscovery`:

- `createSmithyJarManifestUrl` yields this URL: `jar:file:$COURSIER_CACHE_LOCATION/v1/https/s01.oss.sonatype.org/content/repositories/snapshots/org/polyvariant/test-library-core_2.13/0.0.1%2B123-SNAPSHOT/test-library-core_2.13-0.0.1%2B123-SNAPSHOT.jar!/META-INF/smithy/manifest` - so far so good, it seems
- `findModels` fails with `ModelManifestException` caused by `FileNotFoundException`. the path that cause reports is `$COURSIER_CACHE_LOCATION/v1/https/s01.oss.sonatype.org/content/repositories/snapshots/org/polyvariant/test-library-core_2.13/0.0.1+123-SNAPSHOT/test-library-core_2.13-0.0.1+123-SNAPSHOT.jar` - the urlescaping is gone now.

I think this is an issue with `ModelDiscovery.findModels`.

The expected behavior can be achieved with the following workaround: treat the jar as a FileSystem, and use the Path API to get an URL:

```scala
Using.resource(
  FileSystems.newFileSystem(file.toPath())
) { fs =>
  val path = fs.getPath("META-INF", "smithy", "manifest")
  ModelDiscovery.findModels(path.toUri().toURL())
}
```

The `path.toUri()` part stringifies as: `jar:file://$COURSIER_FILE_LOCATION/v1/https/s01.oss.sonatype.org/content/repositories/snapshots/org/polyvariant/test-library-core_2.13/0.0.1%252B123-SNAPSHOT/test-library-core_2.13-0.0.1%252B123-SNAPSHOT.jar!/META-INF/smithy/manifest` - which, to me, sounds correct.

## Request

I think there's more than one issue here:

1. `createSmithyJarManifestUrl` going from URL (a relatively structured type) to a `String` (no structure at all), offering little to no type safety when it's used down the line
2. `findModels` doing something strange with its parameter, discarding the urlencoding of the given path.

so the ideal solution would also be twofold:

- deprecate the Stringly-typed methods and add some safer ones, operating on `URL` or `URI`
- fix the handling of urlencoded paths in `findModels`

Context, if you want to learn more about the issue we had and the workaround: https://github.com/disneystreaming/smithy4s/pull/850

## Notes / questions

- Are there better ways to load a bunch of jars, not all of which necessarily contain Smithy manifests? If not, could one be baked into `ModelAssembler`?

### LINKED ISSUE #2576: Broken Model loading if jar path contain url encoded characters
Hi,

I'm facing following issue when trying to load smithy model. 
I have some local files that were downloaded from my internal, authenticated repository. The files therefore contain something like `user%40my.internal.repo` in the path (where `%40` is url encoded form of `@`). 

When I try to load such model (using regular `Model.assembler().addImport`), I get:

```
 software.amazon.smithy.model.loader.ModelManifestException: Error parsing Smithy model manifest from jar:file:/home/user/.cache/coursier/v1/https/user%40my.internal.repo/maven/com/disneystreaming/alloy/alloy-core/0.3.14/alloy-core-0.3.14.jar!/META-INF/smithy/manifest
```

Below is the minimized example, and the proposed fix. The `toString` on both URLs is exactly the same, however what is attempted to get opened is different. This is some deep java internals, and I don't have time to dig more into it, but looks like the fix I'm proposing will handle Paths that are provided by users butter in some rare scenarios.

Reproduced using [scala-cli](https://scala-cli.virtuslab.org/)

```scala
//> using scala "3.3.5"
//> using dep "software.amazon.smithy:smithy-model:1.56.0"
import java.net.URI
import scala.util.Try
import software.amazon.smithy.model.loader.ModelDiscovery


val locGood = "/home/ssn/Documents/alloy-core-0.3.14.jar"
val locBad = "/home/ssn/Documents/alloy%40core-0.3.14.jar"

def tryOpen(location: String) = Try {
    val url = ModelDiscovery.createSmithyJarManifestUrl(location)
    url.openConnection().getInputStream()  
}

def createSmithyJarManifestUrlFixed(fileOrUrl: String) = {
    val jarEntry = "/META-INF/smithy/manifest"
    val fileUri = new URI("file", null, fileOrUrl, null)
    val jarUri = new URI("jar:" + fileUri.toString + "!" + jarEntry)
    jarUri.toURL
}


def tryOpenFixed(location: String) = Try {
    val url = createSmithyJarManifestUrlFixed(location)
    url.openConnection().getInputStream()
}


println("BROKEN")
List(locGood, locBad).foreach{ l =>
    println(s"Opening $l")
    println(tryOpen(l))
}

println("FIXED")
List(locGood, locBad).foreach{ l =>
    println(s"Opening $l")
    println(tryOpenFixed(l))
}

```

The above prints:

```
BROKEN
Opening /home/ssn/Documents/alloy-core-0.3.14.jar
Success(sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream@246ae04d)
Opening /home/ssn/Documents/alloy%40core-0.3.14.jar
Failure(java.nio.file.NoSuchFileException: /home/ssn/Documents/alloy@core-0.3.14.jar)
FIXED
Opening /home/ssn/Documents/alloy-core-0.3.14.jar
Success(sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream@4157f54e)
Opening /home/ssn/Documents/alloy%40core-0.3.14.jar
Success(sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream@1936f0f5)
```

You can grab the same file I used for testing from
```bash
 curl -L -o alloy-core-0.3.14.jar \
    https://repo1.maven.org/maven2/com/disneystreaming/alloy/alloy-core/0.3.14/alloy-core-0.3.14.jar
```

I can prepare the PR for fix, and think about dropping in some test for it, but I would like to ask if there are any gotchas I should be aware or if the proposed fix looks good to you. 

Thanks!

milesziemer:
#### Background

The model assembler can load Smithy files from a jar - model discovery. This works by looking for a 'META-INF/smithy/manifest' file within the jar, which contains the paths of smithy model files within the jar. You can give the model assembler a class loader to use for model discovery, but you can also add a jar to discover models from using a regular `addImport`. The problem is with the latter case. Because jars are like zip files, you can't read files within the jar using regular file system operations. Instead, you can either use the JarFile api to open the jar, and read specific entries (like a zip file), or you can directly read specific entries via a special URL with the format `jar:file:/path/to/jar.jar!/path/in/jar`. The latter is the way that our model discovery apis work. The problem was that URL doesn't perform any encoding or decoding itself, so callers/consumers have to take care of it, which we were not doing. For example, if the jar file has a path of `/foo/bar%45baz.jar`, we would create a URL like: `jar:file:/foo/bar%45baz.jar!/META-INF/smithy/manifest`, and when the JDK tried to use this URL to read from the file system, it would decode it and attempt to read `/foo/bar@baz.jar`. Since we weren't encoding the URL to begin with, this was only a problem when the path contained `%`.

I tried to preserve the existing model discovery api, and just make it work with these paths. The URL class has a section in its javadoc about this footgun, and in more recent jdk versions, URL constructors are deprecated in favor of `URI::toURL` to avoid it. So that's what I did.

This technically changes the `SourceLocation::filename` for jars that are `addImport`ed, when the jar path has reserved characters (like spaces). Such jars were importable before because URL doesn't validate encoding, and the JDK decoding the URL would be a no-op since it wasn't encoded to begin with. So we could _just_ encode `%`, meaning only the newly-allowed paths would have any encoding. Disregarding the fact that this side-steps the assumed pre-conditions of URL, it would actually exacerbate an existing inconsistency between `SourceLocation::filename` of models discovered on the classpath, and models discovered from imports. The JDK encodes the URLs of resources on the classpath, so their models' source locations will be encoded. Encoding URLs of imported jars means there's no difference. You can assume that a source location with `jar:file:` is a well-formed URL/URI.

#### Testing

Added new ModelAssembler tests for loading a jar via `addImport` when the jar's path has encoded characters.
I added one for adding the import via path, and via URL, in case the way we normalize import paths ever changes.

I also added a ModelDiscovery test to assert the new encoding behavior.

#### Links

Fixes #2576 , Fixes #2103 

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelDiscovery.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/ModelDiscovery.java
@@ -10,6 +10,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.charset.StandardCharsets;
@@ -185,19 +187,25 @@ public final class ModelDiscovery {
      */
     public static URL createSmithyJarManifestUrl(String fileOrUrl) {
         try {
-            return new URL(getFilenameWithScheme(fileOrUrl) + "!/" + MANIFEST_PATH);
-        } catch (IOException e) {
+            // Building a URI from parts (scheme + file path) will make sure the file path
+            // gets properly URI encoded. URL does not do this, instead expecting callers
+            // to encode first, and consumers to decode. Encoding it here means that when
+            // it is decoded by the consumer (e.g. the JDK reading the jar from the file system),
+            // it will point to the correct location.
+            String manifestPath = removeScheme(fileOrUrl) + "!/" + MANIFEST_PATH;
+            return new URI("jar:file", null, manifestPath, null).toURL();
+        } catch (IOException | URISyntaxException e) {
             throw new ModelImportException(e.getMessage(), e);
         }
     }
 
-    private static String getFilenameWithScheme(String filename) {
-        if (filename.startsWith("jar:")) {
-            return filename;
-        } else if (filename.startsWith("file:")) {
-            return "jar:" + filename;
+    private static String removeScheme(String fileOrUrl) {
+        // Index will also cover jar: part of jar:file:
+        int pathStart = fileOrUrl.indexOf("file:");
+        if (pathStart == -1) {
+            return fileOrUrl;
         } else {
-            return "jar:file:" + filename;
+            return fileOrUrl.substring(pathStart + "file:".length());
         }
     }
 
--- a/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelAssemblerTest.java
@@ -1459,4 +1459,36 @@ public class ModelAssemblerTest {
         assertTrue(combinedModel.expectShape(ShapeId.from("smithy.example#MachineData$machineId"), MemberShape.class)
                 .hasTrait(RequiredTrait.ID));
     }
+
+    @Test
+    public void loadsJarFromPathWithEncodedChars() throws Exception {
+        Path source = Paths.get(getClass().getResource("assembler-valid-jar").toURI());
+        Path jarPath = outputDirectory.resolve(Paths.get("path%20with%45encoded%25chars", "test.jar"));
+
+        Files.createDirectories(jarPath.getParent());
+        Files.copy(JarUtils.createJarFromDir(source), jarPath);
+
+        Model model = Model.assembler()
+                .addImport(jarPath)
+                .assemble()
+                .unwrap();
+
+        assertTrue(model.getShape(ShapeId.from("smithy.example#ExampleString")).isPresent());
+    }
+
+    @Test
+    public void loadsJarFromUrlWithEncodedChars() throws Exception {
+        Path source = Paths.get(getClass().getResource("assembler-valid-jar").toURI());
+        Path jarPath = outputDirectory.resolve(Paths.get("path%20with%45encoded%25chars", "test.jar"));
+
+        Files.createDirectories(jarPath.getParent());
+        Files.copy(JarUtils.createJarFromDir(source), jarPath);
+
+        Model model = Model.assembler()
+                .addImport(jarPath.toUri().toURL())
+                .assemble()
+                .unwrap();
+
+        assertTrue(model.getShape(ShapeId.from("smithy.example#ExampleString")).isPresent());
+    }
 }
--- a/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelDiscoveryTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/loader/ModelDiscoveryTest.java
@@ -147,4 +147,14 @@ public class ModelDiscoveryTest {
         assertThat(ModelDiscovery.createSmithyJarManifestUrl("jar:file:/foo.jar"),
                 equalTo(new URL("jar:file:/foo.jar!/META-INF/smithy/manifest")));
     }
+
+    @Test
+    public void encodesSmithyManifestUrl() throws IOException {
+        assertThat(ModelDiscovery.createSmithyJarManifestUrl("/foo bar.jar"),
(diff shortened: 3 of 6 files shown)


smithy-build/src/main/java/software/amazon/smithy/build/PluginContext.java
```diff
@@ -223,6 +234,22 @@ private boolean isSource(FromSourceLocation sourceLocation) {
         String location = sourceLocation.getSourceLocation().getFilename();
         int offsetFromStart = findOffsetFromStart(location);
 
+        if (offsetFromStart > 0) {
+            // Offset means the filename is a URI, so compare to the URI paths to account for encoding and windows.
+            for (String sourceUri : sourceUris) {
+                // Compare starting after the protocol (the source uri will always be "file", because we created it
+                // from a path).
+                int sourceCompareStart = "file:".length();
```
JordonPhillips: this will always be 5

PR #2774, opened 2025-09-10
Add generic depdency graph implementation to smithy-utils and use it for handling topological sorting

JordonPhillips:
This adds a new `DependencyGraph` to smithy-utils, intended for use in the numerous locations where we perform topological sorts.

Currently this only being used in `SmithyIntegration.sort`, but I want to add it to other places where we do topological sorts before I take this out of draft. Results in SmithyIntegration are promising though. For dense dependency graphs I'm seeing a modest 10-20% speed up across the board. For sparse dependency graphs I'm seeing a ***6-20x*** speed up.

I also plan to add the same sort of ordering capabilities to smithy build plugins, which is why I've gone to the trouble to make this generic.

---


.changes/next-release/feature-d368237cdeb1655c7c10969e36c6aa3ca3b3034a.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "feature",
+  "description": "Add a generic dependency graph to smithy-utils to be used for sorting various dependent objects, such as integrations and plugins.",
+  "pull_requests": []
```
JordonPhillips: I gotta update this

smithy-codegen-core/src/jmh/java/software/amazon/smithy/codegen/core/jmh/SmithyIntegrations.java
```diff
@@ -0,0 +1,225 @@
...
+                TestIntegration integration = new TestIntegration(name);
+                if (i < 10) {
+                    independentIntegrations10.put(name, integration);
+                }
+                if (i < 100) {
+                    independentIntegrations100.put(name, integration);
+                }
+                independentIntegrations1000.put(name, integration);
+            }
+            independentIntegrations100 = Collections.unmodifiableMap(independentIntegrations100);
```
kstich, line 61: Why is only the `100` sized map being made unmodifiable?

smithy-codegen-core/src/jmh/java/software/amazon/smithy/codegen/core/jmh/SmithyIntegrations.java
```diff
@@ -0,0 +1,225 @@
...
+import software.amazon.smithy.codegen.core.ImportContainer;
+import software.amazon.smithy.codegen.core.SmithyIntegration;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolWriter;
+
+@Warmup(iterations = 3)
+@Measurement(iterations = 3, timeUnit = TimeUnit.MICROSECONDS)
+@BenchmarkMode(Mode.AverageTime)
+@Fork(1)
+public class SmithyIntegrations {
```
kstich, line 33: I don't see these benchmarks providing a before/after of the sorting in `codegen-core`, and the PR overview doesn't show the difference between before/after of `SmithyIntegration.sort`'s performance. Am I missing something?

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/IntegrationTopologicalSort.java
```diff
@@ -5,119 +5,94 @@
...
     private static final Logger LOGGER = Logger.getLogger(IntegrationTopologicalSort.class.getName());
 
     private final Map<String, I> integrationLookup = new LinkedHashMap<>();
     private final Map<String, Integer> insertionOrder = new HashMap<>();
-    private final Map<String, Set<String>> forwardDependencies = new LinkedHashMap<>();
-    private final Map<String, Set<String>> reverseDependencies = new LinkedHashMap<>();
-
-    private final Queue<String> satisfied = new PriorityQueue<>((left, right) -> {
+    private final DependencyGraph<String> dependencyGraph = new DependencyGraph<>();
+    private final Comparator<String> comparator = (left, right) -> {
```
kstich: This should be a method that can be referenced instead of an anonymous function.

smithy-utils/src/main/java/software/amazon/smithy/utils/DependencyGraph.java
```diff
@@ -0,0 +1,405 @@
...
+        forwardDependencies.get(what).add(dependsOn);
+        reverseDependencies.get(dependsOn).add(what);
+    }
+
+    /**
+     * Adds a set of dependencies to a node.
+     *
+     * <p>If any node is not already present in the graph, it is added.
+     *
+     * @param what The node to add dependencies to to.
```
kstich:
```suggestion
     * @param what The node to add dependencies to.
```

smithy-codegen-core/src/jmh/java/software/amazon/smithy/codegen/core/jmh/SmithyIntegrations.java
JordonPhillips: It can't really do that since the old code is gone. I'll add a table to the description of the results I got.

smithy-build/src/test/java/software/amazon/smithy/build/TimestampPlugin1.java
```diff
@@ -0,0 +1,27 @@
...
+    public void execute(PluginContext context) {
+        context.getFileManifest().writeFile("timestamp", String.valueOf(Instant.now().toEpochMilli()));
+        try {
+            Thread.sleep(1);
+        } catch (InterruptedException ignored) {}
+    }
+
+    @Override
+    public boolean isSerial() {
+        return true;
```
kstich, line 28: I don't think there's a reliance on these plugins being serial, but all of them are listed that way. Can these be non-serial (my read is they can) or is it a secret dependency on this behavior?

smithy-utils/src/main/java/software/amazon/smithy/utils/DependencyGraph.java
```diff
@@ -0,0 +1,405 @@
...
+
+        // Check for cycles.
+        if (result.size() != reverseDependencies.size()) {
+            List<String> remaining = new ArrayList<>(reverseDependencies.size() - result.size());
+            for (T node : reverseDependencies.keySet()) {
+                if (!result.contains(node)) {
+                    remaining.add(node.toString());
+                }
+            }
+            throw new IllegalStateException(
```
kstich: What about using this detailed exception from `TopologicalShapeSort` - https://github.com/smithy-lang/smithy/blob/426c1ef6cb3f5e48ba67f1ddcfe3fb728d9e1878/smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java#L116

smithy-build/src/test/java/software/amazon/smithy/build/TimestampPlugin1.java
JordonPhillips: This is in case we ever decide to run plugins in parallel. We don't now, but we could.

smithy-utils/src/main/java/software/amazon/smithy/utils/DependencyGraph.java
JordonPhillips: I can use something like that, but I can't use that since smithy-model depends on smithy-utils, not the other way around.

JordonPhillips: Actually this isn't so simple because exceptions can't be generic. Best I can do is `Set<?>`, which isn't great. But I can point people to `findCycles` for better info if needed.

.changes/next-release/feature-5918b18d6bad7f0ad2c8061f9910f70dccadce2c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Updated `LoaderShapeMap` to use `DependencyGraph` to sort shapes\ninstead of `TopologicalShapeSort`. This results in some error messages\nchanging:\n\n* There is no longer a special message for missing transitive mixins.\n* The message for genrically missing mixins is now applied by\n  ApplyMixin, which has a slightly different wording and which will\n  trigger for each mising mixin.\n* The error message for detected cycles will now include the entire\n  cycle that a shape is part of, rather than just the edges that it\n  is directly connected to.",
```
kstich, line 3:
```suggestion
  "description": "Updated `LoaderShapeMap` to use `DependencyGraph` to sort shapes\ninstead of `TopologicalShapeSort`. This results in some error messages\nchanging:\n\n* There is no longer a special message for missing transitive mixins.\n* The message for generically missing mixins is now applied by\n  ApplyMixin, which has a slightly different wording and which will\n  trigger for each mising mixin.\n* The error message for detected cycles will now include the entire\n  cycle that a shape is part of, rather than just the edges that it\n  is directly connected to.",
```

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/loader/mixins-detects-cycles2.errors
```diff
@@ -1,4 +1,5 @@
-[ERROR] smithy.example#Mixin1: Unable to resolve mixins; cycles detected between this shape and [smithy.example#Bad] | Model
-[ERROR] smithy.example#Mixin2: Unable to resolve mixins; cycles detected between this shape and [smithy.example#Mixin1] | Model
-[ERROR] smithy.example#Mixin3: Unable to resolve mixins; cycles detected between this shape and [smithy.example#Mixin2] | Model
-[ERROR] smithy.example#Bad: Unable to resolve mixins; cycles detected between this shape and [smithy.example#Mixin3] | Model
+
```
kstich:
```suggestion
```

kstich (approved):

PR #2785, opened 2025-09-25
Fix topological sort NPE when re-enqueueing shapes

Jacoby6000:
#### Background
I have, on several occasions, run in to issues with an NPE coming out of the TopologicalShapeSort class. Specifically:
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Set.remove(Object)" because "dependentDependencies" is null
    at software.amazon.smithy.model.loader.TopologicalShapeSort.dequeueSortedShapes(TopologicalShapeSort.java:83)
    at software.amazon.smithy.model.transform.ReplaceShapes.updateMixins(ReplaceShapes.java:183)
    at software.amazon.smithy.model.transform.ReplaceShapes.transform(ReplaceShapes.java:80)
    at software.amazon.smithy.model.transform.ModelTransformer.replaceShapes(ModelTransformer.java:102)
    ...
```
In some of these situations, I was able to alleviate the problem by changing the order in which I perform my model transformations, but I finally ran in to a case where I could not. I decided to look a little deeper.

When running `dequeueSortedShapes`, each shape that has no dependencies is processed in the order that they are enqueued. Upon dequeuing, they are removed from the forward dependencies map.  This is fine when `satisfiedShapes` only contains satisfied shapes, but presents issues when something with an unsatisfied dependency is marked as satisfied.  In this scenario, a given dependent may be processed before its dependencies and results in the case below:
* B is enqueued with no dependencies
* A is enqueued with no dependencies
* B is then re-enqueued, with a dependency on A
* ...
* `dequeueSortedShapes` is called
* B is processed as a satisfied shape immediately
  * Removes itself from forward dependencies
  * Has no reverse dependencies, so this shape is done processing
* A is processed as a satisfied shape
  * Removes itself from forward dependencies 
  * Detects B as a reverse dependency
  * Attempts to fetch the forward dependencies of B
  * forward dependencies are null, because B was already processed

The solution winds up being very simple. When a shape is enqueued with dependencies, we should check if it was previously marked as satisfied, and remove it from the queue if it was.

A related/opposite case was identified here: https://github.com/smithy-lang/smithy/pull/2785#discussion_r2379883973
This PR fixes that case as well.

#### Testing
* I reproduced the failure using model interactions exclusively, then reproduced the failure directly with the topological sorter.

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
@@ -56,11 +56,11 @@ public final class TopologicalShapeSort {
     public void enqueue(ShapeId shape, Collection<ShapeId> dependencies) {
         if (dependencies.isEmpty()) {
             satisfiedShapes.offer(shape);
-            
+
             // Remove dependencies if this shape was enqueued in the past with them.
             Set<ShapeId> previousDependencies = forwardDependencies.remove(shape);
-            if(previousDependencies != null) {
-                for(ShapeId dependency : previousDependencies) {
+            if (previousDependencies != null) {
+                for (ShapeId dependency : previousDependencies) {
                     reverseDependencies.get(dependency).remove(shape);
                 }
             }
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
@@ -56,6 +56,14 @@ public final class TopologicalShapeSort {
     public void enqueue(ShapeId shape, Collection<ShapeId> dependencies) {
         if (dependencies.isEmpty()) {
             satisfiedShapes.offer(shape);
+            
+            // Remove dependencies if this shape was enqueued in the past with them.
+            Set<ShapeId> previousDependencies = forwardDependencies.remove(shape);
+            if(previousDependencies != null) {
+                for(ShapeId dependency : previousDependencies) {
+                    reverseDependencies.get(dependency).remove(shape);
+                }
+            }
         } else {
             for (ShapeId dependency : dependencies) {
                 reverseDependencies.computeIfAbsent(dependency, unused -> new HashSet<>()).add(shape);
@@ -63,7 +71,8 @@ public final class TopologicalShapeSort {
             forwardDependencies.put(shape, new HashSet<>(dependencies));
 
             if (satisfiedShapes.contains(shape)) {
-                // This shape was previously marked as satisfied, but now we know it has dependencies.
+                // This shape was previously enqueued with no dependencies and marked as satisfied.
+                // Now it has dependencies.
                 satisfiedShapes.remove(shape);
             }
         }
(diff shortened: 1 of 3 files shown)


Jacoby6000:
Another potential solution may be to change the way that ReplaceShapes works, such that a shape with a duplicateId is never enqueued in to the topological sorter twice.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
```diff
@@ -57,10 +57,15 @@ public void enqueue(ShapeId shape, Collection<ShapeId> dependencies) {
...
-                reverseDependencies.computeIfAbsent(dependent, unused -> new HashSet<>()).add(shape);
+            for (ShapeId dependency : dependencies) {
+                reverseDependencies.computeIfAbsent(dependency, unused -> new HashSet<>()).add(shape);
             }
             forwardDependencies.put(shape, new HashSet<>(dependencies));
+
+            if (satisfiedShapes.contains(shape)) {
+                // This shape was previously marked as satisfied, but now we know it has dependencies.
+                satisfiedShapes.remove(shape);
+            }
```
Jacoby6000, line 77: This fix also begs the question... What if a shape with dependencies gets re-enqueued with no dependencies? Should we remove it from the dependency maps?

sugmanue:
Hi Jacob, thank you for the fix, we will get someone assigned to review it, but can you please fix the build failures, should be as easy as running `./gradlew :smithy-model:spotlessApply` and committing the changes.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TopologicalShapeSort.java
Jacoby6000: I locally created a couple of test cases which represent the reverse of the scenario described in the PR description:

```java
    @Test
    public void testReEnqueueWithoutDependencies() {
        List<Pair<ShapeId, Set<ShapeId>>> input = new ArrayList<>();
        input.add(Pair.of(from("test#B"), SetUtils.of(from("test#A"))));
        input.add(Pair.of(from("test#B"), SetUtils.of()));
        input.add(Pair.of(from("test#A"), SetUtils.of()));

        TopologicalShapeSort sort = new TopologicalShapeSort();
        input.forEach(pair -> sort.enqueue(pair.getLeft(), pair.getRight()));
        List<ShapeId> result = sort.dequeueSortedShapes();

        assertThat(result.get(0), equalTo(from("test#B")));
        assertThat(result.get(1), equalTo(from("test#A")));
    }

    @Test
    public void testReEnqueueSwapDependencies() {
        List<Pair<ShapeId, Set<ShapeId>>> input = new ArrayList<>();
        input.add(Pair.of(from("test#B"), SetUtils.of()));
        input.add(Pair.of(from("test#A"), SetUtils.of(from("test#B"))));

        input.add(Pair.of(from("test#B"), SetUtils.of(from("test#A"))));
        input.add(Pair.of(from("test#A"), SetUtils.of()));

        TopologicalShapeSort sort = new TopologicalShapeSort();
        input.forEach(pair -> sort.enqueue(pair.getLeft(), pair.getRight()));
        List<ShapeId> result = sort.dequeueSortedShapes();

        assertThat(result.get(0), equalTo(from("test#A")));
        assertThat(result.get(1), equalTo(from("test#B")));
    }

```

A real world example of this situation might be something like
* A mixes in B
* B is satisfied
* Replace shapes is called with
* B mixes in A
* A is satisfied

These also fail with an NPE, so I suppose we must remove them from the dependency maps. I suspect this will hamper performance somewhat, since the reverse dependencies implementation will be O(n) (before factoring in the performance semantics of operations on Maps)

Jacoby6000: This has been addressed.

JordonPhillips (approved):
I have a pr up that created a centralized dependency graph implementation - perhaps it should be used here. But in the meantime this looks good.

PR #2793, opened 2025-10-11
Add a sufix to distinguish added required traits to structures or unions

sugmanue:
#### Background

Added a new suffix to the changed nullability test for when the `@required` was added to a member that refers to an structure or a union. That case is backwards compatible for SDK generators using the nullability check `CheckMode.CLIENT_CAREFUL`, such as AWS SDKs. This is similar to what we already do [here](https://github.com/smithy-lang/smithy/blob/b51e9fa71787e03dc9473852df2d5d4f93d7045d/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/ChangedNullability.java#L147-L157) for the `@required` trait is removed.

#### Testing

Added a unit test.

#### Links

---

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,6 @@
 [versions]
 junit5 = "5.13.4"
+junit-platform = "1.13.4"
 hamcrest = "3.0"
 jmh = "0.7.2"
 spotbugs = "6.2.4"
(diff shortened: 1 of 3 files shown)


gradle/libs.versions.toml
```diff
@@ -40,7 +41,7 @@ junit-bom = { module = "org.junit:junit-bom", version.ref = "junit5" }
 junit-jupiter-api = { module = "org.junit.jupiter:junit-jupiter-api", version.ref = "junit5" }
 junit-jupiter-engine = { module = "org.junit.jupiter:junit-jupiter-engine", version.ref = "junit5" }
 junit-jupiter-params = { module = "org.junit.jupiter:junit-jupiter-params", version.ref = "junit5" }
-junit-platform-launcher = { module = "org.junit.platform:junit-platform-launcher" }
+junit-platform-launcher = { module = "org.junit.platform:junit-platform-launcher", version.ref = "junit-platform" }
```
sugmanue, line 44: `smithy-jmespath` was failing locally complaining about not able to initialize `junit-platform`. This change fixed it.

PR #2796, opened 2025-10-16
Loosen `ChangedMemberTarget` diff evaluator and make ModelDiff a bit more flexible

JordonPhillips:
This loosens the `ChangedMemberTarget` diff evaluator to not report `ERROR` level events so often. Now it will always report an `ERROR` if the change is expected to result in codegen type errors, but otherwise it will default to `WARNING` unless differing traits between the targets would result in a higher severity event had those trait changes been applied to the original target.

This is accomplished by re-running `ModelDiff` with a set of synthetic changes, comparing the new and old targets as if they were the same shape that just got changed. To facilitate this, `ModelDiff` and `Differences` had to be updated to allow running based on a curated list of changes rather than only allow running based on auto-detected changes. `DiffEvaluator` also had to be updated to allow passing along the `ClassLoader`.

---


JordonPhillips:
Windows line separators strike again

PR #2807, opened 2025-10-23
Fix changlog reading and new-change generation

JordonPhillips:
This fixes some bugs in the changelog tool.

* It fixes an issue where the changelog tool was attempting to read new changelog entries from the base branch instead of the pr branch.
* It fixes a more minor issue where filling out the new-change form without a known pr would result in a pr list value of `[""]` instead of `[]`.
* It fixes an issue where the wrong environment variable was being checked for the pull request number.

---

### DIFF
--- a/.github/workflows/changelog-ci.yml
+++ b/.github/workflows/changelog-ci.yml
@@ -87,5 +87,5 @@ jobs:
                 cd base/.changes
                 uv run amend \
                   --review-comment \
-                  --pull-request-number "$REPO_NUMBER" \
+                  --pull-request-number "$PR_NUMBER" \
                   --repository-dir "$REPO_DIR/pr"


.github/workflows/changelog-ci.yml
```diff
@@ -87,5 +87,5 @@ jobs:
                 cd base/.changes
                 uv run amend \
                   --review-comment \
-                  --pull-request-number "$REPO_NUMBER" \
+                  --pull-request-number "$PR_NUMBER" \
```
JordonPhillips, line 90: Note that this environment variable is set in the `env` block above, and its value will be printed in the logs. If you check the logs for this run, for example, you can see it has the value `2807`.

sugmanue (approved):

PR #2833, opened 2025-11-10
[Bugfix] Correctly compute and enforce blob length values in the NodeValidationVisitor

brandondahler:
#### Background
* What do these changes do? 
  * Fixes the computation logic for BlobLengthPlugin to correctly compute the blob's length
  * Correctly reports values which aren't base64 encoded as errors
  * Adds an opt-out Feature for the NodeValidationVisitor to revert back to existing behavior
* Why are they important?
  * Current validation behavior assumes blob values are simply the utf-8 encoding of the string; however, the effective use and spec states that blob values are base64 strings.

#### Testing
  * Verified unit test that were broken were legitimately wrong
    * Complete, fixed unit tests that were using non-base64 values when testing
  * Added new unit test for the new failure
    * Complete, added new tests specifically to check for the new error

#### Links
  * N/A

---


brandondahler:
This isn't going to work as a bugfix -- it is too breaking.  Will have to introduce a new feature of the node visitor which can require the length be semantically correct.

Issue #2821, opened 2025-11-03
Feature request: Support for defining and verifying positive and negative test vectors for constrained shapes and members


brandondahler:
### Context
Smithy currently allows you to constrain type definitions and structure members based on patterns, length, range, etc via built-in traits.  These constraint traits are particularly useful because they are used by code generators to implement input (and possibly output) validation.  Input validation on the server side is of particular interest because it is used as a means of verifying that untrusted inputs are, in fact, safe to use.  

The only catch with these constraint traits is the consequences of modifying them may not be easily understood in some scenarios -- for example, regular expressions as used by the `@pattern` trait are notoriously difficult to understand when non-trivial.  In order to guarantee that the implementation allows the intended values while disallowing unintended values, service builders must currently define positive and negative test cases and apply them to the final implementation manually.

### Proposal
We should make available and implement a `@testVectors` trait which accepts a list of allowed and disallowed values on a per shape and/or per member basis for which constraint traits can be applied to.  The list of allowed and disallowed values will be checked against the configured constraints at build time and discrepancies will cause validation errors to occur.

For example, a valid usage may look like this:
```
@length(min: 36, max: 36)
@pattern("^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")
@testVectors({
  allowed: [
    "00000000-0000-0000-0000-000000000000"
    "12345678-90ab-cdef-1234-567890abcdef"
    "ffffffff-ffff-ffff-ffff-ffffffffffff"
  ]
  disallowed: [
    ""
    "0-0-0-0-0"
    "A0000000-0000-0000-0000-000000000000"
    "FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF"
  ]
})
string Uuid
```

This new trait helps model authors 1/ express their intent for what values **MUST** and **MUST NOT** look like, 2/ have a mechanism to ensure that emergent "problem" values can be imbued into the interface definition, and 3/ understand the consequences of constraint changes.  Similarly, it helps model consumers understand the intent from the model author and have concrete understanding of the allowed and disallowed value.

This proposal does not seek to eliminate the ultimate need of service builders to verify that their implementation matches their definition; however, it may be useful from a test automation standpoint as these test vectors could be used to auto-generate per-field tests which validate their implementation's behavior against the definition.

#### Changes Required
[...]

PR #2838, opened 2025-11-12
Add NodeValidationVisitor feature that enforces base64 encoding of blob values

brandondahler:
#### Background
* What do these changes do? 
  * Adds a new `REQUIRE_BASE_64_BLOB_VALUES` NodeValidationVisitor Feature which enforces that blob values are valid base64-encoded values
  * Fixes BlobLengthPlugin to enforce length on the base64-decooded byte array's length instead of the base64-encoded string's length when the new feature is enabled
* Why are they important?
  * Per both the [1.0](https://github.com/smithy-lang/smithy/blob/56cca11cf6cd0664090261217eaf9d30b63d2b0b/docs/source-1.0/spec/core/model.rst?plain=1#L2387-L2389) and [2.0](https://github.com/smithy-lang/smithy/blob/56cca11cf6cd0664090261217eaf9d30b63d2b0b/docs/source-2.0/spec/model.rst?plain=1#L756-L758) specs for trait node values, blob values are "A `string` value that is base64 encoded."; however, existing trait validators and related code generators do not necessarily follow this specification.
    * Some trait validators like the [DefaultTraitValidator only warn on invalid values](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/DefaultTraitValidator.java#L107-L113)
    * Some codegen implementations like [smithy-typescript's HttpProtocolTestGenerator explicitly parse the blob value as an array of utf-16 bytes](https://github.com/smithy-lang/smithy-typescript/blob/cc0124e69e2d6943f0960854bcc3abf2defd2c20/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java#L1138C58-L1138C68)

#### Testing
  * Left all existing tests alone
  * Added new parameterized test to verify both positive and negative cases for the blob behavior when the feature is enabled

#### Links
  * Mentioned this feature in the comments of #2821 
  * No direct issue created

---

### DIFF
--- a/.changes/next-release/feature-22e1dd049a18d2fdf13cf205af81f5374e34a46b.json
+++ b/.changes/next-release/feature-22e1dd049a18d2fdf13cf205af81f5374e34a46b.json
@@ -1,6 +1,6 @@
 {
   "type": "feature",
-  "description": "Add NodeValidationVisitor feature that enforces base64 encoding of blob values",
+  "description": "Added a `NodeValidationVisitor` feature that enforces base64 encoding of blob values.",
   "pull_requests": [
     "[#2838](https://github.com/smithy-lang/smithy/pull/2838)"
   ]


.changes/next-release/feature-22e1dd049a18d2fdf13cf205af81f5374e34a46b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add NodeValidationVisitor feature that enforces base64 encoding of blob values",
```
JordonPhillips:
```suggestion
  "description": "Added a `NodeValidationVisitor` feature that enforces base64 encoding of blob values.",
```

PR #2851, opened 2025-11-13
Add new shapeExamples trait that communicates and enforces allowed and disallowed values

brandondahler:
#### Background
* What do these changes do? 
  * Adds a new `@shapeExamples` trait  which defines the set of allowed and disallowed values for a given shape.
  * Updates the `NodeValidationVisitor` to be more accurate in some less-commonly used cases
  * Adds a new  `ShapeExamplesTraitValidator` implementation which  verifies that:
    1. Allowed values are accepted by the shape based on the configured constraints
    2. Disallowed values are not accepted by the shape based on the configured constraints
* Why are they important?
  * Configuring and modifying a shape's constraints is tedious work with potentially large consequences if mistakes are made.
  * This is especially true for `@pattern`-based constrained values
  * This feature offers a mechanism to both express your intent as a shape author about what you *meant* to configure while also enforcing that the intent is a reality during the model validation phase.

#### Testing
  * Significant  effort was put into building out an ErrorFilesTest-based suite of tests to ensure that the validation works across as much of the gamut of constraint and shape type combinations that exist.
  * Documentation updates were best effort -- I do not have an environment setup that is readily capable of building and inspecting the output.

#### Links
  * Implements #2821

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/BlobLengthPlugin.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/validation/node/BlobLengthPlugin.java
@@ -29,7 +29,12 @@ final class BlobLengthPlugin extends MemberAndShapeTraitPlugin<BlobShape, String
         byte[] value = node.getValue().getBytes(StandardCharsets.UTF_8);
 
         if (context.hasFeature(NodeValidationVisitor.Feature.REQUIRE_BASE_64_BLOB_VALUES)) {
-            value = Base64.getDecoder().decode(value);
+            try {
+                value = Base64.getDecoder().decode(value);
+            } catch (IllegalArgumentException e) {
+                // Error will reported by the blobShape method in NodeValidationVisitor
+                return;
+            }
         }
 
         int size = value.length;
--- a/smithy-model/src/test/java/software/amazon/smithy/model/validation/NodeValidationVisitorTest.java
+++ b/smithy-model/src/test/java/software/amazon/smithy/model/validation/NodeValidationVisitorTest.java
@@ -741,6 +741,18 @@ public class NodeValidationVisitorTest {
 
                 // base64 encoded value without padding
                 {"ns.foo#Blob1", "\"Zg\"", null},
+
+                // Member validation
+                {"ns.foo#Structure2", "{\"d\": \"Zg==\"}", null},
+                {"ns.foo#Structure2",
+                        "{\"d\": \"Zm9vbw==\"}",
+                        new String[] {
+                                "d: Value provided for `ns.foo#Structure2$d` must have no more than 3 bytes, but the provided value has 4 bytes"}},
+                {"ns.foo#Structure2",
+                        "{\"d\": \"{}\"}",
+                        new String[] {
+                                "d: Blob value must be a valid base64 string"
+                        }},
         });
     }
 }
--- a/smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/shape-examples-trait.errors
+++ b/smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/shape-examples-trait.errors
@@ -48,7 +48,8 @@
 [ERROR] smithy.example#ConflictingExamplesString: Disallowed shape example `"aa"` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.1
 [ERROR] smithy.example#ConflictingExamplesString: Disallowed shape example `"aaa"` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.2
 [ERROR] smithy.example#ConflictingExamplesBlob: Allowed shape example `""`: Value provided for `smithy.example#ConflictingExamplesBlob` must have at least 1 bytes, but the provided value only has 0 bytes | ShapeExamplesTrait.allowed.0
-[ERROR] smithy.example#ConflictingExamplesBlob: Allowed shape example `"YWFhYQ=="`: Value provided for `smithy.example#ConflictingExamplesBlob` must have no more than 3 bytes, but the provided value has 4 bytes | ShapeExamplesTrait.allowed.1
+[ERROR] smithy.example#ConflictingExamplesBlob: Allowed shape example `"{}"`: Blob value must be a valid base64 string | ShapeExamplesTrait.allowed.1
+[ERROR] smithy.example#ConflictingExamplesBlob: Allowed shape example `"YWFhYQ=="`: Value provided for `smithy.example#ConflictingExamplesBlob` must have no more than 3 bytes, but the provided value has 4 bytes | ShapeExamplesTrait.allowed.2
 [ERROR] smithy.example#ConflictingExamplesBlob: Disallowed shape example `"YQ=="` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.0
 [ERROR] smithy.example#ConflictingExamplesBlob: Disallowed shape example `"YWE="` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.1
 [ERROR] smithy.example#ConflictingExamplesBlob: Disallowed shape example `"YWFh"` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.2
@@ -130,7 +131,8 @@
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesString: Disallowed shape example `"aaa"` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.2
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Allowed shape example `null`: Required structure member `conflictingExamplesBlob` for `smithy.example#DirectMembersStructure` cannot be null | ShapeExamplesTrait.allowed.0
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Allowed shape example `""`: Value provided for `smithy.example#DirectMembersStructure$conflictingExamplesBlob` must have at least 1 bytes, but the provided value only has 0 bytes | ShapeExamplesTrait.allowed.1
-[ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Allowed shape example `"YWFhYQ=="`: Value provided for `smithy.example#DirectMembersStructure$conflictingExamplesBlob` must have no more than 3 bytes, but the provided value has 4 bytes | ShapeExamplesTrait.allowed.2
+[ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Allowed shape example `"{}"`: Blob value must be a valid base64 string | ShapeExamplesTrait.allowed.2
+[ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Allowed shape example `"YWFhYQ=="`: Value provided for `smithy.example#DirectMembersStructure$conflictingExamplesBlob` must have no more than 3 bytes, but the provided value has 4 bytes | ShapeExamplesTrait.allowed.3
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Disallowed shape example `"YQ=="` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.0
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Disallowed shape example `"YWE="` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.1
 [ERROR] smithy.example#DirectMembersStructure$conflictingExamplesBlob: Disallowed shape example `"YWFh"` passed all validations when it shouldn't have | ShapeExamplesTrait.disallowed.2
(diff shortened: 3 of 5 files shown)


smithy-model/src/main/java/software/amazon/smithy/model/validation/testrunner/SmithyTestCase.java
```diff
@@ -295,4 +298,25 @@ public static final class Error extends RuntimeException {
             this.result = result;
         }
     }
+
+    private static final class ErrorsFileValidationEventFormatter implements ValidationEventFormatter {
```
brandondahler, line 302: Updated implementation here to make it easier to iterate on error files -- the previous implementation would output the extra file location information and hints if they exist, which are explicitly omitted from the errors file normally.

You can now copy+paste directly from the failure message in the test result into the errors file.

smithy-model/src/main/java/software/amazon/smithy/model/validation/NodeValidationVisitor.java
```diff
@@ -395,16 +397,19 @@ public List<ValidationEvent> unionShape(UnionShape shape) {
     @Override
     public List<ValidationEvent> memberShape(MemberShape shape) {
         List<ValidationEvent> events = applyPlugins(shape);
-        if (value.isNullNode()) {
+
+        if (!value.isNullNode()) {
+            model.getShape(shape.getTarget()).ifPresent(target -> {
```
brandondahler, line 402: Originally traversing into the shape when the value was null pretty much guaranteed that an error would be reported -- generally of the form "<string, number, object> was expected but value was null" (not exact wording).

Since is the `memberShape` method, this only affects aggregate types which have an explicit `null` value as opposed to the member being omitted.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -770,6 +770,22 @@ string pattern
 )
 structure required {}
 
+/// Defines values which are specifically allowed and/or disallowed for a shape.
+@trait(
+    selector: ":test(number, string, blob, structure, list, map, member)"
+)
+structure shapeExamples {
+  allowed: ShapeExampleList
+  disallowed: ShapeExampleList
```
robin-aws, line 779: nit:

```suggestion
  valid: ShapeExampleList
  invalid: ShapeExampleList
```

AFAICT "valid" appears a whole lot more in the spec/documentation than "allowed"

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -770,6 +770,22 @@ string pattern
 )
 structure required {}
 
+/// Defines values which are specifically allowed and/or disallowed for a shape.
+@trait(
+    selector: ":test(number, string, blob, structure, list, map, member)"
+)
```
JordonPhillips, line 776: This should be able to target unions. Also, gotta make sure to restrict the target for member shapes.
```suggestion
@trait(
    selector: """
        :test(
            number, string, blob, structure, union, list, map,
            member > :test(number, string, blob, structure, union, list, map)
        )"""
)
```

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -770,6 +770,22 @@ string pattern
...
+  allowed: ShapeExampleList
+  disallowed: ShapeExampleList
+}
+
+@private
+@length(min: 1)
+@sparse
+list ShapeExampleList {
+  member: Document
+}
```
JordonPhillips, line 787: I'd like to open this up a bit by making the list value a structure. That'll let us add additional properties aside from the value itself. At the moment I've added a documentation property to describe why a given value does or does not validate. There was also some desire to add a specific error that invalid values trigger. I'm not so sure about that. But opening it up like this makes that possible too.

```suggestion
/// A list of values to be mapped to a shape.
@private
@length(min: 1)
@sparse
list ShapeExampleList {
    member: ShapeExample
}

/// A value that may be mapped to a shape.
@private
structure ShapeExample {
    /// The actual value that may be mapped to the shape. This must match the
    /// expected shape type, regardless of whether this example represents a
    /// valid or invalid value.
    @required
    value: Document

    /// Optional documentation in CommonMark format describing why the example
    /// does or does not match the shape's constraints.
    documentation: String
}
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ShapeExamplesTraitValidator.java
```diff
@@ -0,0 +1,87 @@
...
+                    .build();
+
+            List<ValidationEvent> validationEvents = shape.accept(visitor);
+            List<ValidationEvent> nonErrorValidationEvents = validationEvents.stream()
+                    .filter(validationEvent -> validationEvent.getSeverity() != Severity.ERROR)
+                    .collect(Collectors.toList());
+
+            events.addAll(nonErrorValidationEvents);
+
+            if (validationEvents.size() == nonErrorValidationEvents.size()) {
```
JordonPhillips, line 77: I don't think it's useful to allow examples that don't match the expected type, even for the invalid list. Perhaps we should forward that along.

smithy-model/src/main/java/software/amazon/smithy/model/traits/ShapeExamplesTrait.java
```diff
@@ -0,0 +1,124 @@
...
+        private List<Node> allowed;
+        private List<Node> disallowed;
+
+        public ShapeExamplesTrait.Builder allowed(List<Node> allowed) {
+            this.allowed = allowed;
+            return this;
+        }
+
+        public ShapeExamplesTrait.Builder disallowed(List<Node> disallowed) {
+            this.disallowed = disallowed;
```
JordonPhillips, line 91: Using `BuilderRef` prevents leaking mutations and reduces copies.

```suggestion
        private BuilderRef<List<Node>> allowed = BuilderRef.forList();
        private BuilderRef<List<Node>> disallowed = BuilderRef.forList();

        public ShapeExamplesTrait.Builder allowed(List<Node> allowed) {
            this.allowed.clear();
            this.allowed.get().addAll(allowed);
            return this;
        }

        public ShapeExamplesTrait.Builder disallowed(List<Node> disallowed) {
            this.disallowed.clear();
            this.disallowed.addAll(disallowed);
```

smithy-model/src/main/java/software/amazon/smithy/model/traits/ShapeExamplesTrait.java
```diff
@@ -0,0 +1,124 @@
...
+        this.disallowed = builder.disallowed;
+        if (allowed == null && disallowed == null) {
+            throw new SourceException("One of 'allowed' or 'disallowed' must be provided.", getSourceLocation());
+        }
+        if (allowed != null && allowed.isEmpty()) {
+            throw new SourceException("'allowed' must be non-empty when provided.", getSourceLocation());
+        }
+        if (disallowed != null && disallowed.isEmpty()) {
+            throw new SourceException("'disallowed' must be non-empty when provided.", getSourceLocation());
+        }
```
JordonPhillips, line 38: Empty lists are perfectly fine, so long as at least one is populated.

```suggestion
        this.allowed = builder.allowed.copy();
        this.disallowed = builder.disallowed.copy();
        if (allowed.isEmpty() && disallowed.isEmpty()) {
            throw new SourceException("One of 'allowed' or 'disallowed' must be non-empty.", getSourceLocation());
        }
```

smithy-model/src/main/java/software/amazon/smithy/model/traits/ShapeExamplesTrait.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Gets the disallowed values.
+     *
+     * @return returns the optional disallowed values.
+     */
+    public Optional<List<Node>> getDisallowed() {
+        return Optional.ofNullable(disallowed);
+    }
```
JordonPhillips, line 57:
```suggestion
    public List<Node> getAllowed() {
        return allowed;
    }

    /**
     * Gets the disallowed values.
     *
     * @return returns the optional disallowed values.
     */
    public List<Node> getDisallowed() {
        return disallowed;
    }
```

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
robin-aws: +1. As the person with the desire mentioned :), I think it would be great to also have a `validationEventId` member stating precisely what event is expected, using [the same matching rules as suppressions](https://smithy.io/2.0/spec/model-validation.html#matching-suppression-event-ids). That would enable the validation of this trait much more precise.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ShapeExamplesTraitValidator.java
robin-aws: If you take my suggestion (https://github.com/smithy-lang/smithy/pull/2851/changes#r2834425764), this could be more precise and check that you only get events that match the expected id.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
robin-aws: I'd also suggest adding `tags`, so that there is a way to mark examples that should not be included in documentation. This would support using this trait to thoroughly test individual `@pattern` or `@conditions` applied traits without necessarily including far too many "examples" in docs.

brandondahler: >I think it would be great to also have a validationEventId member stating precisely what event is expected, 

Is the idea that this would:
1. Be specifically for invalid ShapeExample instances to say what failure(s) are expected,
2. Define the segment in the validation event id if this example fails, or
3. Something else?

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
brandondahler: Member targets were not restricted because **all** members can have constraining traits applied to them regardless of whether the underlying value can  be constrained.  For  example, a `boolean`  shape does not have any valid constraint traits available; however, a member pointing at the `boolean` shape (i.e. `Boolean`) can have `@required` applied to it which causes `null` to become invalid.

For unions - I'll have to look closer but I suspect my thought was that union types wouldn't make sense to have constraints.  I can see potential value in defining the examples at the union level instead of having to go deeper to the union's target shapes or higher to the member usage of the union -- I just need to check that the implementation actually works.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ShapeExamplesTraitValidator.java
brandondahler: I don't follow either of your two comments -- can you explain further?

This if statement is saying that if all validation events are non-error events then it **should** log an error event because this disallowed case should have resulted in  an error.

@robin-aws I think what you're alluding to matches my interpretation of your idea being "Be specifically for invalid ShapeExample instances to say what failure(s) are expected"?

JordonPhillips: For example, this would not be useful:

```smithy
@shapeExamples(invalid: [{value: 2}])
string Foo
```

It's a type mismatch. That could never be valid, so there's no confusion to clear up and no reason to have such a test case for validation purposes.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
robin-aws: FWIW `@conditions` is allowed on unions, although only because it's well-formed and not because I could think any valuable uses for it. :)

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
robin-aws: Yes to 1. It would be a way to express more precisely WHY examples are invalid, and for validation to fail if the example raises other unexpected events. For example:

e.g.

```smithy
@conditions({
  MutexAB: {
    documentation: "Only provide A or B, not both"
    expression: "!(A != null && B != null)"
  }
})
@shapeExamples({
  invalid: [
    { value: { A: "Hi", B: "There" }, validationEventId: "MutexAB"
  ]
})
structure FooInput {
  A: String
  B: String
}
```

(Not 100% sure if there needs to be a prefix on the event ID for that to be right. The main point is that the `@conditions` validation adds the ID of the condition that failed to the event id.)

Validation of the @shapeExamples trait would fail if you got some other validation failures, such as a type mismatch on `{ A: "Hi", B: 42 }`.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ShapeExamplesTraitValidator.java
robin-aws: Yeah, see https://github.com/smithy-lang/smithy/pull/2851#discussion_r2898482905 for more details on my suggestion.

This logic here would just assert that all `validationEvents` match the `validationEventId` prefix, because anything else is unexpected, error or not.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
robin-aws: Also probably the most general to support a list of `validationEventIds` in case it's difficult to trigger only one in some complicated cases.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ShapeExamplesTraitValidator.java
brandondahler: I see what each of you are saying now, makes sense

Issue #2852, opened 2025-11-14
Feature request: Support for human-readable blob values in the IDL


brandondahler:
### Context
Smithy currently supports defining blob values for traits as base64 encoded strings in both the IDL and JSON AST.  Per both the [1.0](https://github.com/smithy-lang/smithy/blob/56cca11cf6cd0664090261217eaf9d30b63d2b0b/docs/source-1.0/spec/core/model.rst?plain=1#L2387-L2389) and [2.0](https://github.com/smithy-lang/smithy/blob/56cca11cf6cd0664090261217eaf9d30b63d2b0b/docs/source-2.0/spec/model.rst?plain=1#L756-L758) specs for trait node values, blob values are "A `string` value that is base64 encoded.".

Despite the spec being clear on the intended representation, existing trait validators and related code generators do not necessarily follow this specification due to the reality that most uses of blob values really are better expressed as human-readable strings.  

For example:
* Some trait validators like the [DefaultTraitValidator only warn on invalid values](https://github.com/smithy-lang/smithy/blob/main/smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/DefaultTraitValidator.java#L107-L113)
* Some codegen implementations like [smithy-typescript's HttpProtocolTestGenerator explicitly parse the blob value as an array of utf-16 bytes](https://github.com/smithy-lang/smithy-typescript/blob/cc0124e69e2d6943f0960854bcc3abf2defd2c20/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java#L1138C58-L1138C68)

### Proposal
To support use-cases where human-readable blob values better serve the model author's expression of the value, we will support two new flavors of strings, referred to as byte string and byte text block, which are represented in the IDL as human-readable text but are automatically converted to the equivalent base64 value in the JSON AST.  These two flavors will be differentiated by a prefix of `b` in front of the double quotes of the standard quoted text or text block token respectively.  The parsing, whitespace, and escape character semantics will all work in the same manner as the standard token values.

Usage will look like the following:
``` smithy
$version: "2.0"

namespace smithy.example

@default("e30=")
blob previousString

@default(b"{}")
blob newString

@default("ewogICAgImZvbyI6ICJiYXIiCn0=")
blob previousTextBlock

@default(
    b"""
    {
        "foo": "bar"
    }
    """
)
blob newTextBlock
```

This would translate into the following JSON AST:
``` json
{
[...]

PR #2853, opened 2025-11-14
Introduce concept of byte strings and byte text blocks

brandondahler:
#### Background
* What do these changes do? 
  * Adds the ability to represent human-readable strings as properly encoded blob values
* Why are they important?
  * Blob shapes must be used in places where binary values are supported; however, the only option for representing blob values in the IDL have an inherently human unreadable representation -- base64 encoded strings. 
  * In places we actually use blob literals, the value being used is often a human readable string that has been munged by the base64 encoding.
  * By adding these new byte string flavors, we can represent these potentially binary values as human readable strings in the IDL while keeping the same base64 encoding in the JSON AST.

#### Testing
  * Unit tests updated in several different areas.
  * Primarily found and replicated places where `QUOTED_TEXT` and `TEXT_BLOCK`s were being tested
  * Documentation updates were best effort -- I do not have an environment setup that is readily capable of building and inspecting the output.

#### Links
* Implements #2852

---


sugmanue:
Hey @brandondahler, we are working on a similar feature for Smithy IDL 2.1, see pull request #3163. I'd love you feedback on the approach.

PR #2868, opened 2025-12-01
Add guidance for implementing HTTP interfaces in a Smithy client

JordonPhillips:
This adds a section to the docs to discuss guidance and reccomendations for implementing Smithy clients. Currently this is only covers HTTP interfaces, but it is intended to be expanded broadly.

---

### DIFF
--- a/.changes/next-release/documentation-b6f64a8421e29dc8eea3139ed59a6440f6fed83d.json
+++ b/.changes/next-release/documentation-b6f64a8421e29dc8eea3139ed59a6440f6fed83d.json
@@ -1,0 +1,7 @@
+{
+  "type": "documentation",
+  "description": "Added a section to discuss recommendations for implementing Smithy clients. Currently this section only includes information about HTTP interfaces, but it will expand over time to cover more topics related to implementing clients.",
+  "pull_requests": [
+    "[#2868](https://github.com/smithy-lang/smithy/pull/2868)"
+  ]
+}
--- a/docs/source-2.0/guides/client-guidance/application-protocols/http.md
+++ b/docs/source-2.0/guides/client-guidance/application-protocols/http.md
@@ -1,0 +1,189 @@
+# HTTP
+
+HTTP is the most common application protocol used by Smithy clients. This guide
+provides advice on how to integrate and expose HTTP clients.
+
+## Configuration
+
+Smithy clients should allow their users to configure the HTTP client that the
+Smithy client uses to send requests. Users may want to change the client used to
+something that has different performance characteristics or support for features
+that they need.
+
+## HTTP interfaces
+
+Smithy clients should provide interfaces for HTTP clients that standardize how
+the Smithy client interacts with the HTTP client, allowing any HTTP client to be
+used as long as it implements the interface.
+
+### Clients
+
+It is recommended to make HTTP clients implementations of
+[`ClientTransport`](#transport-clients).
+
+```java
+public interface HttpClient implements ClientTransport<HttpRequest, HttpResponse> {
+    HttpResponse send(Context context, HttpRequest request);
+}
+```
+
+#### Context
+
+HTTP clients don't have many common context parameters, but they should check
+the context for a request timeout setting and use it if it's present.
+
+```java
+/**
+ * This utility class holds shared context key definitions that are useful
+ * for HTTP implementations.
+ */
+public final class HttpContext {
+    public static final Context.Key<Duration> HTTP_REQUEST_TIMEOUT = Context.key("HTTP.RequestTimeout");
+
+    // This is a utility class that is not intended to be constructed, so it
+    // has a private constructor.
+    private HttpContext() {}
+}
+```
+
+### Requests and Responses
+
+{rfc}`9110` discusses HTTP requests and responses collectively as "messages",
+and it can be useful to encode their shared features in a shared interface.
+
+```java
+public interface HttpMessage {
+    /**
+     * Get the headers of the message.
+     *
+     * @return headers.
+     */
+    HttpFields headers();
+
+    /**
+     * Get the body of the message, or null.
+     *
+     * @return the message body or null.
+     */
+    DataStream body();
+}
+```
+
+Requests introduce the `method` and `uri` properties.
+
+```java
+public interface HttpRequest extends HttpMessage {
+    /**
+     * Get the method of the request.
+     *
+     * @return the method.
+     */
+    String method();
+
+    /**
+     * Get the URI of the request.
+     *
+     * @return the request URI.
+     */
+    URI uri();
+}
+```
+
+Responses introduce a status code.
+
+```java
+public interface HttpResponse extends HttpMessage {
+    /**
+     * Get the status code of the response.
+     *
+     * @return the status code.
+     */
+    int statusCode();
+}
+```
+
+### Fields
+
+Most users who have interacted with HTTP directly are familiar with the concept
+of headers. Headers were originally introduced in HTTP/1.0 and, since then, the
+concept of key/value pairs has expanded to include trailers and other arbitrary
+metadata. As of {rfc}`9110`, these key/value pairs are exclusively referred to
+as {rfc}`fields <9110#section-5>`.
+
+When designing HTTP interfaces for Smithy clients, be careful to understand
+field semantics. In particular, it is important to understand that field keys
+are case-insensitive and may appear more than once in an HTTP message. Since
+field keys may appear more than once, it is recommended that they are
+represented as an iterable collection of pairs or as a map whose value type is a
+list. This allows protocol implementations to safely handle joining and
+splitting.
+
+It is recommended to have utilities to convert fields to and from maps. Fields
+are often conceptualized as maps, so providing these utilities allows users to
+access fields in a more comfortable way without sacrificing correctness.
+
+```java
+public interface HttpFields extends Iterable<Map.Entry<String, List<String>>> {
+    /**
+     * Create an HttpFields instance from a map.
+     *
+     * @param fields Field map to use as a data source.
+     * @return the created fields.
+     */
+    static HttpFields of(Map<String, List<String>> fields) {
+        // This constructs a theoretical default implementation of the
+        // HttpFields interface that creates an unmodifiable copy of the given
+        // map.
+        return fields.isEmpty() ? UnmodifiableHttpFields.EMPTY : new UnmodifiableHttpFields(fields);
+    }
+
+    /**
+     * Convert the HttpFields to a map.
+     *
+     * @return the fields as a map.
+     */
+    Map<String, List<String>> toMap();
+
+    /**
+     * Check if the given field is case-insensitively present.
+     *
+     * @param name Name of the field to check.
+     * @return true if the field is present.
+     */
+    default boolean containsField(String name) {
+        return !getAllValues(name).isEmpty();
+    }
+
+    /**
+     * Get the first field value of a specific field by case-insensitive name.
+     * 
+     * Smithy clients know whether a given field should have a single value or
+     * a list value. This helper method simplifies usage for fields with a
+     * single value.
+     *
+     * @param name Name of the field to get.
+     * @return the matching field value, or null if not found.
+     */
+    default String getFirstValue(String name) {
+        var list = getAllValues(name);
+        return list.isEmpty() ? null : list.get(0);
+    }
+
+    /**
+     * Get the values of a specific field by case-insensitive name.
+     *
+     * @param name Name of the field to get the values of.
+     * @return the values of the field, or an empty list.
+     */
+    List<String> getAllValues(String name);
+}
+```
+
+#### Implementation recommendations
+
+It is not recommended to automatically attempt to join values for a given field
+key at the HTTP layer. {rfc}`9110#section-5` allows field values to be joined
+with a comma, but doing so automatically can introduce data corruption if one of
+the field values already includes a comma. {rfc}`Section 5.6 <9110#section-5.6>`
+includes productions that can help to handle those edge cases, but whether they
+are used or not is up to the protocol definition.
--- a/docs/source-2.0/guides/client-guidance/application-protocols/index.md
+++ b/docs/source-2.0/guides/client-guidance/application-protocols/index.md
@@ -1,0 +1,40 @@
+# Application Protocols
+
+This section provides guidance on how to implement and integrate different
+application protocols into a client.
+
+Application protocols define how operations are transmitted over a network. The
+most commonly used application protocol by Smithy clients is HTTP, but other
+protocols like MQTT may also be used by Smithy clients and servers. When
+designing a client, be careful to not couple any components to a particular
+application protocol unless they interact explicitly with that protocol. For
+example, an HTTP request serializer inherently needs to be coupled to HTTP, but
+a JSON serializer does not.
+
+(transport-clients)=
+## Transport clients
+
+Smithy clients and services have a common access pattern regardless of what
+application protocol is being used: a client sends requests to a server and
+receives responses. This can be represented by a simple interface:
+
+```java
+public interface ClientTransport<RequestT, ResponseT> {
+    ResponseT send(Context context, RequestT request);
+}
+```
+
+In addition to the request, it is recommended to introduce a context parameter
+to the `send` method to allow the client to be configured for each request. It
+is recommended to make this a generic context object rather than a type with
+fixed properties. Leaving it unrestricted allows context to be passed into
+custom `ClientTransport` implementations that may not be relevant to other
+implementations.
+
+## Navigation
+
+```{toctree}
+:maxdepth: 1
+
+http
+```
--- a/docs/source-2.0/guides/client-guidance/index.md
+++ b/docs/source-2.0/guides/client-guidance/index.md
@@ -11,45 +11,51 @@ generator, see
 
 ## Goals of this guide
 
-- Give advice on how to design client components.
+- Provide guidance on how to design client components.
 - Provide guidance on how to avoid coupling components to particular transport
-  protocols or to AWS-specific features.
+  protocols or to specific features.
 - Provide guidance on how to make clients extensible.
 
 ## Non-goals of this guide
 
 - Provide guidance on how to design and implement code generators.
-- Force specific implementation details. This guide is non-normative, you are
-  free to deviate from its advice.
+- Force specific implementation details. This guide is non-normative, feel free
+  to deviate from its advice.
 
 ## Tenets for Smithy clients
 
-These are the tenets of Smithy clients (unless you know better ones):
+Smithy clients should follow these tenets:
 
-1. **Smithy implementations adhere to the spec**. The Smithy spec and model
-   are the contract between clients, servers, and other implementations.
-   A Smithy client written in any programming language should be able to
-   connect to a Smithy server written in any programming language without
-   either having to care about the programming language used to implement
-   the other.
-2. **Smithy clients are familiar to developers**. Language idioms
-   and developer experience factor in to how developers and companies
-   choose between Smithy and alternatives.
-3. **Components, not monoliths**. We write modular components that
-   developers can compose together to meet their requirements. Our
-   components have clear boundaries: a client that uses an AWS protocol is
-   not required to use AWS credentials.
-4. **Our code is maintainable because we limit public interfaces**. We
-   limit the dependencies we take on. We don't expose overly open
-   interfaces that hinder our ability to evolve the code base.
-5. **No implementation stands alone**. Test cases, protocol tests, code
-   fixes, and missing abstractions have a greater impact if every Smithy
-   implementation can use them rather than just a single implementation.
-6. **Service teams don't need to know the details of every client
-   that exists or will ever exist**. When modeling a service,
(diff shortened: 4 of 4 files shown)


docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
+# HTTP
+
+HTTP is the most common application protocol used by Smithy clients. This guide
+provides advice on how to integrate and expose HTTP clients.
+
+## Configuration
+
+Smithy clients should allow their users to configure the HTTP client that the
+Smithy client uses to send requests. Users may want to change the client used to
+something that has better performance characteristics or support for features
```
kstich:
```suggestion
something that has different performance characteristics or support for features
```

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+## HTTP interfaces
+
+Smithy clients should provide interfaces for HTTP clients that standardize how
+the Smithy client interacts with the HTTP client, allowing any HTTP client to be
+used as long as it implements the interface.
+
+### Fields
+
+Most users who have interacted with HTTP directly are familiar with the concept
+of headers. Headers were originally introduced in HTTP/1.0, and since then the
```
kstich:
```suggestion
of headers. Headers were originally introduced in HTTP/1.0 and, since then, the
```

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+concept of key/value pairs has expanded to include trailers and other arbitrary
+metadata. As of {rfc}`9110`, these key/value pairs are exclusively referred to
+as {rfc}`fields <9110#section-5>`.
+
+When designing HTTP interfaces for Smithy clients, you should be careful to
```
kstich:
```suggestion
When designing HTTP interfaces for Smithy clients, be careful to
```

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+    Map<String, List<String>> toMap();
+
+    /**
+     * Check if the given field is case-insensitively present.
+     *
+     * @param name Name of the field to check.
+     * @return true if the field is present.
+     */
+    default boolean containsField(String name) {
+        return !allValues(name).isEmpty();
```
kstich:
```suggestion
        return !getAllValues(name).isEmpty();
```

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+     * 
+     * Smithy clients know whether a given header should have a single value or
+     * a list value. This helper method simplifies usage for headers with a
+     * single value.
+     *
+     * @param name Name of the field to get.
+     * @return the matching field value, or null if not found.
+     */
+    default String getFirstValue(String name) {
+        var list = allValues(name);
```
kstich:
```suggestion
        var list = getAllValues(name);
```

.changes/next-release/documentation-b6f64a8421e29dc8eea3139ed59a6440f6fed83d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "documentation",
+  "description": "Added a section to discuss reccomendations for implementing Smithy clients. Currently this section only includes information about HTTP interfaces, but it will expand over time to cover more topics related to implementing clients.",
```
kstich:
```suggestion
  "description": "Added a section to discuss recommendations for implementing Smithy clients. Currently this section only includes information about HTTP interfaces, but it will expand over time to cover more topics related to implementing clients.",
```

docs/source-2.0/guides/client-guidance/application-protocols/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+Application protocols define how operations are transmitted over a network. The
+most commonly used application protocol by Smithy clients is HTTP, but other
+protocols like MQTT may also be used by Smithy clients and servers. When
+designing a client, be careful to not couple any components to a particular
+application protocol unless they interact explicitly with that protocol. For
+example, an HTTP request serializer inherently needs to be coupled to HTTP, but
+a JSON serializer does not.
+
+(transport-clients)=
```
kstich, line 14: Is this `=` part of the syntax?

docs/source-2.0/guides/client-guidance/index.md
```diff
@@ -0,0 +1,63 @@
...
+- Give advice on how to design client components.
+- Provide guidance on how to avoid coupling components to particular transport
+  protocols or to AWS-specific features.
+- Provide guidance on how to make clients extensible.
+
+## Non-goals of this guide
+
+- Provide guidance on how to design and implement code generators.
+- Force specific implementation details. This guide is non-normative, you are
+  free to deviate from its advice.
```
kstich:
```suggestion
- Force specific implementation details. This guide is non-normative, feel
  free to deviate from its advice.
```

docs/source-2.0/guides/client-guidance/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Tenets for Smithy clients
+
+These are the tenets of Smithy clients (unless you know better ones):
```
kstich:
```suggestion
These are the tenets of Smithy clients:
```

docs/source-2.0/guides/client-guidance/index.md
```diff
@@ -0,0 +1,63 @@
...
+While topics in this guide may briefly discuss code generation, this guide
+does not describe how to build a code generator. To learn how to build a code
+generator, see
+[Creating a Code Generator](project:../building-codegen/index.rst).
+
+## Goals of this guide
+
+- Give advice on how to design client components.
+- Provide guidance on how to avoid coupling components to particular transport
+  protocols or to AWS-specific features.
```
kstich:
```suggestion
  protocols or to specific features.
```

docs/source-2.0/guides/client-guidance/application-protocols/index.md
```diff
@@ -0,0 +1,40 @@
...
+    ResponseT send(Context context, RequestT request);
+}
+```
+
+In addition to the request, it is recommended to introduce a context parameter
+to the `send` method to allow the client to be configured for each request. It
+is recommended to make this a generic context object rather than a type with
+fixed properties. Leaving it unrestricted allows context to be passed into
+custom `ClientTransport` implementations that may not be relevant to
+implementations owned by you.
```
kstich:
```suggestion
fixed properties. Leaving it unrestricted allows context to be passed into
custom `ClientTransport` implementations that may not be relevant to
other implementations.
```

docs/source-2.0/guides/client-guidance/application-protocols/index.md
JordonPhillips: Yes, this is how you define an [explicit target for cross-referencing in MyST](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#creating-explicit-targets).

docs/source-2.0/guides/client-guidance/index.md
JordonPhillips: por que

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+
+HTTP is the most common application protocol used by Smithy clients. This guide
+provides advice on how to integrate and expose HTTP clients.
+
+## Configuration
+
+Smithy clients should allow their users to configure the HTTP client that the
+Smithy client uses to send requests. Users may want to change the client used to
+something that has different performance characteristics or support for features
+that they need.
```
ianbotsf, line 11: **Style**: I think we should avoid proscriptive terms like "should" or "tenet". Instead, since this is guidance only, I think we should explore the reasons why certain features could be beneficial (you mostly already did that) and then "recommend" a solution.

For example:

```md
## Configuration

Smithy clients delegate to an HTTP client for low-level communication with remote
web services. Certain users or use cases may benefit from replacing or
customizing the HTTP client, for instance to achieve specific performance
characteristics or to use advanced HTTP features. To facilitate these scenarios,
we recommend making the HTTP client configurable in Smithy clients.
```

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+the Smithy client interacts with the HTTP client, allowing any HTTP client to be
+used as long as it implements the interface.
+
+### Fields
+
+Most users who have interacted with HTTP directly are familiar with the concept
+of headers. Headers were originally introduced in HTTP/1.0 and, since then, the
+concept of key/value pairs has expanded to include trailers and other arbitrary
+metadata. As of {rfc}`9110`, these key/value pairs are exclusively referred to
+as {rfc}`fields <9110#section-5>`.
```
ianbotsf, line 111: This flow feels odd to me. We talk about how HTTP clients should have an interface but then we jump into talking about headers. I think it may read more naturally to go top-down (vs bottom-up) and describe first the Smithy client configuration interface (not present in the current PR), then the `HttpClient` interface, then requests/responses, then fields. What do you think?

ianbotsf, line 111: I presume this `{rfc}` code is some kind of template used by Sphinx but obviously it doesn't render properly when [the file is viewed on GitHub](https://github.com/smithy-lang/smithy/blob/0abad2a4bc9c4d457e79a5f50072cfdb5cc08d43/docs/source-2.0/guides/client-guidance/application-protocols/http.md). How do we see the final rendered output?

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @param fields Field map to use as a data source.
+     * @return the created fields.
+     */
+    static HttpFields of(Map<String, List<String>> fields) {
+        // This constructs a theoretical default implementation of the
+        // HttpFields interface that creates an unmodifiable copy of the given
+        // map.
+        return fields.isEmpty() ? UnmodifiableHttpFields.EMPTY : new UnmodifiableHttpFields(fields);
+    }
```
ianbotsf, line 138: **Disclaimer**: I'm going to pick on this interface a bit since it's our first one and it's a good strawman but I'm mostly wanting to establish precedent and principles for how we approach _all_ interfaces present and future.

This interface seems a little overengineered to me, at least for an introductory read. It makes several choices that seem to have future uses in mind without first laying out why they're necessary, useful, etc. For instance:
* Why extend `Iterable`?
* Why include a static factory method?
* Why offer conversions to/from `Map`?

I'm not necessarily opposed to any of these on principle, but I don't know that I would've chosen to include them if I wrote this.

That said, I _do_ appreciate the well-written doc comments for each method, parameter, and return value. 😄 We should add docs to the interface as well!

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+#### Implementation recommendations
+
+It is not recommended to automatically attempt to join values for a given field
+key at the HTTP layer. {rfc}`9110#section-5` allows field values to be joined
+with a comma, but doing so automatically can introduce data corruption if one of
+the field values already includes a comma. {rfc}`Section 5.6 <9110#section-5.6>`
+includes productions that can help to handle those edge cases, but whether they
+are used or not is up to the protocol definition.
```
ianbotsf, line 189: What should implementors do instead? Are we implying that all HTTP client implementations will be able to _accept_ lists of header values (vs merely a single value)?

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Get the body of the message, or null.
+     *
+     * @return the message body or null.
+     */
+    DataStream body();
+}
+```
+
+Requests introduce the method and URI properties.
```
ianbotsf: **Style**: We should monospace `method` and `uri` here.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+```java
+public interface HttpResponse extends HttpMessage {
+    /**
+     * Get the status code of the response.
+     *
+     * @return the status code.
+     */
+    int statusCode();
+}
+```
```
ianbotsf: Another question that's not necessarily about this exact interface but our approach in general: would it be better to model fields like these as enums? `int` is a superset of HTTP codes so nothing precludes `0` or `-1` or `Integer.MAX_VALUE` from being returned. Additionally, nothing about this value conveys its semantic meaning (e.g., success, failure, other).

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+### Clients
+
+It is recommended to make HTTP clients implementations of
+[`ClientTransport`](#transport-clients).
+
+```java
+public interface HttpClient implements ClientTransport<HttpRequest, HttpResponse> {
+    HttpResponse send(Context context, HttpRequest request);
+}
+```
```
ianbotsf: Is `#transport-clients` a valid link? That heading appears in **index.md** but not in this doc. (I'm a Sphinx noob so maybe this is just how Sphinx works?)

docs/source-2.0/guides/client-guidance/application-protocols/http.md
```diff
@@ -0,0 +1,179 @@
...
+HTTP clients don't have many common context parameters, but they should check
+the context for a request timeout setting and use it if it's present.
+
+```java
+public final class HttpContext {
+    public static final Context.Key<Duration> HTTP_REQUEST_TIMEOUT = Context.key("HTTP.RequestTimeout");
+
+    private HttpContext() {}
+}
+```
```
ianbotsf: In general, how do we feel about access modifiers? I recognize that good library design requires careful attention to which declarations are `public`, `internal`, `private`, etc. but these interfaces are intended to be illustrative, not necessarily copy-paste solutions. I lean towards omitting any unnecessary syntax which isn't useful in conveying the intent of the abstraction, including removing `public` from basically everything.

docs/source-2.0/guides/client-guidance/application-protocols/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+Smithy clients and services have a common access pattern regardless of what
+application protocol is being used: a client sends requests to a server and
+receives responses. This can be represented by a simple interface:
+
+```java
+public interface ClientTransport<RequestT, ResponseT> {
+    ResponseT send(Context context, RequestT request);
+}
+```
```
ianbotsf, line 25: Won't most client transport implementations have resources which live outside of request/response scope (e.g., a thread pool, a task queue, etc.)? Should we capture that in the interface by extending `AutoCloseable` or adding an explicit `close` method?

docs/source-2.0/guides/client-guidance/application-protocols/http.md
ianbotsf: We touch on the notion of context here without ever really defining what it is. I suspect the notion of context varies quite a bit across languages/frameworks so we'll likely need a dedicated topic on this.

docs/source-2.0/guides/client-guidance/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Non-goals of this guide
+
+- Provide guidance on how to design and implement code generators.
+- Force specific implementation details. This guide is non-normative, feel free
+  to deviate from its advice.
+
+## Tenets for Smithy clients
+
+These are the tenets of Smithy clients (unless you know better ones):
```
ianbotsf: I like many of the ideas in this section but it reads too much like a list of requirements or mandates. I think we could rephrase this slightly (or more extensively in some cases) to be advice, desirable outcomes, etc.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
mtdowling: Status code should be an int since there’s a ton of them and the HTTP RFC defines them and groups them. An enum here would be huge and would need a fallback anyway.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: I do not agree that "should" is overly prescriptive. Should is a recommendation, not a requirement. Trying to ban it will lead to awkward phrasing.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: The documentation build CI task uploads the built docs as a zip file that you can download.

JordonPhillips: I think it's subjective. RFCs tend to build up, for example. But I have no strong feelings either way

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: > Why extend `Iterable`

This is covered in the leading paragraphs. HTTP fields aren't maps, they're iterable pairs.

> Why offer conversions to/from `Map`?

Despite the fact that fields aren't maps, people still conceptualize them this way. If you don't offer safe conversions, people will do it themselves and might do it wrong. For example, they may do a naive `map.put(k, v)` which would erase any previous values.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: HTTP implementations should do nothing. This is a serialization decision that is up to the protocol implementation.

An HTTP client that can't take a list of values for a header is not compliant with the relevant RFCs. How they represent that might vary. It's up to the interface implementation to translate into whatever the underlying client uses.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: `index.md` contains the line `(transport-clients)=`, which creates an [explicit target](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#creating-explicit-targets) that can be referenced outside of its file.

docs/source-2.0/guides/client-guidance/application-protocols/http.md
JordonPhillips: I think we need to stick to the idioms and best-practices of the language. Access modifiers have meaning and having no access modifiers has meaning. While we shouldn't venture into concepts that are overly abstract without reason, access modifiers are a fundamental concept to Java.

We talked before about our target audiences for these guidelines. At a minimum, we're expecting people who read this to have a formal education in computer science or equivalent experience. I don't expect such a person to be a Java expert, but I do expect them to be able to learn enough Java to understand these examples.

For developers who are more experienced, I expect them to be familiar with the concept of access modifiers already, if not the minutiae of how they work in Java. I expect that they know enough about their language of expertise to know if and how that concept applies.

> these interfaces are intended to be illustrative, not necessarily copy-paste solutions

While I do agree, I think the closer to copy-pasteable we are the better. For me, I think the lines are anything an IDE would collapse by default (i.e. imports) and most things that would require defining another class.

JordonPhillips: Context is talked about more in the application protocols section. I agree that it needs more... context... but I think it's big enough a topic to be its own topic. I can add that to this PR if you like, otherwise I'll do it in a followup PR

docs/source-2.0/guides/client-guidance/application-protocols/index.md
JordonPhillips: I think resource management is off topic. We might add topics about that later, but there's a lot that goes into it and a lot of it is going to be language specific.

As a Java interface I still think it isn't necessary since the underlying implementation should be free to handle resource management however it likes.

docs/source-2.0/guides/client-guidance/index.md
JordonPhillips: Some of these we can be strict on. 1 and 6 in particular are critical. I've softened the language broadly and made some editorial changes

docs/source-2.0/guides/client-guidance/application-protocols/http.md
ianbotsf: A follow up PR sounds good.

ianbotsf (approved):

docs/source-2.0/guides/client-guidance/application-protocols/http.md
lucix-aws: I am sort of on the fence. It is _leaning_ into prescriptive but I think if there's a general disclaimer on this whole thing that it's non-normative we are probably ok?

PR #2878, opened 2025-12-05
JMESPath evaluation

robin-aws:
#### Background
* What do these changes do? 
  * Adds the ability to evaluate `JmespathExpression`s
    * Defines a `JmespathRuntime<T>` interface to provide the necessary basic operations on JSON values at runtime.
    * The actual `Evaluator` is largely ported over from `smithy-java`'s [`JMESPathDocumentQuery`](https://github.com/smithy-lang/smithy-java/blob/main/jmespath/src/main/java/software/amazon/smithy/java/jmespath/JMESPathDocumentQuery.java), but also taking some inspiration from [`jmespath-java`](https://github.com/burtcorp/jmespath-java), which also has a generic interface for adapting any existing JSON value representation.
  * Adds `JmespathRuntime<T>` implementations for both `LiteralExpression` and `Node`.
  * Adds a separate `smithy-jmespath-tests` module with the compliance tests from https://github.com/jmespath/jmespath.test, and instantiates them for both runtime implementations.
* Why are they important?
  * The existing `smithy-java` evaluator only works on its own `Document` type, and evaluating expressions on generated types such as `SerializableStruct` subtypes currently requires first converting to a `Document`, which is slow. I've also written up implementations of `JmespathRuntime` for `Document` and generated types: see `smithy-java` link below.
  * JMESPath expressions are very easy to get wrong. The existing type checker/linter helps catch egregious errors but is far from complete. Having an interpreter opens the door to features where models can provide examples/tests of their expressions, possibly using the pending `@shapeExamples` trait (https://github.com/smithy-lang/smithy/pull/2851)

#### Testing
  * Mainly existing regression tests and the new `smithy-jmespath-tests` module.

#### Links
* https://github.com/smithy-lang/smithy-java/pull/967

Note for reviewers: I'm not fully satisfied with how I've arranged the code into packages. I want to minimize how much API is exposed publicly, but didn't want to add all the new types into the base package. At the same time I don't want to expose the executable version of the `Function` interface publicly yet - I want to support custom functions in the future, but I'm not yet sure the current definition is optimal for that.
* Edit: I dumped all of the Function-related types into the `evaluation` package and made them package-private. This can be revisited in a future change to support custom functions.

---


smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Override
+    public Node createNull() {
+        return Node.nullNode();
+    }
+
+    @Override
+    public Node createBoolean(boolean b) {
+        return new BooleanNode(b, SourceLocation.none());
```
mtdowling: Use a cached value. I think Node might already handle this for you.

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,184 @@
...
+        try {
+            return value.expectStringNode().getValue();
+        } catch (ExpectationNotMetException e) {
+            throw new JmespathException(JmespathExceptionType.INVALID_TYPE, "Incorrect type", e);
+        }
+    }
+
+    @Override
+    public Node createNumber(Number value) {
+        return new NumberNode(value, SourceLocation.none());
```
mtdowling: Would use Node.of (IIRC). e.g., we might choose to one day intern 0-9 or something

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,184 @@
...
+            return createNull();
+        }
+    }
+
+    @Override
+    public ObjectBuilder<Node> objectBuilder() {
+        return new ObjectNodeBuilder();
+    }
+
+    private static class ObjectNodeBuilder implements ObjectBuilder<Node> {
```
mtdowling: nit - make these inner classes final just in case it helps the JIT

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,184 @@
...
+            case STRING:
+                return EvaluationUtils.codePointCount(value.expectStringNode().getValue());
+            default:
+                throw new IllegalArgumentException();
+        }
+    }
+
+    @Override
+    public Node element(Node array, Node index) {
+        return array.expectArrayNode().get(index.expectNumberNode().getValue().intValue()).orElseGet(this::createNull);
```
mtdowling: This is pretty rough for array indexing. I think we should add a method to ArrayNode to get a value at an index or return null if not found to remove indirection and allocations here.

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,184 @@
...
+    public Node element(Node array, Node index) {
+        return array.expectArrayNode().get(index.expectNumberNode().getValue().intValue()).orElseGet(this::createNull);
+    }
+
+    @Override
+    public Iterable<Node> asIterable(Node value) {
+        if (value.isArrayNode()) {
+            return value.expectArrayNode().getElements();
+        } else {
+            return new MappingIterable<>(x -> x, value.expectObjectNode().getMembers().keySet());
```
mtdowling: Do you need the MappingIterable here? Can you just return `value.expectObjectNode().getMembers().keySet()`?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/TypeChecker.java
```diff
@@ -43,6 +43,13 @@
 import software.amazon.smithy.jmespath.ast.SliceExpression;
 import software.amazon.smithy.jmespath.ast.Subexpression;
 
+// Note this does not yet use the Evaluator with a LiteralExpressionJmespathRuntime,
```
mtdowling, line 46: is it possible to adapt to that?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/MappingIterable.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.jmespath.evaluation;
+
+import java.util.Iterator;
+import java.util.function.Function;
+
+public class MappingIterable<T, R> implements Iterable<R> {
```
mtdowling: final?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/LiteralExpressionJmespathRuntime.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import software.amazon.smithy.jmespath.ast.LiteralExpression;
+import software.amazon.smithy.jmespath.evaluation.EvaluationUtils;
+import software.amazon.smithy.jmespath.evaluation.JmespathRuntime;
+import software.amazon.smithy.jmespath.evaluation.MappingIterable;
+import software.amazon.smithy.jmespath.evaluation.NumberType;
+
+public class LiteralExpressionJmespathRuntime implements JmespathRuntime<LiteralExpression> {
```
mtdowling: final?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+package software.amazon.smithy.jmespath.evaluation;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.Objects;
+import software.amazon.smithy.jmespath.RuntimeType;
+
+public class EvaluationUtils {
```
mtdowling:
```suggestion
public final class EvaluationUtils {
```

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+                    } else {
+                        arrayStringBuilder.append(",");
+                    }
+                    arrayStringBuilder.append(toString(element));
+                }
+                arrayStringBuilder.append("]");
+                return arrayStringBuilder.toString();
+            case OBJECT:
+                StringBuilder objectStringBuilder = new StringBuilder();
+                objectStringBuilder.append("{");
```
mtdowling: for all these appends: using char for these one character appends is faster (avoid string length and array copies)

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+
+    ///////////////////////////////
+    // Common collection operations for ARRAYs and OBJECTs
+    ///////////////////////////////
+
+    /**
+     * Returns the number of elements in an ARRAY or the number of keys in an OBJECT.
+     * Otherwise, throws a JmespathException of type INVALID_TYPE.
+     */
+    Number length(T value);
```
mtdowling: Why not return integer? We will never have a string or array or collection bigger than fits in an integer. We'd OOM.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * Iterate over the elements of an ARRAY or the keys of an OBJECT.
+     * Otherwise, throws a JmespathException of type INVALID_TYPE.
+     * <p>
+     * Does not use Collection to avoid assuming there are fewer than Integer.MAX_VALUE
```
mtdowling: there will always be fewer than Integer.MAX_VALUE in the array, but iterable is a better API here

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+        void put(T key, T value);
+
+        void putAll(T object);
+
+        T build();
+    }
+
+    /**
+     * If the given value is an OBJECT, returns the value mapped to the given key.
+     * Otherwise, returns NULL.
```
mtdowling, line 315: like a Java null? `NULL` reads like it's special. Is it the result of calling createNull on the Runtime?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+        int step = stepNumber.intValue();
+
+        if (start < stop) {
+            if (step > 0) {
+                for (int idx = start; idx < stop; idx += step) {
+                    output.add(element(array, createNumber(idx)));
+                }
+            }
+        } else {
+            if (step < 0) {
```
mtdowling, line 272: else if?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,289 @@
...
+
+        void add(T value);
+
+        void addAll(T array);
+
+        T build();
+    }
+
+    /**
+     * If the given value is an ARRAY, returns the element at the given index.
```
mtdowling, line 245: what if it's null? Return null?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/MapObjectBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public void put(T key, T value) {
+        result.put(runtime.asString(key), value);
+    }
+
+    @Override
+    public void putAll(T object) {
+        for (T key : runtime.asIterable(object)) {
```
mtdowling, line 35: For arrays you had an instanceof check for a fast path of adding a map to this. Is it worth it here too?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+            return a.doubleValue() / b.doubleValue();
+        }
+    }
+
+    public static int codePointCount(String string) {
+        return string.codePointCount(0, string.length());
+    }
+
+    public static <T> boolean equals(JmespathRuntime<T> runtime, T a, T b) {
```
mtdowling, line 99: do we need null checks?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/InheritingClassMap.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Objects;
+
+/**
+ * A map using Class values as keys that accounts for subtyping.
+ * <p>
+ * Useful for external implementations of polymorphism,
+ * such as attaching behavior to an existing type hierarchy you cannot modify.
+ * Can be more efficient than a chain of if statements using instanceof.
+ */
+public class InheritingClassMap<T> {
```
mtdowling: This is only used in EvaluationUtils and I think overkill where a normal map can be used. If kept, please make it package-private.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
```diff
@@ -0,0 +1,136 @@
...
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.Objects;
+import software.amazon.smithy.jmespath.RuntimeType;
+
+public class EvaluationUtils {
+
+    public static final InheritingClassMap<NumberType> numberTypeForClass = InheritingClassMap.<NumberType>builder()
```
mtdowling: these are all final classes, right? We can just use a normal hashmap here

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
```diff
@@ -0,0 +1,346 @@
...
+            return result.build();
+        } else if (literalExpression.isObjectValue()) {
+            JmespathRuntime.ObjectBuilder<T> result = runtime.objectBuilder();
+            for (Map.Entry<String, Object> entry : literalExpression.expectObjectValue().entrySet()) {
+                T key = runtime.createString(entry.getKey());
+                T value = visit(LiteralExpression.from(entry.getValue()));
+                result.put(key, value);
+            }
+            return result.build();
+        } else if (literalExpression.isStringValue()) {
```
mtdowling: nit but string and boolean are likely more common than array and object literals

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
```diff
@@ -0,0 +1,346 @@
...
+        return runtime.value(current, runtime.createString(fieldExpression.getName()));
+    }
+
+    @Override
+    public T visitIndex(IndexExpression indexExpression) {
+        int index = indexExpression.getIndex();
+        if (!runtime.is(current, RuntimeType.ARRAY)) {
+            return runtime.createNull();
+        }
+        // TODO: Capping at int here unnecessarily
```
mtdowling: (repeating from the other comment): It'll always be < max integer, so this all is fine

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/functions/FunctionRegistry.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.jmespath.functions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class FunctionRegistry {
```
mtdowling:
```suggestion
public final class FunctionRegistry {
```

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/functions/FunctionRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static Map<String, Function> builtins = new HashMap<>();
```
mtdowling:
```suggestion
    private static final Map<String, Function> BUILTINS = new HashMap<>();
```

smithy-model/build.gradle.kts
```diff
@@ -14,6 +14,8 @@ extra["displayName"] = "Smithy :: Model"
 extra["moduleName"] = "software.amazon.smithy.model"
 
 dependencies {
+    api(project(":smithy-jmespath"))
```
mtdowling: I don't love the artifact size hit to smithy-model to unconditionally pull in jmespath. Do we need Node-based evaluation? If so, can it be a bridge package?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
```diff
@@ -0,0 +1,346 @@
...
+                continue;
+            }
+            flattened.add(val);
+        }
+        return flattened.build();
+    }
+
+    @Override
+    public T visitFunction(FunctionExpression functionExpression) {
+        Function function = FunctionRegistry.lookup(functionExpression.getName());
```
mtdowling, line 122: It'd be nice if FunctionRegistry also included a way to get all registered functions, and then that could be used to turn runtime function lookup into primitive array index access rather than the current hashmap access. Not sure how though... maybe all builtin functions get assigned a fixed index and functionExpression returns that index or -1 if it's custom. Then the parser can setup the right index, the FunctionExpression getter returns that index, and then maybe add a method to JmesPath runtime called `getFunction(FunctionExpression)`? or something like that. This would also remove the global function registry and support custom functions.

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
sugmanue: Also, similarly to what is done in `LiteralExpressionJmespathRuntime`, it would be better to test `value.isObjectNode()` and throw if it is not.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
robin-aws: No because a runtime may or may not use a Java `null` for a JSON NULL, so passing null to this method and several others will only be an error for some runtimes. I'll add something on this to the `JmespathRuntime` interface documentation though.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/Lexer.java
```diff
@@ -5,17 +5,17 @@
...
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.function.Predicate;
 import software.amazon.smithy.jmespath.ast.LiteralExpression;
+import software.amazon.smithy.jmespath.evaluation.JmespathRuntime;
 
-final class Lexer {
+final class Lexer<T> {
```
sugmanue: The type param is weird, because it returns `T` for both, values and arrays. Probably a bound will be better but for that some kind of marker interface would be needed.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: The intention is that NULL is a distinct type from ARRAY, so you get an error as per the second line. But I'll clarify the notation in the interface documentation.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: Would be technically correct, but I prefer keeping the formatting symmetrical (since you can't combine the two ifs above) to emphasize it's the same reasoning: if the step is in the wrong directly, the result is empty. I added comments to clarify this though.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: Yup the latter, will clarify at the top level

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/MapObjectBuilder.java
robin-aws: Tempting but no, added this comment though:

```java
        // A fastpath for when object is a Map doesn't quite work,
        // because you would need to know that it's specifically a Map<String, T>.
```

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
robin-aws: `Node.nullNode()` and `Node.from(boolean)` were not in fact already using cached values, but I fixed that. :)

smithy-model/src/main/java/software/amazon/smithy/model/node/NodeJmespathRuntime.java
robin-aws: Ah it was to implicitly upcast each `StringNode` to a `Node`. But I now realize that's silly when the method return type can be `Iterable<? extends Node>` instead. :)

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
robin-aws: As in a `getOrNull` method that returns a possibly-null `Node` instead of an `Optional<Node>` as `get` does?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
robin-aws: Technically BigInteger and BigDecimal aren't. Unlikely to be extended, sure, but there should be no penalty to using an `InheritingClassMap<T>` over a `HashMap<Class<?>, T>` if there are never any misses.

smithy-model/build.gradle.kts
JordonPhillips: Yeah I'd rather move this out

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
robin-aws: Yeah I was thinking through this as well. The AST would have to change somehow, perhaps to add a `FunctionIndexExpression` that you can compile `FunctionExpression`s to.

I will move all of the `Function` related types into the same package as the evaluator so that all of them can be package-private, so we can figure this out and support custom functions in a later PR.

smithy-model/build.gradle.kts
robin-aws: Yes I plan to leverage the Node-based evaluation in the next PR, which will add a new trait that uses JMESPath and wants to validate that positive/negative examples are actually correct.

I've moved the NodeJmespathRuntime class and tests to a separate `smithy-model-jmespath` package now though.

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
robin-aws: I ended up adding an `elementAt` method that should always return a non-null value (since the elements of an `ArrayNode` should use `NullNode` rather than `null`) and throws if the index is out of range. That works better here since the evaluator checks the range already for you. Let me know if that's not what you had in mind.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: For the record here was my thinking, so I don't seem nuts :)

JMESPath itself doesn't actually specify that lengths/indexes are limited to 2^32, it only says that numbers at least include integers (in the mathematical sense) and double-precision floating point (and `smithy-java`'s runtime allows for `BigInteger` and `BigDecimal` as well). Plus `ARRAY` values do not necessarily have to be limited to explicitly-allocated lists of values. You could absolutely have a runtime support "virtual arrays" such as "all long values from 0 to Long.MAX_VALUE", where the implementation of `element(i)` just returns `i`.

HOWEVER, I think having this runtime evaluation limited to `int` is perfectly reasonable, since the use case is evaluating expressions on concrete, explicit values. I would still want to make sure that any static analysis of expressions doesn't make the assumption that lengths or indexes must fit into a Java `int` though.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: 👍 https://github.com/smithy-lang/smithy/pull/2878#discussion_r2608279440

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
robin-aws: 👍 https://github.com/smithy-lang/smithy/pull/2878#discussion_r2608279440

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
mtdowling: I think we should just add a method that return the value or null. It's more general purpose and covers your use case too.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
mtdowling: Isn't there still a global and mutable FunctionRegistry being added here? I think ideally we had a way to pass the user's own function registry into the parser (using a pre-made and immutable default registry if not provided), it makes the function ast nodes maybe even with direct references to the function to call, and barring that the registry is no longer needed?

smithy-model/build.gradle.kts
mtdowling: For this PR can we leave out smithy-model-jmespath too?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
mtdowling: Yeah, lets stick to int for this codebase

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
robin-aws: I had hoped to leave custom functions out of scope for this PR and to add support later, hence making FunctionRegistry and all the related classes package-private. 

Overall I'm not yet sure of the right API for custom functions. I think you should be able to provide custom functions independently from providing a `JmespathRuntime<T>` implementation, since a custom function can use the runtime abstraction and hence work for any runtime (although runtimes could selectively override builtin or custom function implementations for efficiency). I'm also not sure if the minimal version of `Function` here is ideal since it doesn't declare any types, but we should support extending the TypeChecker functionality for custom functions as well.

smithy-model/build.gradle.kts
robin-aws: Sure I could move that to the next PR that will actually use it, that makes sense.

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
robin-aws: Sure, it means an extra range check in my case but I'm not concerned about that. Operations that will access lots of array elements will use `asIterable` instead anyway.

robin-aws: Oh, but do you want to return `null` if the array is actually storing a `NullNode` at the index, and not just if the index is out of range? That's less ideal here since I actually want the `NullNode`, but I could work with that.

mtdowling: I was thinking to return a regular Java null, basically just a pass through to Array.get, even throwing IndexOutOfBoundsException if out of range. Thoughts?

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/Evaluator.java
mtdowling: Ok sounds good since it's hidden away and we can change it. For the future implementer: I like the idea of passing them into the parser, and having references to functions from the nodes though. Worth looking at.

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
robin-aws: Yup that's what I had the first time. :) Agreed on just throwing for out of range. But AFAICT an `ArrayNode` should only ever have non-`null` elements and use `NullNode` instead. e.g. [this check in `withValue`](https://github.com/smithy-lang/smithy/blob/a5a34923b835639bd4b8d117e3fdf1dd4baf8617/smithy-model/src/main/java/software/amazon/smithy/model/node/ArrayNode.java#L147). So that means passing through to `List.get` should never return a Java `null`. I think that's the right call though.

robin-aws: Discussed offline and I'm going to go with returning a NullNode if the element is a NullNode or if the index is OOB. Will pull the Node changes into the next PR along with smithy-model-jmespath though.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/TypeChecker.java
robin-aws: Discussed more offline - it might be technically possible but it would require weakening the contract on `JmespathRuntime<T>`, e.g. to allow the special `LiteralExpression.ANY` value that isn't actually a JSON value. Instead, in the follow-up PR to support custom functions, I'm going to add a more precise type system similar to [the one for the rules engine](https://github.com/smithy-lang/smithy/tree/main/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/evaluation/type) and replace the TypeChecker code with more precise logic not based on interpretation on dummy values. So I'll leave this class alone in this PR for now.

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/JmespathExceptionType.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package software.amazon.smithy.jmespath;
+
+public enum JmespathExceptionType {
+    SYNTAX,
+    INVALID_TYPE,
+    INVALID_VALUE,
+    UNKNOWN_FUNCTION,
+    INVALID_ARITY,
+    OTHER;
```
JordonPhillips, line 43: Needs class and member docs

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,306 @@
...
+                return asNumber(value).toString();
+            case ARRAY:
+                StringBuilder arrayStringBuilder = new StringBuilder();
+                arrayStringBuilder.append('[');
+                boolean first = true;
+                for (T element : asIterable(value)) {
+                    if (first) {
+                        first = false;
+                    } else {
+                        arrayStringBuilder.append(',');
```
JordonPhillips, line 120: nit:
```suggestion
                        arrayStringBuilder.append(', ');
```

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
```diff
@@ -0,0 +1,306 @@
...
+     * <p>
+     * Runtimes may or may not use a Java null value to represent a JSON null value.
+     */
+    T createNull();
+
+    ///////////////////////////////
+    // BOOLEANs
+    ///////////////////////////////
+
+    T createBoolean(boolean b);
```
JordonPhillips, line 165: Missing docs on several of these create methods

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/LiteralExpressionJmespathRuntime.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import software.amazon.smithy.jmespath.ast.LiteralExpression;
+import software.amazon.smithy.jmespath.evaluation.EvaluationUtils;
+import software.amazon.smithy.jmespath.evaluation.JmespathRuntime;
+import software.amazon.smithy.jmespath.evaluation.MappingIterable;
+import software.amazon.smithy.jmespath.evaluation.NumberType;
+
+public final class LiteralExpressionJmespathRuntime implements JmespathRuntime<LiteralExpression> {
```
JordonPhillips, line 23: Missing docs

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/JmespathExpression.java
```diff
@@ -81,4 +108,25 @@ public LinterResult lint(LiteralExpression currentNode) {
...
     }
+
+    /**
+     * Evaluate the expression for the given current node.
+     *
+     * @param currentNode The value to set as the current node.
+     * @return Returns the result of evaluating the expression.
+     */
+    public LiteralExpression evaluate(LiteralExpression currentNode) {
+        return evaluate(currentNode, new LiteralExpressionJmespathRuntime());
```
JordonPhillips:
```suggestion
        return evaluate(currentNode, LiteralExpressionJmespathRuntime.INSTANCE);
```

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/EvaluationUtils.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+package software.amazon.smithy.jmespath.evaluation;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.Objects;
+import software.amazon.smithy.jmespath.RuntimeType;
+
+public final class EvaluationUtils {
```
JordonPhillips, line 16: Missing docs

JordonPhillips (changes requested):

smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/evaluation/JmespathRuntime.java
robin-aws: Believe it or not one of the JMESPath compliance tests requires this. :P But I added a comment.

PR #2894, opened 2025-12-16
Update smithy-docgen to consume shared snippet files to render examples

JordonPhillips:
This updates smithy-docgen to pull examples from shared snippet files generated by other build plugins. It pulls these from both explicit configuration as well as from the projection-local shared space.

---

### DIFF
--- a/smithy-docgen/src/it/java/software/amazon/smithy/docgen/PluginTest.java
+++ b/smithy-docgen/src/it/java/software/amazon/smithy/docgen/PluginTest.java
@@ -9,6 +9,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.net.URL;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -40,7 +41,7 @@ public class PluginTest {
     }
 
     @Test
-    public void pluginGeneratesMarkdown() {
+    public void pluginGeneratesMarkdown() throws Exception {
         ObjectNode settings = Node.objectNodeBuilder()
                 .withMember("service", "com.example#DocumentedService")
                 .withMember("format", "markdown")
@@ -49,7 +50,7 @@ public class PluginTest {
                         Node.objectNodeBuilder()
                                 .withMember("com.example#ExternalResource", "https://aws.amazon.com")
                                 .build())
-                .withMember("snippetConfigs", Node.fromStrings(SNIPPETS.getFile()))
+                .withMember("snippetConfigs", Node.fromStrings(Paths.get(SNIPPETS.toURI()).toString()))
                 .build();
         PluginContext context = getPluginContext(model, settings);
 
@@ -64,7 +65,7 @@ public class PluginTest {
     }
 
     @Test
-    public void pluginGeneratesSphinxMarkdown(@TempDir Path tempDir) {
+    public void pluginGeneratesSphinxMarkdown() throws Exception {
         Model model = getModel("main.smithy");
         ObjectNode settings = Node.objectNodeBuilder()
                 .withMember("service", "com.example#DocumentedService")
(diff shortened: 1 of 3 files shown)


smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/docs/SnippetConfig.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Loads a SnippetConfig from a JSON file.
+     *
+     * @param path The path to the JSON snippet config.
+     * @return Returns a SnippetConfig based on the given file.
+     */
+    public static SnippetConfig load(Path path) {
+        return fromNode(Node.parse(IoUtils.readUtf8File(path)));
```
yasmewad, line 144: Question: Do we need to add some failure checking here with a more elegant message in terms of what went wrong exactly, was it incorrect path, or the file had parsing errors? Or is it just handled in the `fromNode` and `Node.parse`

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/docs/SnippetConfig.java
```diff
@@ -0,0 +1,266 @@
...
+        private static Map<ShapeId, Map<ShapeId, List<Snippet>>> copySnippetMap(
+                Map<ShapeId, Map<ShapeId, List<Snippet>>> snippets
+        ) {
+            LinkedHashMap<ShapeId, Map<ShapeId, List<Snippet>>> copy = new LinkedHashMap<>();
+            for (Map.Entry<ShapeId, Map<ShapeId, List<Snippet>>> entry : snippets.entrySet()) {
+                LinkedHashMap<ShapeId, List<Snippet>> serviceCopy = new LinkedHashMap<>();
+                for (Map.Entry<ShapeId, List<Snippet>> serviceEntry : entry.getValue().entrySet()) {
+                    serviceCopy.put(serviceEntry.getKey(), new ArrayList<>(serviceEntry.getValue()));
+                }
+                copy.put(entry.getKey(), serviceCopy);
```
yasmewad: This and the `immutableWrapSnippetMap` share same logic other than the last part. Let's move them out to a new method and use it instead?

smithy-codegen-core/src/test/java/software/amazon/smithy/codegen/core/docs/SnippetConfigTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    public void parsesNode() throws Exception {
+        SnippetConfig actual = SnippetConfig.load(Paths.get(getClass().getResource("snippet-config.json").toURI()));
+        assertEquals(actual, snippetConfig);
+    }
+
+    @Test
+    public void roundTrips() throws Exception {
+        Node expected = Node.parse(getClass().getResource("snippet-config.json").openStream());
```
yasmewad: Question for my learning: Since we open this as a stream do we need to close the stream too?

yasmewad (commented):
The CI is failing too.

smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/docs/SnippetConfig.java
JordonPhillips: `fromNode` and `Node.parse` will handle error messaging

smithy-codegen-core/src/test/java/software/amazon/smithy/codegen/core/docs/SnippetConfigTest.java
JordonPhillips: Probably. I just switched it to not use a stream instead

JordonPhillips:
fuckin windows

yasmewad (approved):

Issue #2921, opened 2026-01-09
Smithy docgen generates incorrect links for @paginated trait


joewyz:
Smithy docgen generates incorrect links for input / output tokens in @paginated trait.

1. Links direct to .md files for html documentations.
2. Links direct to non-existing file for markdown documentations.

PR #2922, opened 2026-01-13
Fix incorrect links for operation input / output members in docgen

joewyz:
#### Background
Current Smithy Docgen generated incorrect links for operation input / output members. 
For example:
```smithy
@paginated(inputToken: "nextToken", outputToken: "nextToken", pageSize: "pageSize", items: "items")
operation PaginatedOperation {
    input := {
        nextToken: String
    
        pageSize: Integer
    }

    output := {
        items: Items

        nextToken: String
    }
}
```

The generated link for paginated trait's token would be something like:`./../shapes/PaginatedOperationInput.md#paginatedoperationinput-nexttoken`. This is incorrect because we do not generate separate files for operation input / output structures.

With this fix, Smithy Docgen can produce correct link: `./PaginatedOperation.md#paginatedoperation-nexttoken`
This should fix issue #2921 

---

### DIFF
--- a/smithy-docgen/src/main/java/software/amazon/smithy/docgen/DocSymbolProvider.java
+++ b/smithy-docgen/src/main/java/software/amazon/smithy/docgen/DocSymbolProvider.java
@@ -210,7 +210,8 @@ public final class DocSymbolProvider extends ShapeVisitor.Default<Symbol> implem
             builder.definitionFile(getDefinitionFile(serviceShape, operation));
             builder.putProperty(OPERATION_PROPERTY, operation);
             // Input and output structures should use the operation as its link_id
-            builder.putProperty(LINK_ID_PROPERTY, getLinkId(getShapeName(serviceShape, operation)));
+            String suffix = operation.getInputShape().equals(shape.getId()) ? "-request-members" : "-response-members";
+            builder.putProperty(LINK_ID_PROPERTY, getLinkId(getShapeName(serviceShape, operation)) + suffix);
         }
         return builder.build();
     }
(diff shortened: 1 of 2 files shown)


smithy-docgen/src/test/java/software/amazon/smithy/docgen/generators/OperationGeneratorTest.java
```diff
@@ -142,4 +142,50 @@ public void testGeneratesSnippetsFromDiscoveredConfig(@TempDir Path tempDir) {
...
+    public void testPaginatedOperation(@TempDir Path tempDir) {
+        MockManifest manifest = new MockManifest();
+        FileManifest sharedManifest = FileManifest.create(tempDir);
+        ObjectNode settings = settings().toBuilder()
+                .withoutMember("snippetConfigs")
+                .build();
+        sharedManifest.writeFile("snippets/snippets.json", IoUtils.readUtf8Url(SNIPPETS_FILE));
+        execute(manifest, sharedManifest, settings);
+        var operationDocs = manifest.expectFileString("/content/operations/PaginatedOperation.md");
+        System.out.println(operationDocs);
```
kstich:
```suggestion
```

smithy-docgen/src/main/java/software/amazon/smithy/docgen/DocSymbolProvider.java
```diff
@@ -209,6 +209,8 @@ public Symbol structureShape(StructureShape shape) {
             var operation = ioToOperation.get(shape.getId());
             builder.definitionFile(getDefinitionFile(serviceShape, operation));
             builder.putProperty(OPERATION_PROPERTY, operation);
+            // Input and output structures should use the operation as its link_id
+            builder.putProperty(LINK_ID_PROPERTY, getLinkId(getShapeName(serviceShape, operation)));
```
JordonPhillips: They should have a `-request-members` or `-response-members` suffix, since those targets are getting generated already. That should probably be shortened to `-input` and `-output` but it's not critical.

joewyz: Added `-request-members` and `-response-members suffix`, it would require more changes in MemberGenerator to align with the shortened suffix

JordonPhillips (commented):
This is better since we're going to the right page now, but I wonder if the links still work

JordonPhillips (approved):

kstich (approved):

Issue #2933, opened 2026-01-16
Security: Fix Shell Injection in IoUtils and Harden Docker Container


RinZ27:
### Summary
This issue details the security improvements proposed in PR #2932, specifically addressing a potential shell injection vulnerability in `IoUtils` and hardening the Docker container configuration.

### 1. Shell Injection in `IoUtils.runCommand(String)`

**The Vulnerability:**
The existing implementation of `IoUtils.runCommand(String command)` wraps the input command in a shell execution:
```java
// IoUtils.java
if (System.getProperty("os.name").toLowerCase(Locale.ENGLISH).startsWith("windows")) {
    finalizedCommand = new String[]{"cmd.exe", "/c", command};
} else {
    finalizedCommand = new String[]{"sh", "-c", command};
}
```

This pattern is susceptible to **Shell Injection** if any part of the `command` string is derived from untrusted user input. Because the command is passed as a single string to the shell (`sh -c` or `cmd.exe /c`), shell metacharacters (like `;`, `|`, `&&`, `$()`) are interpreted by the shell.

**Exploitation Example (PoC):**
Imagine a scenario where `IoUtils.runCommand` is used to echo a user-provided value:

```java
String userInput = "test; cat /etc/passwd"; // Malicious input
// Intended: echo "test; cat /etc/passwd"
// Actual execution: echo test; cat /etc/passwd
IoUtils.runCommand("echo " + userInput); 
```

*   **Result:** The shell executes `echo test` successfully, and then proceeds to execute `cat /etc/passwd`, leaking sensitive system file contents.
*   **Impact:** Arbitrary Code Execution (ACE) with the privileges of the Java process.

**The Fix:**
The PR introduces a new overload `IoUtils.runCommand(String[] command)` that utilizes `ProcessBuilder` directly with an argument array. This bypasses the shell entirely, treating arguments as literal strings.

```java
// Safe usage
String[] cmd = {"echo", userInput};
IoUtils.runCommand(cmd, ...);
```
In this case, `ProcessBuilder` passes `"test; cat /etc/passwd"` as a single argument to `echo`, resulting in the safe output of the literal string, preventing the injection.

### 2. Docker Container Hardening

**Root User Risk:**
[...]

PR #2934, opened 2026-01-17
chore: run container as non-root user and add healthcheck

RinZ27:
I've split out the Docker hardening changes as discussed in #2933. 

This PR updates the Dockerfile to:
1. Create and switch to a non-root `smithy` user. I also made sure to `chown` the `/smithy` directory so the new user has the right permissions to run the CLI and manage its class data sharing archive.
2. Add a `HEALTHCHECK` using `smithy --version` to monitor container health.

I think this covers what we talked about for the container side of things. Let me know if you want any further tweaks to the setup!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,4 +42,10 @@ RUN echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen
 RUN echo "LANG=en_US.UTF-8" > /etc/locale.conf
 ENV JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 
+# Add a non-root user and set permissions
+RUN useradd -m smithy && \
+    chown -R smithy:smithy /smithy
+
+USER smithy
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -42,4 +42,14 @@ RUN echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen
...
 
-ENTRYPOINT [ "/smithy/bin/smithy" ]
+# Add a non-root user and set permissions
+RUN useradd -m smithy && \
+    chown -R smithy:smithy /smithy
+
+USER smithy
+
+# Use a basic version check as a health indicator
+HEALTHCHECK --interval=30s --timeout=5s --start-period=5s --retries=3 \
```
mtdowling: What is the point of the health check here? This is a CLI that starts and stops, so I'm not sure what the value is in checking if it's installed every 30 seconds.

kstich:
Flickering the PR to force CI to rerun

PR #2935, opened 2026-01-20
Add @conditions, smithy-model-jmespath, NodeValidationPlugin SPI

robin-aws:
#### Background
* What do these changes do? 
  * The main contribution is `@conditions`, the first "contract trait", which specifies JMESPath expressions that must evaluate to `true`.
  * To support this:
    * Adds `smithy-model-jmespath`, to contain a `JmespathRuntime<T>` adaptor for `Node` values.
    * Adds an SPI for `NodeValidatorPlugin` so that packages outside of `smithy-model` can provide additional node validation.
    * Optimizes `NodeValidationVisitor` to index `NodeValidatorPlugin`s by `ShapeType`, to improve on the current approach that invokes all 10 existing plugins for every node regardless of type.
    * Moved the two existing copies of `ModelRuntimeTypeGenerator` and tests into `smithy-model-jmespath`, to avoid having to define a third copy.
* Why are they important?
  * This trait helps both humans and LLMs understand correct use of APIs better, especially constraints that are already manually checked in a service but not otherwise encoded/documented in the model. We reviewed an internal design doc on this that I could add to this PR if desired.

#### Testing
  * Existing `NodeValidationVisitor` tests for the optimization.
  * New tests in `smithy-model-jmespath` and `smithy-contracts`

---

### DIFF
--- a/.changes/next-release/feature-b1ef9c0e3988dbc5f87e4ac3b39201bdf1306604.json
+++ b/.changes/next-release/feature-b1ef9c0e3988dbc5f87e4ac3b39201bdf1306604.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Added a service provider interface for `NodeValidationVisitor` plugins, and optimized to index plugins by the `ShapeType` they apply to.",
+  "pull_requests": [
+    "[#2935](https://github.com/smithy-lang/smithy/pull/2935)"
+  ]
+}


docs/source-2.0/spec/contract-traits.rst
```diff
@@ -0,0 +1,75 @@
+.. _contract-traits:
+
+-----------------
+Contract traits
+-----------------
```
kstich:
```suggestion
---------------
Contract traits
---------------
```

docs/source-2.0/spec/contract-traits.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Contract traits are used to further constrain the valid values and behaviors of a model.
+Like constraint traits, contract traits are for validation only and SHOULD NOT
+impact the types signatures of generated code.
+
+
+Contract trait enforcement
+============================
```
kstich:
```suggestion
Contract trait enforcement
==========================
```

.changes/next-release/feature-077b7f1cd574ee78042363f8d8c3afb20d56f081.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add @conditions contract trait",
```
kstich: This is going to end up in a changelog, can it include more of a real description of the trait?

smithy-contract-traits/src/main/resources/META-INF/smithy/contracts.smithy
```diff
@@ -0,0 +1,23 @@
...
+
+@pattern("^[A-Z]+[A-Za-z0-9]*$")
+string ConditionName
+
+structure Condition {
+    /// JMESPath expression that must evaluate to true.
+    @required
+    expression: String
+
+    /// Description of the condition. Used in error messages when violated.
```
kstich: Any reason to not also make this required while we have the chance?

docs/source-2.0/spec/contract-traits.rst
```diff
@@ -0,0 +1,75 @@
...
+.. list-table::
+    :header-rows: 1
+    :widths: 10 23 67
+
+    * - Property
+      - Type
+      - Description
+    * - expression
+      - ``string``
+      - JMESPath expression that must evaluate to true.
```
kstich:
```suggestion
      - **Required**. JMESPath expression that must evaluate to true.
```

smithy-contract-traits/src/main/resources/META-INF/smithy/contracts.smithy
```diff
@@ -0,0 +1,23 @@
...
+@trait(selector: ":not(:test(service, operation, resource))")
+map conditions {
+    key: ConditionName
+    value: Condition
+}
+
+@pattern("^[A-Z]+[A-Za-z0-9]*$")
+string ConditionName
+
+structure Condition {
```
kstich, line 21: Please add documentation for both of these shapes as well.

smithy-contracts/src/main/resources/META-INF/smithy/contracts.smithy
```diff
@@ -0,0 +1,23 @@
+$version: "2.0"
+
+namespace smithy.contracts
+
+/// Restricts shape values to those that satisfy one or more JMESPath expressions.
+/// Each expression must produce 'true'.
```
kstich:
```suggestion
/// Each expression must produce `true`.
```

smithy-contracts/src/main/resources/META-INF/smithy/contracts.smithy
```diff
@@ -0,0 +1,23 @@
...
+map conditions {
+    key: ConditionName
+    value: Condition
+}
+
+@pattern("^[A-Z]+[A-Za-z0-9]*$")
+string ConditionName
+
+structure Condition {
+    /// JMESPath expression that must evaluate to true.
```
kstich:
```suggestion
    /// JMESPath expression that must evaluate to `true`.
```

smithy-contracts/src/main/resources/META-INF/services/software.amazon.smithy.model.validation.Validator
```diff
@@ -0,0 +1 @@
+software.amazon.smithy.contracts.ConditionsTraitValidator
```
kstich:
```suggestion
software.amazon.smithy.contracts.ConditionsTraitValidator

```

smithy-contracts/src/main/resources/META-INF/smithy/manifest
```diff
@@ -0,0 +1 @@
+contracts.smithy
```
kstich:
```suggestion
contracts.smithy

```

smithy-contract-traits/src/main/resources/META-INF/services/software.amazon.smithy.model.traits.TraitService
```diff
@@ -0,0 +1 @@
+software.amazon.smithy.contracts.ConditionsTrait$Provider
```
kstich, line 1:
```suggestion
software.amazon.smithy.contracts.ConditionsTrait$Provider

```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeFilter.java
```diff
@@ -0,0 +1,55 @@
...
+    public boolean test(Model model, Shape shape) {
+        ShapeType shapeType = shape.getType();
+
+        // Is the shape an expected shape type?
+        if (directShapeTypes.contains(shapeType)) {
+            return true;
+        }
+
+        // Is the targeted member an expected shape type?
+        return shape.asMemberShape()
```
kstich: This would be faster and easier to grok if it used a traditional `if (shape.isMemberShape())` check to bifurcate the behavior.

(Yes, I see it in the code where this came from and we probably should have done it there)

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeType.java
```diff
@@ -181,4 +182,18 @@ public static Optional<ShapeType> fromString(String text) {
                 throw new IllegalStateException("Invalid shape type: " + this);
         }
     }
+
+    public static final EnumSet<ShapeType> NUMBER_TYPES = assignableTo(NumberShape.class);
+    public static final EnumSet<ShapeType> STRING_TYPES = assignableTo(StringShape.class);
+    public static final EnumSet<ShapeType> COLLECTION_TYPES = assignableTo(CollectionShape.class);
```
kstich, line 199: This should only be `ListShape` since `set` has been removed in favor of `@uniqueItems`, would it work without this complication?

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
```diff
@@ -0,0 +1,56 @@
...
+package software.amazon.smithy.model.shapes;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.function.BiPredicate;
+import software.amazon.smithy.model.Model;
+
+public class ShapeTypeMap<T> {
```
kstich: Missing javadoc throughout this class.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeFilter.java
```diff
@@ -0,0 +1,55 @@
...
+package software.amazon.smithy.model.shapes;
+
+import java.util.EnumSet;
+import java.util.function.BiPredicate;
+import software.amazon.smithy.model.Model;
+
+/**
+ * A simple shape filter based on shape types.
+ */
+public class ShapeTypeFilter implements BiPredicate<Model, Shape> {
```
kstich: Missing javadoc throughout the innards of this class.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
```diff
@@ -0,0 +1,56 @@
...
+            for (ShapeType shapeType : ShapeType.values()) {
+                targetValues.get(shapeType).add(value);
+            }
+        }
+    }
+
+    public List<T> get(Model model, Shape shape) {
+        ShapeType shapeType = shape.getType();
+        if (shapeType.equals(ShapeType.MEMBER)) {
+            return shape.asMemberShape()
```
kstich: This can be simplified and optimized using `.expectMemberShape()` since you've done a shape type test already.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
```diff
@@ -0,0 +1,56 @@
...
+            for (ShapeType shapeType : ShapeType.values()) {
+                directValues.get(shapeType).add(value);
+            }
+            for (ShapeType shapeType : ShapeType.values()) {
+                targetValues.get(shapeType).add(value);
+            }
+        }
+    }
+
+    public List<T> get(Model model, Shape shape) {
```
kstich: This should likely return immutable copies.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
```diff
@@ -30,15 +31,38 @@
 public interface NodeValidatorPlugin {
     String[] EMPTY_STRING_ARRAY = new String[0];
 
+    /**
+     * The shapes this plugin applies to.
+     * MemberShapes that target matching shapes are also considered matching.
```
kstich: Missing an `@return`

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/BlobLengthPlugin.java
```diff
@@ -18,10 +19,10 @@
  * Validates length trait on blob shapes and members that target blob shapes.
  */
 @SmithyInternalApi
-final class BlobLengthPlugin extends MemberAndShapeTraitPlugin<BlobShape, StringNode, LengthTrait> {
+public final class BlobLengthPlugin extends MemberAndShapeTraitPlugin<StringNode, LengthTrait> {
```
kstich: What caused these all to be `public` now? I don't see them accessed outside the package.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/RangeTraitPlugin.java
```diff
@@ -4,25 +4,34 @@
...
+import software.amazon.smithy.model.shapes.ShapeTypeFilter;
 import software.amazon.smithy.model.traits.RangeTrait;
 import software.amazon.smithy.model.validation.NodeValidationVisitor;
 import software.amazon.smithy.model.validation.Severity;
 
 /**
  * Validates the range trait on number shapes or members that target them.
  */
-class RangeTraitPlugin implements NodeValidatorPlugin {
+public class RangeTraitPlugin implements NodeValidatorPlugin {
```
kstich: Should have been `final`, not adding a suggestion pending the `public` question

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/UniqueItemsPlugin.java
```diff
@@ -5,22 +5,23 @@
...
 
 /**
  * Validates that items in lists with the `@uniqueItems` trait are unique.
  */
-final class UniqueItemsPlugin extends MemberAndShapeTraitPlugin<CollectionShape, ArrayNode, UniqueItemsTrait> {
-    UniqueItemsPlugin() {
-        super(CollectionShape.class, ArrayNode.class, UniqueItemsTrait.class);
+public final class UniqueItemsPlugin extends MemberAndShapeTraitPlugin<ArrayNode, UniqueItemsTrait> {
+    public UniqueItemsPlugin() {
+        super(EnumSet.of(ShapeType.LIST, ShapeType.SET), ArrayNode.class, UniqueItemsTrait.class);
```
kstich: `@uniqueItems` was never applicable to `set` types, and it currently doesn't exist as a class implementation to extend `CollectionShape`.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
```diff
@@ -4,22 +4,40 @@
...
 abstract class FilteredPlugin<S extends Shape, N extends Node> implements NodeValidatorPlugin {
+    private final EnumSet<ShapeType> shapeTypes;
     private final Class<S> shapeClass;
     private final Class<N> nodeClass;
 
-    FilteredPlugin(Class<S> shapeClass, Class<N> nodeClass) {
+    FilteredPlugin(EnumSet<ShapeType> shapeTypes, Class<S> shapeClass, Class<N> nodeClass) {
+        for (ShapeType shapeType : shapeTypes) {
+            if (!shapeClass.isAssignableFrom(shapeType.getShapeClass())) {
+                throw new IllegalArgumentException();
```
kstich: This turns a skipped `check` call into a terminal exception throw. Is the assertion that this is safe because this is a package-private implementation?

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
```diff
@@ -4,22 +4,40 @@
...
+            }
+        }
+        this.shapeTypes = shapeTypes;
         this.shapeClass = shapeClass;
         this.nodeClass = nodeClass;
     }
 
+    @Override
+    public BiPredicate<Model, Shape> shapeMatcher() {
+        // Only direct shapes, not member shapes pointing to them
```
kstich: Where did this assertion come from? From `member`s having no `Node` that can match them?

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/ModelJmespathUtilities.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package software.amazon.smithy.model.jmespath.node;
+
+import software.amazon.smithy.jmespath.JmespathExpression;
+import software.amazon.smithy.jmespath.LinterResult;
+import software.amazon.smithy.jmespath.ast.LiteralExpression;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.Shape;
+
+public class ModelJmespathUtilities {
```
kstich:
```suggestion
public final class ModelJmespathUtilities {
```

smithy-contracts/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+plugins {
+    id("smithy.module-conventions")
+}
+
+description = "Smithy contracts."
```
kstich: Should have a better description.

smithy-contracts/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+
+extra["displayName"] = "Smithy :: Contracts"
+extra["moduleName"] = "software.amazon.smithy.contracts"
```
kstich: Should this package name contain `traits`?

docs/source-2.0/spec/index.rst
```diff
@@ -43,6 +43,7 @@ Table of contents
     service-types
     mixins
     constraint-traits
+    contract-traits
```
kstich: I think this belongs in the additional specifications like `waiters` and `protocols`, not in the core index.

smithy-contracts/src/test/resources/software/amazon/smithy/contracts/test.smithy
```diff
@@ -0,0 +1,34 @@
...
+    @required
+    end: Timestamp
+}
+
+@conditions({
+    NoKeywords: {
+        expression: "!contains(@, 'id') && !contains(@, 'name')"
+    }
+})
+string Foo
```
kstich:
```suggestion
string Foo

```

smithy-contract-traits/src/test/resources/software/amazon/smithy/contracts/errorfiles/node-validator-plugin.smithy
```diff
@@ -0,0 +1,30 @@
+$version: "2"
+
+namespace smithy.example
+
+use smithy.contracts#conditions
+
+@conditions({
+    NonEmpty: {
+        expression: "min == null || max == null || min <= max"
```
kstich, line 10: nit: should just be `<`

smithy-contracts/src/test/resources/software/amazon/smithy/contracts/errorfiles/node-validator-plugin.smithy
```diff
@@ -0,0 +1,30 @@
+$version: "2"
+
+namespace smithy.example
+
+use smithy.contracts#conditions
+
+@conditions({
+    NonEmpty: {
```
kstich: Wrong name?

smithy-contracts/src/test/resources/software/amazon/smithy/contracts/errorfiles/node-validator-plugin.smithy
```diff
@@ -0,0 +1,30 @@
...
+@range2(max: 2)
+integer Yup
+
+@range2(min: 1, max: 2)
+integer Alrighty
+
+@range2(min: 2, max: 1)
+integer Whoops
+
+
```
kstich:
```suggestion
integer Whoops

```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTraitValidator.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        return model.shapes()
+                .filter(shape -> shape.hasTrait(ConditionsTrait.ID))
+                .flatMap(shape -> validateShape(model, shape).stream())
+                .collect(Collectors.toList());
+    }
+
+    private List<ValidationEvent> validateShape(Model model, Shape shape) {
+        List<ValidationEvent> events = new ArrayList<>();
```
kstich: a) guard this with a `model.isTraitApplied` test.
b) unroll this to a `model.getShapesWithTrait`-based loop that adds to an `events` list that's passed into `validateShape`.


```suggestion
        if (!model.isTraitApplied(ConditionsTrait.class)) {
            return ListUtils.of();
        }
        
        List<ValidationEvent> events = new ArrayList<>();
        for (Shape shape : model.getShapesWithTrait(ConditionsTrait.ID)) {
            validateShape(model, shape, events);
        }
        return events;
    }

    private void validateShape(Model model, Shape shape, List<ValidationEvent> events) {
```

(and then remove the `return` in `validateShape`.)

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTraitPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+import software.amazon.smithy.jmespath.evaluation.Evaluator;
+import software.amazon.smithy.model.jmespath.node.NodeJmespathRuntime;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeType;
+import software.amazon.smithy.model.validation.NodeValidationVisitor;
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.node.MemberAndShapeTraitPlugin;
+
+public class ConditionsTraitPlugin extends MemberAndShapeTraitPlugin<Node, ConditionsTrait> {
```
kstich:
```suggestion
public final class ConditionsTraitPlugin extends MemberAndShapeTraitPlugin<Node, ConditionsTrait> {
```

kstich: Missing javadoc.

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+import software.amazon.smithy.utils.BuilderRef;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.SmithyGenerated;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+/**
+ * Restricts shape values to those that satisfy one or more JMESPath expressions.
+ * Each expression must produce 'true'.
+ */
+@SmithyGenerated
```
kstich: This is no longer generated.

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+        Map<StringNode, Node> members = node.expectObjectNode().getMembers();
+        for (Map.Entry<StringNode, Node> entry : members.entrySet()) {
+            Condition condition = Condition.fromNode(entry.getValue());
+            String name = entry.getKey().expectStringNode().getValue();
+            builder.put(name, condition);
+        }
+        return builder.build();
+    }
+
+    public Map<String, Condition> getConditions() {
```
kstich, line 66: Missing javadoc

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Creates a builder used to build a {@link ConditionsTrait}.
+     */
+    public SmithyBuilder<ConditionsTrait> toBuilder() {
+        return builder().sourceLocation(getSourceLocation())
+                .replace(getConditions());
+    }
+
+    public static Builder builder() {
```
kstich, line 81: Missing javadoc

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+            this.conditions.get().putAll(conditions);
+            return this;
+        }
+
+        public Builder clear() {
+            this.conditions.get().clear();
+            return this;
+        }
+
+        public Builder put(String name, Condition condition) {
```
kstich:
```suggestion
        public Builder putCondition(String name, Condition condition) {
```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+            this.conditions.get().put(name, condition);
+            return this;
+        }
+
+        public Builder remove(String name) {
```
kstich:
```suggestion
        public Builder removeCondition(String name) {
```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Builder for {@link ConditionsTrait}.
+     */
+    public static final class Builder extends AbstractTraitBuilder<ConditionsTrait, Builder> {
+        private final BuilderRef<Map<String, Condition>> conditions = BuilderRef.forOrderedMap();
+
+        private Builder() {}
+
+        public Builder replace(Map<String, Condition> conditions) {
```
kstich:
```suggestion
        public Builder conditions(Map<String, Condition> conditions) {
```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
```diff
@@ -0,0 +1,126 @@
...
+            clear();
+            this.conditions.get().putAll(conditions);
+            return this;
+        }
+
+        public Builder clear() {
```
kstich:
```suggestion
        public Builder clearConditions() {
```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/Condition.java
```diff
@@ -0,0 +1,146 @@
...
+import software.amazon.smithy.model.SourceException;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.ExpectationNotMetException;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ToNode;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.SmithyGenerated;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+@SmithyGenerated
```
kstich: This is no longer generated.

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/Condition.java
```diff
@@ -0,0 +1,146 @@
...
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.ExpectationNotMetException;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ToNode;
+import software.amazon.smithy.utils.SmithyBuilder;
+import software.amazon.smithy.utils.SmithyGenerated;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+@SmithyGenerated
+public final class Condition implements ToNode, ToSmithyBuilder<Condition>, FromSourceLocation {
```
kstich, line 21: Missing or invalid javadoc in various places

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/NodeJmespathRuntime.java
```diff
@@ -0,0 +1,188 @@
...
+import software.amazon.smithy.jmespath.evaluation.JmespathRuntime;
+import software.amazon.smithy.jmespath.evaluation.NumberType;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.ArrayNode;
+import software.amazon.smithy.model.node.ExpectationNotMetException;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+
+public class NodeJmespathRuntime implements JmespathRuntime<Node> {
```
kstich: Missing javadoc

kstich:
```suggestion
public final class NodeJmespathRuntime implements JmespathRuntime<Node> {
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeFilter.java
kstich:
```suggestion
public final class ShapeTypeFilter implements BiPredicate<Model, Shape> {
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
kstich:
```suggestion
public final class ShapeTypeMap<T> {
```

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTraitValidator.java
```diff
@@ -0,0 +1,116 @@
...
+import software.amazon.smithy.jmespath.LinterResult;
+import software.amazon.smithy.jmespath.RuntimeType;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.jmespath.node.ModelJmespathUtilities;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.validation.AbstractValidator;
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.ValidationEvent;
+
+public class ConditionsTraitValidator extends AbstractValidator {
```
kstich:
```suggestion
public final class ConditionsTraitValidator extends AbstractValidator {
```

kstich: Missing javadoc

smithy-model-jmespath/src/main/java/software/amazon/smithy/model/jmespath/node/ModelJmespathUtilities.java
kstich: Missing javadoc throughout

docs/source-2.0/spec/index.rst
robin-aws: I'm certainly cool with that.

Just for my own information, is the core specification meant to cover exactly the contents of `smithy-model`? Is there meant to be a 1-1 correspondence between spec sections and the individual Java packages in general?

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
robin-aws: Ah thanks for catching that. I started out using the `smithy-trait-codegen` plugin and was super pleased with it, but then decided I really wanted to parse the expression at trait construction time, and I couldn't figure out a clean way of extending it so I had to switch to manual code. FWIW it would be awesome for that plugin to support extending the result with manual code somehow as well.

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
robin-aws: FWIWI I imitated the waiters code naming style after I turned off the trait codegen plugin, but fair enough if that's not the recommended style. Is there documentation/better examples for this I missed somewhere?

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTraitValidator.java
robin-aws: Thanks, I'd had that thought but wasn't sure it was worth it given how much other validation code uses this style instead.

It could be very nice to have a Guide on implementing traits in the current Java implementation of Smithy somewhere.

smithy-contract-traits/src/main/resources/META-INF/smithy/contracts.smithy
robin-aws: I think we should. I was slightly worried about it being annoying for internal-only conditions in some cases, but I'd rather make sure anything externally visible has a description.

And just to confirm, I picked "description" rather than "documentation" to bias things towards shorter text that's reasonable to put in error messages. I couldn't find precedent elsewhere in the spec, so I'm open to feedback on that choice.

smithy-contracts/src/test/resources/software/amazon/smithy/contracts/errorfiles/node-validator-plugin.smithy
robin-aws: As in "the set of elements allowed by the range can't be empty", but I can do better :)

smithy-contract-traits/src/test/resources/software/amazon/smithy/contracts/errorfiles/node-validator-plugin.smithy
robin-aws: Doesn't the actual `@range` trait allow min and max to be the same, since they're both inclusive?

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/BlobLengthPlugin.java
robin-aws: I'd initially made them use the SPI as well which AFAICT required that, but since I reverted that I can revert these changes too.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
robin-aws: Is `@return` expected for every method?

My personal preference has always been to avoid Javadoc that doesn't add any information that isn't already in the type system, so I often omit `@param` and `@return`. I'm more than happy to conform with the guidelines for this project, but I don't know what they are.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/RangeTraitPlugin.java
robin-aws: Yup I'll revert the `public`s and then add `final` to all the plugins.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
robin-aws: It's a new error case that is only possible because of the new `EnumSet<ShapeType>` parameter, and if it's inconsistent with the `shapeClass` that's a purely internal error when constructing the plugin, not when applying it. But I'll document things better.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
robin-aws: FYI `Shape` doesn't have any `.expect...` methods, but I'm happy to just plain cast to `MemberShape` :)

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/UniqueItemsPlugin.java
robin-aws: `SetShape` does exist still though? It extends `ListShape` which extends `CollectionShape`. I wanted to make sure the `ShapeType` check was identical to the old `instanceof` check.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
robin-aws: The original test in `apply` only checks if the shape is an instance of the `shapeClass`, and doesn't check for a MemberShape that points to such a shape. That makes sense since the two subclasses are only checking restrictions that are baked into the shape type and aren't controlled by a trait.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeType.java
robin-aws: Given it's only deprecated and not literally removed from the codebase, I think it's safer to keep it. Plus the other cases are needed for sure, so it doesn't really add more complexity to use it here.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/UniqueItemsPlugin.java
robin-aws: Ah you're right that applying it to a `SetShape` was pointless before though, so I removed it from the EnumSet.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
robin-aws: Added a bit more to the comment

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/FilteredPlugin.java
robin-aws: I actually realized it was cleaner to calculate the correct EnumSet in the constructor instead of accepting it as an extra parameter, which made it no longer possible to get it wrong. :)

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
robin-aws: I decided this class wasn't really generically useful anyway and refactored it into a package private `CompositeNodeValidatorPlugin` instead. That also avoided exposing any lists.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeType.java
robin-aws: Note `CollectionLengthPlugin` uses this even though `UniqueItemsPlugin` doesn't any more.

smithy-contract-traits/src/main/resources/META-INF/smithy/contracts.smithy
kstich: I believe we use `message` for cases like this, as seen in the `TraitValidator` and others.

smithy-model/src/main/java/software/amazon/smithy/model/shapes/ShapeTypeMap.java
kstich: I was looking at the `Model` and crossed the streams; the cast works fine here.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
kstich: We publish up-to date Javadoc for each Smithy release. In furtherance of that, we add conforming javadoc (description, `param`s, `return`, `throw`s, etc.) for the whole _public_ interface (classes, interfaces, methods, etc.) aside from `Builder`s - and we do those when there's anything non-standard about the Builder.

smithy-contract-traits/src/main/resources/META-INF/smithy/contracts.smithy
```diff
@@ -0,0 +1,28 @@
...
+map conditions {
+    /// Name of the condition
+    key: ConditionName
+
+    /// Definition of the condition
+    value: Condition
+}
+
+@pattern("^[A-Z]+[A-Za-z0-9]*$")
+string ConditionName
```
kstich, line 18: Missing documentation for this.

docs/source-2.0/spec/index.rst
kstich: >is the core specification meant to cover exactly the contents of smithy-model?

Yes.

>Is there meant to be a 1-1 correspondence between spec sections and the individual Java packages in general?

Specs for things not in `smithy-model` get their own page(s) outside of the core spec. This is currently the additional specs or the AWS specs.

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTraitValidator.java
kstich: >It could be very nice to have a Guide on implementing traits in the current Java implementation of Smithy somewhere.

Agreed :)

smithy-contracts/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
kstich: I think the waiters was a miss in keeping this convention. There's not guidance on this at the moment, but there probably should be.

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/ConditionsTrait.java
robin-aws: I see javadoc here on GitHub now at least, not sure when this comment landed :)

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/Condition.java
```diff
@@ -0,0 +1,149 @@
...
+     * @return Return the documentation.
+     */
+    public String getDocumentation() {
+        return documentation;
+    }
+
+    /**
+     * Creates a builder used to build a {@link Condition}.
+     */
+    public SmithyBuilder<Condition> toBuilder() {
```
kstich:
```suggestion
    @Override
    public Builder toBuilder() {
```

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/Condition.java
```diff
@@ -0,0 +1,149 @@
...
+        return builder()
+                .expression(expressionText)
+                .documentation(documentation);
+    }
+
+    /**
+     * Creates a builder used to build an equivalent {@link Condition}.
```
kstich, line 94:
```suggestion
     * Creates a builder used to build an equivalent {@link Condition}.
     * @return the builder.
```

smithy-contract-traits/src/main/java/software/amazon/smithy/contracts/ConditionsTraitValidator.java
```diff
@@ -0,0 +1,122 @@
...
+    private void validateShape(Model model, Shape shape, List<ValidationEvent> events) {
+        ConditionsTrait conditions = shape.expectTrait(ConditionsTrait.class);
+
+        for (Map.Entry<String, Condition> entry : conditions.getConditions().entrySet()) {
+            events.addAll(validateCondition(model, shape, entry.getKey(), entry.getValue()));
+        }
+    }
+
+    private List<
+            ValidationEvent> validateCondition(Model model, Shape shape, String conditionName, Condition condition) {
```
kstich: Did the code formatter really do this?

robin-aws: It did! I put the parameters on separate lines and that avoided it.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/MemberAndShapeTraitPlugin.java
```diff
@@ -4,46 +4,41 @@
...
+import java.util.function.BiPredicate;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.node.Node;
 import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.ShapeType;
+import software.amazon.smithy.model.shapes.ShapeTypeFilter;
 import software.amazon.smithy.model.traits.Trait;
 
-abstract class MemberAndShapeTraitPlugin<S extends Shape, N extends Node, T extends Trait>
+public abstract class MemberAndShapeTraitPlugin<N extends Node, T extends Trait>
```
mtdowling, line 17: Does this need to be public for this PR?

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
```diff
@@ -30,15 +31,41 @@
...
+     *
+     * @return A {@link BiPredicate} that returns true if this plugin
+     *         has an effect on Node values of the given shape.
+     */
+    default BiPredicate<Model, Shape> shapeMatcher() {
```
mtdowling: You made a ShapeTypeFilter plugin. Why not replace this method with `ShapeTypeFilter getFilter()`?

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NonNumericFloatValuesPlugin.java
```diff
@@ -16,8 +21,13 @@ final class NonNumericFloatValuesPlugin implements NodeValidatorPlugin {
     private static final Set<String> NON_NUMERIC_FLOAT_VALUES = NonNumericFloat.stringRepresentations();
 
     @Override
-    public void apply(Shape shape, Node value, Context context, Emitter emitter) {
-        if (!(shape.isFloatShape() || shape.isDoubleShape()) || !value.isStringNode()) {
+    public BiPredicate<Model, Shape> shapeMatcher() {
+        return new ShapeTypeFilter(EnumSet.of(ShapeType.FLOAT, ShapeType.DOUBLE), EnumSet.noneOf(ShapeType.class));
```
mtdowling: All of these methods that return a new ShapeTypeFilter could probably store them in a static final and return the static value

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/MemberAndShapeTraitPlugin.java
robin-aws: It's beneficial because `ConditionsTraitPlugin` extends it, and doesn't live in `smithy-model`. I'd expect most plugins to be based on some trait so most will also want to extend this.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
robin-aws: I was intending to leave the door open to optimize further in the future, indexing by individual shape rather than just shape type. With this signature plugins can filter more precisely if they want.

I could make the return type `ShapeTypeFilter` for now, since I could replace that with a supertype later without breaking folks. But I wouldn't want to make the name more specific since I can't change that later.

robin-aws: Or rather if I make it `getFilter()` now, I might end having to add a `default BiPredicate<Model, Shape> shapeMatcher()` method later.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NonNumericFloatValuesPlugin.java
robin-aws: Pushed a commit to do this as much as possible. Ended up removing the `default` for `shapeMatcher` because it made it much easier to forget to override that. :)

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/MemberAndShapeTraitPlugin.java
robin-aws: Oh but it needs Javadoc in that case, so I added that.

smithy-model/src/main/java/software/amazon/smithy/model/validation/node/NodeValidatorPlugin.java
robin-aws: Discussed offline and the opportunity cost is just not that high realistically, so I made the change. :)

mtdowling (approved):

PR #2937, opened 2026-01-20
Negative get attr s3 tree fix

mtdowling:
#### Background
* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/.changes/next-release/feature-f8f220029d625913ed04f319494ff9fed5cad7d8.json
+++ b/.changes/next-release/feature-f8f220029d625913ed04f319494ff9fed5cad7d8.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add support for negative GetAttr indexing",
+  "pull_requests": [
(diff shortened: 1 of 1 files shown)


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/expressions/functions/GetAttr.java
```diff
@@ -100,10 +101,6 @@ private static List<Part> parse(String path, FromSourceLocation sourceLocation)
                 try {
                     String number = slicePart.substring(1, slicePart.length() - 1);
                     int slice = Integer.parseInt(number);
-                    if (slice < 0) {
```
kstich, line 103: Implementations are not going to expect negative values at the moment, we'll need some sort of version validation.

mtdowling: Added 1.1 validaton

PR #2941, opened 2026-01-22
Add client guidance for endpoint resolution

JordonPhillips:
---

### DIFF
--- a/docs/source-2.0/guides/client-guidance/endpoints.md
+++ b/docs/source-2.0/guides/client-guidance/endpoints.md
@@ -1,0 +1,199 @@
+# Endpoint Resolution
+
+Smithy clients need to know which server to connect to when making requests.
+This server location is specified by a URI. That URI, bundled with any
+additional context needed to make requests, is called an **endpoint**. This
+guide shows you how to implement flexible endpoint resolution that can handle
+different configuration options and request parameters.
+
+## Progressive flexibility
+
+For simple services using HTTP protocols, there may only be one URI to address.
+When creating a Smithy client intended for such services, the obvious solution
+to endpoint resolution is to expose a simple configuration option that takes the
+entire URI.
+
+As services scale, that simple configuration strategy may become cumbersome to
+customers. Services with geographically-distributed deployment regions may wish
+to have a simplified way to refer to those regions that customers can use as an
+alternative. There may additionally be other factors that complicate endpoint
+resolution, such as alternative endpoints dedicated to IPv6, cellular
+architecture, operation-specific endpoints, and so on. To support these needs,
+it is recommended that Smithy clients allow endpoint resolution to be flexible,
+without sacrificing the ability to easily specify a URI.
+
+## Interfaces
+
+### Endpoint
+
+A resolved endpoint should consist of a URI, context, and an optional list of
+auth schemes that the endpoint supports.
+
+```java
+public interface Endpoint {
+    /**
+     * The endpoint URI.
+     *
+     * @return URI of the endpoint.
+     */
+    URI uri();
+
+    /**
+     * @return Context relevant to the endpoint.
+     */
+    Context context();
+
+    /**
+     * @return a list of auth scheme overrides for the endpoint.
+     */
+    List<EndpointAuthScheme> authSchemes();
+}
+```
+
+#### Context
+
+Endpoints may require additional context that impacts the request outside of
+setting the URI. This context carries metadata that tells the client how to
+modify requests for this specific endpoint. For example, some AWS use cases
+require that additional headers be added to HTTP requests for certain endpoints.
+
+```java
+public static final Context.Key<Map<String, List<String>>> HEADERS = Context.key("Endpoint headers");
+```
+
+#### Auth schemes
+
+Services may need to differentiate auth schemes based on the endpoint, so
+endpoints should be able to specify a more restricted subset of auth schemes
+that are valid for that endpoint. For example, a service may need to require a
+specific auth scheme in a given region due to regulatory requirements. In other
+regions it may prefer to use a less expensive auth scheme by default.
+
+Even if a given endpoint supports all the same auth schemes that the service
+supports, it may need to provide endpoint-specific configuration for those auth
+schemes.
+
+To support these cases, it is recommended that an endpoint contain an optional
+list of supported auth schemes along with context. An empty or unset list
+indicates that the endpoint has no special auth settings.
+
+```java
+/**
+ * An authentication scheme supported for the endpoint.
+ */
+public interface EndpointAuthScheme {
+    /**
+     * The ID of the auth scheme (e.g., "aws.auth#sigv4").
+     *
+     * @return the auth scheme ID.
+     */
+    String authSchemeId();
+
+    /**
+     * @return Context relevant to the auth scheme.
+     */
+    Context context();
+}
+```
+
+(endpoint-resolver)=
+### Endpoint Resolver
+
+It is recommended that the endpoint resolver itself be a simple function that
+takes a predefined set of input parameters and returns an endpoint. This allows
+implementations to be as simple or complex as they like.
+
+```java
+/**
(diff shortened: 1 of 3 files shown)


docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+It is recommended to make the endpoint resolver params a separate object so that
+they may be constructed and passed around more simply. They should include
+information about the operation being invoked, the actual input to the
+operation, and any context available when resolving the endpoint.
+
+```java
+/**
+ * Encapsulates endpoint resolver parameters.
+ */
+public final class EndpointResolverParams {
```
kstich, line 144: nit: the `@return` docs for this could be better.

kstich (approved):

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
+# Endpoint Resolution
+
+Smithy clients need to know which server to connect to when making requests.
+This server location is specified by a URI, called an **endpoint**. This guide
```
ianbotsf: **Nit**: This phrasing makes it sound like URI and endpoint are synonymous but the below interface shows that endpoints are _more_ that simplify URIs.

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+#### Context
+
+Endpoints may require additional context that impacts the request outside of
+setting the URI. This context carries metadata that tells the client how to
+modify requests for this specific endpoint. For example, some AWS use-cases
+require that additional headers be added to HTTP requests for certain endpoints.
+
+```java
+public static final Context.Key<Map<String, List<String>>> HEADERS = Context.key("Endpoint headers");
+```
```
ianbotsf, line 62: **Question**: This is a departure from SRA where we recommended that HTTP headers be a first-class property of an endpoint. Why the change here?

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+entire URI.
+
+As services scale, that simple configuration strategy may become cumbersome to
+customers. Services with geographically-distributed deployment regions may wish
+to have a simplified way to refer to those regions that customers can use as an
+alternative. There may additionally be other factors that complicate endpoint
+resolution, such as alternative endpoints dedicated to IPv6, cellular
+architecture, operation-specific endpoints, and so on. To support these needs,
+it is recommended that Smithy clients allow endpoint resolution to be flexible,
+without sacrificing the ability to easily specify a URI.
```
ianbotsf, line 23: **Opinion**: This section doesn't recommend or even mention using separate resolver/params interfaces for each service, which SRA does. I think we should maintain that recommendation and call it out in this document.

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+     */
+    List<EndpointAuthScheme> authSchemes();
+}
+```
+
+#### Context
+
+Endpoints may require additional context that impacts the request outside of
+setting the URI. This context carries metadata that tells the client how to
+modify requests for this specific endpoint. For example, some AWS use-cases
```
ianbotsf: **Nit**: "use-cases" → "use cases"

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+It is recommended to invoke the `EndpointResolver` in every operation call,
+rather than only once upon client construction. This is because the endpoint can
+vary based on the operation, or the parameters in the operation.
+
+## Configuration
+
+It is recommended that Smithy clients provide a configuration option to set an
+`EndpointResolver`.
+
+Depending on the language or use-case of the client, a static `endpointUri`
```
ianbotsf: **Nit**: "use-case" → "use case"

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+     * @param uri The URI to always return.
+     * @return a new endpoint resolver that always returns the given URI.
+     */
+    static EndpointResolver staticUri(URI uri) { ... }
+}
+```
+
+To further simplify the common case of defining a static URI, it is recommended
+to provide helper methods that implement the interface for common URI or URL
+types.
```
ianbotsf: **Nit**: URLs are a strict subset of URIs and we don't mention the term URL anywhere else so mentioning it here feels out of place.

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### Resolver params
+
+It is recommended to make the endpoint resolver params a separate object so that
+they may be constructed and passed around more simply. They should include
+information about the operation being invoked, the actual input to the
+operation, and any context available when resolving the endpoint.
```
ianbotsf: **Nit**: While we often use `params` in code as a shortcut, we should generally spell out the complete word "parameters" in section headings and prose.

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+To further simplify the common case of defining a static URI, it is recommended
+to provide helper methods that implement the interface for common URI or URL
+types.
+
+#### Resolver params
+
+It is recommended to make the endpoint resolver params a separate object so that
```
ianbotsf: **Nit**:
* "separate" → "dedicated"
* "object" → "interface", "class", or "type"

docs/source-2.0/guides/client-guidance/endpoints.md
```diff
@@ -0,0 +1,195 @@
...
+`ApiOperation` will be defined later in a separate document. At a minimum, it
+should contain the operation's ID as well as the list of effective auth schemes
+that the operation supports.
+:::
+
+## Usage
+
+It is recommended to invoke the `EndpointResolver` in every operation call,
+rather than only once upon client construction. This is because the endpoint can
+vary based on the operation, or the parameters in the operation.
```
ianbotsf: **Opinion**: Given the interface and requirements defined above, it seems _impossible_ to invoke an endpoint resolver only once upon client construction since there's no operation, resolver params, or context. I recommend dropping this section.

ianbotsf (changes requested):

docs/source-2.0/guides/client-guidance/endpoints.md
JordonPhillips: Because Smithy is protocol agnostic and there should therefore be no coupling to protocols, even transport protocols like http. Smithy, for example, already has traits to support mqtt.

docs/source-2.0/guides/client-guidance/endpoints.md
JordonPhillips: It isn't necessary, and it makes everything more complex than it needs to be. When I was working on that for smithy-python, it was a nightmare. You can make it extensible and safe at the cost of it being fiendishly tedious to extend, or you can give up and just add a blank-check codegen hook. Perhaps tolerable for AWS SDK teams, but nobody else.

Instead this recommends providing the entire request, the schema for that request (simplified version of the model), and the request context. With all of that there's really no need to have service-specific interfaces. This is how smithy-java works, and it's perfectly capable of handling even S3's complex endpoint resolution.

docs/source-2.0/guides/client-guidance/endpoints.md
JordonPhillips: True. Perhaps there's a better place for this, since I do think the need for per-operation resolution must be justified as it isn't the obvious path.

PR #2953, opened 2026-01-30
Add new transform compileBdd and compileBddForAws

joewyz:
#### Background
These two newly added transforms will help customers to compile Binary Decision Diagram (BDD) from service shape's `@endpointRuleSet` trait and generate new trait `@endpointBdd` for the service.

---


smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/s3/CompileBddForS3.java
```diff
@@ -0,0 +1,74 @@
...
+import software.amazon.smithy.rulesengine.logic.cfg.Cfg;
+import software.amazon.smithy.rulesengine.traits.EndpointBddTrait;
+import software.amazon.smithy.rulesengine.traits.EndpointRuleSetTrait;
+import software.amazon.smithy.rulesengine.traits.EndpointTestCase;
+import software.amazon.smithy.rulesengine.traits.EndpointTestsTrait;
+
+/**
+ * A dedicated transform to compile Binary Decision Diagram (BDD) from S3.
+ */
+public class CompileBddForS3 implements ProjectionTransformer {
```
sugmanue:
```suggestion
public final class CompileBddForS3 implements ProjectionTransformer {
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/transforms/CompileBdd.java
```diff
@@ -0,0 +1,64 @@
...
+        return "compileBdd";
+    }
+
+    @Override
+    public Model transform(TransformContext transformContext) {
+        Model model = transformContext.getModel();
+        Collection<Shape> shapes = new HashSet<>();
+        Set<ServiceShape> serviceShapes = model.getServiceShapes();
+        for (ServiceShape serviceShape : serviceShapes) {
+            if (serviceShape.hasTrait(EndpointRuleSetTrait.ID)) {
```
sugmanue: Avoid the repeated work by checking that the BDD trait is not already present.
```suggestion
            if (serviceShape.hasTrait(EndpointRuleSetTrait.ID) && !serviceShape.hasTrait(EndpointBddTrait.ID)) {
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/transforms/CompileBdd.java
```diff
@@ -0,0 +1,64 @@
...
+                EndpointRuleSetTrait endpointRuleSetTrait = serviceShape.expectTrait(EndpointRuleSetTrait.class);
+                EndpointRuleSet rules = endpointRuleSetTrait.getEndpointRuleSet();
+                EndpointBddTrait bdd = compileBdd(rules);
+                shapes.add(serviceShape.toBuilder().addTrait(bdd).build());
+            }
+        }
+        return ModelTransformer.create().replaceShapes(model, shapes);
+    }
+
+    private EndpointBddTrait compileBdd(EndpointRuleSet rules) {
```
sugmanue: Add a comment here to warn the reader that the changes to this code might require changes to the other transform.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/transforms/CompileBdd.java
```diff
@@ -0,0 +1,64 @@
...
+    public Model transform(TransformContext transformContext) {
+        Model model = transformContext.getModel();
+        Collection<Shape> shapes = new HashSet<>();
+        Set<ServiceShape> serviceShapes = model.getServiceShapes();
+        for (ServiceShape serviceShape : serviceShapes) {
+            if (serviceShape.hasTrait(EndpointRuleSetTrait.ID)) {
+                EndpointRuleSetTrait endpointRuleSetTrait = serviceShape.expectTrait(EndpointRuleSetTrait.class);
+                EndpointRuleSet rules = endpointRuleSetTrait.getEndpointRuleSet();
+                EndpointBddTrait bdd = compileBdd(rules);
+                shapes.add(serviceShape.toBuilder().addTrait(bdd).build());
```
sugmanue, line 42: Any reason not to validate the test cases _if_ the `EndpointTestsTrait` is present? If there's no reason in particular, let's add the validation here as well.

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/s3/CompileBddForS3.java
```diff
@@ -0,0 +1,74 @@
...
+import software.amazon.smithy.rulesengine.logic.bdd.CostOptimization;
+import software.amazon.smithy.rulesengine.logic.bdd.SiftingOptimization;
+import software.amazon.smithy.rulesengine.logic.cfg.Cfg;
+import software.amazon.smithy.rulesengine.traits.EndpointBddTrait;
+import software.amazon.smithy.rulesengine.traits.EndpointRuleSetTrait;
+import software.amazon.smithy.rulesengine.traits.EndpointTestCase;
+import software.amazon.smithy.rulesengine.traits.EndpointTestsTrait;
+
+/**
+ * A dedicated transform to compile Binary Decision Diagram (BDD) from S3.
```
sugmanue: Let's make this transform AWS generic, and single out the S3 case internally. We don't have the use case but it's possible that we might need to special case other services (e.g., DDB). Making this transformer AWS specific guards against that case. This way we can also have a single transform for all AWS services, instead of having to special case S3 internally.

PR #2954, opened 2026-02-02
Add client retry guidance

JordonPhillips:
Notably I don't concretely define any retry strategies as AWS uses them. Should I? I started writing such a section, but a lot of it comes down to specifics of the service and tuning of the parameters.

---

### DIFF
--- a/docs/source-2.0/guides/client-guidance/retries.md
+++ b/docs/source-2.0/guides/client-guidance/retries.md
@@ -61,6 +61,11 @@ public interface RetryStrategy {
     /**
      * Invoked before the first request attempt.
      *
+     * <p>An initial request will be made even if a token cannot be acquired.
+     * It is recommended to always return a token so that the retry strategy
+     * may continue to track each request and be informed when they succeed or
+     * fail.
+     *
      * @throws TokenAcquisitionFailedException if a token cannot be acquired.
      */
     RetryToken acquireInitialToken();
@@ -99,7 +104,9 @@ to prevent race conditions.
 
 An initial retry token should be acquired at the beginning of a request, before
 the first attempt is made. If an initial token cannot be acquired, the client
-should still make an attempt.
+should still make an attempt. This initial attempt is always made because the
+purpose of the retry strategy is to manage retries, not to gate access to a
+service entirely.
 
 If an attempt fails, the retry strategy is passed the retry token for the
 attempt and given the exception raised by the attempt. If the retry strategy
@@ -407,7 +414,7 @@ demonstrate some of the potential needs of a retry system.
 
 The following is an example retry strategy that implements exponential backoff
 with jitter alongside a token bucket. This strategy adds extra cost for timeout
-errors since they may indicate a more highly degraded service.
+errors since they may indicate a more degraded service.
 
 Aside from delay, the retry token also tracks the number of attempts that have
 been made. This is necessary because this strategy imposes a maximum attempt
@@ -557,7 +564,7 @@ public final class AwsStandardRetryStrategy implements RetryStrategy {
         synchronized (tokensLock) {
             // When a successful request is made, refill the token bucket unless it
             // is already at maximum capacity.
-            if (this.tokens < MAX_CAPACITY) {
+            if (this.tokens <= MAX_CAPACITY - SUCCESS_REFUND) {
                 this.tokens += SUCCESS_REFUND;
             }
         }
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,6 +24,12 @@ todo_include_todos = False
 smartquotes = False
 nitpicky = True
 
+# -- Markdown configuration -----------------------------------------------
+
+myst_enable_extensions = [
+    "colon_fence"
+]
+
 # -- Options for HTML output ----------------------------------------------
 
 html_theme = "furo"


docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,272 @@
...
+
+If transient failures are surfaced to Smithy client users, they will likely
+implement their own retry behavior. This hand-written retry behavior may not
+include important risk mitigations that reduce the impact of outages.
+
+When a service begins to degrade, clients may notice behavior changes, such as
+an elevated number of server errors or connection failures. Clients will
+naturally want to retry these failed requests. Unfortunately, these retries have
+the effect of increasing the load on the service, which may cause the service to
+further degrade, resulting in even more retries.
```
ianbotsf, line 19: **Nit**: This paragraph uses the word "clients" but I think it should be "users".

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,272 @@
...
+are longer each time balances the desire to get a quick success with the desire
+to give the service more time to recover. Adding some randomness to that delay
+(known as **jitter**) can result in a smoother request load. This strategy,
+called **exponential backoff with jitter**, is relatively common but it isn't
+perfect.
+
+There is no perfect retry implementation. Strategies will inevitably improve
+over time as the scale of systems grows and new cascading failure conditions are
+observed. However, the right interface reflecting the problem domain can make
+sure the right extension points are available for future expansion.
```
ianbotsf: **Style**: These paragraphs discuss retry strategies and so I don't believe they belong under the header "Why is a retry system recommended?". I suggest a new section for these.

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,272 @@
...
+         */
+        OTHER;
+    }
+}
+```
+
+## Example request loop
+
+The following is a simplified example of what it looks like to use a
+`RetryStrategy` to implement a retryable request loop.
```
ianbotsf: **Nit**: This example doesn't include any of the retryability features you discussed in the previous section.

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,272 @@
...
+  [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After)
+  header. A retry strategy may choose a delay that is longer than this, but
+  should not choose a delay that is shorter than this since it is information
+  directly from the service.
+
+```java
+/**
+ * Provides retry-specific information about an error.
+ */
+public interface RetryInfo {
```
ianbotsf, line 212: **Correctness**: It's not clear to me how `RetryInfo` implementations will work (do they wrap an `Exception` and derive the relevant fields?) nor where they come from (is there some `RetryInfoFactory` that provides these instances?).

docs/source-2.0/guides/client-guidance/retries.md
JordonPhillips: This is an implementation of the part of the request pipeline that uses retry strategies. It is only concerned with creating the parameters and using the `RetryStrategy` it is not concerned with what the `RetryStrategy` does under the hood, including examining error metadata.

docs/source-2.0/guides/client-guidance/retries.md
JordonPhillips: It's an interface that errors extend. They come from code generation primarily, but can be settable properties on a more generic error.

docs/source-2.0/guides/client-guidance/retries.md
ianbotsf: Perhaps your standard retry mode example will tie it all together more strongly but the ordering of these sections feels confusing. In one section we introduce the retry strategy API, then the next section discusses possible implementation details without connecting them to that API, then the following section shows how to _use_ the strategy public API. At the very least we should reorder these sections so that we finish talking about the retry strategy API before introducing orthogonal concepts.

docs/source-2.0/guides/client-guidance/retries.md
ianbotsf: Will your standard retry strategy example show a clearer example of how they fit into a retry strategy? If not, I suggest expanding on that here.

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,566 @@
...
+}
+```
+
+:::{note}
+
+While the state of a request is intended to be included in the retry token, a
+retry strategy may still need to manage some state that is shared across the
+client. Be careful to ensure that access to that state is synchronized in order
+to prevent race conditions.
+:::
```
ianbotsf, line 101: **Nit**: This note doesn't appear to be rendering properly in the built docs:

Image

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,566 @@
...
+
+#### Using retry strategies
+
+An initial retry token should be acquired at the beginning of a request, before
+the first attempt is made. If an initial token cannot be acquired, the client
+should still make an attempt.
```
ianbotsf: **Question**: Why should the client still make an attempt if an initial token cannot be fetched? Doesn't that indicate the retry strategy thinks we should _not_ make the initial attempt?

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    @Override
+    public void recordSuccess(RetryToken token) {
+        synchronized (tokensLock) {
+            // When a successful request is made, refill the token bucket unless it
+            // is already at maximum capacity.
+            if (this.tokens < MAX_CAPACITY) {
+                this.tokens += SUCCESS_REFUND;
+            }
```
ianbotsf: **Correctness**: This code is only correct when `SUCCESS_REFUND = 1`. For larger values, this may put too many tokens back into the bucket.

docs/source-2.0/guides/client-guidance/retries.md
```diff
@@ -0,0 +1,566 @@
...
+performed. In AWS SDKs, this is how the `standard` retry mode works.
+
+These are only a few possible retry behaviors a client may have, but they
+demonstrate some of the potential needs of a retry system.
+
+### Example retry strategy
+
+The following is an example retry strategy that implements exponential backoff
+with jitter alongside a token bucket. This strategy adds extra cost for timeout
+errors since they may indicate a more highly degraded service.
```
ianbotsf: **Nit**: "a more highly degraded service" → "a more degraded service"

ianbotsf (changes requested):
Nice, I like the flow of these sections and the example retry strategy at the end really ties things all together. Just a few minor things left.

docs/source-2.0/guides/client-guidance/retries.md
JordonPhillips: This system is about managing *retry* behavior, not about gating access to the service on the client side.

On a technical level, the retry strategy may not be able to recover if no attempts are being made at all. I'll make this a bit more clear.

ianbotsf: If that's the case, why do we even need an initial token?

JordonPhillips: It's still useful for tracking request information. Some strategies might want to track success rate, for example. Or they can impose an initial delay.

kstich (approved):

PR #2955, opened 2026-02-04
Added new ERROR event to EndpointTestsTraitValidator when builtin params are inconsistent

rchache:
#### Background
* Added an ERROR level validator that checks for invalid EP tests where the param value is inconsistent with the corresponding builtInParam value.

#### Testing
* Added test cases

#### Links

---

### DIFF
--- a/.changes/next-release/feature-c6ca466b9d6f375fd8c739fce3768c560bcc4dbd.json
+++ b/.changes/next-release/feature-c6ca466b9d6f375fd8c739fce3768c560bcc4dbd.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Added new ERROR event to EndpointTestsTraitValidator when builtin params are inconsistent",
+  "pull_requests": [
+    "[#2955](https://github.com/smithy-lang/smithy/pull/2955)"
+  ]
+}


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/RuleSetParameterValidator.java
```diff
@@ -271,7 +271,7 @@ private void validateTestsParameters(
 
             // All required params from a ruleset must be present in all test cases.
             if (parameter.isRequired() && !parameter.getDefault().isPresent()
-                    && (!testSuiteHasParam || testSuiteParams.get(name).size() != trait.getTestCases().size())) {
+                    && (!testSuiteHasParam || testSuiteParams.get(name).size() < trait.getTestCases().size())) {
```
rchache, line 274: this is only a partial/bandaid fix. there is still a bug in `extractTestSuiteParameters` that double counts and causes testSuiteParams to be larger than it should be.

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
```diff
@@ -0,0 +1,82 @@
...
+import software.amazon.smithy.rulesengine.language.EndpointRuleSet;
+import software.amazon.smithy.rulesengine.traits.EndpointRuleSetTrait;
+import software.amazon.smithy.rulesengine.traits.EndpointTestCase;
+import software.amazon.smithy.rulesengine.traits.EndpointTestOperationInput;
+import software.amazon.smithy.rulesengine.traits.EndpointTestsTrait;
+
+/**
+ * Validator that the params used in endpoint test cases are valid.
+ */
+public class EndpointTestCasesParamsValidator extends AbstractValidator {
```
sugmanue:
```suggestion
public final class EndpointTestCasesParamsValidator extends AbstractValidator {
```

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private List<ValidationEvent> validateTestParamsAreConsistent(
+            ServiceShape serviceShape,
+            EndpointTestsTrait testsTrait,
+            EndpointRuleSet ruleSet
+    ) {
+        List<ValidationEvent> events = new ArrayList<>();
+
+        HashMap<String, String> builtInParameters = new HashMap<>();
```
sugmanue:
```suggestion
        Map<String, String> builtInParameters = new HashMap<>();
```

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
```diff
@@ -0,0 +1,82 @@
...
+            if (parameter.getBuiltIn().isPresent()) {
+                builtInParameters.put(parameter.getName().getName().getValue(), parameter.getBuiltIn().get());
+            }
+        });
+
+        for (EndpointTestCase testCase : testsTrait.getTestCases()) {
+            if (testCase.getOperationInputs() == null
+                    || testCase.getOperationInputs()
+                            .stream()
+                            .allMatch(opInputs -> opInputs.getBuiltInParams() == null)) {
```
sugmanue: Let's remove the use of streams, which tend to decrease performance while also increasing pressure in the GC. A utility that looks over the operation inputs can do get the work done.

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
```diff
@@ -0,0 +1,82 @@
...
+                continue;
+            }
+            for (Entry<StringNode, Node> testCaseParam : testCase.getParams().getMembers().entrySet()) {
+                if (builtInParameters.containsKey(testCaseParam.getKey().getValue())) {
+                    testCase.getOperationInputs()
+                            .stream()
```
sugmanue: Same as above, let's replace the use of streams by plain old loops.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/RuleSetParameterValidator.java
sugmanue: Do we have a plan to get the root cause fixed?

rchache: I took a look and it seemed fairly complicated at least to me. I didn't want to accidentally create a regression, so I did this minimal fix for now

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
rchache: got it, will update both

smithy-aws-endpoints/src/main/java/software/amazon/smithy/rulesengine/aws/validators/EndpointTestCasesParamsValidator.java
```diff
@@ -0,0 +1,92 @@
...
+
+        Map<String, String> builtInParameters = new HashMap<>();
+        ruleSet.getParameters().forEach(parameter -> {
+            if (parameter.getBuiltIn().isPresent()) {
+                builtInParameters.put(parameter.getName().getName().getValue(), parameter.getBuiltIn().get());
+            }
+        });
+
+        for (EndpointTestCase testCase : testsTrait.getTestCases()) {
+            if (testCase.getOperationInputs() == null || testCase.getOperationInputs().isEmpty()) {
```
sugmanue: This collection cannot be `null`.

```suggestion
            if (testCase.getOperationInputs().isEmpty()) {
```

sugmanue:
Ok, I look a bit more into this, the problem you’re to avoid is *not* AWS-specific, so I think it would be more appropriate to just do it in the class that validates the tests, [EndpointTestsTraitValidator](https://github.com/smithy-lang/smithy/blob/e713cd8c9e1624fc4d20d90f2679934799e64aa1/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/EndpointTestsTraitValidator.java#L34), there we already do similar validations but not this case, also there is already some of the information you need already computed.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/EndpointTestsTraitValidator.java
```diff
@@ -186,6 +193,45 @@ private void validateBuiltInsWithoutDefaultsHaveValues(
...
+    private void validateBuiltInsUseInconsistentValues(
+            ServiceShape serviceShape,
+            Map<String, String> builtInParameters,
+            EndpointTestCase testCase,
+            EndpointTestOperationInput testOperationInput,
+            List<ValidationEvent> events
+    ) {
+        if (testOperationInput == null) {
+            return;
+        }
```
sugmanue: `testOperationInput` cannot be `null` when the method is called.
```suggestion
```

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/EndpointTestsTraitValidator.java
```diff
@@ -186,6 +193,45 @@ private void validateBuiltInsWithoutDefaultsHaveValues(
...
+        for (Entry<StringNode, Node> testCaseParam : testCase.getParams().getMembers().entrySet()) {
+            if (!builtInParameters.containsKey(testCaseParam.getKey().getValue())) {
+                continue;
+            }
+
+            ObjectNode operationInputBuiltInParams = testOperationInput.getBuiltInParams();
+            if (operationInputBuiltInParams == null) {
+                continue;
+            }
```
sugmanue: This one cannot be null either, it's initialized to an empty `ObjectNode`.
```suggestion
```

PR #2966, opened 2026-02-12
Fix selector in RawIntegerWithoutRange examples

### LINKED ISSUE #2965: RawIntegerWithoutRange validator examples are invalid
Reference: https://smithy.io/2.0/guides/model-validation-examples.html#require-integers-to-have-a-range-constraint

The three RawIntegerWithoutRange examples do not work as-is because the selector navigates to the target shape, which is the built-in Integer type that doesn't have the trait. It should check for the trait on the member itself, not the target.

tlindsay42:
#### Background
* What do these changes do? 
  Fixes the selector in each example
* Why are they important?
  Examples should work as advertised

##### Existing documentation excerpt

Ref: <https://smithy.io/2.0/guides/model-validation-examples.html#require-integers-to-have-a-range-constraint>

> This example shows how to ***require all integers used in an operation input have a range constraint with both a minimum and maximum value.*** The first validator checks that the range trait exists on the shape, while the other two validators check that both the maximum and minimum values of the range are both filled out.

##### Existing experience

screenshot1

##### Proposed experience

screenshot2

screenshot3

#### Testing
  In my current project. AWS employee, so feel free to DM me if you have questions.

#### Links
  <https://smithy.io/2.0/guides/model-validation-examples.html#require-integers-to-have-a-range-constraint> 
  * Fixes #2965

---

### DIFF
--- a/docs/source-2.0/guides/model-validation-examples.rst
+++ b/docs/source-2.0/guides/model-validation-examples.rst
@@ -226,7 +226,7 @@ to have clear, actionable error messages.
             """,
             selector: """
                 operation -[input]-> $structure(*) > member
-                :test(> number:not([trait|range|min]):not([trait|range|max]))
+                :test(> number):not([trait|range|min]):not([trait|range|max])
                 """
         }
     },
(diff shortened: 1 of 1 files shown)


tlindsay42:
Per our discussion, you will not accept this PR because the desired outcome that description & message template in this example is intended to communicate is to create a custom integer separate with the `@range` trait and reference that. You will consider submitting your own PR to update the documentation and/or message template to make this desired outcome clearer.

Disagree & commit

##### Example

```smithy
operation GetFoo {
    input := {
        something: SomethingValue
    }
}

@range(min: 1, max, 10)
integer SomethingValue
```

PR #2992, opened 2026-03-10
Add client guidance for interceptors

JordonPhillips:
---

### DIFF
--- a/docs/source-2.0/guides/client-guidance/interceptors.md
+++ b/docs/source-2.0/guides/client-guidance/interceptors.md
@@ -31,6 +31,9 @@ to only the properties that are relevant to that hook. This makes each hook
 easier to reason about, and helps to ensure that a change in execution order
 doesn't result in different behavior.
 
+Hooks are intended to be lightweight, so blocking operations should not be
+supported.
+
 ### Hook sequence
 
 The following is an ordered list of recommended hooks. It is also recommended to
@@ -86,24 +89,28 @@ protocols, these are HTTP requests and HTTP responses.
 8. **readAfterExecution** *(immutable)* — The last thing called during an
    execution.
 
-### Error behavior
+### Error handling
 
-Errors raised in hooks are handled consistently. The behavior depends on where
-in the pipeline the hook is called:
+Error handling behavior depends on where in the pipeline the hook is called and
+whether or not the hook is mutable.
 
-- **Hooks called once per execution** (`readBeforeExecution`,
-  `readAfterExecution`): Errors are collected across all interceptors before any
-  further action is taken. If multiple interceptors raise errors, the last error
-  wins and earlier ones are logged and dropped.
+#### Error accumulation
 
-- **Most other hooks**: An error immediately jumps execution to
-  `modifyBeforeAttemptCompletion` (if inside the retry loop) or
-  `modifyBeforeCompletion` (if outside), with the error set as the result.
+When an immutable hook is being executed, all interceptors are always invoked
+and any errors thrown by those interceptors are accumulated before any further
+action is taken. The last error thrown by an interceptor is re-thrown. If the
+language supports suppressed errors, the other errors should be suppressed. If
+not, the other errors should be logged and dropped. The may also be attached to
+the raised error as metadata.
 
-- **`readAfterAttempt`**: Errors are collected the same way as
-  `readBeforeExecution`. After all interceptors have been called, the
-  [retry strategy](#client-guidance-retries) decides whether to retry or proceed
-  to `modifyBeforeCompletion`.
+When a mutable hook is being executed, the first error thrown is immediately
+forwarded on and no other interceptors are invoked.
+
+#### Control flow
+
+When an error is raised in a hook outside the retry loop, execution immediately
+jumps to `modifyBeforeCompletion`. When an error is raised outside the retry
+loop, execution immediately jumps to `modifyBeforeAttemptCompletion`.
 
 ## Interfaces
 
@@ -138,7 +145,22 @@ public class ResponseHook<I, O, RequestT, ResponseT> extends RequestHook<I, O, R
 // Available from readAfterDeserialization and modifyBeforeAttemptCompletion onward.
 // Adds the deserialized output (may be null if the attempt failed).
 public class OutputHook<I, O, RequestT, ResponseT> extends ResponseHook<I, O, RequestT, ResponseT> {
-    public O output() { ... }
+    public Optional<O> output() { ... }
+
+    public Optional<RuntimeException> error() { ... }
+    
+    /**
+     * If an exception {@code e} is provided, throw it, otherwise return the output value.
+     *
+     * @param e Error to potentially rethrow.
+     * @return the output value.
+     */
+    public O forward(RuntimeException e) {
+        if (e != null) {
+            throw e;
+        }
+        return output;
+    }
 }
 ```
 
@@ -213,6 +235,19 @@ public interface Interceptor {
 }
 ```
 
+:::{admonition} Error modeling
+:class: important
+
+In these interfaces, errors are provided separately from the `OutputHook` type.
+This is a minor optimization that eliminates the need to allocate new `OutputHook`
+instances when errors are thrown.
+
+In languages that have a native union type (such as Python) or a native `Result`
+type (such as Rust), the usability benefits of integrating the error into the
+`OutputHook` may be preferrable.
+
+:::
+
 ## Example
 
 The following interceptor adds a tracing header to HTTP requests when running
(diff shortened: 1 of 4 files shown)


docs/source-2.0/guides/client-guidance/interceptors.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Error behavior
+
+Errors raised in hooks are handled consistently. The behavior depends on where
+in the pipeline the hook is called:
+
+- **Hooks called once per execution** (`readBeforeExecution`,
+  `readAfterExecution`): Errors are collected across all interceptors before any
+  further action is taken. If multiple interceptors raise errors, the last error
+  wins and earlier ones are logged and dropped.
```
ianbotsf: **Nit**: I agree the last error wins but I don't agree the others should be dropped. I believe that languages which support suppressed errors or error metadata should still attach those prior failures to the eventual error that gets thrown.

docs/source-2.0/guides/client-guidance/interceptors.md
```diff
@@ -0,0 +1,303 @@
...
+8. **readAfterExecution** *(immutable)* — The last thing called during an
+   execution.
+
+### Error behavior
+
+Errors raised in hooks are handled consistently. The behavior depends on where
+in the pipeline the hook is called:
+
+- **Hooks called once per execution** (`readBeforeExecution`,
+  `readAfterExecution`): Errors are collected across all interceptors before any
```
ianbotsf: **Nit**: This parenthetical list is not exhaustive so should start with "for example". For example:

```markdown
- **Hooks called once per execution** (for example, `readBeforeExecution` and `readAfterExecution`): Errors are collected...
```

docs/source-2.0/guides/client-guidance/interceptors.md
```diff
@@ -0,0 +1,303 @@
...
+  further action is taken. If multiple interceptors raise errors, the last error
+  wins and earlier ones are logged and dropped.
+
+- **Most other hooks**: An error immediately jumps execution to
+  `modifyBeforeAttemptCompletion` (if inside the retry loop) or
+  `modifyBeforeCompletion` (if outside), with the error set as the result.
+
+- **`readAfterAttempt`**: Errors are collected the same way as
+  `readBeforeExecution`. After all interceptors have been called, the
```
ianbotsf: **Nit**: There's nothing special about `readBeforeExecution` that makes it a template for other error handling. It's the same pattern for all read-only hooks which I think should be closer to how we phrase this. I might suggest:

```markdown
- **`readAfterAttempt`**: Errors are collected the same way as other immutable hooks (for example, `readBeforeExecution`). After all interceptors have been called...
```

docs/source-2.0/guides/client-guidance/interceptors.md
ianbotsf: **Open discussion**: In my opinion, the SRA does not offer sufficient guidance on error handling within interceptor hooks. None of the prose mentions errors and only the interface JavaDocs offer any hints about how to handle errors in specific hooks. The pattern you've identified here ("hooks called once per execution") doesn't precisely match those JavaDocs:

| Hook                            | Mutability | Execution count | SRA error behavior | Draft SCG error behavior |
|---------------------------------|------------|-----------------|--------------------|--------------------------|
| `readBeforeExecution`           | Immutable  | 1 per call      | Accumulate         | Accumulate               |
| `modifyBeforeSerialization`     | Mutable    | 1 per call      | Throw first        | Accumulate               |
| `readBeforeSerialization`       | Immutable  | 1 per call      | Throw first        | Accumulate               |
| `readAfterSerialization`        | Immutable  | 1 per call      | Throw first        | Accumulate               |
| `modifyBeforeRetryLoop`         | Mutable    | 1 per call      | Throw first        | Accumulate               |
| `readBeforeAttempt`             | Immutable  | 1 per attempt   | Accumulate         | Throw first              |
| `modifyBeforeSigning`           | Mutable    | 1 per attempt   | Throw first        | Throw first              |
| `readBeforeSigning`             | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `readAfterSigning`              | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `modifyBeforeTransmit`          | Mutable    | 1 per attempt   | Throw first        | Throw first              |
| `readBeforeTransmit`            | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `readAfterTransmit`             | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `modifyBeforeDeserialization`   | Mutable    | 1 per attempt   | Throw first        | Throw first              |
| `readBeforeDeserialization`     | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `readAfterDeserialization`      | Immutable  | 1 per attempt   | Throw first        | Throw first              |
| `modifyBeforeAttemptCompletion` | Mutable    | 1 per attempt   | Throw first        | Throw first              |
| `readAfterAttempt`              | Immutable  | 1 per attempt   | Accumulate         | Throw first              |
| `modifyBeforeCompletion`        | Mutable    | 1 per call      | Throw first        | Accumulate               |
| `readAfterExecution`            | Immutable  | 1 per call      | Accumulate         | Accumulate               |

I'd like to propose a more straightforward rule: immutable hooks accumulate errors, mutable hooks do not. It doesn't matter whether the hook is per call or per attempt, it doesn't matter where/when in the pipeline it happens, it's just down to mutability.

What do we think about adopting that stance?

docs/source-2.0/guides/client-guidance/interceptors.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Interfaces
+
+### Hook input types
+
+Each hook receives a typed input object that contains only the data available at
+that stage of the pipeline, along with a [context object](#typed-context). Using
+typed inputs prevents interceptors from accidentally accessing data that doesn't
+exist yet (for example, trying to read the transport response before a request
+has been sent).
```
ianbotsf, line 123: **Correctness**: These interfaces don't comprehend error tracking, only success tracking. The SRA used a fictitious `Result<SuccessType, ErrorType>` to handle this.

docs/source-2.0/guides/client-guidance/interceptors.md
```diff
@@ -0,0 +1,303 @@
...
+
+The following interceptor adds a tracing header to HTTP requests when running
+inside AWS Lambda. It uses `modifyBeforeTransmit` because the header needs to be
+added to the transport request after signing. Adding it before signing would
+cause the signature to include the header. That would be fine, but this
+particular header is added after signing in practice.
+
+```java
+public class AddTraceHeader implements ClientInterceptor {
+    private final String traceId;
```
ianbotsf, line 261: **Nit**: This example calls the interface `ClientInterceptor` but your spec above calls it just `Interceptor`. We should standardize on one (I like just `Interceptor`).

docs/source-2.0/guides/client-guidance/interceptors.md
ianbotsf, line 1: **Nit**: This draft omits SRA's guidance on the type of code that should run in hooks:

> Ensure that customers do not perform blocking operations in interceptor hooks. Where this is impossible, they are documented that the hooks **must not** perform blocking operations.

I think we should carry this forward with some light rewording.

ianbotsf (changes requested):

docs/source-2.0/guides/client-guidance/interceptors.md
JordonPhillips: Rust-style result types aren't idiomatic in Java. I can add some more details about that option though.

ianbotsf: `Result` is one possible way of handling but not the only one...we could add `@Nullable Throwable` members to the various hook interfaces. Whatever interface we show must be able to handle prior successful responses _and_ failed ones.

docs/source-2.0/guides/client-guidance/interceptors.md
JordonPhillips: My translation here definitely wasn't correct, and I'll fix that. The intent of the SRA accumulating hooks was to guarantee that each interceptor is called because those hooks in particular represent key lifecycle checkpoints where resources are likely to be created or destroyed.

I think that accumulating on all the immutable hooks is probably fine. It can let interceptors know how far a request got if they care to track that. And since they can't modify anything anyway, it should be fine.

mtdowling: I think there are different types of mutability. The current input/output/request/response -- we can constrain mutability. The content -- we should not constrain mutability. It's the thing that can accumulate state over the lifetime of a call and allows interceptors to be stateless. I don't think there's a strong reason to mandate immutability of context.

mtdowling (approved):

PR #3001, opened 2026-03-17
Change the transform to add new shapes before removing old ones

sugmanue:
#### Background

Change the `CreateDedicatedInputAndOutput` transform to add new shapes before removing old ones. If the shapes are removed first as it was done before this change, when the operation has the `clientDiscoveredEndpoint` the transform
`CleanClientDiscoveryTraitTransformer` that handles the removal of shapes throws an exception because the operation that's expecting is not longer in the model. To avoid this issue, we change the order of the transformation operations, first we add the new shapes, and after that we remove the old ones. This ensures that when the `CleanClientDiscoveryTraitTransformer` is run the expected shapes are present.

#### Testing

Added a test to the transform.

#### Links

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
@@ -87,8 +87,7 @@ final class CreateDedicatedInputAndOutput {
             }
             // Handle a corner case when the operation uses the same structure for input and output and
             // it is named as expected for one of them, e.g., GetFoo with input and output GetFooInput
-            // In that case we don't want to delete it or the newly introduced input as will be deleted
-            // as well.
+            // In that case we don't want to delete it or the newly introduced input will be deleted as well.
             if (removeInput && removeOutput) {
                 toRemove.add(input);
                 toRemove.add(output);


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
```
sugmanue, line 3: This change seems to fix the issue with spotless.

smithy-model/src/main/java/software/amazon/smithy/model/transform/CreateDedicatedInputAndOutput.java
```diff
@@ -79,18 +81,30 @@ Model transform(ModelTransformer transformer, Model model) {
...
                 if (idChanged && isSingularReference(reverse, output, operation)) {
-                    toRemove.add(output);
+                    removeOutput = true;
                     LOGGER.fine("Removing now unused output shape " + output.getId());
                 }
             }
+            // Handle a corner case when the operation uses the same structure for input and output and
+            // it is named as expected for one of them, e.g., GetFoo with input and output GetFooInput
+            // In that case we don't want to delete it or the newly introduced input as will be deleted
+            // as well.
```
kstich:
```suggestion
            // In that case we don't want to delete it or the newly introduced input will be deleted as well.
```

PR #3004, opened 2026-03-18
Fix issues handling high-precision numbers

kstich:
The Double.isFinite method was incorrectly used to detect if a value would fit in a double. The method handles this properly, but it's not actually what we want - we want to know if the value will be accurately represented by a double.

To resolve this, a NumberUtils class is added with a method that validates precision compatibility. Tests are added for this and all call sites to ensure the behavior remains consistent.

---

### DIFF
--- a/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/Lexer.java
+++ b/smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/Lexer.java
@@ -10,6 +10,7 @@ import java.util.Objects;
 import java.util.function.Predicate;
 import software.amazon.smithy.jmespath.ast.LiteralExpression;
 import software.amazon.smithy.jmespath.evaluation.JmespathRuntime;
+import software.amazon.smithy.utils.NumberUtils;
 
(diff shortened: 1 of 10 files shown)


smithy-jmespath/src/main/java/software/amazon/smithy/jmespath/Lexer.java
```diff
@@ -470,7 +472,23 @@ private Token parseNumber() {
         try {
             Number number;
             if (lexeme.contains(".") || lexeme.toLowerCase().contains("e")) {
-                number = Double.parseDouble(lexeme);
+                Double doubleValue = Double.parseDouble(lexeme);
+                if (doubleValue.isInfinite() || doubleValue.isNaN()) {
+                    // Doubles parse values outside their range, but set their values
+                    // to infinity. See if we got a real infinity or just a big value.
+                    try {
```
JordonPhillips: Why not check the string? Also, can this entire logic block be put into `NumberUtils`? Otherwise it's being repeated thrice.

JordonPhillips (changes requested):

PR #3008, opened 2026-03-21
Fix cloudWatchNamespace's doc in aws.api#service

sorawee:
#### Background
`CloudWatchMetricNamespace` is the type, but the field name is actually called `cloudWatchNamespace`.

#### Testing
No need to test. Documentation change only.

#### Links
* https://github.com/smithy-lang/smithy/blob/9b6aa2a8f417cf89f7a19ac497a4cdce6d74279c/smithy-aws-traits/src/main/resources/META-INF/smithy/aws.api.smithy#L219

---

### DIFF
--- a/docs/source-2.0/aws/aws-core.rst
+++ b/docs/source-2.0/aws/aws-core.rst
@@ -27,7 +27,7 @@ Value type
     * :ref:`service-cloudtrail-event-source`
     * :ref:`service-doc-id`
     * :ref:`service-endpoint-prefix`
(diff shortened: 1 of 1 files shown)


docs/source-2.0/aws/aws-core.rst
```diff
@@ -245,10 +245,10 @@ purposes. Additionally, this value can be used to attempt to resolve endpoints.
 
 .. _service-cloudwatch-metric-namespace:
```
sugmanue: Would you mind also fixing this label and its reference in the service section?

PR #3010, opened 2026-03-24
Add llms.txt generation to docs build

yasmewad:
Enable AI coding assistants to discover Smithy documentation by generating a structured `llms.txt` manifest. Without this, tools produce generic or incorrect answers about Smithy.

Generate the file from existing RST sources as part of every docs release so it stays current without manual upkeep.

#### Background
* Add a build step that generates an [llms.txt](https://github.com/AnswerDotAI/llms-txt) manifest from existing RST sources. This gives AI coding assistants a 
structured index of all Smithy documentation pages so they can discover, search, and reference them.
* Integrate the generation into the html Makefile target so every docs release produces a current manifest at smithy.io/2.0/llms.txt without manual maintenance.

#### Testing
* Ran the generation script locally and verified all 117 URLs return HTTP 200.

#### Links
* llms.txt proposal: https://github.com/AnswerDotAI/llms-txt

---

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -22,7 +22,7 @@ build-landing-page:
 	cd landing-page && npm run build
 
 llms-txt: html2
-	@source build/venv/bin/activate && python3 generate_llms_txt.py
+	python3 generate_llms_txt.py
 
 html: html2 html1 build-landing-page merge-versions llms-txt
 


docs/Makefile
```diff
@@ -21,7 +21,10 @@ html2:
 build-landing-page:
 	cd landing-page && npm run build
 
-html: html2 html1 build-landing-page merge-versions 
+llms-txt: html2
+	@source build/venv/bin/activate && python3 generate_llms_txt.py
+
+html: html2 html1 build-landing-page llms-txt merge-versions
```
sugmanue: Let's make this change additive by keeping the html targets together and this target at the end.
```suggestion
html: html2 html1 build-landing-page merge-versions llms-txt
```

docs/Makefile
```diff
@@ -21,7 +21,10 @@ html2:
 build-landing-page:
 	cd landing-page && npm run build
 
-html: html2 html1 build-landing-page merge-versions 
+llms-txt: html2
+	@source build/venv/bin/activate && python3 generate_llms_txt.py
```
sugmanue: This one depends on html2 that already does the activation of the venv, this can be removed.
```suggestion
	python3 generate_llms_txt.py
```

PR #3047, opened 2026-04-14
Implement tagged literals for IDL 2.1

sugmanue:
#### Background

These changes are expected to be part of the IDL 2.1 changes planed. For now, it adds support for "tagged literals" which allows a literal to be prefixed with a tag that will modify the way the literal is parsed and the resulting token. This change implements:

* `#b <string>` Similar to b-strings in Python signals that the string contains binary data that can be composed by hex or octal escapes or literal values. The resulting token will be the string base64 encoded.
* `#hex <string>` Similar to a hex dump string that contains hexadecimal values, spaces are ignored and strings starting with `#` are consider comments and ignored until the end of line.
* `#timestamp <string>` This tag converts that given ISO encoded date to Smithy's epoch second with millis precision.
* `#re <string>` Takes backslashes as literal chars instead of escape sequences allowing writing regular expressions with character classes without having to escape the slashes.

A model that uses all of that is shown below:

```smithy
$version: "2.1"

structure Example {

    @default(#timestamp "1994-07-05T00:00:00Z")
    startDate: Timestamp

    @pattern(#re "^\d{3}-\d{2}-\d{4}$")
    founderSsn: String

    @default(#b ".PNG\x0D\x0A\x1A\x0A")
    avatar: Blob

    @default(#hex """
          # The first 4 hex values are the literal .PNG, the magic value for PNG images
          89 50 4E 47 0D 0A 1A 0A
          """)
    bigAvatar: Blob
}
    

* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
@@ -21,10 +21,15 @@ class DefaultTokenizer implements IdlTokenizer {
     private Number currentTokenNumber;
     private CharSequence currentTokenStringSlice;
     private String currentTokenError;
+    private Version version;
 
     DefaultTokenizer(String filename, CharSequence model) {
         this.filename = filename;
         this.parser = new SimpleParser(model, 64);
+    }
+
+    void setVersion(Version version) {
+        this.version = version;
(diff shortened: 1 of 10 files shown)


sugmanue:
Closing in favor of #3163 3163

PR #3059, opened 2026-04-17
Add Version 2.1

JordonPhillips:
This adds the 2.1 version to Smithy. Currently there is no difference between 2.0 and 2.1, this is just to get the process started.

The default version is now 2.1, so serialized models will reflect that.

---


sugmanue:
Something that we review later, but I still think that keeping the AST format on 2.0 would be beneficial.

sugmanue (approved):

PR #3078, opened 2026-04-30
Add trait to declare types for metadata

JordonPhillips:
This adds a new trait to declare types for metadata keys. Values for these keys will be validated against the shape targeted by the trait.

While this enables modeling of the suppressions and validators keys, actually adding those to the model will be done later.

---


smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/metadata/map-typed-metadata.errors
joewyz, line 1: nit: should we totally blank file for valid test error files?

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/metadata/no-metadata-value.errors
joewyz, line 1: nit: same as the previous comment on valid test error files.

smithy-model/src/test/java/software/amazon/smithy/model/traits/MetadataTraitTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Optional;
+import org.junit.jupiter.api.Test;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StringShape;
+
+public class MetadataTraitTest {
```
joewyz, line 23: nit: Should we have a test for conflict key defined in different files?

JordonPhillips: I don't think that really makes a difference

joewyz (approved):

PR #3086, opened 2026-05-01
Preserve typed traits when renaming shapes

BDQ:
The `renameShapes` projection transform downgrades typed traits to `DynamicTrait`. After running `renameShapes`, `operation.getTrait(SomeTrait.class)` returns empty even though the trait is still on the shape, because what's stored is now an untyped `DynamicTrait`.

  I hit this from the smithy-typescript codegen path: with `renameShapes` in the projection, lookups for `@staticContextParams` returned empty, so the generated client emitted incomplete endpoint instructions, and endpoint resolution failed at runtime.

#### Background
Two issues contribute here:

1. `smithy-build/transforms/RenameShapes` calls `ModelTransformer.create()` instead of using `context.getTransformer()`. Every other transform in smithy-build/transforms/ already uses the context's transformer; `RenameShapes` seems to be the lone holdout?

2. `smithy-model/transform/ModelTransformer` even when callers construct a transformer via `createWithServiceProviders(cl)`, the classloader is used to discover plugins and then thrown away. The 2-arg `renameShapes(model, renamed)` overload calls `Model.assembler()` with no classloader, so trait reassembly uses smithy-model's classloader and providers from sibling jars (e.g. rules-engine traits) are never found. Anything not in smithy-model itself falls through to `DynamicTrait`.

#### Testing

Adds a regression test in each module: an empty URLClassLoader (no providers) is passed to the transformer/context, and we assert that a typed DeprecatedTrait becomes a DynamicTrait after renameShapes. 

---

### DIFF
--- a/.changes/next-release/bugfix-02fe7a562eb287afa32967a77ee5d80a33419e15.json
+++ b/.changes/next-release/bugfix-02fe7a562eb287afa32967a77ee5d80a33419e15.json
@@ -1,0 +1,7 @@
+{
+  "type": "bugfix",
+  "description": "Preserve typed traits when renaming shapes",
+  "pull_requests": [
+    "[#3086](https://github.com/smithy-lang/smithy/pull/3086)"
+  ]
+}
--- a/.changes/next-release/bugfix-02fe7a562eb287afa32967a77ee5d80a33419e15.json
+++ b/.changes/next-release/bugfix-02fe7a562eb287afa32967a77ee5d80a33419e15.json
@@ -1,6 +1,6 @@
 {
   "type": "bugfix",
-  "description": "Preserve typed traits when renaming shapes",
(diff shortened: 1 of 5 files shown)


.changes/next-release/bugfix-02fe7a562eb287afa32967a77ee5d80a33419e15.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "bugfix",
+  "description": "Preserve typed traits when renaming shapes",
```
kstich:
```suggestion
  "description": "Fixed an issue where traits would lose their types, reverting to DynamicTrait, when renaming shapes",
```

kstich (approved):

PR #3087, opened 2026-05-01
Fixed trait codegen to generate more specific return type for trait provider

timocov:
#### Background

Currently, if you have a trait for custom protocol, when you run the trait-codegen for it the provider for such trait will be generated with `public Trait createTrait(ShapeId target, Node value)` signature. This is generally fine, but when you integrate this trait into smithy-java client codegen, the compilation fails because when [it generates `protocolTrait` property](https://github.com/smithy-lang/smithy-java/blob/016a777697023e90f7899275bc91aa7103bf3feb/codegen/codegen-plugin/src/main/java/software/amazon/smithy/java/codegen/client/generators/ClientInterfaceGenerator.java#L451) by default [it is using provider](https://github.com/smithy-lang/smithy-java/blob/91afa891651dd3cccab4418ba04c165703f7815c/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/integrations/core/GenericTraitInitializer.java#L40-L43) that essentially gets resolved into something like this:

```java
        private static final CustomTrait protocolTrait = new CustomTrait.Provider().createTrait(
            ShapeId.from("customTraitShapeId"),
            Node.objectNode()
        );
```

Where `CustomTrait.Provider().createTrait` currently returns `Trait` and the compilation fails.

This change changes the return type so that the signature now looks like `public CustomTrait createTrait(ShapeId target, Node value)`.

#### Testing

- Added tests to validate the provider's signature.
- Tested locally with internal traits I have that the generated code is correct

#### Links

---

### DIFF
--- a/.changes/next-release/feature-2abbbb589bd53cb39804f4df52e06a7773a7c038.json
+++ b/.changes/next-release/feature-2abbbb589bd53cb39804f4df52e06a7773a7c038.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Fixed trait codegen to generate more specific return type for trait provider",
+  "pull_requests": [
+    "[#3087](https://github.com/smithy-lang/smithy/pull/3087)"
+  ]
+}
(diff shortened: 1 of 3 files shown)


yasmewad (approved):
Thank you for your contribution and making these changes.

PR #3094, opened 2026-05-04
Add --check flag to smithy format

### LINKED ISSUE #2317: Add option to fail the build on formatting violations
Hi folks, thanks for the great work on Smithy. 

Is there an option to fail the Gradle build on formatting violations? It would be great to be able to be able to set options for the smithyFormat task to fail on violations, if not already an option to do so.

If this isn't an option, is it on the roadmap, or is there existing information that would make implementing it via a PR easier, if this is aligned with the maintainers' vision?

yasmewad:
The `smithy format` command previously only supported in-place formatting, with no way to verify formatting without modifying files. This made CI enforcement impossible without risking unintended writes in read-only pipelines.

Add a `--check` mode that compares each file's current content against its formatted output and exits with code 2 if any files differ. Refactor the file-walking logic into a shared `walkSmithyFiles` helper used by both `formatFile` and the new `checkFile`, fixing a `Files.find` stream leak with try-with-resources in the process.

Fixes #2317 

---

### DIFF
--- a/smithy-cli/src/it/java/software/amazon/smithy/cli/FormatCommandTest.java
+++ b/smithy-cli/src/it/java/software/amazon/smithy/cli/FormatCommandTest.java
@@ -75,4 +75,33 @@ public class FormatCommandTest {
                             + "string MyString2%n")));
         });
     }
+
+    @Test
+    public void checkFormattedFile() {
+        // Format the files first, then run --check on the same project copy.
+        IntegUtils.withProject("bad-formatting", path -> {
+            RunResult formatResult = IntegUtils.run(path, ListUtils.of("format", "model"));
+            assertThat(formatResult.getExitCode(), equalTo(0));
+
+            RunResult checkResult = IntegUtils.run(path, ListUtils.of("format", "--check", "model"));
+            assertThat(checkResult.getExitCode(), equalTo(0));
+        });
+    }
+
+    @Test
+    public void checkUnformattedFile() {
+        IntegUtils.run("bad-formatting", ListUtils.of("format", "--check", "model"), result -> {
+            assertThat(result.getExitCode(), equalTo(2));
+            assertThat(result.getOutput(), containsString("not formatted"));
+        });
+    }
+
+    @Test
+    public void checkNoArgs() {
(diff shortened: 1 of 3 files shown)


smithy-cli/src/main/java/software/amazon/smithy/cli/commands/FormatCommand.java
```diff
@@ -85,31 +106,62 @@ private int run(Arguments arguments, Env env) {
...
+                throw new CliError("Error walking " + path + ": " + e.getMessage());
             }
-        } else if (Files.isRegularFile(file) && file.toString().endsWith(".smithy")) {
-            TokenTree tree = parse(file);
+        } else if (Files.isRegularFile(path) && path.toString().endsWith(".smithy")) {
+            action.accept(path);
+        }
+    }
+
+    private void checkFile(Path file, List<Path> violations) {
```
JordonPhillips: There's a lot of overlap between this and check. How about instead you have something like:

```java
private int run(Arguments arguments, Env env) {
    ...
    List<Path> formatted = formatFiles(paths, !options.check)
    if (options.check && !formatted.isEmpty()) {
        // error
    }
    return 0;
}

// returns list of files with changes
private List<Path> formatFiles(List<Path> roots, bool flush) {}

// only write if flush == true
// returns true if there were changes
private bool formatFile(Path file, bool flush) {}
```

smithy-cli/src/it/java/software/amazon/smithy/cli/FormatCommandTest.java
```diff
@@ -75,4 +75,33 @@ public void formatsDirectory() {
...
+
+            RunResult checkResult = IntegUtils.run(path, ListUtils.of("format", "--check", "model"));
+            assertThat(checkResult.getExitCode(), equalTo(0));
+        });
+    }
+
+    @Test
+    public void checkUnformattedFile() {
+        IntegUtils.run("bad-formatting", ListUtils.of("format", "--check", "model"), result -> {
+            assertThat(result.getExitCode(), equalTo(2));
```
kstich, line 94:
```suggestion
            assertThat(result.getExitCode(), equalTo(1));
```

The CliError thrown for this will get turned into an exit 1

PR #3085, opened 2026-05-01
Add providerARNs to authorizer definition

mee-aa:
The `x-amazon-apigateway-authorizer` extension supports a `providerARNs` field that lists Amazon Cognito user pool ARNs for `COGNITO_USER_POOLS` authorizers. This field was missing from the Smithy `AuthorizerDefinition` structure.

Add the optional `providerARNs` list field to `AuthorizerDefinition` in both the Smithy model and Java class. The `AddAuthorizers` mapper now includes `providerARNs` in the generated OpenAPI security scheme when present.

#### Background

* **What do these changes do?**
  * Add an optional `providerARNs` field (list of string) to the `AuthorizerDefinition` structure in `smithy-aws-apigateway-traits`. This field specifies the Amazon Cognito user pool ARNs for `COGNITO_USER_POOLS` authorizers.
  * Update the `AddAuthorizers` mapper in `smithy-aws-apigateway-openapi` to include `providerARNs` in the generated `x-amazon-apigateway-authorizer` extension when the field is present.
  * Update the authorizer properties table in `amazon-apigateway.rst` and add a note in `converting-to-openapi.rst`.
* **Why are they important?**
  * Without this field, customers using Cognito user pool authorizers had no way to specify `providerARNs` in their Smithy model and had to use `jsonAdd` workarounds instead. This is the only missing field on `AuthorizerDefinition` for REST API authorizers.

#### Testing

* Unit test added to `AuthorizersTraitTest`: verifies `providerARNs` round-trips through the factory and is accessible via `getProviderARNs()`
* Integration test added to `AddAuthorizersTest`: loads a Cognito authorizer model with `providerARNs`, converts to OpenAPI, and verifies both `type: "cognito_user_pools"` and the `providerARNs` array appear in the generated `x-amazon-apigateway-authorizer` extension
* All existing authorizer tests pass with no regressions

#### Links

* [[x-amazon-apigateway-authorizer extension (providerARNs property)](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html)](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html)
* [[Integrate a REST API with an Amazon Cognito user pool](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-enable-cognito-user-pool.html)](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-enable-cognito-user-pool.html)

---

[...]

PR #3095, opened 2026-05-05
Add smithy-cfn-json build plugin

ericliu03:
#### Background
* Add a new `smithy-cfn-json` build plugin (in the `smithy-aws-apigateway-cfn` subproject) that serializes a Smithy model to JSON AST with CloudFormation `Fn::Sub` intrinsic function wrapping at known trait paths. The output is could be used in CFN template. 
* Without this plugin, customers deploying via CloudFormation must use the OpenAPI conversion path to get `Fn::Sub` support for deploy-time resource references like Lambda ARNs and VPC Link IDs. This plugin provides the same capability for the native Smithy JSON AST import path.
* the code following the existing [CloudFormationSubstitution](https://github.com/smithy-lang/smithy/blob/a8766999c9af395ded6f8946bff3ad831b59b5ef/smithy-aws-apigateway-openapi/src/main/java/software/amazon/smithy/aws/apigateway/openapi/CloudFormationSubstitution.java#L33) which is for openapi mapper

#### Testing
* Unit tests for `CloudFormationFnSubInjector`: verifies substitution at each known path, no substitution without `${...}` pattern, no substitution at non-known paths, wildcard matching, non-string node preservation
* Unit tests for `SmithyCfnJson`: plugin name, `requiresValidModel`, SPI discovery via `ServiceLoader`
* Unit tests for `SmithyCfnJsonConfig`: deserialization with full config and defaults
* Integration tests: end-to-end plugin execution with a real Smithy model, verifying `Fn::Sub` wrapping on `uri`, `credentials`, `connectionId`, `integrationTarget`; disabled mode matches raw `ModelSerializer` output
* Golden-file integration test comparing full plugin output against expected JSON

#### Links
* [x-amazon-apigateway-integration](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html)
* [CloudFormation Fn::Sub](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html)

---


smithy-aws-apigateway-cfn/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+description = "This module provides a smithy-build plugin that serializes a Smithy model " +
+    "to JSON AST with CloudFormation Fn::Sub substitution for use as a CFN RestApi Body."
+
+extra["displayName"] = "Smithy :: Amazon API Gateway CloudFormation JSON"
+extra["moduleName"] = "software.amazon.smithy.aws.apigateway.cfn"
+
+dependencies {
+    api(project(":smithy-build"))
+    api(project(":smithy-model"))
+    testImplementation(project(":smithy-aws-apigateway-traits"))
```
ericliu03, line 18: this is used for testing, but i guess we can use another trait for testing and remove this dependency

smithy-aws-apigateway-cfn/src/test/java/software/amazon/smithy/aws/apigateway/cfn/SmithyCfnJsonTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Test
+    public void requiresValidModel() {
+        assertTrue(new SmithyCfnJson().requiresValidModel());
+    }
+
+    @Test
+    public void isDiscoverableViaSpi() {
+        boolean found = false;
+        for (SmithyBuildPlugin plugin : ServiceLoader.load(SmithyBuildPlugin.class)) {
```
ericliu03, line 30: IIUC, when this test passes, it means we can use this plugin in `smithy-build.json`

smithy-aws-apigateway-cfn/src/test/java/software/amazon/smithy/aws/apigateway/cfn/SmithyCfnJsonGoldenFileTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+        new SmithyCfnJson().execute(context);
+
+        String output = manifest.getFileString("MyService.smithy.json").get();
+        ObjectNode actual = Node.parse(output).expectObjectNode();
+        ObjectNode expected = Node.parse(
+                IoUtils.readUtf8Resource(getClass(), "integration-model.expected.json"))
+                .expectObjectNode();
+
+        Node.assertEquals(actual, expected);
```
ericliu03, line 48: this test is a bit unnecessary since we have the integration test, but this covers the file -> file generation

docs/source-2.0/guides/model-translations/smithy-cfn-json.rst
```diff
@@ -0,0 +1,147 @@
...
+The ``smithy-cfn-json`` plugin serializes a Smithy model to its JSON AST
+representation with automatic CloudFormation ``Fn::Sub`` substitution wrapping.
+The output is intended for use as the ``Body`` property of an
+``AWS::ApiGateway::RestApi`` CloudFormation resource, enabling direct Smithy
+model import without OpenAPI conversion.
+
+String values containing ``${...}`` variable syntax at known trait paths are
+automatically wrapped in ``{"Fn::Sub": "..."}`` objects so that CloudFormation
+resolves the references at deploy time before passing the body to the API
+Gateway SmithyImporter.
```
mee-aa, line 24: what is SmithyImporter? (˵ ¬ᴗ¬˵)

docs/source-2.0/guides/model-translations/smithy-cfn-json.rst
```diff
@@ -0,0 +1,147 @@
+.. _smithy-cfn-json:
+
+==========================================================
+Converting Smithy to CloudFormation JSON
+==========================================================
+
```
mee-aa, line 6: I like this!! The NodeVisitor approach and path-matching with wildcards are clean. 

One note, the core Fn::Sub injection logic is generic, it's not really API Gateway-specific. Any trait with ARN-valued fields could benefit. Would it make sense to put this in a more general module (e.g., smithy-aws-cloudformation) and make the paths configurable via the plugin config?

The API Gateway paths could be defaults, but users could add their own for custom traits. This would also avoid needing to modify the class every time we add a new trait with an ARN field (like we just did with providerARNs on authorizers)!

smithy-aws-apigateway-cfn/src/main/java/software/amazon/smithy/aws/apigateway/cfn/CloudFormationFnSubInjector.java
```diff
@@ -0,0 +1,114 @@
...
+ * Walks a Smithy JSON AST node tree and wraps string values containing
+ * CloudFormation variable syntax in Fn::Sub intrinsic function objects.
+ */
+final class CloudFormationFnSubInjector extends NodeVisitor.Default<Node> {
+
+    private static final Logger LOGGER = Logger.getLogger(CloudFormationFnSubInjector.class.getName());
+    private static final String SUBSTITUTION_KEY = "Fn::Sub";
+    private static final Pattern SUBSTITUTION_PATTERN = Pattern.compile("\\$\\{.+}");
+
+    static final List<String> PATHS = Arrays.asList(
```
mee-aa, line 35: These paths are hardcoded to API Gateway traits. If someone adds a new trait with an ARN field, they will have to modify this class. can we make these configurable via `SmithyCfnJsonConfig`? something like `additionalSubstitutionPaths` that users can set in smithy-build.json. The defaults could still be these API Gateway paths, but it future-proofs the plugin.

smithy-aws-apigateway-cfn/build.gradle.kts
mee-aa: If we want this plugin to work beyond API Gateway, we could test with a simple inline trait instead of depending on the apigateway-traits module. That also proves it works for any trait with ARN fields.

smithy-aws-apigateway-cfn/src/main/java/software/amazon/smithy/aws/apigateway/cfn/SmithyCfnJsonConfig.java
```diff
@@ -0,0 +1,54 @@
...
+    public void setService(ShapeId service) {
+        this.service = service;
+    }
+
+    /**
+     * Gets whether CloudFormation substitution is disabled.
+     *
+     * @return Returns true if substitution is disabled.
+     */
+    public boolean getDisableCloudFormationSubstitution() {
```
mee-aa, line 50: `isDisable...` right? since its a boolean

smithy-aws-apigateway-cfn/src/main/java/software/amazon/smithy/aws/apigateway/cfn/SmithyCfnJson.java
```diff
@@ -0,0 +1,61 @@
...
+        ShapeId serviceId = config.getService();
+        if (serviceId == null) {
+            throw new SmithyBuildException(
+                    "smithy-cfn-json plugin requires a 'service' configuration property");
+        }
+
+        Model model = context.getModel();
+        ServiceShape service = model.expectShape(serviceId, ServiceShape.class);
+
+        ObjectNode astNode = ModelSerializer.builder().build().serialize(model);
```
mee-aa, line 53: This serializes every shape in the model, right? and not just the ones the service uses?

 Should it filter to only the service's shapes so the output doesn't include unrelated stuff?

docs/source-2.0/guides/model-translations/smithy-cfn-json.rst
```diff
@@ -0,0 +1,147 @@
...
+------------------------------
+
+When ``disableCloudFormationSubstitution`` is ``false`` (the default), string
+values containing ``${...}`` variable syntax at the following trait paths are
+automatically wrapped in ``{"Fn::Sub": "..."}`` objects:
+
+* ``aws.apigateway#integration`` — ``uri``, ``credentials``, ``connectionId``,
+  ``integrationTarget``
+* ``aws.apigateway#authorizers`` — ``*/uri``, ``*/credentials``
+* ``aws.auth#cognitoUserPools`` — ``providerArns/*``
```
mee-aa, line 112: this should be apigateway cognito authorizer providerarns: https://github.com/smithy-lang/smithy/pull/3085

I am yet to discuss with the smithy team re: aws auth's cognito user pools trait: https://smithy.io/2.0/aws/aws-auth.html#aws-auth-cognitouserpools-trait

There's a lot of confusion between these two authorizing traits.

PR #3097, opened 2026-05-06
Add ability to publish our docs in markdown format

yasmewad:
#### Background

This adds markdown versions of all documentation pages alongside the existing HTML output. AI agents and LLMs can now consume Smithy docs directly at `/2.0/markdown/<path>.md` without parsing HTML, which is more token-efficient and eliminates the need to navigate through `<div>` tags and CSS classes to find actual content.

The `llms.txt` index is updated to point to markdown URLs so agents can programmatically discover all available content. A `<link rel="alternate">` tag is added to the landing page so agents parsing `smithy.io` can find `llms.txt` without guessing paths.

This also fixes a pre-existing crash in `redirects.py` where `app.warn` ([deprecated in 1.6, removed in 2.0](https://www.sphinx-doc.org/en/master/extdev/deprecated.html)) caused the build to fail when using any non-HTML builder. Custom node handlers are added for `tip`, `danger`, `caution`, and `admonition` directives so their content is preserved in the markdown output rather than silently dropped.

#### Testing

Built locally with Python 3.12. The markdown builder produces 123 files for Smithy 2.0 and 56 for 1.0. Code blocks, tables, cross-references, and links all render correctly. The HTML build continues to work unchanged.

#### Links

* [llmstxt.org spec](https://llmstxt.org): the standard for `/llms.txt` discovery
* [sphinx-markdown-builder](https://pypi.org/project/sphinx-markdown-builder/): the Sphinx extension used
* [Sphinx deprecated APIs](https://www.sphinx-doc.org/en/master/extdev/deprecated.html): `app.warn` removal reference
* Builds on #3010 which added `llms.txt` generation

### DIFF
--- a/docs/landing-page/index.html
+++ b/docs/landing-page/index.html
@@ -31,6 +31,12 @@
     />
 
     <title>Smithy</title>
+    <link
+      rel="alternate"
+      type="text/plain"
+      href="/2.0/llms.txt"
+      title="LLM-readable documentation"
+    />
   </head>
   <body class="dark">
     <div id="root"></div>
--- a/.changes/next-release/feature-fcb484ef5c3589dd9a8ec2ae20d1975e25d51fb6.json
+++ b/.changes/next-release/feature-fcb484ef5c3589dd9a8ec2ae20d1975e25d51fb6.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Dev/yasmewad/markdown docs publishing",
+  "pull_requests": [
+    "[#3097](https://github.com/smithy-lang/smithy/pull/3097)"
+  ]
+}
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -18,13 +18,21 @@ html1:
 html2:
 	@source build/venv/bin/activate && $(SPHINXBUILD) -M html "source-2.0" "build/2.0" $(SPHINXOPTS) -W --keep-going -n $(O)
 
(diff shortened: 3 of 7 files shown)


docs/landing-page/index.html
```diff
@@ -31,6 +31,12 @@
     />
 
     <title>Smithy</title>
+    <link
```
yasmewad, line 34: This allows implicit discovery if an LLM or agent comes across the documentation through our HTML pages

JordonPhillips:
Why not just convert all the source to markdown? Some of our pages already use it. That would make offboarding to some other thing easier

yasmewad:
> Why not just convert all the source to markdown? Some of our pages already use it. That would make offboarding to some other thing easier

Hmm, my intention was to make it additive as a change. Would we lose those !TIP and !IMPORTANT sections if we move away from .rst? 

What about we ship this so we get benefit right away for the intention of providing value right now with current conversion and then convert in batches to `.md`?

JordonPhillips:
> Hmm, my intention was to make it additive as a change. Would we lose those !TIP and !IMPORTANT sections if we move away from .rst?

No, we'd be converting to myst, which is fully compatible with rst directives.

yasmewad:
Closing as I think I might have found a way to re-utilize our existing .rst source as well for agents and serve llms.txt. given we have .rst source it might be similar in terms of token efficiency.

PR #3107, opened 2026-05-09
Add copy-on-write optimization for `toBuilder()` via `BuilderRef.setBorrowed`

sugmanue:
## Summary

Adds a `setBorrowed(T)` method to `BuilderRef` that enables zero-copy `toBuilder()` calls. When a builder is created from an existing built object, the already-immutable collections are borrowed directly instead of being copied through setter methods. A copy only occurs if the builder actually mutates the collection.

## Motivation

Previously, calling `toBuilder()` on any built object would copy every collection field, even if the resulting builder was never mutated or only a single field was changed. For objects with many collection fields (like `ServiceShape`, `Model`, `IntegrationTrait`), this meant multiple unnecessary allocations on every `toBuilder()` call.

## Changes

**`BuilderRef` interface**, Added `default void setBorrowed(T value)` with a default implementation that throws `UnsupportedOperationException` for backwards compatibility.

**`DefaultBuilderRef`**, Implemented `setBorrowed` to set the borrowed (immutable) reference directly, clearing any owned state. Also optimized `copy()` to return the borrowed value directly when no mutation has occurred, avoiding a redundant copy+wrap cycle.

**30 classes across the project**, Added private `Builder(BuiltObject)` constructors that use `setBorrowed` for collection fields, and simplified `toBuilder()` to `return new Builder(this)`.

## Performance characteristics

| Scenario | Before | After |
|----------|--------|-------|
| `toBuilder()` call | O(n) copy per collection field | O(1), just sets a reference |
| `toBuilder()` then `build()` (no mutation) | 2× O(n) copies per field | Zero copies, same instance reused |
| `toBuilder()`, mutate one field, `build()` | 2× O(n) copies for ALL fields | O(n) copy only for the mutated field |

## Backwards compatibility

- The existing `BuilderRef` API (`get`, `peek`, `copy`, `hasValue`, `clear`) is unchanged.
- `setBorrowed` has a default implementation that throws, so external implementations of `BuilderRef` are not broken.
- The `copy()` optimization (returning borrowed directly) is safe because `borrowed` is always immutable in all codepaths.

## Megamorphic call site fix

Replaced all collection copy constructors (`new HashMap<>(map)`, `new ArrayList<>(list)`, etc.) in `BuilderRef` factory methods with explicit for-each iteration. This avoids the megamorphic performance penalty described in [JDK-8368292](https://bugs.openjdk.org/browse/JDK-8368292), where bulk methods like `putAll`/`addAll` suffer from (3 + 4n) virtual method calls that the JIT cannot inline due to the many different collection types flowing through these paths (HashMap, TreeMap, LinkedHashMap, unmodifiableMap, emptyMap, etc.).

### DIFF
--- a/smithy-utils/src/main/java/software/amazon/smithy/utils/BuilderRef.java
+++ b/smithy-utils/src/main/java/software/amazon/smithy/utils/BuilderRef.java
@@ -81,6 +81,20 @@ public interface BuilderRef<T> extends CopyOnWriteRef<T> {
     void clear();
 
     /**
+     * Sets a borrowed (immutable) value directly, clearing any owned state.
+     *
+     * <p>This enables the {@code toBuilder()} optimization: when creating a
+     * builder from an existing built object, the already-immutable collections
+     * can be set directly without copying. A copy will only be made if the
+     * builder actually mutates the value via {@link #get()}.
+     *
+     * @param value The immutable value to borrow. May be {@code null} to clear.
+     */
+    default void setBorrowed(T value) {
+        throw new UnsupportedOperationException("setBorrowed");
+    }
+
+    /**
      * Creates a builder reference to an unordered map.
      *
      * @param <K> Type of key of the map.
@@ -89,7 +103,13 @@ public interface BuilderRef<T> extends CopyOnWriteRef<T> {
      */
     static <K, V> BuilderRef<Map<K, V>> forUnorderedMap() {
         return new DefaultBuilderRef<>(HashMap::new,
-                HashMap::new,
+                source -> {
+                    Map<K, V> copy = new HashMap<>((int) (source.size() / 0.75f + 1));
+                    for (Map.Entry<K, V> entry : source.entrySet()) {
+                        copy.put(entry.getKey(), entry.getValue());
+                    }
+                    return copy;
+                },
                 Collections::unmodifiableMap,
                 Collections::emptyMap);
     }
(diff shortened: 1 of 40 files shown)


smithy-utils/src/main/java/software/amazon/smithy/utils/BuilderRef.java
```diff
@@ -89,7 +103,13 @@ public interface BuilderRef<T> extends CopyOnWriteRef<T> {
      */
     static <K, V> BuilderRef<Map<K, V>> forUnorderedMap() {
         return new DefaultBuilderRef<>(HashMap::new,
-                HashMap::new,
+                source -> {
+                    Map<K, V> copy = new HashMap<>((int) (source.size() / 0.75f + 1));
```
yasmewad, line 107: A small comment on the 0.75f division + 1 might be helpful. Having said that TIL about the optimization to avoid re-hashing.

sugmanue: I don't think that's needed, it's a known common pattern, it appears in `ShapeIdConflictValidator.java` without a comment and in three different places in Smithy-java, also without comment.

yasmewad: Always an opportunity to improve the code base for new users. But this is minor. So approving.

yasmewad (approved):

PR #3088, opened 2026-05-01
Add endpointConfiguration trait

mee-aa:
API Gateway REST APIs have an endpoint configuration that specifies the endpoint type (EDGE, REGIONAL, or PRIVATE), optional VPC endpoint IDs, and whether the default execute-api endpoint is disabled.

Add the `aws.apigateway#endpointConfiguration` structure trait with a required `types` enum list, optional `vpcEndpointIds` list, and optional `disableExecuteApiEndpoint` boolean. No OpenAPI mapper is included; `vpcEndpointIds` and `disableExecuteApiEndpoint` exist in the `x-amazon-apigateway-endpoint-configuration` extension but a mapper can be added separately if needed.

#### Background

* **What do these changes do?**
  * Add a new `endpointConfiguration` structure trait to `smithy-aws-apigateway-traits` with a required `types` field (enum list: `EDGE`, `REGIONAL`, `PRIVATE`), optional `vpcEndpointIds` (string list), and optional `disableExecuteApiEndpoint` (boolean).
  * Add trait documentation in `amazon-apigateway.rst`.
  * No OpenAPI mapper is included. The `x-amazon-apigateway-endpoint-configuration` extension supports `vpcEndpointIds` and `disableExecuteApiEndpoint` for REST APIs, but `types` is not part of the extension (it comes from import parameters). A mapper for the supported fields can be added separately.
* **Why are they important?**
  * Without this trait, customers had no way to specify endpoint configuration directly in their Smithy model and had to use `jsonAdd` workarounds or set these values outside the model.

#### Testing

* Unit tests for the trait class: builder construction with all fields (PRIVATE + VPC endpoints + disable flag), required-field-only (REGIONAL), round-trip via Provider (EDGE), factory registration
* `toBuilder()` round-trip verified in each builder test
* All existing tests pass

#### Links

* [[API endpoint types for REST APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-endpoint-types.html)](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-endpoint-types.html)
* [[x-amazon-apigateway-endpoint-configuration extension](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-endpoint-configuration.html)](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-endpoint-configuration.html)

---

[...]

PR #3110, opened 2026-05-12
Add ipAddressType and OpenAPI mapper to endpointConfiguration

mee-aa:
API Gateway REST APIs support an `ipAddressType` property on the endpoint configuration that controls whether an API accepts IPv4, IPv6, or both. The `@endpointConfiguration` trait did not model this property, and the trait had no OpenAPI mapper, so `vpcEndpointIds` and `disableExecuteApiEndpoint` had to be set via `jsonAdd` workarounds.

Add `ipAddressType` as a string member with supported values `ipv4` and `dualstack`. Add the `AddEndpointConfiguration` OpenAPI mapper, which writes `vpcEndpointIds` and `disableExecuteApiEndpoint` to the `x-amazon-apigateway-endpoint-configuration` extension. `types` and `ipAddressType` are omitted because API Gateway configures them outside of the OpenAPI document at import time. Add a CloudFormation substitution path for `vpcEndpointIds` so values using `Fn::Sub` variable syntax are wrapped correctly.

#### Background

* **What do these changes do?**
  * Add the `ipAddressType` string member to `aws.apigateway#endpointConfiguration`. Modeled as a plain string so new values (for example, an `ipv6`-only mode) can ship without a Smithy release.
  * Add the `AddEndpointConfiguration` OpenAPI mapper. Emits the extension when at least one mapped field is set; skips it otherwise to avoid empty extensions.
  * Add `x-amazon-apigateway-endpoint-configuration/vpcEndpointIds/*` to the `CloudFormationSubstitution` path list.
  * Update trait docs in `amazon-apigateway.rst` and add a new "Endpoint configuration" section to `converting-to-openapi.rst`.

* **Why are they important?**
  * `ipAddressType` is the last missing public `endpointConfiguration` property, bringing the trait to parity with CloudFormation's `AWS::ApiGateway::RestApi` `EndpointConfiguration`.
  * Without the mapper, customers had to use `jsonAdd` workarounds in `smithy-build.json` to attach `vpcEndpointIds` or `disableExecuteApiEndpoint`. The CFN substitution path keeps typical deployment flows working.

#### Testing

* Extended `EndpointConfigurationTraitTest` to cover `ipAddressType` in all four existing cases (builder, round-trip from node, TraitFactory registration, required-fields-only).
* Added `AddEndpointConfigurationTest` with three scenarios: full config (asserts only `vpcEndpointIds` and `disableExecuteApiEndpoint` are emitted), minimal config (only `types` set, no extension emitted), and CloudFormation variable in `vpcEndpointIds` (asserts the value is wrapped in `{"Fn::Sub": "${...}"}`).
* Ran `./gradlew clean build`: all 574 tasks pass, including spotless, spotbugs, checkstyle, javadoc, and tests.

#### Links

* Follow-up to #3088, which added the `@endpointConfiguration` trait without the OpenAPI mapper.

---


smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.smithy
```diff
@@ -191,6 +191,12 @@ structure endpointConfiguration {
     /// Whether clients can invoke the API using the default execute-api
     /// endpoint.
     disableExecuteApiEndpoint: Boolean
+
+    /// The IP address type that can invoke the API. Use `ipv4` to allow
+    /// only IPv4 addresses. Use `dualstack` to allow both IPv4 and IPv6
```
kstich, line 187: If this is going to be an open string, it should emit WARNING events in validation if the value isn't one of these known values.

smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.smithy
```diff
@@ -191,6 +191,12 @@ structure endpointConfiguration {
...
+    /// addresses. For the `PRIVATE` endpoint type, only `dualstack` is
+    /// supported.
```
kstich, line 189: This should emit an ERROR event in validation if not met.

kstich (changes requested):

smithy-aws-apigateway-traits/src/main/resources/META-INF/smithy/aws.apigateway.smithy
mee-aa: I kept it as a string for forward compatibility in case if AWS adds a new value, existing Smithy JARs keep working. A WARNING validator would emit false positives until the allow-list is updated here, reintroducing the coupling we were avoiding. API Gateway already rejects unknown values at import time, shall we leave the validation there or add it here?

sugmanue: Do we really need to worry about "forward compatibility" here? The rollout of IPv6 is been going for 26 years now and these changes do not happen without a lot of planning which give use leeway for change it.

As for the next question, a local validation is still best for the UX, you don't want to wait until deploy time for this to fail, you need to know as early as possible in the development cycle.



> shall we leave the validation there or add it here?

I don't think this is either-or situation, but rather it’s both-and.

smithy-aws-apigateway-traits/src/test/java/software/amazon/smithy/aws/apigateway/traits/EndpointConfigurationValidatorTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.Test;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.validation.Severity;
+import software.amazon.smithy.model.validation.ValidatedResult;
+import software.amazon.smithy.model.validation.ValidationEvent;
+
+public class EndpointConfigurationValidatorTest {
```
kstich: These should use the package's errorfiles convention

smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/EndpointConfigurationValidator.java
```diff
@@ -0,0 +1,53 @@
...
+@SmithyInternalApi
+public final class EndpointConfigurationValidator extends AbstractValidator {
+
+    private static final String DUALSTACK = "dualstack";
+    private static final String PRIVATE = "PRIVATE";
+
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        return model.shapes(ServiceShape.class)
+                .filter(service -> service.hasTrait(EndpointConfigurationTrait.class))
```
kstich: `model.getServiceShapesWithTrait(EndpointConfigurationTrait.class)`

smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/EndpointConfigurationValidator.java
```diff
@@ -0,0 +1,53 @@
...
+                                    .severity(Severity.ERROR)
+                                    .message(String.format(
+                                            "The `PRIVATE` endpoint type requires `ipAddressType` to be "
+                                                    + "`dualstack`, but found `%s`.",
+                                            ipAddressType))
+                                    .build())
+                            .map(Stream::of)
+                            .orElseGet(Stream::empty);
+                })
+                .collect(Collectors.toList());
```
kstich: Create a `List<ValdiationEvent>` up front and use traditional loops instead of a sequence of intermediate streams and flatmaps.

smithy-aws-apigateway-traits/src/main/java/software/amazon/smithy/aws/apigateway/traits/EndpointConfigurationValidator.java
```diff
@@ -0,0 +1,53 @@
...
+                .filter(service -> service.hasTrait(EndpointConfigurationTrait.class))
+                .flatMap(service -> {
+                    EndpointConfigurationTrait trait = service.expectTrait(EndpointConfigurationTrait.class);
+                    return trait.getIpAddressType()
+                            .filter(ipAddressType -> trait.getTypes().contains(PRIVATE)
+                                    && !DUALSTACK.equals(ipAddressType))
+                            .map(ipAddressType -> ValidationEvent.builder()
+                                    .id(getName())
+                                    .shape(service)
+                                    .severity(Severity.ERROR)
```
kstich: This reads like it can use the `error(...)` method on the abstract class.

kstich (changes requested):

kstich (approved):

PR #3130, opened 2026-05-15
Add apiConfig to aws.api#tagEnabled for renamed tagging operations

taylorlo-aws:
#### Background
* What do these changes do?

Allow services that model tagging via non-default operation names (e.g., AddTagsToResource instead of TagResource) to keep validation and tag discovery on. The new TaggableServiceApiConfig structure has three optional ShapeID slots (tagApi, untagApi, listTagsApi); each unset slot falls back to the default-named operation. Resource-level apiConfig still wins over service-level for resource-bound resolution.

AwsTagIndex now resolves the per-service tagging op slots via the service-level apiConfig before falling back to default-named lookups. A new TagEnabledServiceValidator check warns when an apiConfig override coexists with the default-named operation in the service operations list, so modelers either drop the override or remove the dead op.

* Why are they important?

Services like RDS use service level tagging operations that are not named the default. This change allows the usage of `tagEnabled` to semantically associate these non-standard tagging APIs in the model so that consumers can find them while still applying the existing validation.

#### Testing

All existing model test cases pass; this change is backwards compatible. New model test cases were added to validate the new override + fallback behavior. 

#### Links

N/A

N/A

---

### DIFF
--- a/.changes/next-release/feature-d0d9dd98dfc2595a63c97eca17c24604e402c1e5.json
+++ b/.changes/next-release/feature-d0d9dd98dfc2595a63c97eca17c24604e402c1e5.json
@@ -1,5 +1,7 @@
 {
   "type": "feature",
   "description": "Add apiConfig to the aws.api#tagEnabled trait so services can declare non-default operation names (e.g., AddTagsToResource) for the TagResource, UntagResource, and ListTagsForResource slots; tag API discovery and validation honor the override before falling back to default-named operations.",
-  "pull_requests": []
+  "pull_requests": [
+    "[#3130](https://github.com/smithy-lang/smithy/pull/3130)"
+  ]
 }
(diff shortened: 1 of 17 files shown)


.changes/next-release/feature-d0d9dd98dfc2595a63c97eca17c24604e402c1e5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add apiConfig to the aws.api#tagEnabled trait so services can declare non-default operation names (e.g., AddTagsToResource) for the TagResource, UntagResource, and ListTagsForResource slots; tag API discovery and validation honor the override before falling back to default-named operations.",
```
kstich:
```suggestion
  "description": "Added support for services to declare non-default operation names (e.g., AddTagsToResource) for tagging operations at the service level in the `tagEnabled` trait. Tag API discovery and validation honor the override before falling back to default-named operations.",
```

kstich (approved):

PR #3134, opened 2026-05-16
Add inline collection declarations for Smithy IDL 2.1

sugmanue:
## Summary

This PR introduces inline collection syntax for Smithy IDL 2.1, allowing list and map shapes to be declared directly in member target positions:

```smithy
$version: "2.1"

structure MyStructure {
    names: [String]
    tags: {String: String}
    nested: {String: [Integer]}
}
```

This is syntactic sugar: the assembler generates synthetic shapes (`_SyntheticListOfString`, `_SyntheticMapOfStringToString`, etc.) that are shared across structurally equivalent usages.

### Core model (`smithy-model`)

- **`Version.java`**: Refactored to use a bitmask for feature flags. Added `INLINE_COLLECTIONS` flag for 2.1.
- **`Trait.java`**: Added `SerializationMode` enum (`ALL`, `AST_ONLY`, `NONE`) and `serializationMode()` method to control where traits are serialized.
- **`SyntheticShapeTrait`**: New trait in `smithy.synthetic#generated` namespace. Uses `AST_ONLY` serialization, persists in JSON AST but not IDL. Registered via TraitService SPI.
- **`SyntheticShapeNaming`**: Utility for generating deterministic synthetic shape names from target types.
- **`SyntheticShapeValidator`**: Critical validator that errors if traits are applied to synthetic shapes, and warns if user shapes use the `_Synthetic` prefix.
- **`IdlModelLoader`**: Parses `[Target]` and `{Key: Value}` in member target positions. Creates synthetic shapes with deduplication. Version-gated to 2.1+. Nesting capped at 3 levels.
- **`SmithyIdlModelSerializer`**: Emits inline syntax for members targeting synthetic shapes. Excludes synthetic shapes from top-level serialization.
- **`ModelSerializer`**: Uses `serializationMode()` instead of `isSynthetic()` for trait filtering.

### Syntax tree (`smithy-syntax`)

- **`TreeType.java`**: Added `MEMBER_TARGET`, `INLINE_LIST_TARGET`, `INLINE_MAP_TARGET` tree types.
- **`FormatVisitor.java`**: Formats inline collections as `[Target]` and `{Key: Value}`.

### Diff tool (`smithy-diff`)

- **`AddedShape`/`RemovedShape`**: Suppress events for synthetic shapes.
- **`ModifiedTrait`**: Added `SyntheticShapeTrait` to `IGNORED_TRAITS`.

### CLI (`smithy-cli`)

- **`MigrateCommand`**: Updated to emit version `"2.1"` and skip any 2.x files.

### Documentation

- `spec/idl.rst`: ABNF grammar update, new "Inline collection declarations" section, version statement note, examples in List/Map/Structure/Union sections.
- `spec/aggregate-types.rst`: Inline syntax notes in List and Map sections.
- `spec/model.rst`: Member shapes note, Shape ID `_Synthetic` prefix reservation, trait restriction note.
- `spec/json-ast.rst`: New "Synthetic shapes in the JSON AST" section.
- `spec/type-refinement-traits.rst`: Note that `@sparse` cannot apply to synthetics.
- `spec/selectors.rst`: Example for selecting synthetic shapes.
- `guides/style-guide.rst`: Recommendation to prefer inline for single-use collections.
- `guides/evolving-models.rst`: Migration guidance for inline vs explicit.
- `guides/building-codegen/mapping-shapes-to-languages.rst`: Synthetic shape naming guidance.
- `guides/building-codegen/using-the-semantic-model.rst`: ShapeVisitor note.
- `guides/model-validation-examples.rst`: Selector note for excluding synthetics.
- `guides/smithy-build-json.rst`: Tip for selecting synthetics in projections.
- `quickstart.rst`: Forward-looking tip about inline syntax.
- `designs/inline-collections.md`: Design document.

## Testing

- **Valid model tests**: Basic inline collections, grouping (shared shapes), union members, use-statement resolution, nested collections.
- **Error tests**: Version gating (2.0 rejects inline syntax), nesting depth limit (>3 levels).
- **IDL round-trip test**: Parse, serialize, compare matches original.
- **JSON AST round-trip test**: IDL to JSON to model comparison.
- **Diff tests**: Add/remove inline collections, inline to explicit migration, explicit to inline migration, explicit with constraint traits, no-change scenario.

## Limitations

- Inline collections cannot carry shape-level traits (`@sparse`, `@uniqueItems`, `@length` on the collection). Use explicit shape definitions for these.
- `apply` statements cannot target synthetic shapes.
- Nesting is limited to 3 levels.


designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
+# Inline collection declarations
+
+* **Author**: Manuel Sugawara
+* **Created**: 2026-05-14
+
+## Abstract
+
+This proposal introduces syntactic sugar for declaring list and map shapes inline within member definitions. Instead of requiring a separate top-level shape definition, members can declare their collection type directly using `[Target]` for lists and `{Key: Value}` for maps. The assembler generates synthetic shapes with stable, content-derived names that are grouped by structural equivalence.
```
JordonPhillips: plz line wrap

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+member_target = shape_id / inline_list / inline_map
+
+inline_list = "[" ws member_target ws "]"
+
+inline_map = "{" ws member_target ws ":" ws member_target ws "}"
+```
+
+### Synthetic shape naming
+
+Inline collections produce synthetic shapes with content-derived names. The naming scheme uses a `_Synthetic` prefix followed by a description of the collection type and its targets.
```
JordonPhillips: Is the `_Synthetic` prefix even necessary? I don't think the collision problem is real. If there's a `ListOfString` already in the namespace then it's probably what they should be using anyway. Or that thing should be renamed.

You could do a `$inlineListPrefix` directive or something like that to add it I guess.

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+
+- Identifies the shape as assambler-generated.
+- Allows IDL serializers to reconstruct the inline syntax when converting from AST back to IDL.
+- Is not user-applicable, only the assambler may attach it.
+
+### Traits on synthetic shapes
+
+No traits may be applied to synthetic shapes, either at definition time or via `apply` statements. The `apply` statement MUST NOT target a shape that has the `@synthetic` trait.
+
+Traits that would normally apply to a collection shape (such as `@sparse`, `@uniqueItems`, or `@length` when constraining collection size) cannot be expressed using inline syntax. Authors requiring these traits MUST use explicit shape definitions.
```
JordonPhillips: nit: length can be applied to members

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+
+Permitting traits like `@sparse`, `@uniqueItems`, or `@length` on synthetic
+shapes, either via inline syntax or `apply`.
+
+This was rejected because:
+
+- It complicates grouping — traits on the shape would need to be part of the equivalence check.
+- Applying traits to a shared/grouped shape affects all usages, which is rarely the intent.
+- It introduces the new concept of "traits declared on a member that semantically apply to the target shape."
+- The explicit shape definition path already handles these cases cleanly.
```
JordonPhillips: An important item to add to this list: the vast majority of traits applied to lists/maps today across aws services and smithy-owned traits can be applied to a member.  In our internal plan doc thing I put in a table that shows it that I'm too lazy to copy here.

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+- Synthetic shapes cannot be targeted by `apply` statements.
+- The inline syntax is only available in member target positions, not as top-level definitions.
+
+## Migration
+
+### Adopting inline syntax
+
+Existing models using explicit list and map definitions can adopt inline syntax incrementally. Replacing an explicit shape with inline syntax changes the shape ID (from `ns#MyList` to `ns#_SyntheticListOfString`), which is a breaking change for:
+
+- Other models referencing the shape by name
```
JordonPhillips, line 366: This is only a break if you remove the previous shape.

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+- Selectors targeting the shape by ID
+- Code generators that derive type names from shape names
```
JordonPhillips, line 368: Does anybody generate concrete types for lists/maps? I don't think we even consider it a backwards incompatible change in smithy-diff.

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+
+Authors should only migrate shapes that are not referenced externally.
+
+### Outgrowing inline syntax
```
JordonPhillips, line 372: We should have validators for this. If we see that a significant number of identical inline collections share the same member traits, we should recommend creating a dedicated shape instead. Particularly relevant for the length trait, which could be considered to have security implications.l

designs/inline-collections.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Limitations
+
+- Inline collections cannot carry shape-level traits. Authors needing `@sparse`, `@uniqueItems`, `@length` (on the collection), or other shape-level traits must use explicit shape definitions.
+- Synthetic shapes cannot be targeted by `apply` statements.
+- The inline syntax is only available in member target positions, not as top-level definitions.
+
+## Migration
+
+### Adopting inline syntax
```
JordonPhillips, line 359: We should add tooling for this. The IDL serializer can infer inline IO, for example, and we could do similar here.

docs/source-2.0/guides/building-codegen/mapping-shapes-to-languages.rst
```diff
@@ -64,6 +64,13 @@ built-in type is available hurts the developer experience of the generator
 and should be avoided. Furthermore, changing the names of these types in the
 Smithy model should not impact generated code.
 
+.. versionadded:: 2.1
+   :ref:`Inline collection declarations <idl-inline-collections>` produce
+   synthetic list and map shapes with auto-generated names (e.g.,
+   ``_SyntheticListOfString``). These names are implementation details and
+   SHOULD NOT be used as generated type names. Generators can detect synthetic
+   shapes by checking for the ``smithy.synthetic#generated`` trait.
```
JordonPhillips, line 72: This isn't the same trait described in the design doc. This version is referenced in a lot of places.

docs/source-2.0/spec/aggregate-types.rst
```diff
@@ -72,6 +82,17 @@ The following example defines a map of strings to integers:
         value: Integer
     }
 
+.. versionadded:: 2.1
+   maps can be declared inline in member target positions using ``{Key: Value]``
```
JordonPhillips:
```suggestion
   maps can be declared inline in member target positions using ``{Key: Value}``
```

docs/source-2.0/spec/idl.rst
```diff
@@ -1423,6 +1479,67 @@ The suffixes for the generated names can be customized using the
...
+
+**Limitations**
+
+- Inline collections require ``$version: "2.1"`` or later.
+- Traits cannot be applied to synthetic shapes. Collections requiring
+  shape-level traits (such as ``@sparse``, ``@uniqueItems``, or ``@length``)
+  must use explicit shape definitions.
+- The ``apply`` statement cannot target synthetic shapes.
+- Shape names starting with ``_Synthetic`` are reserved for
+  assembler-generated shapes.
```
JordonPhillips: Can't do that. That's backwards-incompatible.

docs/source-2.0/spec/json-ast.rst
```diff
@@ -289,6 +289,55 @@ The following example defines a map of strings to numbers:
...
+                "traits": {
+                    "smithy.synthetic#generated": {}
+                }
+            }
+        }
+    }
+
+The ``smithy.synthetic#generated`` trait:
+
+- Is not user-applicable: only the assembler attaches it.
```
JordonPhillips: This isn't true. It isn't *intended* to be user-applicable. You can't stop it from being applied manually.

docs/source-2.0/spec/type-refinement-traits.rst
```diff
@@ -516,6 +516,11 @@ The following example defines a :ref:`map <map>` shape that MAY contain
         value: String
     }
 
+.. versionadded:: 2.1
+   The ``@sparse`` trait cannot be applied to synthetic shapes created by
+   :ref:`inline collection declarations <idl-inline-collections>`. To define
+   a sparse list or map, use an explicit shape definition.
+
```
JordonPhillips: This is unnecessary clutter. No traits can be applied to synthetic collections.

JordonPhillips (changes requested):

designs/inline-collections.md
sugmanue: If there's a `ListOfString` already we cannot merge those together, one can have traits that will inadvertently be applied to inline declarations. 

`$inlineListPrefix` is a good idea though.

designs/inline-collections.md
sugmanue: Yes, I wil change to clarify. That case is OK as the constraint is applied to the member and not the shape.

designs/inline-collections.md
sugmanue: Yes!, the IDL serializer is changed in the PR to create the inline collections declarations I think, I will double check that that is included.

designs/inline-collections.md
sugmanue: Yes, at least one internal SDK is known to do it.

designs/inline-collections.md
sugmanue: That's a good idea!

docs/source-2.0/guides/building-codegen/mapping-shapes-to-languages.rst
sugmanue: I will update the design doc to match this name.

docs/source-2.0/spec/idl.rst
sugmanue: I will clarify, we just emit a warning.

docs/source-2.0/spec/json-ast.rst
sugmanue: Well, it's kind of true as the code is written, yes, you can crate an AST with it but it cannot be applied to the IDL as of this change.

designs/inline-collections.md
sugmanue: I researched into this and it will work, similar to what `$operationInputSuffix` does, but I'd push back on the value of this. The `$operationInputSuffix` exists because teams had pre-existing naming conventions (e.g., Request/Response instead of Input/Output). For synthetic collections, there's no pre-existing convention to accommodate since the feature is new. The prefix is an implementation detail that users shouldn't need to think about, and it's there solely to prevent conflicts between already existing or added shapes after the inline declaration was introduced. We want clear semantics for those that are not easy to get wrong or cimcumbert.

designs/inline-collections.md
sugmanue: Yes, it does, `renderInlineCollectionTarget` in `SmithyIdlModelSerializer` takes care of it. The `idl-serialization/cases/inline-collections.smithy` test covers the IDL round-trip.

docs/source-2.0/spec/idl.rst
sugmanue: Changed the wording to say "SHOULD NOT"

docs/source-2.0/spec/json-ast.rst
sugmanue: Clarified this point.

docs/source-2.0/spec/type-refinement-traits.rst
sugmanue: removed

designs/inline-collections.md
sugmanue: done

designs/inline-collections.md
sugmanue: Sorry, I misunderstood, yeah, a tool that detects and suggest to convert it would be nice.

designs/inline-collections.md
sugmanue: Yes, I will clarify that.

designs/inline-collections.md
```diff
@@ -0,0 +1,426 @@
...
+Note: It is theoretically possible for a user-defined shape to
+conflict with a synthetic name (e.g., a user defines a shape named
+`_SyntheticListOfString`). In practice this is unlikely, and the
+assembler SHOULD emit a validation error if a conflict is detected,
+instructing the user to rename their shape.
+
+The synthetic shapes are placed in the same namespace as the structure
+that contains them.
+
+### Grouping
```
JordonPhillips, line 155: This section basically repeats information from the above section. This is actually somewhat confusing since it doesn't bring up namespacing.

designs/inline-collections.md
```diff
@@ -0,0 +1,426 @@
...
+`@sparse` to `_SyntheticListOfString` would make every `[String]`
+member in the namespace sparse, which is almost never the intent. To
+prevent this, traits cannot be applied to synthetic shapes.
+
+No traits may be applied to synthetic shapes, either at definition
+time or via `apply` statements. The `apply` statement MUST NOT target
+a shape that has the `@generated` trait.
+
+Traits that would normally apply to a collection shape (such as
+`@sparse`, `@uniqueItems`, or `@length` when constraining collection
```
JordonPhillips: length needs to be removed here.

designs/inline-collections.md
```diff
@@ -0,0 +1,426 @@
...
+
+Note that some traits like `@length` can also be applied to the
+containing member. When applied to a member, the trait constrains only
+that specific member usage, not the shared synthetic shape. For
+example, `@length(min: 1) names: [String]` constrains the `names`
+member but does not affect other members that reference the same
+`_SyntheticListOfString` shape.
+
+Traits on the containing member (e.g., `@documentation`,
+`@deprecated`) apply to the member itself, not to the synthetic shape.
```
JordonPhillips: The second paragraph is repeated information

designs/inline-collections.md
```diff
@@ -0,0 +1,426 @@
...
+
+### Referencing synthetic shapes
+
+Synthetic shape names exist as valid shape IDs in the semantic model
+and AST. However, since traits cannot be applied to synthetic shapes
+and their members cannot be independently referenced, there is no
+practical use case for referencing them directly in IDL `apply`
+statements or member target positions.
```
JordonPhillips: Nit: apply also exists in the AST

smithy-model/src/main/java/software/amazon/smithy/model/traits/Trait.java
```diff
@@ -65,6 +65,33 @@ default boolean isSynthetic() {
...
+    /**
+     * Controls where a trait is serialized.
+     */
+    enum SerializationMode {
+        /** Trait is serialized in both JSON AST and IDL. */
+        ALL,
+        /** Trait is serialized in the JSON AST only, not in the IDL. */
+        AST_ONLY,
+        /** Trait is never serialized (purely in-memory). */
+        NONE
```
JordonPhillips, line 92: This should probably be applied to the synthetic enum trait

smithy-model/src/main/java/software/amazon/smithy/model/traits/synthetic/SyntheticShapeTrait.java
```diff
@@ -0,0 +1,49 @@
...
+ * Marks a shape as assembler-generated (synthetic).
+ *
+ * <p>This trait is applied by the assembler to shapes created from inline
+ * collection syntax (e.g., {@code [String]} or {@code {String: String}}).
+ * It is not user-applicable and does not need a definition in the prelude.
+ *
+ * <p>Unlike other synthetic traits, this trait is persisted in the JSON AST
+ * so that IDL serializers can reconstruct inline syntax from it.
+ */
+public final class SyntheticShapeTrait extends AnnotationTrait {
```
JordonPhillips, line 22: This trait isn't described in a model but it serialized in models. That's a problem, because now an average model is not going to be able to describe itself. I'd rather this were just in the prelude.

docs/source-2.0/guides/evolving-models.rst
```diff
@@ -144,6 +144,39 @@ The following changes to trait definitions are backward compatible:
...
+``_SyntheticListOfString``). This may be a breaking change for:
+
+- Other models referencing the shape by name.
+- Code generators that derive type names from shape IDs.
+
+**Migrating from inline to explicit**
+
+Extracting an inline collection into a named shape is also wire-compatible
+but changes the shape ID. This is necessary when the collection needs
+shape-level traits such as ``@sparse``, ``@uniqueItems``, or ``@length``.
```
JordonPhillips:
```suggestion
shape-level traits such as ``@sparse`` or ``@uniqueItems``.
```

docs/source-2.0/spec/aggregate-types.rst
```diff
@@ -27,6 +27,17 @@ The following example defines a list with a string member from the
         member: String
     }
 
+.. versionadded:: 2.1
+   Lists can be declared inline in member target positions using ``[Target]``
```
JordonPhillips:
```suggestion
   Lists can be declared inline using ``[Target]``
```

docs/source-2.0/spec/aggregate-types.rst
```diff
@@ -72,6 +82,17 @@ The following example defines a map of strings to integers:
         value: Integer
     }
 
+.. versionadded:: 2.1
+   maps can be declared inline in member target positions using ``{Key: Value}``
```
JordonPhillips:
```suggestion
   maps can be declared inline using ``{Key: Value}``
```

docs/source-2.0/spec/aggregate-types.rst
```diff
@@ -108,6 +129,18 @@ followed by ``$value``. For example, the shape ID of the ``key`` member in
...
+   ``{Key: Value}`` syntax. Inline maps produce a synthetic shape that is
+   shared across all members using the same key and value types.
+   See :ref:`idl-inline-collections`.
+
+   .. code-block:: smithy
+
+       structure Example {
+           tags: {String: String}
+       }
+
```
JordonPhillips: dupe

docs/source-2.0/spec/idl.rst
```diff
@@ -953,6 +965,17 @@ Traits can be applied to the list shape and its member:
         }
 
 
+.. versionadded:: 2.1
+   Lists can also be declared inline in member target positions using
```
JordonPhillips:
```suggestion
   Lists can also be declared inline using
```

docs/source-2.0/spec/idl.rst
```diff
@@ -1049,6 +1072,17 @@ Traits can be applied to the map shape and its members:
         }
 
 
+.. versionadded:: 2.1
+   Maps can also be declared inline in member target positions using
```
JordonPhillips:
```suggestion
   Maps can also be declared inline using
```

docs/source-2.0/spec/json-ast.rst
```diff
@@ -289,6 +289,57 @@ The following example defines a map of strings to numbers:
     }
 
 
+.. _ast-synthetic-shapes:
```
JordonPhillips: I don't think this section is necessary.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/MigrateCommand.java
```diff
@@ -60,7 +60,7 @@ final class MigrateCommand implements Command {
 
     private static final Logger LOGGER = Logger.getLogger(MigrateCommand.class.getName());
     private static final Pattern VERSION_1 = Pattern.compile("(?m)^\\s*\\$\\s*version:\\s*\"1(\\.0)?\"\\s*$");
-    private static final Pattern VERSION_2 = Pattern.compile("(?m)^\\s*\\$\\s*version:\\s*\"2(\\.0)?\"\\s*$");
+    private static final Pattern VERSION_2 = Pattern.compile("(?m)^\\s*\\$\\s*version:\\s*\"2(\\.\\d)?\"\\s*$");
```
JordonPhillips, line 63: This should be split off, at least to its own commit.

smithy-diff/src/test/resources/software/amazon/smithy/diff/diffs/synthetic/explicit-to-inline.events
```diff
@@ -0,0 +1,3 @@
+[ERROR] smithy.example#TagList: Removed list `smithy.example#TagList` | RemovedShape
+-----
+[WARNING] smithy.example#MyStructure$tags: The shape targeted by the member `smithy.example#MyStructure$tags` changed from `smithy.example#TagList` (list) to `smithy.example#_SyntheticListOfString` (list). This was determined backward compatible. | ChangedMemberTarget
```
JordonPhillips: We are gonna get so many tickets. Just an unbelievable amount. Why not just flatten this out in this case?

smithy-model/src/main/java/software/amazon/smithy/model/loader/SyntheticShapeNaming.java
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.model.loader;
+
+/**
+ * Generates deterministic names for synthetic shapes created from inline
+ * collection declarations.
+ */
+final class SyntheticShapeNaming {
```
JordonPhillips: These should be moved to `LoaderUtils` or back into `IdlModelLoader` since they're only needed there anyway.

smithy-model/src/main/java/software/amazon/smithy/model/loader/Version.java
```diff
@@ -20,193 +20,37 @@
  */
 enum Version {
 
-    UNKNOWN {
```
JordonPhillips, line 23: nit: This change doesn't need to be part of this pr

JordonPhillips (changes requested):
Looking good! Mostly doc stuff

docs/source-2.0/spec/json-ast.rst
sugmanue: Fair, removed.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/MigrateCommand.java
sugmanue: It's already in its own commit, see [50efb0](https://github.com/smithy-lang/smithy/pull/3134/changes/50efb0be0b2dce9cc124f2b9e57027aa31039028).

smithy-diff/src/test/resources/software/amazon/smithy/diff/diffs/synthetic/explicit-to-inline.events
sugmanue: I guess we could special case this in diff, but not sure how is this different from any other similar change.

smithy-model/src/main/java/software/amazon/smithy/model/traits/Trait.java
sugmanue: Yes, that's how it works for synthetic shapes by default, see default implementation above.

smithy-model/src/main/java/software/amazon/smithy/model/traits/synthetic/SyntheticShapeTrait.java
sugmanue: Not adding it to the prelude was intentional to prevent people from using it. I guess we can make it `@private`, thoughts?

smithy-model/src/main/java/software/amazon/smithy/model/loader/SyntheticShapeNaming.java
sugmanue: I will move it into `LoaderUtils`.

smithy-model/src/main/java/software/amazon/smithy/model/traits/synthetic/SyntheticShapeTrait.java
JordonPhillips: I don't think omitting it from the prelude will prevent people from using it. Private won't work because all the applications will trigger errors. If you want to block it from the IDL you'll need to specifically block it in the loader.

smithy-diff/src/test/resources/software/amazon/smithy/diff/diffs/synthetic/explicit-to-inline.events
JordonPhillips: We have historically fielded a TON of confusion from people switching semantically equivalent shape types, like from one string to another, usually from `String` to `MyString` or something similarly trivial. That's why the shape target change is a warning: because we specifically changed that diff validator to have a lower severity if the new target is effectively compatible. Even then though it was usually a case of switching from some prelude target to a local target, which wouldn't trigger the error since the prelude is never really removed.

Now we're going to be encouraging people to make changes that trigger an error diff. We should not be doing that. This is a detectable scenario. iirc there's some interface you can use to decorate events. You should use that to lower the severity on the removed shape here and add some context like "All previous references to this shape changed to use inline collection syntax. See the ChangedMemberTarget evaluator to see any effective trait changes.

sugmanue: I have pushed a commit to fix this. Hopefully it addresses your concerns.

JordonPhillips (approved):

PR #3148, opened 2026-05-29
Add a shapeClosures metadata key

JordonPhillips:
This adds a shapeClosures metadata key that allows users to define closures of shapes that are not necessarily rooted in a service shape.

This is primarily motivated by the desire to code-generate types without service or client framing, but is left open enough to be used for other purposes.

---


smithy-build/src/test/java/software/amazon/smithy/build/transforms/FlattenClosureNamespacesTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // service rename keys round-trip through the same machinery.
+        ArrayNode shapeClosures = result.getMetadata().get("shapeClosures").expectArrayNode();
+        List<Node> closures = shapeClosures.getElements();
+        ObjectNode closure = closures.get(0).expectObjectNode();
+        ObjectNode rename = closure.expectObjectMember("rename");
+        assertTrue(rename.getStringMember("com.flat#RenamedFoo").isPresent());
+        assertFalse(rename.getStringMember("com.example#Foo").isPresent());
+    }
+
+    @Test
```
yefrig: this test doesn't seem to belong here

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/shapeClosures/shape-closure-empty-selector.errors
```diff
@@ -0,0 +1 @@
+[ERROR] -: metadata.shapeClosures.0.includeBySelector: String value provided for `smithy.api#ShapeClosure$includeBySelector` must be | TypedMetadata.shapeClosures
```
yefrig: I wonder why the string is getting truncated here

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/shapeClosures/shape-closure-rename-rules.errors
```diff
@@ -0,0 +1,11 @@
...
+[ERROR] -: metadata.shapeClosures.2.rename.com.example#WidgetResource (map-key): Renames must target shapes that are in the model and must not target services, resources, operations, or members. | TypedMetadata.shapeClosures
+[ERROR] -: metadata.shapeClosures.3.rename.com.example#WidgetService (map-key): Renames must target shapes that are in the model and must not target services, resources, operations, or members. | TypedMetadata.shapeClosures
+[ERROR] -: Shape closure `com.example#NonRenamedConflict` rename for `com.example#Foo` ("BAR") case-insensitively conflicts with non-renamed shape `com.example#Bar`. | ShapeClosure
+[ERROR] -: Shape closure `com.example#RenameRenameConflict` rename for `com.example#Bar` ("CONFLICT") case-insensitively conflicts with the rename for `com.example#Foo`. | ShapeClosure
+[ERROR] -: metadata.shapeClosures.6.rename.com.example#Foo: String value provided for `smithy.api#Identifier` must match regular expression: | TypedMetadata.shapeClosures
+[ERROR] -: Shape closure `com.example#BadSelector` has an invalid `includeBySelector`: | ShapeClosure
```
yefrig: It seems like the SelectorException message is coming up empty :/

JordonPhillips:
@yefrig I did some re-orgnaizing of the tests in general. There were some artifacts from when I initially had them all in a single file that I hadn't addressed.

Additionally, I refactored the validator a bit to avoid having to run the selector twice. Selectors can be expensive, so I really don't want to be doing that. I ended up just adding a `getRootShapesInClosure` method to the index, which enables me to filter out root shapes that are already covered by included namespaces. If anything is left, the selector is (probably) empty. This actually changes behavior a bit because an empty selector can be reported even if the selector does match things. But in that instance it means the selector is redundant, so it's probably worth calling out anyway.

Also, I added the clarification to the docs that the closure id has no formal meaning. Michael had asked for that, but I forgot to include it.

yefrig (approved):

PR #3151, opened 2026-06-02
Add message property to waiter acceptors

KaramVSingh:
Add an optional `message` field to waiter acceptors that allows extracting a human-readable message from the operation output via a JMESPath expression. This enables waiter consumers (e.g., SDKs, CLIs) to surface actionable failure reasons to users.

The expression must resolve to a string or array of strings and is only valid on acceptors with a failure state.

Validations:
- ERROR if message is set on a non-failure acceptor
- ERROR if the JMESPath expression has invalid syntax
- ERROR if the expression resolves to a type other than string or array of strings

#### Background
* What do these changes do? Adds message to Smithy waiters
* Why are they important? Helps give waiter consumers better insight about next steps

#### Testing
* How did you test these changes? built waiter sub-project
* 
---


smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaitableTraitValidator.java
```diff
@@ -52,6 +52,19 @@ private List<ValidationEvent> validateOperation(Model model, OperationShape oper
...
+                            String.format(
+                                    "Waiter `%s`, acceptor %d: The `message` property is only valid on "
+                                            + "acceptors with a `failure` state, but found state `%s`.",
+                                    waiterName,
+                                    i,
+                                    acceptor.getState())));
+                }
+                if (acceptor.getMessage().isPresent() && acceptor.getState() == AcceptorState.FAILURE) {
+                    events.addAll(visitor.validateMessage(acceptor.getMessage().get()));
+                }
```
sugmanue, line 67:
```suggestion
                if (acceptor.getMessage().isPresent()) {
                    if (acceptor.getState() == AcceptorState.FAILURE) {
                        events.addAll(visitor.validateMessage(acceptor.getMessage().get()));
                    } else {
                        events.add(error(operation,
                                         trait,
                                         String.format(
                                                       "Waiter `%s`, acceptor %d: The `message` property is only valid on "
                                                       + "acceptors with a `failure` state, but found state `%s`.",
                                                       waiterName,
                                                       i,
                                                       acceptor.getState())));
                    }
                }
```

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
```diff
@@ -204,4 +204,60 @@ private void addEvent(Severity severity, String message, String... eventIdParts)
...
+                                    "The `message` JMESPath expression must resolve to a string or array of "
+                                            + "strings, but this expression resolves to an array of `%s`.",
+                                    new LiteralExpression(array.get(0)).getType()),
+                            JMESPATH_PROBLEM,
+                            "MessageTypeMismatch",
+                            waiterName,
+                            String.valueOf(acceptorIndex));
+                }
+            } else if (returnType != RuntimeType.STRING && returnType != RuntimeType.ANY) {
+                addEvent(Severity.DANGER,
```
sugmanue: This being a new feature can be strict. Why this is not an `ERROR` instead? what's the use case for having it resolve to anything other than a `String`

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
```diff
@@ -204,4 +204,60 @@ private void addEvent(Severity severity, String message, String... eventIdParts)
...
+
+        try {
+            JmespathExpression expression = JmespathExpression.parse(messageExpression);
+            LinterResult result = expression.lint(input);
+            for (ExpressionProblem problem : result.getProblems()) {
+                addJmespathEvent(messageExpression, problem);
+            }
+
+            RuntimeType returnType = result.getReturnType();
+            if (returnType == RuntimeType.ARRAY) {
```
sugmanue, line 226: What's the use case for accepting arrays of strings?

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
KaramVSingh: Youre right this is not the right decision - updating

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
KaramVSingh: Some APIs return Reasons as a list, I can scope this to strings for now and we can extend later

sugmanue (approved):

smithy-waiters/src/main/java/software/amazon/smithy/waiters/WaiterMatcherValidator.java
mtdowling: If you have a use case for this, keep it a list of strings IMO

KaramVSingh: Updating - thanks :D

PR #3156, opened 2026-06-04
Support directed type and closure codegen

JordonPhillips:
This adds support for codegen based on shape closures rather than service closures in directed codegen. It also adds support for a mode that only generates data shapes, which can be used in either mode.

---


docs/source-2.0/guides/building-codegen/configuring-the-generator.rst
```diff
@@ -58,11 +58,22 @@ example, ``"smithy.example#Weather"``). Smithy models can contain
...
+
+The ``closure`` property defines the ID of the
+:ref:`shape closure <shape-closures>` to generate (for example,
+``"smithy.example#Events"``).
+
+``service`` and ``closure`` optionality
+---------------------------------------
+
+Generators should make both ``service`` and ``closure`` optional. The
+generator must fail if both are set. If neither are set, the
```
joewyz: nit: neither is set

joewyz (approved):

PR #3157, opened 2026-06-04
Added test cases for errors in AWS JSON protocols using different namespaces

kuhe:
#### Background
Adds test cases for the AWS JSON protocols that cover the case where service returned error identifier namespace is different from those in the SDK's known model.

The AWS SDK for JS had a bug caused by my misunderstanding of the protocol spec detailed in https://github.com/aws/aws-sdk-js-v3/pull/8031.

That bug made JSON 1.0 error parsing not strip the namespace when performing error lookup. Because the test namespace is the same as the SDK's model namespace, this did not fail in protocol tests, which only assert that a particular error type was resolved.

In practice, most JSON 1.0 services return a different namespace than what exists in their models.

#### Testing
Tested by generating AWS protocol tests for the AWS SDK JSv3, which passed.

#### Links
* Bugfix in JSv3 https://github.com/aws/aws-sdk-js-v3/pull/8031
* Error parsing https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html

---

### DIFF
--- a/.changes/next-release/feature-115f276aaf230e5e0b9502b6d0bcc9594962b529.json
+++ b/.changes/next-release/feature-115f276aaf230e5e0b9502b6d0bcc9594962b529.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add test cases for AWS JSON protocols using different namespaces",
+  "pull_requests": [
+    "[#3157](https://github.com/smithy-lang/smithy/pull/3157)"
+  ]
+}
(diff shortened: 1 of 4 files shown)


.changes/next-release/feature-115f276aaf230e5e0b9502b6d0bcc9594962b529.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Add test cases for AWS JSON protocols using different namespaces",
```
sugmanue:
```suggestion
  "description": "Added test cases for errors in AWS JSON protocols using different namespaces",
```

sugmanue (approved):

PR #3158, opened 2026-06-07
Relax MemberShouldReferenceResource to one-of-N matching

shduke:
Previously the validator warned about every matching resource not listed in `@references`, which created noise whenever an identifier name (e.g. `userId`) collided across multiple resources. Now declaring a reference to at least one matching resource silences the warning for the rest; the warning still fires (listing all matches) when none are referenced.

#### Testing
* Added `multi-match-member-references.smithy` errorfile test validating the new one-of-N behavior works
* Full smithy test suite passes.

---


shduke:
@sugmanue any thoughts on this?

PR #3163, opened 2026-06-10
Sugmanue/idl 2.1/tagged literals

sugmanue:
#### Background

Implements tagged string literals for Smithy IDL 2.1. A tagged string literal (`#tag "content"`) changes how string content is interpreted at parse time, producing a regular value. Four tags are supported:

- `#re` regex patterns without double-escaping
- `#b` binary byte sequences (base64 output)
- `#hex` hex dumps with comments (base64 output)
- `#timestamp` RFC 3339 to epoch-seconds conversion

The tokenizer emits `TAG` + `RAW_STRING`/`RAW_TEXT_BLOCK` tokens; the parser validates the tag, checks version 2.1 support, and delegates to tag-specific handlers. The smithy-syntax package provides a `TAGGED_LITERAL` tree type for LSP support and the formatter normalizes spacing.

#### Testing

- Parameterized unit tests for each tag handler (escape rules, error cases)
- Tokenizer-level tests (TAG + RAW_STRING emission, disambiguation, fallback to POUND)
- Valid model tests (IDL + JSON pairs via ValidSmithyModelLoaderRunnerTest)
- Error model tests (version gating, invalid hex, invalid timestamp)
- IDL round-trip serialization test
- Formatter tests (preservation and spacing normalization)
- Shape ID conflict test (e.g., `smithy.example#re` is not mistaken for a tagged literal)

#### Links

- Design: `designs/tagged-string-literals.md`
- Draft prototype: #3047

---


designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+* **Author**: Manuel Sugawara
+* **Created**: 2026-05-14
+
+## Abstract
+
+This proposal introduces tagged string literals to the Smithy IDL. A tagged string literal is a string prefixed with
+`#<tag>` that changes how the string content is interpreted at parse time. The tag instructs the parser to use
+alternative rules for processing escape sequences and encoding, producing a regular string value. This is purely
+syntactic sugar; the AST and semantic model remain unchanged. Four initial tags are proposed: `#re` (regex), `#b`
+(binary), `#hex` (hex dump), and `#timestamp` (ISO date to epoch seconds).
```
JordonPhillips: How about just `#t` for timestamp?

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+    """
+```
+
+The tag identifier follows the same rules as Smithy identifiers (letters, digits, underscores; must start with a letter
+or underscore).
+
+#### ABNF
+
+```
+tagged_string_literal = "#" identifier (quoted_text / text_block)
```
JordonPhillips: This doesn't allow for whitespace between the identifier and the string, but the above sample includes it.

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+A tagged string literal can appear anywhere a regular string literal can appear in the IDL (trait values, node values,
+etc.).
+
+### Disambiguation
+
+The `#` character is already used in shape IDs (e.g., `com.example#Shape`). The tokenizer disambiguates by checking two
+conditions: the identifier after `#` must be a known tag name, and it must be followed by a string literal (`"` or
+`"""`). If either condition fails, `#` is emitted as a plain `POUND` token and subsequent characters are tokenized
+normally. Because the set of known tags is small and fixed (`re`, `b`, `hex`, `timestamp`), collisions with shape member
+names are unlikely in practice.
```
JordonPhillips, line 71: Is this a real issue? It doesn't seem like it. The formations are incompatibly different and they appear in different contexts so I don't see how a collision is conceptually possible.

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Processing model
+
+Tagged string literals are processed across two layers:
```
JordonPhillips, line 75: This says "two layers" followed by a four element list

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+|Tagged literal|Equivalent plain string|Meaning|
+|-|-|-|
+|`#b "Hello world"`|`"SGVsbG8gd29ybGQ="`|UTF-8 bytes, base64-encoded|
+|`#b "\x00\x01\x02"`|`"AAEC"`|Three bytes: 0, 1, 2|
+|`#b "\xff"`|`"/w=="`|Single byte: 255|
+
+#### `#hex` Hex dump literals
+
+The `#hex` tag interprets the string as a hex dump. The string contains hexadecimal values where spaces are ignored and
```
JordonPhillips, line 142: Is there some standard that we can link to here? (Or rather, in the docs)

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### `#timestamp` Timestamp literals
+
+The `#timestamp` tag converts an ISO 8601 date/time string into its `epoch-seconds` representation, following the same
+rules as the `epoch-seconds` timestamp format defined by Smithy. This makes timestamp default values and test data more
+readable.
+
+**Rules:**
+
+* Input must be a valid ISO 8601 date/time string (as defined by the `date-time` production in RFC 3339 Section 5.6)
```
JordonPhillips: Best not to even mention ISO 8601 because it could introduce confusion. Just say RFC3339, as we do in the timestamp format trait docs.

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+* Output is the number of seconds that have elapsed since 00:00:00 Coordinated Universal Time (UTC),
+  Thursday, 1 January 1970, with optional millisecond precision. Values that are more granular than
+  millisecond precision SHOULD be truncated to fit millisecond precision. This follows the same rules
```
JordonPhillips: I don't think should is appropriate here. We're doing it, so what are we doing. This seems to be largely copied from the timestamp format docs, but this particular part doesn't make sense in this context.

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+**Examples:**
+
+|Tagged literal|Equivalent value|Meaning|
+|-|-|-|
+|`#timestamp "2024-01-01T00:00:00Z"`|`1704067200`|Midnight Jan 1, 2024 UTC|
+|`#timestamp "2024-06-15T12:30:00.500Z"`|`1718451000.5`|With milliseconds|
+
+### Text block support
+
+Both tags work with text blocks. The text block formatting rules (leading whitespace removal, line normalization) are
```
JordonPhillips: Both?

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+
+Using a single-letter prefix directly attached to the string, like Python's `b'...'` or Rust's `r"..."`.
+
+This was rejected because:
+
+* It only solves one use case at a time. Binary gets `b"..."`, but regex would need a separate
+  mechanism (e.g., `r"..."`) — each one is a new special case.
+* A letter followed by a string is ambiguous with an identifier followed by a string value,
+  requiring context-sensitive parsing.
+* Not extensible without grammar changes — each new prefix is a new production.
```
JordonPhillips: I'm pretty sure you can define this generically in abnf just fine - it's just that the parsing would suck.

Also, point 1 and 3 are the same thing

designs/tagged-string-literals.md
```diff
@@ -0,0 +1,329 @@
...
+* Preserving them would complicate every tool that processes the AST.
+* The AST already has a well-defined string representation; adding variants
+  would break existing consumers.
+
+## Limitations
+
+* Tagged string literals are IDL-only. They cannot be expressed in the JSON AST format.
+* The set of tags is fixed and hardcoded. New tags require changes to the Smithy parser.
+* IDL round-tripping: converting AST back to IDL cannot reconstruct tagged literals since the tag
+  information is not preserved. The plain string equivalent is emitted instead.
```
JordonPhillips, line 285: This is a show-stopper. We format by default in the gradle plugin, so we'd be wiping this all away, defeating the readability goals.

docs/source-2.0/spec/idl.rst
```diff
@@ -165,6 +165,7 @@ string support defined in :rfc:`7405`.
     TextBlock           :`ThreeDquotes` [`SP`] `NL` *`TextBlockContent` `ThreeDquotes`
     TextBlockContent    :`QuotedChar` / (1*2DQUOTE 1*`QuotedChar`)
     ThreeDquotes        :DQUOTE DQUOTE DQUOTE
+    TaggedStringLiteral :"#" `Identifier` (`QuotedText` / `TextBlock`)
```
JordonPhillips: Same issue mentioned above - there's no affordance for whitespace between the identifer and text. I don't mind that to be honest, but calling it out since all the examples i've seen have a space.

docs/source-2.0/spec/idl.rst
```diff
@@ -2516,4 +2523,201 @@ example is interpreted as ``Foo\nBaz Bam``:
...
+    * - ``#b``
+      - Binary literal. Interprets the string as a sequence of bytes using
+        hex escapes (``\xHH``), octal escapes (``\OOO``), and named escapes.
+        Produces a base64-encoded string.
+    * - ``#hex``
+      - Hex dump literal. Interprets the string as hexadecimal byte values where
+        spaces are ignored and ``#`` begins a comment until end of line.
+        Produces a base64-encoded string.
+    * - ``#timestamp``
+      - Timestamp literal. Converts an ISO 8601 date/time string into its
```
JordonPhillips: RFC3339

docs/source-2.0/spec/idl.rst
```diff
@@ -2516,4 +2523,201 @@ example is interpreted as ``Foo\nBaz Bam``:
...
+      - Description
+    * - ``#re``
+      - Regular expression literal. Backslash sequences are passed through
+        literally, allowing regex patterns to be written without double-escaping.
+    * - ``#b``
+      - Binary literal. Interprets the string as a sequence of bytes using
+        hex escapes (``\xHH``), octal escapes (``\OOO``), and named escapes.
+        Produces a base64-encoded string.
+    * - ``#hex``
+      - Hex dump literal. Interprets the string as hexadecimal byte values where
```
JordonPhillips, line 2586: Definitely need to link to some docs or something here.

JordonPhillips (changes requested):

designs/tagged-string-literals.md
sugmanue: The issue is that the tokenizer doesn't have context to know if it should produce POUND + IDENTIFIER or TAG, and this matters for the smithy-syntax package.

designs/tagged-string-literals.md
sugmanue: It refers to the tokenizer and parser, let me rephrase to make clear that.

designs/tagged-string-literals.md
sugmanue: No, there's no standard that I'm aware of.

designs/tagged-string-literals.md
sugmanue: I will remove the second sentence.

designs/tagged-string-literals.md
sugmanue: I will change by "All ..."

designs/tagged-string-literals.md
sugmanue: > I'm pretty sure you can define this generically in abnf just fine - it's just that the parsing would suck.

Yes, you can. The point is that it will change the grammar, just that.

designs/tagged-string-literals.md
sugmanue: The formatter works with the syntax tree, which already supports and handles the new grammar. Formatting won't remove the tagged literals.

docs/source-2.0/spec/idl.rst
sugmanue: Yes, let me add it.

docs/source-2.0/spec/idl.rst
sugmanue: I will cross reference the section below that documents the format.

designs/tagged-string-literals.md
sugmanue: I'm not sure about it, I think that `#timestamp` works better. Unlike the others, `t` is not a well established abbreviation for "timestamps".

designs/tagged-string-literals.md
JordonPhillips: Why would it ever produce POUND + IDENTIFIER though? There is no other production that satisfies that.

docs/source-2.0/spec/idl.rst
```diff
@@ -2516,4 +2523,216 @@ example is interpreted as ``Foo\nBaz Bam``:
...
+
+The content of the string is parsed according to the following grammar:
+
+.. code-block:: none
+
+    hex_content = *(hex_pair / WS / comment / NL)
+    hex_pair    = Hex Hex
+    comment     = "#" *(%x00-09 / %x0B-0D / %x0E-10FFFF) ; # until end of line
+    WS          = SP / HTAB
+    NL          = LF / CRLF
```
JordonPhillips: nit: productionlist

Also, `CamelCase`

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
```diff
@@ -357,6 +363,96 @@ private IdlToken parseIdentifier() {
...
+            currentTokenEnd = parser.position();
+            return currentTokenType = IdlToken.POUND;
+        }
+
+        // Read the identifier.
+        parser.consumeWhile(ParserUtils::isValidIdentifierCharacter);
+        int afterTag = parser.position();
+        String tag = parser.sliceFrom(savedPos + 1);
+
+        // Only recognize known tags to avoid conflicting with shape IDs.
```
JordonPhillips, line 383: Again, I really don't think this is an issue. I haven't got to the tests yet, but please add a test case that exercises the issue you see here.

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
```diff
@@ -357,6 +363,96 @@ private IdlToken parseIdentifier() {
...
+        }
+
+        CharSequence rawContent = parseRawStringContents(isTextBlock);
+        currentTokenEnd = parser.position();
+        currentTokenStringSlice = rawContent;
+        return currentTokenType = isTextBlock ? IdlToken.RAW_TEXT_BLOCK : IdlToken.RAW_STRING;
+    }
+
+    /**
+     * Reads raw content between quotes without applying escape processing.
```
JordonPhillips: You still need to allow escaping double quotes to a degree. If you don't, then dquotes are unrepresentable in raw strings and triple-dquotes are unrepresentable in raw text blocks.

It's a tricky situation. Python handles this in [raw strings](https://docs.python.org/3/reference/lexical_analysis.html#raw-string-literals) by letting you escape the quote for the purpose of not terminating the string, but leaves the slash in the parsed value. That is:

```python
r"foo\"bar" == "foo\\\"bar"
```

The individual tag handlers can decide what to do with that sequence, but the tokenizer needs to know whether to go on or not.

For raw strings, I think it's fine that we carry forward python behavior.

There's also the question of whether we allow escaping newlines in raw strings. It's not a matter of grammar ambiguity so I think it doesn't clear the bar. Python also doesn't allow this.

docs/source-2.0/spec/idl.rst
```diff
@@ -165,6 +165,7 @@ string support defined in :rfc:`7405`.
     TextBlock           :`ThreeDquotes` [`SP`] `NL` *`TextBlockContent` `ThreeDquotes`
     TextBlockContent    :`QuotedChar` / (1*2DQUOTE 1*`QuotedChar`)
     ThreeDquotes        :DQUOTE DQUOTE DQUOTE
+    TaggedStringLiteral :"#" `Identifier` [`SP`] (`QuotedText` / `TextBlock`)
```
JordonPhillips: I don't think the `QuotedText` or `TextBlock` productions are actually valid here. Both define escapes that will not (necessarily) be honored. I think you need `RawTextBlock` and `RawQuotedText` productions. I mention this in another comment, but think about what escapes need to be honored. I think that a dquote escape should be honored, even for raw strings, otherwise the grammar is ambiguous. (See the other comment about raw strings in particular.)

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
```diff
@@ -0,0 +1,264 @@
...
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Handles tagged string literals in the Smithy IDL.
+ *
+ * <p>A tagged string literal is a string prefixed with {@code #tag} that changes
+ * how the string content is interpreted. The result can be a string or number token.
```
JordonPhillips:
```suggestion
 * <p>A tagged string literal is a string prefixed with a tag (e.g. {@code #b}) that changes
 * how the string content is interpreted. The result can be a string or number token.
```

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
```diff
@@ -0,0 +1,264 @@
...
+
+    static Result scan(String tag, CharSequence lexeme, boolean isTextBlock) {
+        lexeme = IdlStringLexer.normalizeAndFormat(lexeme, isTextBlock);
+        return HANDLERS.get(tag).apply(lexeme);
+    }
+
+    /**
+     * Scans regex string contents. Backslash sequences are passed through literally
+     * except for {@code \"} and {@code \\}.
+     */
```
JordonPhillips, line 70: I think we should be passing through all backslash sequences literally. I've got another comment about it that goes in more detail.

designs/tagged-string-literals.md
sugmanue: I mean the sequence of POUND followed by IDENTIFIER (used in shape id) should be distinguishable from the single token TAG.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
```diff
@@ -0,0 +1,264 @@
...
+    private static Result scanTimestampContents(CharSequence lexeme) {
+        try {
+            Instant instant = Instant.parse(lexeme);
+            long epochSecond = instant.getEpochSecond();
+            int millis = instant.getNano() / 1_000_000;
+            if (millis == 0) {
+                return Result.ofNumber(epochSecond);
+            }
+            double value = epochSecond + millis / 1000.0;
```
JordonPhillips: How about returning a BigDecimal instead? I'd prefer not to loose anything.

JordonPhillips (changes requested):

docs/source-2.0/spec/idl.rst
sugmanue: I will remove `WS` and `NL` (all upper case) are already defined at the beginning of the doc (see [here](https://smithy.io/2.0/spec/idl.html#smithy-idl-abnf)).

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
sugmanue: "without applying escape processing" in this case means that the behavior is what you described, see below, if a backslash is found then we skip the char following it, which will be passed verbatim.

docs/source-2.0/spec/idl.rst
sugmanue: Yeah, good catch. I will push an update to correct this.

docs/source-2.0/spec/idl.rst
```diff
@@ -165,7 +165,13 @@ string support defined in :rfc:`7405`.
     TextBlock           :`ThreeDquotes` [`SP`] `NL` *`TextBlockContent` `ThreeDquotes`
     TextBlockContent    :`QuotedChar` / (1*2DQUOTE 1*`QuotedChar`)
     ThreeDquotes        :DQUOTE DQUOTE DQUOTE
-    TaggedStringLiteral :"#" `Identifier` [`SP`] (`QuotedText` / `TextBlock`)
+    TaggedStringLiteral :"#" `Identifier` [`SP`] (`RawQuotedText` / `RawTextBlock`)
+    RawQuotedText       :DQUOTE *`RawQuotedChar` DQUOTE
+    RawQuotedChar       :%x09 / %x20-21 / %x23-5B / %x5D-10FFFF
+                        :/ `Escape` %x00-10FFFF
```
JordonPhillips: I think we still don't want to claim support for control codes that can't really be represented.

```suggestion
                        :/ `Escape` (%x09 / %x20-7E / %x80-10FFFF)
```

We probably shouldn't be claiming support for DEL (7F) in normal strings either....

docs/source-2.0/spec/idl.rst
```diff
@@ -2687,11 +2693,9 @@ The content of the string is parsed according to the following grammar:
...
 .. code-block:: none
 
-    hex_content = *(hex_pair / WS / comment / NL)
-    hex_pair    = Hex Hex
-    comment     = "#" *(%x00-09 / %x0B-0D / %x0E-10FFFF) ; # until end of line
-    WS          = SP / HTAB
-    NL          = LF / CRLF
+    HexContent = *(HexPair / WS / Comment / NL)
+    HexPair    = Hex Hex
+    Comment     = "#" *(%x00-09 / %x0B-0D / %x0E-10FFFF) ; # until end of line
```
JordonPhillips: Why the support for all those control characters here too? (Also, the last two of these are one contiguous segment.)

smithy-model/src/test/java/software/amazon/smithy/model/loader/TaggedStringLiteralTest.java
```diff
@@ -354,6 +355,40 @@ public void poundFollowedByNonIdentifier() {
         assertThat(t.getCurrentToken(), is(IdlToken.POUND));
     }
 
+    // --- Raw string escape handling ---
+
+    @Test
+    public void rawStringPreservesEscapedQuote() {
```
JordonPhillips: These tests you added tell me that the tokenizer doesn't omit these values, but it says nothing about how they're processed.

smithy-model/src/test/java/software/amazon/smithy/model/loader/TaggedStringLiteralTest.java
```diff
@@ -0,0 +1,401 @@
...
+        assertThat(TaggedStringLiteral.hasHandler("hex"), is(true));
+        assertThat(TaggedStringLiteral.hasHandler("timestamp"), is(true));
+    }
+
+    @Test
+    public void hasHandlerReturnsFalseForUnknown() {
+        assertThat(TaggedStringLiteral.hasHandler("unknown"), is(false));
+    }
+
+    // --- Tokenizer-level tests ---
```
JordonPhillips: These tests don't belong here. They aren't testing TaggedStringLiteral, they're testing the tokenizer, which doesn't invoke TaggedStringLiteral.

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
```diff
@@ -357,6 +363,96 @@ private IdlToken parseIdentifier() {
...
+        int start = parser.position();
+
+        while (!parser.eof()) {
+            char next = parser.peek();
+            if (next == '"' && (!triple || (parser.peek(1) == '"' && parser.peek(2) == '"'))) {
+                break;
+            }
+            parser.skip();
+            if (next == '\\') {
+                parser.skip();
```
JordonPhillips: I think the comment on this is somewhere between confusing and wrong. It's processing escapes right at the bottom here. Maybe "without evaluating" would be better wording?

smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/tagged-string-literals-spacing.formatted.smithy
JordonPhillips, line 1: We need some text block test cases here

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
```diff
@@ -0,0 +1,265 @@
...
+                char next = lexeme.charAt(i + 1);
+                if (next == '"') {
+                    result.append('"');
+                    i++;
+                } else if (next == '\\') {
+                    result.append('\\');
+                    i++;
+                } else if (next == '\n') {
+                    // Escaped newline: skip both backslash and newline.
+                    i++;
```
JordonPhillips: To be clear, I'd prefer that these are removed. The first isn't necessary because regex has generic escaping, so `\"` and `"` are semantically equivalent. The second isn't necessary because we *don't* have generic escaping so a `\` followed by a non-`"` character isn't treated as a single character.

For newlines, I'd rather they're just always skipped. without needing an escape. Regex has `\n` if you want to have a semantic newline. If I want to have a compelx regex I can just split it up without some dangling `\` adding confusion.

```
@pattern(#re """
    [A-Z]:\\\\
    ([\w\s\d]+\\)*
    ([\w\s\d]+(\.[\w\d]+)?)?
    """)
string WindowsPath
```

Now that I think of it, if we want to really make regex text blocks useful we could steal even more from Python and implement their [verbose syntax](https://docs.python.org/3/library/re.html#re.VERBOSE). So that example above could become:

```
@pattern(#re """
    [A-Z]:\\\\               # The Windows drive, e.g. C:\\
    ([\w\s\d]+\\)*           # Any number of path segments, e.g. \foo\bar\baz\
    ([\w\s\d]+(\.[\w\d]+)?)? # An optional file name with an optional extension
    """)
string WindowsPath
```

That'd be pretty sick. (We could also do `//` comments i suppose.)

JordonPhillips (changes requested):

docs/source-2.0/spec/idl.rst
sugmanue: I will remove this an reuse `NotNL` defined above `(%x09 / %x20-10FFFF)`.

smithy-model/src/test/java/software/amazon/smithy/model/loader/TaggedStringLiteralTest.java
sugmanue: I've added a change to `software/amazon/smithy/model/loader/valid/2.1/tagged-literals.smithy` and `software/amazon/smithy/model/loader/valid/2.1/tagged-literals.json` to show how these are processed.

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
sugmanue: Yeah, your wording reads best. I will change it.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
sugmanue: verbose syntax is cool, but I think it will go in detriment of regular regex, what about adding yet another tag for it? e.g., `#rex`

JordonPhillips: How about `raw` and `re`?

JordonPhillips (commented):
Looks good! I think we could do a split of `raw` and `re`, with the latter having the verbose syntax... but I'm also not so set on it that I'd block things. Having a separate `rex` would also be fine. It can go in another pr if you like, or added here, whichever you prefer

smithy-model/src/main/java/software/amazon/smithy/model/loader/DefaultTokenizer.java
sugmanue: discussed elsewhere, we need to distinguish tags elements.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
sugmanue: I'm not sure to be honest, we do represent timestamps using doubles anyway. I'm not sure why we should do it differently here. Help me understand if I'm missing something else.

smithy-model/src/test/java/software/amazon/smithy/model/loader/TaggedStringLiteralTest.java
sugmanue: I'll move these to `TokenizerTest`.

smithy-syntax/src/test/resources/software/amazon/smithy/syntax/formatter/tagged-string-literals-spacing.formatted.smithy
sugmanue: Indeed, I will add one as well.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
sugmanue: I feel like raw seems too generic for this re-specific use case

sugmanue: Oh, reading more carefully your comment below I think I get your meaning, `#raw` is just unprocessed string whereas `#re` will get the verbose syntax. I'm still not fully set on that idea.

smithy-model/src/main/java/software/amazon/smithy/model/loader/TaggedStringLiteral.java
sugmanue: On a second thought that's fair, one of the main use cases for this would be in protocol tests and we should be as precise as possible. Done.

JordonPhillips (approved):

PR #3167, opened 2026-06-14
Make JSON AST loader streaming

mtdowling:
Removes the old vendored JSON parsing code that required creating the entire model as a Node before loading. Now we stream from JSON and only create Nodes for traits and metadata and things that need Nodes. Loading a model from a Node or unparsed string still work correctly.

#### Background
* What do these changes do? 
* Why are they important?

#### Testing

#### Links

---


smithy-model/src/fuzz/java/software/amazon/smithy/model/loader/JsonAstLoaderFuzzTest.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>Also exercises the already-parsed-node path ({@code addDocumentNode}) so both {@code AstReader}
+ * implementations — {@code JsonAstReader} and {@code NodeAstReader} — are driven by the same loader.
+ */
+class JsonAstLoaderFuzzTest {
+
+    private static final int MAX_INPUT = 64 * 1024;
+
+    @FuzzTest
+    void loadUnparsedJsonModel(byte[] data) {
```
JordonPhillips, line 29: I wonder what the odds are that we accidentally get a valid model

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstReader.java
```diff
@@ -0,0 +1,781 @@
...
+    private long[] dedupKeys;
+    private String[] dedupVals;
+
+    private JsonAstReader(String filename, Reader reader, boolean allowComments, int bufferSize) {
+        if (reader == null) {
+            throw new NullPointerException("reader is null");
+        }
+        this.filename = filename;
+        this.reader = reader;
+        this.allowComments = allowComments;
```
JordonPhillips, line 107: Is there a need to disallow comments?

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstLoader.java
```diff
@@ -0,0 +1,780 @@
...
+            ShapeId id = expectShapeId(idText, reader.lastKeyLocation());
+            SourceLocation shapeLocation = reader.currentLocation();
+            if (reader.currentType() != AstReader.Type.OBJECT) {
+                throw new SourceException("Expected object, but found "
+                        + AstReader.describe(reader.currentType()) + ".", shapeLocation);
+            }
+            int shapeBodyDepth = reader.depth() + 1;
+            reader.startObject();
+            // The AST serializes `type` first.
+            String firstKey = reader.nextKey();
```
JordonPhillips: I bet this will break people. Probably doesn't matter though

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstLoader.java
```diff
@@ -0,0 +1,780 @@
...
+            String type = reader.expectStringValue("`type` member");
+            try {
+                loadShape(id, type, shapeLocation);
+            } catch (SourceException e) {
+                emit(ValidationEvent.fromSourceException(e).toBuilder().shapeId(id).build());
+                // Unwind to this shape's body and drain it so the cursor is ready for the next shape.
+                reader.skipToDepth(shapeBodyDepth);
+                while (reader.nextKey() != null) {
+                    reader.skipValue();
+                }
```
JordonPhillips: Can you not just use skipToDepth here to skip all the way down to the end of the shape def without the loop?

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstLoader.java
```diff
@@ -0,0 +1,780 @@
...
+    }
+
+    private boolean parse() {
+        if (reader.currentType() != AstReader.Type.OBJECT) {
+            return false;
+        }
+        SourceLocation modelLocation = reader.currentLocation();
+        reader.startObject();
+
+        List<String> topLevelKeys = new ArrayList<>();
```
JordonPhillips: Why the list here when you use `Keys` everywhere else?

JordonPhillips (approved):
Only a few small questions

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstReader.java
mtdowling: We support it now, so I kept it.

smithy-model/src/main/java/software/amazon/smithy/model/loader/JsonAstLoader.java
mtdowling: Yep. Fixed.

PR #3182, opened 2026-06-19
Add OpenAPI option to use operation input and output schema names

### LINKED ISSUE #2469: Retain operation input/output Smithy symbol (or allow customization) upon conversion to OpenAPI
This relates to [a previous issue](https://github.com/smithy-lang/smithy/issues/2443) where we asked about the possibility of inlining map shapes. This has been quickly addressed, which we are very thankful for.

Is this currently supported? For example, Consider the following operation shape:
```smithy
/// Do something given its name
@http(
    method: "POST"
    uri: "/do-something"
)
operation DoSomething {
    input := {
        @required
        nameOfSomething: String
    }
    output := {
        comment: String
    }
}
```

More or less, this would currently get converted to the following JSON Schema:
```json
{
            "post": {
                "description": "Do something given its name",
                "operationId": "DoSomething",
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "$ref": "#/components/schemas/DoSomethingRequestContent"
                            }
                        }
                    },
                    "required": true
                },
                "responses": {
                    "200": {
                        "description": "DoSomething 200 response",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/DoSomethingResponseContent"
                                }
                            }
                        }
                    },
                    // ...
```

This is problematic because our build process cares about the schema name of an operation's input and output contents (for code generation purposes).

Is there a way to change these schema names to follow the original Smithy input and output structure names?

arnabnandy7:
#### Background
* Adds an opt-in OpenAPI setting, `useOperationInputOutputShapeNames`, to name synthesized operation request and response content schemas after the Smithy input/output structures when possible.
* This lets generated OpenAPI use component names like `DoSomethingInput` and `DoSomethingOutput` instead of `DoSomethingRequestContent` and `DoSomethingResponseContent`.
* The existing naming behavior remains the default. If an operation reuses the same structure for both input and output, the converter falls back to the existing suffix-based names to avoid component collisions.
* This is important for OpenAPI consumers and code generators that depend on stable schema names matching the original Smithy operation structures.

#### Testing
* Ran `./gradlew :smithy-openapi:spotlessApply --no-daemon --max-workers=1`
* Ran `./gradlew :smithy-openapi:test --tests software.amazon.smithy.openapi.fromsmithy.protocols.AwsRestJson1ProtocolTest --tests software.amazon.smithy.openapi.fromsmithy.OpenApiConfigTest -PnoFormat --no-daemon --max-workers=1`

#### Links
* Closes #2469

---

### DIFF
--- a/.changes/next-release/feature-53b24a08a91d33c36656fcc88da53e6994761c41.json
+++ b/.changes/next-release/feature-53b24a08a91d33c36656fcc88da53e6994761c41.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "description": "Added an OpenAPI setting to name synthesized operation input and output content schemas after their Smithy structure names.",
+  "pull_requests": []
+}
--- a/.changes/next-release/feature-53b24a08a91d33c36656fcc88da53e6994761c41.json
+++ b/.changes/next-release/feature-53b24a08a91d33c36656fcc88da53e6994761c41.json
@@ -1,5 +1,7 @@
 {
   "type": "feature",
   "description": "Added an OpenAPI setting to name synthesized operation input and output content schemas after their Smithy structure names.",
-  "pull_requests": []
+  "pull_requests": [
+    "[#3182](https://github.com/smithy-lang/smithy/pull/3182)"
+  ]
 }
--- a/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
+++ b/smithy-openapi/src/main/java/software/amazon/smithy/openapi/OpenApiConfig.java
@@ -99,6 +99,7 @@ public class OpenApiConfig extends JsonSchemaConfig {
     private ErrorStatusConflictHandlingStrategy onErrorStatusConflict;
     private OpenApiVersion version = OpenApiVersion.VERSION_3_0_2;
     private boolean useStringsForArbitraryPrecision = false;
+    private boolean useOperationInputOutputShapeNames = false;
 
     public OpenApiConfig() {
         super();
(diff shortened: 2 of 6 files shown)


JordonPhillips:
Can I ask what the need for this is?

One reason why we don't do this is that these OpenAPI structures actually don't match the IO shapes, just the body content. Headers, uri params, and query params are all separate. So if you're trying to map the Smithy structures to the OpenAPI structures, a simple name mapping isn't going to work regardless.

arnabnandy7:
> Can I ask what the need for this is?

Thanks for the feedback, @JordonPhillips.

The motivation comes from the original issue #2469 downstream OpenAPI consumers and code generators that use the schema component names as stable identifiers. When the synthesised names are `DoSomethingRequestContent` / `DoSomethingResponseContent`, those names carry no direct relationship to the Smithy model's `DoSomethingInput` / `DoSomethingOutput` shapes, making traceability harder.

You're right that the OpenAPI body-content schema is not a 1:1 mirror of the Smithy IO shape; headers, URI params, and query params are stripped out. This PR doesn't claim otherwise; the option is purely about naming the synthesised body-content schema after the original IO shape (when there's no collision), not about structural equivalence. The intent is that downstream consumers get a predictable, model-traceable name rather than a generated suffix.

That said, if there's a concern that this naming could mislead consumers into assuming structural equivalence, I'm happy to add explicit documentation/javadoc on the option, warning about this distinction.

Is there an approach you'd recommend instead to surface stable, model-traceable names for the synthesised body schemas? Happy to revise the implementation based on your guidance.

PR #3184, opened 2026-06-21
Smithy model binary format design and implementation

sugmanue:
#### Background
* Adds Smithy Model Format (SMF), a compact binary serialization format for Smithy models (developer preview).
* SMF eliminates JSON parsing overhead by interning strings into a symbol table, using positional encoding for known structure, and supporting selective loading of model subsets.
* Benchmarks show 3-6x faster full-model loading and up to 23x faster selective loading (1 operation from EC2) vs JSON AST.

#### Testing
* Unit tests for all components (LEB128, writer, reader, round-trip, selective loading).
* Integration test round-tripping all 424 AWS models from the BOM (`./gradlew :smithy-model:integ -PawsModelsTests`).
* Fuzz testing via Jazzer (`./gradlew :smithy-model:fuzz`).
* JMH benchmarks comparing JSON, SMF assembler, SMF direct, and SMF selective paths.

#### Links
* Design doc: `designs/compact-binary-model-format.md`
* Spec: `docs/source-2.0/spec/smf.rst`

---


designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+   (7.7MB JSON, 4,904 shapes), this is substantial work that produces no
+   semantic value.
+
+2. **Repeated string processing**: The string `"smithy.api#documentation"`
+   appears 11,541 times in the EC2 model. Each occurrence must be read,
+   allocated, and compared against known keys. Structural keys like `"traits"`,
+   `"target"`, and `"members"` appear tens of thousands of times.
+
+3. **Intermediate representation**: JSON loading first builds a `Node` tree,
+   then walks it to construct shapes. This doubles the allocation cost.
```
JordonPhillips, line 30: iirc we actually don't do that anymore

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+4. Maintain full fidelity with a fully resolved model, any assembled model
+   is representable in SMF, and loading an SMF file produces an equivalent
+   model without requiring trait merging or mixin application.
+5. Keep the format simple enough that a complete reader is implementable in
+   under 1,000 lines of code in any language.
+
+## Non-goals
+
+1. Human readability. A companion tool can convert SMF to/from JSON AST.
+2. Minimal wire size. The format optimizes for loading speed, not compression.
```
JordonPhillips, line 68: Nit: we're not sending this over the wire, so this is just minimal size. 

Also: if compression isn't a concern then why the VarInt / VarUInt instead of fixed-width ints that are easier to parse?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+
+## Proposal
+
+### Overview
+
+SMF is a binary encoding of a **fully resolved** Smithy model that:
+
+- Represents the model after all apply statements have been merged and all
+  mixin members have been flattened into their target shapes. Apply statements
+  are not preserved. Mixin relationships are preserved as metadata on each
```
JordonPhillips, line 83: If apply statements aren't preserved then I don't think this can be consumed as a dependency. Apply statements exist in the JSON format because Package A can write an apply statement targeting a shape in Package B.

Why not just include dangling apply statements?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+  for each shape type, so keys like `"type"`, `"traits"`, `"members"`, and
+  `"target"` are never encoded.
+- Encodes shape types as a single-byte enum instead of a string.
+- Preserves member ordering (members are written and read sequentially).
+- Length-prefixes each shape to enable skip-scanning.
+- Includes a shape index for random-access selective loading.
+- Uses a minimal self-describing encoding for dynamic trait values.
+
+Because the model is fully resolved, the reader does not need to perform
+trait merging, mixin application, or forward reference resolution. It can
```
JordonPhillips, line 99: So are we duplicating things? I'm not sure how you get rid of forward references otherwise.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+- Text ↔ binary transcoding is built in, useful for debugging.
+- Ion 1.1 macros could encode shape definitions as compact template
+  invocations, approaching the density of a custom format.
+
+**Cons:**
+
+- Generality tax. Ion supports timestamps, decimals, s-expressions,
+  annotations, clobs, blobs, and multiple integer encodings. Smithy needs 6
+  value types. The reader must handle (or skip) features that will never
+  appear.
```
JordonPhillips, line 170: This doesn't seem like a real issue. You'd be using a library to parse it, that's one of the points of a standard tool like this.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+- Ion structs are specified as unordered. Relying on field order for member
+  semantics is fragile and implementation-dependent.
```
JordonPhillips, line 172: This is also true for JSON, I don't think it's an actual issue.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+  construct shapes, the same two-pass pattern as JSON (parse to tree, then
+  interpret tree).
+- ion-java is a substantial dependency (~50K LOC). A custom reader for the
+  Smithy-specific subset is ~500-800 LOC.
+- Ion 1.1 is still maturing. The macro system adds significant complexity to
+  both the format and the implementation.
+
+Ion with a shared symbol table would yield roughly a 3-4x improvement over
+JSON. The custom format matches or exceeds this while being simpler, smaller,
+and tailored to the exact access patterns of model loading.
```
JordonPhillips, line 183: Your claims for the performance improvement from the custom format vary in this doc, which doesn't give me confidence. Here you say it beats a 3-4x speedup, earlier you claim it's a 2-4x speedup, later you say 1.7x-3.6x.

Did you actually implement prototypes in cbor or ion? If so, give actual numbers rather than vague claims.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+uses positional encoding for the known structure, and employs a minimal
+self-describing encoding only for the dynamic parts (trait values, metadata).
+
+**Pros:**
+
+- Significant loading performance improvement (1.7x-3.6x across all model sizes). The reader is a tight loop: read symbol ID →
+  switch on shape type byte → read fields positionally. No generic dispatch,
+  no string comparisons, no intermediate tree.
+- Minimal dependency. The reader is self-contained (~500-800 LOC) with no
+  external library requirements.
```
JordonPhillips, line 197: This isn't really true. Every implementation is going to need to re-implement this, so in some senses it's actually a con. Unlike the IDL, this format is intended to be consumed by dynamic clients, which will then all need to fully implement this instead of handing off most of the work to a library that already exists.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+- Member ordering is an explicit guarantee of the format, not an
+  implementation detail of a map type.
+- Tailored optimizations: annotation traits in 1 byte, pre-parsed ShapeIds,
+  strippable documentation, selective loading via shape index.
+- Simpler than Ion for this use case, only the 6 JSON value types, no
+  timestamps/decimals/annotations/s-expressions.
```
JordonPhillips, line 203: No timestamps or decimals? Explain.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+The choice comes down to: how much of the loading hot path can you optimize?
+
+| Format | Repeated strings | Structural keys | Intermediate tree | ShapeId re-parsing | Selective loading |
+|--------|-----------------|-----------------|-------------------|-------------------|-------------------|
+| CBOR | ✗ repeated (Packed CBOR draft exists but immature) | ✗ still encoded | ✗ still built | ✗ still re-parsed | ✗ not supported |
+| Ion + shared symtab | ✓ interned | ✗ still encoded (as SIDs) | ✗ still built | ✓ once per symbol | ✗ not natively |
+| Custom (SMF) | ✓ interned | ✓ eliminated | ✓ partially (structural only) | ✓ once per symbol | ✓ built-in |
+
+The custom format is the only option that addresses all five sources of
+overhead. The implementation cost is modest (a reader is under 1,000 LOC)
```
JordonPhillips, line 225: > all five sources of overhead

Citation needed. You mention these as some examples of overhead earlier, but this statement says that they are the only sources. I don't doubt that these are big, low-hanging fruit but I do doubt that they're the only things.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+and the format is simple enough that other languages can implement it from
+this specification without reverse-engineering a complex general-purpose
+library.
+
+Ion remains a reasonable pragmatic choice if implementation bandwidth is
+constrained, it gets 60-70% of the benefit with zero format design work.
```
JordonPhillips, line 231: Citation needed.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+### Header
+
+The header is exactly 8 bytes:
+
+```
+Offset  Size  Field
+0       4     Magic number: 0x53 0x4D 0x46 0x00 (ASCII "SMF\0")
+4       1     Format version (MUST be 0x01 for this specification)
+5       1     Smithy version major (e.g., 2 for Smithy 2.x)
+6       1     Smithy version minor (e.g., 0 for Smithy 2.0, 1 for Smithy 2.1)
```
JordonPhillips, line 282: This separates format version and "smithy version" in a way that neither the idl nor the ast do.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+The symbol table maps integer IDs to strings. It consists of a reference to
+an optional shared (pre-defined) symbol table followed by locally-defined
+symbols.
+
+```
+VarUInt   sharedTableId
+            0 = no shared table
+            1 = smithy-core shared table (see below)
+            Other values reserved for future shared tables.
+VarUInt   sharedTableVersion (present only if sharedTableId != 0)
```
JordonPhillips, line 309: This format has no version for this table, so where is that version defined?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+  entries in the shared symbol table.
+- IDs S+1 through S+L (where L is localSymbolCount) correspond to the local
+  symbols in order.
+
+The **smithy-core shared table** (sharedTableId=1) is a well-known table
+defined by this specification. It contains commonly used Smithy strings that
+appear across most models. Its contents are versioned; version N is always a
+prefix of version N+1 (new entries are appended, never reordered or removed).
+
+The initial contents of the smithy-core shared table (version 1) are:
```
JordonPhillips, line 328: Where is this stored, how is it managed, and how are other shared tables stored and managed?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+13  smithy.api#Timestamp
+14  smithy.api#Document
+15  smithy.api#PrimitiveBoolean
+16  smithy.api#PrimitiveByte
+17  smithy.api#PrimitiveShort
+18  smithy.api#PrimitiveInteger
+19  smithy.api#PrimitiveLong
+20  smithy.api#PrimitiveFloat
+21  smithy.api#PrimitiveDouble
+```
```
JordonPhillips, line 353: Why are traits not in here?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+- All multi-byte integers are little-endian unless stated otherwise.
+- `VarUInt` is an unsigned 32-bit integer encoded as LEB128 (7 payload bits
+  per byte, most significant bit is the continuation flag: 1 = more bytes
+  follow, 0 = final byte). The maximum encoded value is 2^32 - 1 (at most
+  5 bytes). Encodings longer than 5 bytes are invalid. This type is used for
+  all structural fields: counts, lengths, and symbol references.
+- `VarInt` is a signed 64-bit integer encoded as zigzag + LEB128:
+  encode as `(n << 1) ^ (n >> 63)`, then write as unsigned LEB128. The range
+  is -2^63 to 2^63 - 1 (at most 10 bytes). This type is used exclusively
+  within DynamicValue encodings for integer trait values.
```
JordonPhillips, line 250: What's `DynamicValue`

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+an optional shared (pre-defined) symbol table followed by locally-defined
+symbols.
+
+```
+VarUInt   sharedTableId
+            0 = no shared table
+            1 = smithy-core shared table (see below)
+            Other values reserved for future shared tables.
+VarUInt   sharedTableVersion (present only if sharedTableId != 0)
+VarUInt   localSymbolCount
```
JordonPhillips, line 310: It'd be nice to drop a length in here so this could be skipped. If you're doing a selective parse then you probably only want to read the symbols you care about. Same goes for some of these other symbol tables.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+In the shapes section, each trait is encoded as:
+
+```
+SymRef    traitId
+VarUInt   traitValueRef (index into the trait value table)
+```
+
+This replaces inline `DynamicValue` encoding with a compact integer reference.
+Annotation traits (`{}`) that appear thousands of times are stored once and
+referenced by a 1-2 byte VarUInt instead of being encoded inline each time.
```
JordonPhillips, line 385: I'm not saying this doesn't have value, but the example of replacing 2 bytes with 2 bytes plus a lookup isn't very motivating.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+
+### Shape Index
+
+The shape index maps shape IDs to byte offsets within the shapes section and
+includes a neighbor list (dependency graph) for each shape. This enables
+computing the transitive closure of any shape without parsing shape bodies.
```
JordonPhillips, line 391: Why is computing closures something we care about in this context?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+1. Read the symbol table and shape index.
+2. Build a lookup from shape ID to index entry.
+3. Starting from the desired root shapes (e.g., a service + operations),
+   compute the transitive closure by following neighbor lists, this is pure
+   integer set operations with no shape parsing.
+4. Collect the byte offsets of all shapes in the closure.
+5. Sort offsets to enable a single sequential forward pass through the shapes
+   section.
+6. Scan the shapes section, parsing only shapes whose offsets are in the
+   sorted set, skipping all others via `byteLength`.
```
JordonPhillips, line 442: If the point of this is to only load one operation, this isn't going to do that. It's going to read every operation on the service because every operation will be connected to the service shape.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+├────────────────────────────────┤
+│ Symbol Table                   │
+├────────────────────────────────┤
+│ Trait Value Table              │
+├────────────────────────────────┤
+│ Shape Index (optional)         │
+├────────────────────────────────┤
+│ Metadata Section               │
+├────────────────────────────────┤
+│ Shapes Section                 │
```
JordonPhillips, line 269: Nit: The shape index seems oddly placed here. shouldn't it come right before the shapes section?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+DynamicValue  value
+```
+
+#### Type-Specific Payloads
+
+**Simple shapes** (0x00 through 0x0C):
+
+```
+VarUInt   mixinCount
+SymRef[]  mixinIds (mixinCount entries; typically 0)
```
JordonPhillips, line 531: since every shape can have mixins, shouldn't this be above with the standard trait info block?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+```
+
+**Named-member shapes** (enum 0x0D, intEnum 0x0E, structure 0x11, union 0x12):
```
JordonPhillips, line 534: I know compacting isn't a goal, but nevertheless the member target could be removed for enums/intenums

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+VarUInt   valueTraitCount
+Trait[]   valueTraits
+VarUInt   mixinCount
+SymRef[]  mixinIds
+```
+
+**Operation shape** (0x13):
+
+```
+byte      flags:
```
JordonPhillips, line 585: These flags only seems to be relevant for compacting. IIRC unset input / output is equivalent to targeting the unit type

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+Dynamic values represent arbitrary JSON-compatible data used for trait values
+and metadata values. Each value is prefixed with a single-byte type tag.
+
+| Tag  | Type        | Payload |
+|------|-------------|---------|
+| 0x00 | null        | (none) |
+| 0x01 | false       | (none) |
+| 0x02 | true        | (none) |
+| 0x03 | integer     | VarInt value |
+| 0x04 | double      | 8 bytes, IEEE 754 double, little-endian |
```
JordonPhillips, line 670: If we aren't caring about file size, why not get rid of doubles and just represent them as big decimal? I'd prefer to keep as much precision as possible. We read from the IDL into big decimal anyway. And realistically we won't have many floating point numbers in the model anyway so file size impact from this would be minimal anyway.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+- Tag 0x0A (big decimal) is for decimal values that require arbitrary
+  precision. The value is encoded as a length-prefixed UTF-8 string
+  containing the decimal number (e.g., "273.15").
+  Both tags use string representation for simplicity since they are extremely
+  rare in practice.
+- Object keys (tag 0x07) are always inline strings, not symbol references.
+  This keeps the dynamic value encoding self-contained and avoids coupling
+  trait value internals to the symbol table.
+
+### Selective Loading
```
JordonPhillips, line 701: This seems like conflicting information with the previously stated single-pass selective loading.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+
+Values 0x16-0xFF are reserved for future shape types.
+
+#### Trait Encoding
+
+Each trait:
+
+```
+SymRef        traitId (shape ID of the trait definition)
+DynamicValue  value
```
JordonPhillips, line 522: How does this reference a value from the duplicates table?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+2. Read the shape index.
+3. Determine the initial set of needed shape IDs (e.g., the service shape
+   and desired operation shapes).
+4. Compute the transitive closure: for each needed shape, read it from the
+   shapes section (using the index to locate it), extract all SymRef values
+   that reference other shapes, and add those to the needed set. Repeat until
+   no new shapes are discovered.
+5. The resulting set of shapes constitutes the minimal model subset.
+
+For streaming I/O (where seeking is not possible), the reader can perform
```
JordonPhillips, line 718: Why would you ever be reading a model from a non-seekable stream? It's always going to be a file on disk, outside of perhaps test cases.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+   integer set operations with no shape parsing.
+4. Collect the byte offsets of all shapes in the closure.
+5. Sort offsets to enable a single sequential forward pass through the shapes
+   section.
+6. Scan the shapes section, parsing only shapes whose offsets are in the
+   sorted set, skipping all others via `byteLength`.
+
+This approach ensures that closure computation is O(closure_size) with no
+I/O beyond the index, and shape reading is a single sequential pass with
+no seeking.
```
JordonPhillips, line 446: I think you could achieve the same single-pass parsing without storing neighbors by simply writing the file with shapes sorted in reverse-topological order

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+│ Shapes Section                 │  ← CRC-32C computed over these bytes
+├────────────────────────────────┤
+│ CRC-32C (4 bytes, LE)          │
+└────────────────────────────────┘
+```
+
+A conforming writer MUST compute and append the CRC-32C after writing all
+sections. A conforming reader MUST verify the CRC-32C before processing any
+data and MUST reject the file with a clear error if the checksum does not
+match.
```
JordonPhillips, line 839: Validating this checksum before processing data eliminates the "no seeking" goal stated in one of the single-pass parsing sections.

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+
+CRC-32C is chosen over CRC-32 because it has hardware acceleration on modern
+x86 (SSE 4.2) and ARM (CRC extension) processors, making verification
+effectively free on the critical loading path.
+
+## Open Questions
+
+1. **Source locations**: The format does not encode source location information
```
JordonPhillips, line 847: We could include source locations. It wouldn't be terribly useful to look at though. We'd need to update the error renderer to show the bytes alongside parsed values in that case. Doable

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+   (file, line, column). Validation errors on SMF-loaded models will not have
+   useful locations. Is this acceptable for the target use cases (dynamic
+   client, codegen), or should an optional source map section be defined?
+
+2. **Integrity checking**: ~~The format has no checksum or CRC. A corrupted file
```
JordonPhillips, line 852: It does though?

designs/compact-binary-model-format.md
```diff
@@ -0,0 +1,965 @@
...
+Based on the EC2 model (4,904 shapes, 7.7MB pretty-printed JSON):
+
+| Representation | Size |
+|---|---|
+| Pretty-printed JSON AST | 7,683 KB |
+| Minified JSON AST | 4,905 KB |
+| SMF (full, with docs) | ~2,940 KB |
+| SMF (stripped, no docs) | ~755 KB |
+| Minified JSON + gzip | 599 KB |
+| SMF (stripped) + gzip | ~350 KB |
```
JordonPhillips, line 886: This needs to include json without docs

JordonPhillips (changes requested):

PR #3186, opened 2026-06-24
Use AOT cache for CLI startup instead of AppCDS

mtdowling:
#### Background
* What do these changes do?

Replace the AppCDS shared archive (smithy.jsa) with an AOT cache, which caches classes loaded *and* linked.

- Bundle a JDK 25 JRE in the runtime image (imageJreVersion 17 -> 25) and require JDK 25+ to build the CLI module (the rest of the monorepo still builds on 17+).
- Launcher loads the cache via -XX:AOTCache; -Xlog:aot=off makes a missing or incompatible cache degrade silently instead of printing [error][aot] noise. (just like how CDS worked).
- WarmupCommand keeps its existing two-phase orchestration and is still run at install time (and by the build's optimize task); only the JVM flags change, from the AppCDS classlist/-Xshare:dump steps to AOTMode=record then AOTMode=create. Generating at install time is required because an AOT cache is platform-specific, like the AppCDS archive it replaces.

The install scripts and the optimize task are unchanged in shape; they still invoke 'smithy warmup'.

* Why are they important?

Measured ~42% faster cold start (~0.24s -> ~0.14s for 'validate --help').

#### Testing

All tests pass. Built and used CLI locally.

You will want to ensure the CI the builds the Smithy CLI has JDK 25.

#### Links

---


JordonPhillips (approved):

JordonPhillips:
Turns out our build automation DOES have 25 bug doesn't actually use it. Sad. Gonna have to revert for now to unblock the release, then we'll merge this back in.

PR #3196, opened 2026-06-30
Add a union trait getter with covariant return

### LINKED ISSUE #3189: Union cases expose value with erased type
Currently when generating classes for traits using smithy-trait-codegen, union shapes
```smithy
@trait
union Foo {
  string: String
  int: Int
}
```
 are generated this way:
```java
@SmithyGenerated
public abstract class FooTrait extends AbstractTrait {
    // ...
    public abstract <T> T getValue();

    @SmithyGenerated
    public static final class StringMember extends FooTrait {
        private final String value;
        // ...

        @Override
        @SuppressWarnings("unchecked")
        public <T> T getValue() {
            return (T) value;
        }
    }

    @SmithyGenerated
    public static final class IntegerMember extends FooTrait {
        private final Integer value;
        // ...

        @Override
        @SuppressWarnings("unchecked")
        public <T> T getValue() {
            return (T) value;
        }
    }
    // ...
}
```
As we can see, I cannot access the original, well-typed `value` of a union member, because the only getter, `getValue`, erases its type, and even if I perform type matching (java `instanceof` / kotlin`when`) on the object type, I still cannot access the value without specifying its  type on the caller side, hence if Smithy definition is changed - I won't get compilation error, I'll get `ClassCastException` in runtime.

There should be another, case-specific getter (that's not an override) method OR base class (`FooTrait` in this case) have to be generic over value type, but I guess it'd break a lot of existing code and would be harder to deal with

JordonPhillips:
This adds a new getter for union values in trait codegen that has a covariant return type. The old getter had the unforutnate effect of type erasure. The old getter will still be generated, but it's deprecated and can be omitted via a plugin setting.

Fixes #3189

---

### DIFF
--- a/.changes/next-release/bugfix-922e5164c811cb68317d8291ed84e40981b144ff.json
+++ b/.changes/next-release/bugfix-922e5164c811cb68317d8291ed84e40981b144ff.json
@@ -1,0 +1,7 @@
+{
+  "type": "bugfix",
+  "description": "Added a new getter for generated union trait values with a covariant return type to address the issue of the old getters erasing value types. The old getters are still around by default, but are deprecated. They may be omitted entirely via a plugin setting.",
+  "pull_requests": [
+      "[#3196](https://github.com/smithy-lang/smithy/pull/3196)"
+  ]
+}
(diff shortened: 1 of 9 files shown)


sugmanue:
I'd be great to get an example of the new generated code, that would help the reviewer to have a better sense of what's changed.

sugmanue:
I feel like using `value()` without the bean-style `get` breaks the consistency with other classes that we codegen. A different option would be to use a switch to allow users to generate `getValue()` as `value()` is defined, or use a different name, e.g., `getVariantValue()`. I don't feel super strong about it, so approving for you to decide.

sugmanue (approved):

PR #3208, opened 2026-07-09
Add smithy ai command and publishable smithy-ai skills module

yasmewad:
## Demo

End-to-end walkthrough of the new `smithy ai` command built from this branch — discover bundled skills, preview with `--dry-run`, install into the Claude harness, and confirm the install is idempotent.

---

## What & why

Teach any AI agent authoritative Smithy knowledge without a lossy web-fetch, and make that knowledge a standalone artifact instead of being trapped in the CLI.

- **`smithy ai install`** writes a thin `smithy-docs-navigator` skill that points agents at the published `llms.txt` index and each page's raw `.rst` source (read with a non-summarizing tool). References URLs, so it never goes stale.
- **New `smithy-ai` module** owns the skill content + a discovery API (`AiContent`, `AiSkill`). Skills live as browsable `skills/<name>/` dirs (a `SKILL.md` plus any supporting files), staged under `META-INF/smithy-ai/` and discovered at runtime via `ClassLoader.getResources()` (the `ModelDiscovery` idiom). Ships as its own JAR in the CLI runtime image. The CLI drops its hardcoded skill registry.
- **`generate_llms_txt.py`** now emits raw `.rst` links instead of `.html`. It also includes some other links to our smithy ecosystem repositories similar to how strands does it.

## CLI surface

```
smithy ai install [--harness <name>] [--skill <name|path>] [--dir <path>] [--force] [--dry-run]
smithy ai list    [--dir <path>]
```

`--harness` targets one or (omitted) every harness detected under the root; `claude`/`kiro` supported, `codex`/`copilot`/`gemini` rejected with a reason. `--skill` takes a bundled name or an external `SKILL.md` path. Installs a skill's whole directory.

## Testing

- `AiContentTest` (module discovery) + `AiInstallCommandTest`/`AiListCommandTest` (CLI); full `smithy-cli` suite green.
- Runtime image built; `smithy-ai-*.jar` in every platform `lib/`; real binary exercised end-to-end (list, dry-run, install, idempotent re-run, auto-detect, external multi-file skill, `--force`, unsupported-harness error).

**Value of having a skill v/s not**: I used the mechanism that aws-sdk-skills use using LLM as a judge against a hand written model , smithy-java  server using its own smithy-build.json and ts client with its own smithy-build.json with appropriate plugin application. local A/B (skill vs bare agent, same modeling prompt of building a service and having a code generated java server and typescript client, the output is expected to use correct plugins and setup directories appropriately).

## Follow ups

Wire up the benchmarking suite in a separate PR so all future skills can be validated.

| Metric | with-skill | without-skill | Delta |
|---|---|---|---|
| Model validates | PASS | PASS | there was fidelity difference in terms of output details, and java v/s typescript client setup with and without skill) |
| Credits (token/cost proxy) | 15.38 | 28.21 | ~-45% |
| Wall-clock | 8m 22s | 12m 44s | ~-34% |
| Tool turns | 97 | 146 | ~-34% |
| Authoritative doc fetches | 6 | 0 | n/a |

Both validate (basic modeling is within the base model's competence); the skill's win is **cost and fidelity**: it read the real `.rst`/quickstart docs and wrote correct build wiring early, while the baseline trial-and-errored from memory. Directional (n=1); the benchmark harness is not in this PR and will land separately.

## Demo

smithy-ai-demo


yasmewad:
Closing in favor of: https://github.com/smithy-lang/smithy/pull/3209 so that the changelog workflow works.

PR #3209, opened 2026-07-10
Add reviewable Smithy agent skills

yasmewad:
## What & why

Teach any AI agent authoritative Smithy knowledge without a lossy web-fetch. This ships three things: a design doc, plain reviewable skill files, and a versioned Maven artifact for consumers who want a dependency instead of a copy.

- **Design doc** at `designs/agent-skills.md` records the motivation, the trust boundary (provenance and packaging integrity vs. content review), the goals per stakeholder, the alternatives considered, and future work.
- **Skill files** live under `smithy-agent-skills/src/main/resources/skills/`, organized into component subdirectories that mirror the smithy.io docs layout (`core/` for the IDL and spec, and one subtree per component later). Each skill is an [Agent Skills](https://agentskills.io) folder: a `SKILL.md` plus optional `references/`/`assets/`/`scripts/`. The first skill, `core/smithy-docs-navigator`, points agents at the published `llms.txt` index and each page's raw `.rst` source, read with a non-summarizing tool.
- **Naming convention.** A skill's identity is its `SKILL.md` `name` (which must equal its leaf folder), so the name is a flat global identifier. Every skill name carries a `smithy-` prefix, and a component skill carries its component too (`smithy-ts-<capability>`), so names stay unique when composed downstream. The subdirectory is organizational and for `CODEOWNERS` scoping only.
- **Install by pointing your agent at the repo.** `skills/README.md` tells users to have their agent read the `SKILL.md` and copy it into their own skills directory. Claude Code, Codex, and Kiro all consume the same agentskills.io format, and the per-agent destination is documented.
- **`generate_llms_txt.py`** now emits raw `.rst` links instead of `.html`, plus links to other Smithy ecosystem repositories.

## Publishing `skills/` as a versioned artifact

The same skill files are also published as `software.amazon.smithy:smithy-agent-skills`, versioned with each Smithy release, so consumers can depend on the reviewed skills through the same Maven pipeline as every other Smithy artifact. The artifact is packaged and unpacked at build time; nothing loads skill content at agent runtime.

- **`smithy-agent-skills` Gradle module** — no Java source, no dependencies, standard Apache-2.0 POM via `smithy.module-conventions`. The standard Gradle resource pipeline packages the skills under a top-level `skills/` entry path in the JAR. `verifySkillsVerbatim` walks the tree recursively (so the component subdirectories work with no build change) and fails the build if the JAR's `skills/` entries differ byte-for-byte from source.
- **Bundled-file review policy.** `references/` and `assets/` are reviewed like the instructions. `scripts/` get closer scrutiny for what the code can do and stay thin wrappers over the Smithy CLI. One rule is absolute: no file may fetch and execute remote content at runtime.
- **`.github/CODEOWNERS`** — the skills directory requires core-team review.
- **Changelog** — one `feature` entry covering the skills and the artifact.

## Testing

- The `smithy-docs-navigator` skill was validated end-to-end against a hand-written model (benchmark below).
- Build verified: `publishToMavenLocal` publishes cleanly; `unzip -l` shows exactly the `skills/` entries plus standard `META-INF`; `diff -r` of the extracted `skills/` against source is byte-identical; `verifySkillsVerbatim` fails when a `skills/` file changes after the JAR is built; `./gradlew :smithy-agent-skills:verifySkillsVerbatim` is green with the component-subdirectory layout.

**Value of the skill vs. no skill.** Using the same LLM-as-judge setup the aws-sdk-skills use, I ran a local A/B against a hand-written model: a smithy-java server and a TypeScript client, each with its own `smithy-build.json` and plugin application. Same modeling prompt both trials (build the service, generate the Java server and TypeScript client, using correct plugins and directory setup), skill vs. bare agent.

| Metric | with-skill | without-skill | Delta |
|---|---|---|---|
| Model validates | PASS | PASS | fidelity differed in output detail and in Java vs. TypeScript client setup |
| Credits (token/cost proxy) | 15.38 | 28.21 | ~-45% |
| Wall-clock | 8m 22s | 12m 44s | ~-34% |
| Tool turns | 97 | 146 | ~-34% |
| Authoritative doc fetches | 6 | 0 | n/a |

Both validate, since basic modeling is within the base model's competence. The skill's win is cost and fidelity: it read the real `.rst`/quickstart docs and wrote correct build wiring early, while the baseline trial-and-errored from memory. This is one exploratory run (n=1) measuring efficiency, not answer correctness; the benchmark harness is not in this PR and will land separately.


docs/generate_llms_txt.py
JordonPhillips, line 1: This should have types

docs/llms_extra.json
```diff
@@ -0,0 +1,53 @@
+{
+  "_comment": "Curated overlay appended to llms.txt by generate_llms_txt.py. Hand-maintained links the mechanical source-2.0 walk cannot produce: prominent in-spec references (including backward compatibility), example projects, per-language code generators, and tooling. Links prefer raw README/markdown (LLM-readable text) over repo landing pages, and use each repo's real default branch (smithy-python is 'develop', others 'main'). Notes are deliberately coarse and durable (lifecycle role, language, client/server scope, maturity) so they do not drift with each release; point at the repo for current feature detail. All URLs verified to resolve (HTTP 200).",
```
JordonPhillips: > Hand-maintained links the mechanical source-2.0 walk cannot produce

wut

smithy-ai/skills/smithy-docs-navigator/SKILL.md
```diff
@@ -0,0 +1,84 @@
+---
+name: smithy-docs-navigator
+description: Answer any Smithy question from the authoritative source instead of memory or a lossy web-fetch. Use whenever a task involves Smithy - writing, reading, reviewing, or fixing a Smithy model; the IDL syntax or grammar; a trait's exact rules (@length, @pattern, @range, @required, @http, @paginated, etc.); selectors; shapes, resources, operations, or services; HTTP bindings, protocols, or auth; whether a model change is backward compatible; or which code generator to use for a target language. Reads the published llms.txt index and the reStructuredText (.rst) source of each page - the complete, version-matched text a summary would drop. Trigger phrases include "in Smithy", "Smithy model", "smithy.io", "what does the Smithy spec say", "how do I define ... in Smithy", "is this a breaking change", "which Smithy codegen".
```
JordonPhillips: I get that this is yaml frontmatter, but wrapping would still be appreciated

smithy-agent-skills/src/main/resources/skills/core/smithy-docs-navigator/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+properties, and normative rules. This skill sends you to the `.rst` source and
+has you read it with a non-summarizing tool, so every answer is grounded in the
+complete, version-matched text rather than memory or a summary.
+
+## Usage
+
+Use this skill whenever a request touches Smithy - it is the default way to get
+an authoritative answer. Concretely:
+
+- Authoring or editing a model: shape, resource, operation, or service syntax;
```
JordonPhillips, line 34: I feel like this is breaking down and tossing in semicolons where commas should be

smithy-ai/skills/smithy-docs-navigator/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Usage
+
+Use this skill whenever a request touches Smithy - it is the default way to get
+an authoritative answer. Concretely:
+
+- Authoring or editing a model: shape, resource, operation, or service syntax;
+  the IDL grammar; mixins; shape-ID resolution.
+- A trait's exact rules and properties (`@length`, `@pattern`, `@range`,
+  `@required`, `@http`, `@paginated`, protocol and auth traits, ...).
```
JordonPhillips: is this listing of traits necessary

smithy-ai/src/main/java/software/amazon/smithy/ai/AiContent.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Discovers Smithy AI skills bundled on the classpath.
```
JordonPhillips: I'm a bit worried about this from a security angle. We're asking customers to trust this skill, but it'd be pretty easy for somebody to anonymously add some malicious prompt as part of a supply chain attack and we'd have no way to detect that or limit exposure.

smithy-ai/src/main/java/software/amazon/smithy/ai/AiContent.java
```diff
@@ -0,0 +1,144 @@
...
+        if (url == null) {
+            throw new AiContentException("Bundled AI resource not found: " + path);
+        }
+        return url;
+    }
+
+    /** Package-private: defensive copy of a file list, exposed to consumers as unmodifiable. */
+    static List<String> unmodifiable(List<String> src) {
+        return Collections.unmodifiableList(new ArrayList<>(src));
+    }
```
JordonPhillips: This seems to just be an alias to a standard library method.

smithy-ai/build.gradle.kts
```diff
@@ -0,0 +1,78 @@
...
+extra["displayName"] = "Smithy :: AI"
+extra["moduleName"] = "software.amazon.smithy.ai"
+
+dependencies {
+    // IoUtils for reading bundled resources. api because AiSkill surfaces reads to consumers.
+    api(project(":smithy-utils"))
+}
+
+// The browsable content roots. These live at the module root (not under src/main/resources) so the
+// tree is publishable to GitHub as-is and mirrors github.com/aws/agent-toolkit-for-aws. The build
```
JordonPhillips: What does publishable to github mean? And does it matter that it follows that structure? It's not clear that it actually does anything to follow it.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/AiHarness.java
```diff
@@ -0,0 +1,139 @@
...
+                "codex",
+                "codex CLI has no SKILL.md skills directory; it reads AGENTS.md project instructions");
+        UNSUPPORTED_REASONS.put(
+                "copilot",
+                "GitHub Copilot has no SKILL.md skills directory; it reads "
+                        + ".github/copilot-instructions.md, .github/instructions/*.instructions.md, or AGENTS.md");
+        UNSUPPORTED_REASONS.put(
+                "gemini",
+                "Gemini CLI has no SKILL.md skills directory; it reads a GEMINI.md context file");
+    }
```
JordonPhillips: `MapUtils.of`

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/AiInstallCommand.java
```diff
@@ -0,0 +1,247 @@
...
+                    env.stdout()
+                            .println("Would " + (exists ? "overwrite " : "install ") + skill.getName()
+                                    + " (" + files.size() + " file" + (files.size() == 1 ? "" : "s")
+                                    + ") at " + skillMd);
+                }
+            }
+            return;
+        }
+
+        if (exists && !options.force) {
```
JordonPhillips: Is this necessary? If we're doing a straight copy then the cost is basically nothing and I think users will want to be able to upgrade by default if we make some changes, which we certainly will.

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/AiListCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private static final class Options implements ArgumentReceiver {
+        private String dir;
+
+        @Override
+        public Consumer<String> testParameter(String name) {
+            switch (name) {
+                case "--dir":
+                case "-C":
```
JordonPhillips: Why `-C`?

smithy-cli/src/main/java/software/amazon/smithy/cli/commands/AiSkill.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /** Reads the UTF-8 content of one relative file. */
+    String readFile(String relativePath) {
+        if (bundled != null) {
+            return bundled.readFile(relativePath);
+        }
+        try {
+            return new String(Files.readAllBytes(externalRoot.resolve(relativePath)),
+                    java.nio.charset.StandardCharsets.UTF_8);
```
JordonPhillips: We should be using ioutils

JordonPhillips (changes requested):

smithy-agent-skills/src/main/resources/skills/core/smithy-docs-navigator/SKILL.md
mtdowling: semicolons are often used as super-commas when elements of a list themselves have commas.

smithy-ai/src/main/java/software/amazon/smithy/ai/AiContent.java
yasmewad: This is a valid concern, and I looked around and noticed this kind of content is vended by doing 2 things:
1. maintaining a strict CODEOWNERS permission entry for skills.
2. Every skill authored in this repository, must be from the repository owner and must go through PR reviewed with the owning team.

This constraint applies only to this module. That being said, we can pursue the process of creating a separate repository to limit the blast radius as an alternative and remove discovery of knowledge and/or plugin authoring capability to repositories in smithy-lang organization admins. Similar approach seems to be followed by the agentic toolkit for aws repository. In either case, I will have an offline discussion for the same.

JordonPhillips: repo hygiene is entirely unrelated to this issue. The issue is having an open door for anybody to inject anything via an unrestricted classpath search.

If you want to bundle things directly into the cli, that's best done explicitly at build time rather than a runtime classpath search.

docs/generate_llms_txt.py
yasmewad: I have updated the code-review, added types and made this a simple skill.md to get us moving. This should resolve security concerns for users, and they can just point to our README and install the skill.

JordonPhillips (approved):

docs/generate_llms_txt.py
```diff
@@ -25,17 +32,25 @@
     ("languages", "Languages"),
 ]
 
-RST_TITLE_RE = re.compile(r"^([=\-~`#\"^+*:.!'_]{2,})\s*$")
+RST_TITLE_RE: re.Pattern[str] = re.compile(r"^([=\-~`#\"^+*:.!'_]{2,})\s*$")
+MARKDOWN_TITLE_RE: re.Pattern[str] = re.compile(r"^#\s+(.+?)\s*#*\s*$")
+SOURCE_EXTENSIONS: tuple[str, ...] = (".rst", ".md")
```
JordonPhillips:
```suggestion
SOURCE_EXTENSIONS: tuple[str, ...] = (".rst", ".md", ".rst.template", ".md.template")
```

smithy-agent-skills/src/main/resources/skills/core/smithy-docs-navigator/SKILL.md
```diff
@@ -0,0 +1,124 @@
...
+## Core Concepts
+
+- **The source file is authoritative. A summary is not.** The rendered HTML,
+  when passed through a summarizing fetch, drops the exact rules you need. Read
+  the source.
+- **`llms.txt` is the one index.** Published at `https://smithy.io/2.0/llms.txt`,
+  its generated documentation links point at each page's canonical `.rst` or
+  `.md` source, including resolved symlink targets. It also lists curated
+  **Key references** (including backward compatibility), **Examples and
+  resources**, **Code generators by language** (with each generator's language
```
JordonPhillips:
```suggestion
  **Key references** (including backward compatibility), **Examples**,
  **Code generators by language** (with each generator's language
```

This makes it match the actual section names

smithy-agent-skills/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+val verifySkillsVerbatim by tasks.registering {
+    description = "Verifies the packaged skills/ entries are byte-identical to the module's skills/ source."
+    group = "verification"
+    dependsOn(tasks.jar)
+
+    val sourceDir = skillsSourceDir.asFile
+    val jarFile = tasks.jar.flatMap { it.archiveFile }
+
+    inputs.dir(sourceDir)
+    inputs.file(jarFile)
```
JordonPhillips, line 41: Declaring inputs without declaring outputs means you still never get cached. If you want caching you could toss in some signalling output.

smithy-agent-skills/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+
+    doLast {
+        // Map every source file under skills/ to its relative "skills/<path>" entry name and content hash.
```
JordonPhillips: Nit: you're actually loading the whole file here, not just a hash

JordonPhillips (changes requested):

smithy-agent-skills/build.gradle.kts
yasmewad: That's a good point. I will update that.

smithy-agent-skills/build.gradle.kts
yasmewad: Ahh will tell Codex to correct this logic.

smithy-agent-skills/build.gradle.kts
yasmewad: Can you help me double check and see if this is what you mentioned when adding a marker and using outputs? I tried reading through the docs and realized this is one way to do it.

JordonPhillips: Yeah gradle is tricky. The other way would be to reorganize this as a class, but if this works it works.

JordonPhillips (approved):

PR #3214, opened 2026-07-15
Fix trait codegen for default values of members with `@idRef` trait applied

fan-tom:
Closes #3212 

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3218, opened 2026-07-20
Extend @idempotent trait with response behavior properties

thehungryturnip:
Add optional `exists` and `notFound` properties to the `@idempotent` trait, allowing operations to declare the errors returned when encountering unexpected resource states (already exists for creates, not found for deletes). Each property is a flat list of error shape IDs. An empty list indicates the operation returns a successful response; when omitted, the behavior is unspecified.

The two properties are mutually exclusive — only one may be specified on a given operation.

Changes:
- Extend idempotent struct in prelude.smithy with exists and notFound list members
- Add IdempotentErrors private list shape
- Rewrite IdempotentTrait.java from AnnotationTrait to AbstractTrait with Optional<List<ShapeId>> for each property
- Add IdempotentTraitValidator: validates mutual exclusivity and that referenced errors exist on the operation's errors list
- Add model-based test scenarios (errorfiles)
- Update public documentation (behavior-traits.rst)

Fully backwards compatible: bare @idempotent (no args) continues to work unchanged.

# Background

## What do these changes do?

Extends the `@idempotent` trait with two optional, mutually exclusive properties: `exists` and `notFound`. Each is a list of error shape IDs that the operation may return when encountering the respective state. An empty list means the operation returns success. This allows operations to declare their response behavior when encountering unexpected resource states (e.g., a create operation called on an already-existing resource, or a delete operation called on a non-existent resource).

## Why are they important?

Smithy's `@idempotent` trait (per RFC 9110 Section 9.2.2) indicates that repeating a request has the same intended effect — but the RFC explicitly states the response might differ. This creates ambiguity for orchestration systems that need to know whether a repeated call returns success or an error. This extension resolves that ambiguity at the model level, enabling downstream consumers (e.g., CloudFormation resource type generation) to determine whether pre-existence checks are needed without requiring out-of-band documentation. The change is fully backwards compatible — bare `@idempotent` (no arguments) continues to work unchanged.

# Testing

## How did you test these changes?

`./gradlew clean build -Plog-tests` passes with no failures. Added model-based test scenarios covering:
- Bare `@idempotent` (backwards compatibility)
- `exists` with valid error list
- `notFound` with valid error list
- `exists` with empty list (returns success)
- `notFound` with empty list (returns success)
- Multiple errors in the list
- Invalid: both `exists` and `notFound` specified (mutual exclusivity)
- Invalid: error not in operation's errors list (for both properties)

# Links

- RFC 9110 Section 9.2.2 (Idempotent Methods): https://datatracker.ietf.org/doc/html/rfc9110.html#section-9.2.2
- Smithy idempotency docs: https://smithy.io/2.0/spec/behavior-traits.html#idempotency


thehungryturnip:
Note on CI: The Validate staged changelogs check fails because it uses a pull_request_target workflow that [refuses to check out fork PR code](https://gh.io/securely-using-pull_request_target) for security reasons. This is expected for all fork-based PRs (e.g., #3214 has the same failure). The [CONTRIBUTING.md](https://github.com/smithy-lang/smithy/blob/main/CONTRIBUTING.md#contributing-via-pull-requests) instructs contributors to fork the repository, so this check will always fail for external contributions.

All other CI checks (build, test, formatting, docs, AWS API models) are passing. I've included a changelog entry at .changes/next-release/feature-idempotent-response-behavior.json — happy to adjust the format if needed once a maintainer can validate it.

smithy-jsonschema/src/test/resources/software/amazon/smithy/jsonschema/default-values-disabled.jsonschema.v07.json
```diff
@@ -26,6 +26,9 @@
                 "FOO",
                 "BAR"
             ]
+        },
+        "Unit": {
```
sugmanue: Why are those files in a different package got changed?

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
```diff
@@ -4,27 +4,191 @@
...
+import software.amazon.smithy.utils.ToSmithyBuilder;
 
 /**
  * Indicates that an operation is idempotent.
+ *
+ * <p>Optionally describes the response behavior when encountering unexpected
+ * resource states (already exists for creates, not found for deletes).
  */
-public final class IdempotentTrait extends AnnotationTrait {
+public final class IdempotentTrait extends AbstractTrait implements ToSmithyBuilder<IdempotentTrait> {
```
sugmanue, line 23: Changes to traits in the smithy package need to be reflected in the public documentation, [the behavior traits](https://github.com/smithy-lang/smithy/blob/main/docs/source-2.0/spec/behavior-traits.rst?plain=1#L5) page at the very least for this change.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -618,7 +618,33 @@ structure readonly {}
...
-structure idempotent {}
+structure idempotent {
+    /// Describes the response when the operation is called and the resource
+    /// already exists. Applicable to create lifecycle operations.
+    alreadyExistsResponse: IdempotentResponse
+
+    /// Describes the response when the operation is called and the resource
+    /// does not exist. Applicable to delete lifecycle operations.
+    notFoundResponse: IdempotentResponse
+}
```
sugmanue: Let's have a single member to describe what would happen when the call would be consider a no-op, IOW, the side-effects of the operation won't be repeated. I suggest we use `onNoop` for that, such that the trait will be used as:

```smithy
/// Indicates that the operation will return a success response, when the 
/// its expected side effects were already realized.
@idempontent(onNoop: {success: {}})
```

or 

```smithy
/// Indicates that the operation will return a an error response, when expected
/// its expected side effects were already realized.
@idempontent(onNoop: {errors: [NotFound]})
```

sugmanue (changes requested):

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -618,7 +618,33 @@ structure readonly {}
...
+
+/// Describes how an idempotent operation responds when encountering
+/// an unexpected resource state (already exists or not found).
+@private
+union IdempotentResponse {
+    /// The operation returns a successful response.
+    success: Unit
+
+    /// The operation returns one of the specified errors.
+    errors: IdempotentResponseErrors
```
KaramVSingh: Is there a reason we want this to be Errors rather than Error?

smithy-jsonschema/src/test/resources/software/amazon/smithy/jsonschema/default-values-disabled.jsonschema.v07.json
thehungryturnip: It's because we introduced `success: Unit`. The `Unit == object` relationship needs to be defined everywhere.

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
thehungryturnip: I see. So we want to update the docs in the same PR.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
thehungryturnip: Further discussion is leading up to change the shape to:

```
@idempotent(exists: [])
```

and 

```
@idempotent(notFound: [])
```

Where `[]` indicates that no errors will be thrown.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
thehungryturnip: The Smithy team wants to leave room for scenarios where multiple errors would indicate not found/already exists.

e.g. A delete operation could potentially throw both `NotFound` and `InvalidInput` depending on the situation.

.changes/next-release/feature-22efa2bb8e26e3101895d3416035808dd6fbe6ee.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "\"Extended the `@idempotent` trait with optional `exists` and `notFound` properties to declare the errors returned when encountering unexpected resource states.\"",
```
sugmanue:
```suggestion
  "description": "Extended the `@idempotent` trait with optional `exists` and `notFound` properties to declare the errors returned when encountering unexpected resource states.",
```

docs/source-2.0/spec/behavior-traits.rst
```diff
@@ -85,6 +117,30 @@ Conflicts with
         output: DeleteSomethingOutput
     }
 
+The following example uses ``exists`` to indicate that a create operation
+returns a ``ConflictException`` when the resource already exists:
+
+.. code-block:: smithy
+
+    @idempotent(exists: ["smithy.example#ConflictException"])
```
sugmanue:
```suggestion
    @idempotent(exists: [ConflictException])
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotentTraitValidator.java
```diff
@@ -0,0 +1,79 @@
...
+        List<ValidationEvent> events = new ArrayList<>();
+        List<ShapeId> operationErrors = operation.getErrors();
+
+        for (ShapeId errorId : errors) {
+            if (!operationErrors.contains(errorId)) {
+                events.add(error(operation,
+                        trait,
+                        String.format(
+                                "The `%s` property references error `%s`, but it is not in the "
+                                        + "operation's errors list. Expected one of: [%s]",
```
sugmanue:
```suggestion
                                        + "operation's errors list. Expected one of: %s",
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotentTraitValidator.java
```diff
@@ -0,0 +1,79 @@
...
+                                propertyName,
+                                errorId,
+                                operationErrors.stream()
+                                        .map(ShapeId::toString)
+                                        .sorted()
+                                        .collect(Collectors.joining(", ")))));
```
sugmanue: Use instead `ValidationUtils.tickedList`.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotentTraitValidator.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private List<ValidationEvent> validateErrors(
+            OperationShape operation,
+            IdempotentTrait trait,
+            List<ShapeId> errors,
+            String propertyName
+    ) {
+        List<ValidationEvent> events = new ArrayList<>();
+        List<ShapeId> operationErrors = operation.getErrors();
```
sugmanue: The service can also declare common errors, you need to use instead `OperationIndex#getErrors(ToShapeId service, ToShapeId operation)`. We could, for now, just disallow using common errors declared by the service, I think that makes sense but we need to clarify that in the documentation. Otherwise the logic gets complicated, two different services can use the same operation, and we don't have an easy way to map back from operation to services.

Also, since you're using this to lookup a `Set` would be better.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/IdempotentTraitValidator.java
```diff
@@ -0,0 +1,79 @@
...
+        return events;
+    }
+
+    private List<ValidationEvent> validateErrors(
+            OperationShape operation,
+            IdempotentTrait trait,
+            List<ShapeId> errors,
+            String propertyName
+    ) {
+        List<ValidationEvent> events = new ArrayList<>();
```
sugmanue: Let's pass in the events to avoid creating a new list here.

docs/source-2.0/spec/behavior-traits.rst
```diff
@@ -73,10 +73,42 @@ Summary
...
+        a shape with the :ref:`error-trait` that is also in the operation's
+        ``errors`` list. An empty list indicates the operation returns a
+        successful response. When omitted, the behavior is unspecified.
+
+The ``exists`` and ``notFound`` properties are mutually exclusive; only one
+may be specified on a given operation.
+
+When applied without any properties, the trait indicates idempotency without
+specifying the response behavior for unexpected resource states. This is fully
+backwards compatible with existing uses of the trait:
```
sugmanue: Let's remove the "This is fully backwards compatible with existing uses of the trait:", bit. That can go in the PR description and/or change log message.
```suggestion
specifying the response behavior for unexpected resource states. 
```

smithy-model/src/test/resources/software/amazon/smithy/model/errorfiles/validators/idempotent-trait-both-properties.smithy
```diff
@@ -0,0 +1,25 @@
+$version: "2"
+
+namespace smithy.example
+
+@idempotent(
+    exists: ["smithy.example#ConflictException"]
+    notFound: ["smithy.example#NotFoundException"]
```
sugmanue: Using bare names is more idiomatic. Here and below.
```suggestion
    exists: [ConflictException]
    notFound: [NotFoundException]
```

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
```diff
@@ -4,27 +4,142 @@
...
+        public Trait createTrait(ShapeId target, Node value) {
+            IdempotentTrait result = new IdempotentTrait(value.expectObjectNode());
+            result.setNodeCache(value);
+            return result;
+        }
+    }
+
+    public static final class Builder extends AbstractTraitBuilder<IdempotentTrait, Builder> {
+        private List<ShapeId> exists;
+        private List<ShapeId> notFound;
```
sugmanue: We need to use a `BuilderRef` instead. See [here](https://github.com/smithy-lang/smithy/blob/4cfbf4728c1dcb3e4749489b59e9d10bec330ff7/smithy-utils/src/main/java/software/amazon/smithy/utils/BuilderRef.java#L185)

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
```diff
@@ -4,27 +4,142 @@
...
+
+        private Builder() {}
+
+        @Override
+        public IdempotentTrait build() {
+            return new IdempotentTrait(this);
+        }
+
+        public Builder exists(List<ShapeId> exists) {
+            this.exists = exists;
```
sugmanue: We need to copy the list instead of keeping a reference to it.

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
```diff
@@ -4,27 +4,142 @@
...
+    private static final String EXISTS = "exists";
+    private static final String NOT_FOUND = "notFound";
+
+    private final List<ShapeId> exists;
+    private final List<ShapeId> notFound;
+
+    private IdempotentTrait(Builder builder) {
+        super(ID, builder.sourceLocation);
+        this.exists = builder.exists;
+        this.notFound = builder.notFound;
```
sugmanue: The list needs to be made immutable. That can be solved if the builder uses a `BuilderRef`.

sugmanue (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/traits/IdempotentTrait.java
```diff
@@ -4,27 +4,142 @@
...
+                .map(id -> Node.from(id.toString()))
+                .collect(ArrayNode.collect());
+    }
+
+    @Override
+    public Builder toBuilder() {
+        return builder()
+                .sourceLocation(getSourceLocation())
+                .exists(exists)
+                .notFound(notFound);
```
sugmanue: Create instead a `Builder` constructor that takes the `IdempotentTrait` instead and inside uses `BuilderRef#setBorrowed` to avoid eagerly copying the collection.

sugmanue (approved):

PR #3219, opened 2026-07-20
Clean orphaned endpoint ruleset params after shape removal

### LINKED ISSUE #3190: Models transformed with `filterShapes` may no longer pass endpoint ruleset validation
The [AWS SDK for Kotlin](https://github.com/aws/aws-sdk-kotlin) uses [a Smithy transformer to remove operations from service models](https://github.com/smithy-lang/smithy-kotlin/blob/9a463900ec5a32953eb2170754e93eb0bbca0846/codegen/codegen/src/main/kotlin/aws/smithy/kotlin/codegen/transforms/IncludeOperations.kt#L16-L34) in order to [generate minimized service clients](https://github.com/aws/aws-sdk-kotlin/blob/72f3c275a8b2b48752fd27dfebf1fe41ad2c3ab1/aws-runtime/aws-config/build.gradle.kts#L89-L228) for use in credentials providers such as [STS](https://github.com/aws/aws-sdk-kotlin/blob/72f3c275a8b2b48752fd27dfebf1fe41ad2c3ab1/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProvider.kt#L104-L111), [SSO](https://github.com/aws/aws-sdk-kotlin/blob/72f3c275a8b2b48752fd27dfebf1fe41ad2c3ab1/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/SsoCredentialsProvider.kt#L101-L108), [Signin](https://github.com/aws/aws-sdk-kotlin/blob/72f3c275a8b2b48752fd27dfebf1fe41ad2c3ab1/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/LoginCredentialsProvider.kt#L91-L94), etc. Removing operations via [Smithy's `ModelTransformer.filterShapes` API](https://github.com/smithy-lang/smithy/blob/808398d66666134ad4026fee385f3496862234f0/smithy-model/src/main/java/software/amazon/smithy/model/transform/ModelTransformer.java#L184-L197) **does not** update the endpoint ruleset (or its BDD equivalent) in the service model. This can lead to a situation where ruleset parameters defined in a service's `endpointRuleSet` trait are not resolvable against the set of parameters defined in operations' `staticContextParams`, `operationContextParams`, etc. Essentially, ruleset parameters may become orphaned by shape removals. This is turn [causes Smithy's `RuleSetParameterValidator` to fail](https://github.com/smithy-lang/smithy/blob/808398d66666134ad4026fee385f3496862234f0/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/validators/RuleSetParameterValidator.java#L264-L269).

Smithy's built-in APIs for transforming models should be ruleset-aware so that they do not cause validators to fail on otherwise-valid transformations. Alternatively, Smithy could provide additional APIs for transforming rulesets which could be incorporated into calling code to achieve the same effect.

joewyz:
#### Background
#3190  reported validation failure caused by the orphaned paramters in `@endpointRuleSet` after the `removeShape` transform. This PR introduces a new model transform plugin `CleanEndpointRuleSetParameters` to clean up the orphaned paramters in `@endpointRuleSet`, `@endpointTests` and `@endpointBdd`.

After merge, this PR will fix #3190 

---

### DIFF
--- a/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
+++ b/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
@@ -1,0 +1,221 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.smithy.rulesengine.traits;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.transform.ModelTransformer;
+import software.amazon.smithy.model.transform.ModelTransformerPlugin;
+import software.amazon.smithy.rulesengine.language.EndpointRuleSet;
+import software.amazon.smithy.rulesengine.language.syntax.parameters.Parameter;
+import software.amazon.smithy.rulesengine.language.syntax.parameters.Parameters;
+import software.amazon.smithy.rulesengine.language.syntax.rule.Condition;
+import software.amazon.smithy.rulesengine.language.syntax.rule.Rule;
+import software.amazon.smithy.rulesengine.language.syntax.rule.TreeRule;
+import software.amazon.smithy.rulesengine.transforms.CompileBdd;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Removes orphaned endpoint ruleset parameters and their associated rules from a service's
+ * {@code @endpointRuleSet} trait when operations are removed from the model.
+ *
+ * <p>When operations are removed via {@link ModelTransformer#filterShapes} or
+ * {@link ModelTransformer#removeShapes}, parameters that were only bound through
+ * the removed operations' {@code @staticContextParams}, {@code @operationContextParams},
+ * or {@code @contextParam} traits may become orphaned. To prevent validation failures, this plugin
+ * removes those orphaned parameters and prunes any rules that reference them, recompiles the
+ * {@code @endpointBdd} trait from the cleaned ruleset, and drops {@code @endpointTests} cases that
+ * reference them.
+ */
+@SmithyInternalApi
+public final class CleanEndpointRuleSetParameters implements ModelTransformerPlugin {
+    @Override
+    public Model onRemove(ModelTransformer transformer, Collection<Shape> removed, Model model) {
+        if (!hasRelevantRemovals(removed)) {
+            return model;
+        }
+
+        Set<Shape> servicesToUpdate = new HashSet<>();
+        TopDownIndex topDownIndex = TopDownIndex.of(model);
+
+        for (ServiceShape service : model.getServiceShapesWithTrait(EndpointRuleSetTrait.class)) {
+            ServiceShape updated = cleanRuleSetParameters(model, topDownIndex, service);
+            if (updated != null) {
+                servicesToUpdate.add(updated);
+            }
+        }
+
+        if (servicesToUpdate.isEmpty()) {
+            return model;
+        }
+        return transformer.replaceShapes(model, servicesToUpdate);
+    }
+
+    private boolean hasRelevantRemovals(Collection<Shape> removed) {
+        for (Shape shape : removed) {
+            if (shape.isOperationShape() || shape.isMemberShape()) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private ServiceShape cleanRuleSetParameters(Model model, TopDownIndex topDownIndex, ServiceShape service) {
+        Set<String> boundParams = computeBoundParameters(model, topDownIndex, service);
+        EndpointRuleSetTrait endpointRuleSetTrait = service.expectTrait(EndpointRuleSetTrait.class);
+        EndpointRuleSet ruleSet = endpointRuleSetTrait.getEndpointRuleSet();
+
+        Set<String> orphanedParams = new HashSet<>();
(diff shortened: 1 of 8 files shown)


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+        return result;
+    }
+
+    private Rule pruneRule(Rule rule, Set<String> orphanedParams) {
+        // Drop the rule if any of its conditions reference an orphaned parameter.
+        if (conditionsReferenceOrphanedParam(rule.getConditions(), orphanedParams)) {
+            return null;
+        }
```
JordonPhillips, line 155: I think you've got the meaning of "pruned" confused here. When you prune a tree, you remove unwanted branches. The branches that are removed in this way are "pruned". You wouldn't say that the branches that are retained are pruned, however. It's a bit strange I'll grant you. Nevertheless I'd prefer you use the word "retained" or similar for things that aren't being removed.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
```diff
@@ -0,0 +1,221 @@
...
+            if (pruned != null) {
+                result.add(pruned);
+            }
+        }
+        return result;
+    }
+
+    private Rule pruneRule(Rule rule, Set<String> orphanedParams) {
+        // Drop the rule if any of its conditions reference an orphaned parameter.
+        if (conditionsReferenceOrphanedParam(rule.getConditions(), orphanedParams)) {
```
JordonPhillips, line 153: Conditions are the only thing you're checking, but parameters can also appear in endpoint URLs, headers, properties, and error expressions.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
```diff
@@ -0,0 +1,221 @@
...
+
+        // Compile BDD based on updated endpoint ruleset
+        if (service.hasTrait(EndpointBddTrait.ID)) {
+            serviceBuilder.addTrait(CompileBdd.compileBdd(updatedRuleSet));
+        }
+
+        // Clean endpoint tests that reference orphaned parameters.
+        if (service.hasTrait(EndpointTestsTrait.ID)) {
+            EndpointTestsTrait testsTrait = service.expectTrait(EndpointTestsTrait.class);
+            serviceBuilder.addTrait(cleanEndpointTests(testsTrait, orphanedParams));
```
JordonPhillips: This will add a trait even if every test is pruned.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
```diff
@@ -0,0 +1,221 @@
...
+        List<EndpointTestCase> cleanedCases = new ArrayList<>();
+        for (EndpointTestCase testCase : testsTrait.getTestCases()) {
+            if (!testCaseReferencesOrphanedParam(testCase, orphanedParams)) {
+                cleanedCases.add(testCase);
+            }
+        }
+        return testsTrait.toBuilder().testCases(cleanedCases).build();
+    }
+
+    private boolean testCaseReferencesOrphanedParam(EndpointTestCase testCase, Set<String> orphanedParams) {
```
JordonPhillips, line 250: This checks for explicit references, but test cases can rely on implicit references also. If a parameter has a default value, for example, the test expectations can rely on that without referencing the parameter explicitly.

I don't think there's a great way to determine this statically. You might have to just evaluate the tests to see if they still pass. I'd still do this static check first since it's deterministic and cheaper for the cases it covers.

JordonPhillips (changes requested):

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
joewyz: ~~I think if one parameter is used, then it would appear in the nearest condition arrary.~~ Will add check for other rules.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
joewyz: I found the other plugins specificlly retain the empty trait in this [commit](https://github.com/smithy-lang/smithy/commit/d8aa29df288cd034ac43857cd540441b52b75a64). Maybe we should follow the pattern?

JordonPhillips: Maybe ask kevin about that. I bet its to avoid errors from the trait not being there or something.

JordonPhillips (approved):
It's possible to remove all the test cases and then get a failure because there's a client context param that's required. But that's probably fine, since the root cause of that is that there's no dedicated test case for it.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
```diff
@@ -0,0 +1,283 @@
...
+import software.amazon.smithy.rulesengine.language.syntax.rule.Rule;
+import software.amazon.smithy.rulesengine.language.syntax.rule.TreeRule;
+import software.amazon.smithy.rulesengine.transforms.CompileBdd;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Removes orphaned endpoint ruleset parameters and their associated rules from a service's
+ * {@code @endpointRuleSet} trait when operations are removed from the model.
+ *
+ * <p>When operations are removed via {@link ModelTransformer#filterShapes} or
```
mtdowling, line 44: This strategy is sound for cases when a model has both an endpointRuleset and a BBD. But we plan to eventually drop the ruleset from published models. We need a strategy for that too. e.g., recompiling the BDD using only the BDD, publishing the endpointRuleset in a dedicated Smithy model alongside the model that has everything else, etc.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/CleanEndpointRuleSetParameters.java
joewyz: Found it was to solve one internal package build failure due to missing test trait back in the release of 1.39.0. I think we should just keep it in this case.

PR #3238, opened 2026-07-29
Add traits to declare stable wire order of members with syntax sugar

JordonPhillips:
This adds traits to stably declare the wire order of members. See the included design doc for more details.

The little things I invite people to yell about are the specific format of the syntax sugar and the way the formatter aligns it. I'm not really that bothered about either.

Oh and another weird thing about this is that the traits are in `smithy.protocols`, not the prelude. It's not hard to add the sugar inside smithy-model anyway, but I thought I'd call that out. I'm willing to consider moving the new traits to the prelude.

---


docs/source-2.0/spec/idl.rst
```diff
@@ -226,6 +227,15 @@ string support defined in :rfc:`7405`.
...
    productions were added to support inline collection declarations.
 
+.. versionadded:: 2.1
+   The ``MemberIndex`` production was added to provide shorthand for the
+   :ref:`idx trait <idx-trait>`.
+
+   The grammar does not limit the number of digits in a member index, so
+   syntax-only tools can parse and format values of any magnitude. When
+   assembling a semantic model, the value MUST fit in the Smithy ``integer``
+   type used by the ``idx`` trait and therefore MUST NOT exceed 2147483647.
```
sugmanue: I feel uneasy about allowing the full signed 32bit int range here. I'm pretty sure things will start breaking if people start using that many members. Can we consider restricting this further (say, 65536) and loosening if the need arises?

docs/source-2.0/spec/protocol-traits.rst
```diff
@@ -257,6 +257,160 @@ The following steps are taken to determine the serialization format of a
...
+
+``idx`` trait
+-------------
+
+Summary
+    Defines the stable wire index of a structure or union member.
+Trait selector
+    ``:is(structure, union) :not([trait|mixin]) > member``
+
+    *Any member of a non-mixin structure or union*
```
sugmanue:
```suggestion
    *Any member of a structure or union that is not a mixin or trait*
```

docs/source-2.0/spec/protocol-traits.rst
```diff
@@ -257,6 +257,160 @@ The following steps are taken to determine the serialization format of a
...
+Value type
+    ``integer`` greater than or equal to 1
+
+If any member of a structure or union has the ``idx`` trait, every member of
+that shape MUST have a unique ``idx`` value. Values MUST start at 1 and
+increase without gaps.
+
+Changing an existing member's ``idx`` value is a backward-incompatible change
```
sugmanue: I feel that "non backwards-compatible" is easier to parse than "backwards-incompatible".

```suggestion
Changing an existing member's ``idx`` value is a non backward-compatible change
```

smithy-model/src/main/java/software/amazon/smithy/model/loader/ParserUtils.java
```diff
@@ -39,7 +47,7 @@ public static String parseNumber(SimpleParser parser) {
 
         // Consume decimals.
         char peek = parser.peek();
-        if (peek == '.') {
+        if (peek == '.' && (!allowTrailingDotDelimiter || isDigit(parser.peek(1)))) {
```
sugmanue: Applying De Morgan's here helped me.
```suggestion
        if (peek == '.' && !(allowTrailingDotDelimiter && !isDigit(parser.peek(1)))) {
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -667,6 +671,23 @@ private void writeShapeMembers(Collection<MemberShape> members) {
...
+            Trait trait = traits.get(IDX_TRAIT);
+            if (trait == null || !traitFilter.test(trait) || !trait.toNode().isNumberNode()) {
+                return Optional.empty();
+            }
+            try {
+                int value = trait.toNode()
+                        .expectNumberNode()
+                        .asBigDecimal()
+                        .orElseThrow(ArithmeticException::new)
+                        .intValueExact();
```
sugmanue: No need to create a `BigDecimal`, the number should be already valid.
```suggestion
                        .getValue()
                        .intValue();
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -667,6 +671,23 @@ private void writeShapeMembers(Collection<MemberShape> members) {
...
+                return value > 0 ? Optional.of(value) : Optional.empty();
```
sugmanue: Is it ever possible for value to be `<= 0`? And if so, why would this corner of the code would change it if the value got this far?
```suggestion
                return Optional.of(value);
```

smithy-model/src/main/java/software/amazon/smithy/model/shapes/SmithyIdlModelSerializer.java
```diff
@@ -667,6 +671,23 @@ private void writeShapeMembers(Collection<MemberShape> members) {
             }
         }
 
+        private Optional<Integer> getMemberIndex(Map<ShapeId, Trait> traits) {
```
sugmanue: Personal preference: use a nullable `Integer` instead of `Optional`.

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IdxTrait.java
```diff
@@ -0,0 +1,85 @@
...
+            BigDecimal number = numberNode.asBigDecimal()
+                    .orElseThrow(() -> new ExpectationNotMetException(
+                            "Expected idx to be a finite integer",
+                            numberNode));
+            final int exactValue;
+            try {
+                exactValue = number.intValueExact();
+            } catch (ArithmeticException e) {
+                throw new ExpectationNotMetException(
+                        "Expected idx to be an integer between -2147483648 and 2147483647",
```
sugmanue, line 76: Consider adding the actual value here.

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTrait.java
```diff
@@ -0,0 +1,55 @@
...
+import software.amazon.smithy.model.FromSourceLocation;
+import software.amazon.smithy.model.SourceLocation;
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AbstractTrait;
+
+/**
+ * Requires structures and unions in a service closure to use member indexes.
+ */
+public final class IndexedTrait extends AbstractTrait {
```
sugmanue: Let's extend instead `AnnotationTrait`, and simplify this class.
```suggestion
public final class IndexedTrait extends AnnotationTrait {
```

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
```diff
@@ -0,0 +1,153 @@
...
+        int lastIdx = 0;
+        for (MemberShape member : members) {
+            if (!member.hasTrait(IdxTrait.class)) {
+                if (isIndexExempt(model, member)) {
+                    continue;
+                }
+                String reason = inServiceClosure
+                        ? String.format("Structure \"%s\" is in the closure of an indexed service",
+                                member.getContainer())
+                        : String.format("Shape \"%s\" contains an idx trait", member.getContainer());
```
sugmanue:
```suggestion
                        ? String.format("Structure `%s` is in the closure of an indexed service",
                                member.getContainer())
                        : String.format("Shape `%s` contains an idx trait", member.getContainer());
```

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
```diff
@@ -0,0 +1,153 @@
...
+                events.add(error(member,
+                        member.getSourceLocation(),
+                        String.format(
+                                "%s, but its member \"%s\" is missing an idx trait",
```
sugmanue:
```suggestion
                                "%s, but its member `%s` is missing an idx trait",
```

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
```diff
@@ -0,0 +1,153 @@
...
+            if (value == lastIdx) {
+                events.add(error(member, idx, String.format("Duplicate idx value \"%d\"", value)));
+            } else {
+                int expected = lastIdx + 1;
+                if (value != expected) {
+                    if (value == expected + 1) {
+                        events.add(error(member,
+                                idx,
+                                String.format(
+                                        "idx must increase monotonically starting at 1, no members found for idx %d",
```
sugmanue: [Monotonically increasing](https://mathworld.wolfram.com/MonotoneIncreasing.html) only imply that the values must be increasing not that there are no gaps. Furthermore, I don't see the need for the sequence of indexes to match the member order. You can add the a member that logically group together with an existing one, and while you cannot reshuffle the indexes to make room for it, you certainly can group them together in the model. 
Actually, this is explicitly stated on the design doc:

> The declaration order of members has no effect on their indexes

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
sugmanue: Moving the logic to create and report the event to a method would improve the readability of this method.

sugmanue (changes requested):

docs/source-2.0/spec/idl.rst
JordonPhillips: Sure. Realistically, even like 63 would stretch the bounds of belief for a real api

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
JordonPhillips: I don't agree. If it was the same error message repeated sure, but it isn't.

docs/source-2.0/spec/idl.rst
JordonPhillips: Since it's an integer, I went ahead and left the parser as-is. So it can still parse up to max int. But the range trait validator will reject anything above 2^16-1

designs/indexed-services.md
```diff
@@ -0,0 +1,386 @@
...
+The shorthand does not require a `use` statement.
+
+#### Grammar updates
+
+The shape member grammar is updated as follows:
+
+```abnf
+ShapeMember = TraitStatements [MemberIndex]
+              (ExplicitShapeMember / ElidedShapeMember) [ValueAssignment]
+MemberIndex = %x31-39 *(DIGIT) [`SP`] "." [`SP`]
```
yefrig: what's the reasoning behind allowing spacing between the digits and the `.`?

smithy-protocol-traits/src/main/java/software/amazon/smithy/protocol/traits/IndexedTraitValidator.java
sugmanue: Sure, up to you, but it's not about refactoring, is about creating the error messages take too much visual space to understand what the logic is, and that's what matters.

sugmanue (approved):

yefrig (approved):

designs/indexed-services.md
sugmanue: Good point, and what's the point of making the trailing space optional, now that I think about I'd prefer not to allow spaces between the index number and require one after the member index.

```
MemberIndex = %x31-39 *(DIGIT) "." `SP`
```

Otherwise all of the following are allowed:

```
  1. member
  1.member
  1 . member
  1 .member
```

sugmanue (approved):

yefrig (approved):

PR #3239, opened 2026-08-03
Add max selector parse depth of 100

kstich:


yefrig (approved):

smithy-model/src/main/java/software/amazon/smithy/model/selector/SelectorParser.java
```diff
@@ -111,22 +117,32 @@ List<InternalSelector> parse() {
...
+            while (!eof() && !BREAK_TOKENS.contains(peek())) {
+                selectors.add(createSelector());
+                // Always skip ws after calling createSelector.
+                ws();
+            }
+
+            return selectors;
+        } finally {
+            nestedSelectorDepth--;
+        }
```
sugmanue, line 145: Why do we care? Do we somehow use exceptions to control the flow that will require us to ensure that the depth counter gets decremented regardless?

PR #3244, opened 2026-08-05
Note when removed service operation remains in closure

yefrig:
Adds coverage for moving an operation out of direct service.operations while keeping it reachable through the same service closure, ensuring the diff emits a warning instead of treating it like a hard removal.

---

### DIFF
--- a/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
+++ b/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
@@ -1,6 +1,6 @@
 {
   "type": "feature",
-  "description": "The `RemovedOperationBinding` diff event is now a `WARNING` instead of an `ERROR` when a service operation is moved to a resource within the same service closure",
+  "description": "The `RemovedOperationBinding` diff event is now a `NOTE` instead of an `ERROR` when a service operation is moved to a resource within the same service closure",
   "pull_requests": [
     "[#3244](https://github.com/smithy-lang/smithy/pull/3244)"
   ]
--- a/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
+++ b/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
@@ -1,6 +1,6 @@
 {
   "type": "feature",
-  "description": "Update RemovedEntityBinding diff evaluator to warn when removed service operation remains in closure",
+  "description": "The `RemovedOperationBinding` diff event is now a `WARNING` instead of an `ERROR` when a service operation is moved to a resource within the same service closure",
   "pull_requests": [
     "[#3244](https://github.com/smithy-lang/smithy/pull/3244)"
   ]
--- a/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
+++ b/.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
@@ -1,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Update RemovedEntityBinding diff evaluator to warn when removed service operation remains in closure",
+  "pull_requests": [
+    "[#3244](https://github.com/smithy-lang/smithy/pull/3244)"
+  ]
+}
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
@@ -66,10 +66,11 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
 
         for (OperationShape operation : topDownIndex.getContainedOperations(entity)) {
             ShapeId currentId = operation.getId();
-            if (currentId.equals(operationId)) {
+            int result = currentId.compareTo(operationId);
+            if (result == 0) {
                 return true;
             }
-            if (currentId.compareTo(operationId) > 0) {
+            if (result > 0) {
                 return false;
             }
         }
--- a/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
+++ b/smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
@@ -10,8 +10,10 @@ import java.util.List;
 import java.util.Set;
 import software.amazon.smithy.diff.ChangedShape;
 import software.amazon.smithy.diff.Differences;
+import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.TopDownIndex;
 import software.amazon.smithy.model.shapes.EntityShape;
+import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.validation.Severity;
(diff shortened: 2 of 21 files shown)


smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/RemovedEntityBindingTest.java
```diff
@@ -37,6 +38,31 @@ public void detectsRemovedOperationFromService() {
         assertThat(events.get(0).getSourceLocation(), equalTo(source));
     }
 
+    @Test
+    public void warnsWhenOperationMovesFromServiceToContainedResource() {
```
kstich: How does this code handle the opposite direction of this test, if an operation is moved from a resource to the service?

yefrig: Should it? AFAIK the common use case is adding resources to a service. It seems like the opposite direction could also have (more?) impact for integrations/consumers :/

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -32,17 +33,35 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
...
+
+        for (ShapeId removed : findRemoved(oldEntity.getResources(), newEntity.getResources())) {
+            events.add(createRemovedEvent(REMOVED_RESOURCE, newEntity, removed, Severity.ERROR));
+        }
+    }
+
+    private boolean isStillContainedInService(EntityShape entity, ShapeId operationId, TopDownIndex topDownIndex) {
+        return entity.isServiceShape() && topDownIndex.getContainedOperations(entity)
+                .stream()
+                .anyMatch(operation -> operation.getId().equals(operationId));
```
sugmanue: Please avoid using streams, change this logic to use a plain old loop, also notice the javadoc comment on `getContainedOperations` _Operations returned are be sorted by id._. You can use that fact to short-circuit the loop.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -32,17 +33,35 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
...
+            TopDownIndex topDownIndex
+    ) {
+        EntityShape oldEntity = change.getOldShape();
+        EntityShape newEntity = change.getNewShape();
 
-        findRemoved(change.getOldShape().getResources(), change.getNewShape().getResources())
-                .forEach(removed -> events.add(createRemovedEvent(REMOVED_RESOURCE, change.getNewShape(), removed)));
+        for (ShapeId removed : findRemoved(oldEntity.getOperations(), newEntity.getOperations())) {
+            Severity severity =
+                    isStillContainedInService(newEntity, removed, topDownIndex) ? Severity.WARNING : Severity.ERROR;
```
sugmanue: Is there any reason not to downgrade this one all the way to `NOTE`?

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -32,17 +33,35 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
 
     @Override
     public List<ValidationEvent> evaluate(Differences differences) {
+        TopDownIndex topDownIndex = TopDownIndex.of(differences.getNewModel());
```
sugmanue: Let's move this one to `validateBindings` that's the only place where is used.

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -32,17 +33,35 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
...
+                    isStillContainedInService(newEntity, removed, topDownIndex) ? Severity.WARNING : Severity.ERROR;
+            events.add(createRemovedEvent(REMOVED_OPERATION, newEntity, removed, severity));
+        }
+
+        for (ShapeId removed : findRemoved(oldEntity.getResources(), newEntity.getResources())) {
+            events.add(createRemovedEvent(REMOVED_RESOURCE, newEntity, removed, Severity.ERROR));
+        }
+    }
+
+    private boolean isStillContainedInService(EntityShape entity, ShapeId operationId, TopDownIndex topDownIndex) {
```
sugmanue: Nit, "still" seems redundant, what about `isContainedInService(EntityShape, ShapeId, TopDownIndex)`

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -51,7 +70,12 @@ private Set<ShapeId> findRemoved(Set<ShapeId> oldShapes, Set<ShapeId> newShapes)
         return removed;
     }
 
-    private ValidationEvent createRemovedEvent(String typeOfRemoval, EntityShape parentEntity, ShapeId childShape) {
+    private ValidationEvent createRemovedEvent(
```
sugmanue: The same error message is used for both, the warning and error cases which I think is confusing, what about creating a dedicated error for the moved case? E.g., another method like:

```java
    private ValidationEvent createMovedEvent(EntityShape service, ShapeId operationId) {
        String message = String.format(
                "Operation binding of `%s` was removed from service shape, `%s`, "
                        + "but the operation is still bound within the service closure through a resource",
                operationId,
                service.getId());
        return ValidationEvent.builder()
                .id(REMOVED_OPERATION + FROM_SERVICE + operationId.getName())
                .severity(Severity.WARNING)
                .shape(service)
                .message(message)
                .build();
    }
```

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/RemovedEntityBindingTest.java
```diff
@@ -37,6 +38,65 @@ public void detectsRemovedOperationFromService() {
         assertThat(events.get(0).getSourceLocation(), equalTo(source));
     }
 
+    @Test
+    public void warnsWhenOperationMovesFromServiceToNestedResource() {
```
sugmanue: Can we also add tests for the test runner? Those are nice as can show clearly the before and after and it corresponding event. I can think of the following case:

* Operation fully removed from service (not in closure)    -> ERROR 
* Operation moved from service to a direct resource        -> WARNING
* Operation moved from service to a deeply nested resource -> WARNING
* Operation moved from resource up to service              -> ERROR (on resource)

sugmanue (changes requested):

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
yefrig: Yes, I like this!!

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/RemovedEntityBindingTest.java
yefrig: ohh I think this one doesn't have any. let me add some basic ones as well

.changes/next-release/feature-693344fa645e321a5cf5076fa3ce892fbe252215.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Update RemovedEntityBinding diff evaluator to warn when removed service operation remains in closure",
```
sugmanue: The description is developer-facing, not user-facing A more user-friendly description would be something like
```suggestion
  "description": "The `RemovedOperationBinding` diff event is now a `WARNING` instead of an `ERROR` when a service operation is moved to a resource within the same service closure",
```

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
```diff
@@ -33,16 +36,44 @@ public final class RemovedEntityBinding extends AbstractDiffEvaluator {
...
+        }
+
+        for (OperationShape operation : topDownIndex.getContainedOperations(entity)) {
+            ShapeId currentId = operation.getId();
+            if (currentId.equals(operationId)) {
+                return true;
+            }
+            if (currentId.compareTo(operationId) > 0) {
+                return false;
+            }
```
sugmanue: You can use a local variable to avoid two method calls.
```suggestion
            int result = currentId.compareTo(operationId);
            if (result == 0) {
                return true;
            }
            if (result > 0) {
                return false;
            }
```

smithy-diff/src/main/java/software/amazon/smithy/diff/evaluators/RemovedEntityBinding.java
yefrig: We all agree that a `NOTE` severity is more accurate. Updating!

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/RemovedEntityBindingTest.java
```diff
@@ -37,6 +38,97 @@ public void detectsRemovedOperationFromService() {
...
+        List<ValidationEvent> events = ModelDiff.compare(modelA, modelB);
+        List<ValidationEvent> removedEvents =
+                TestHelper.findEvents(events, "RemovedOperationBinding.FromService");
+
+        assertThat(removedEvents.size(), equalTo(2));
+        assertThat(TestHelper.findEvents(removedEvents, Severity.ERROR).size(), equalTo(2));
+    }
+
+    @Test
+    public void warnsWhenOperationMovesFromServiceToNestedResource() {
```
JordonPhillips: nit: `warns`

JordonPhillips (approved):

smithy-diff/src/test/java/software/amazon/smithy/diff/evaluators/RemovedEntityBindingTest.java
yefrig: haha I just had the same thought!

sugmanue (approved):

JordonPhillips (approved):

PR #3254, opened 2026-08-13
add recorpused rpcv2json protocol tests

lucix-aws:
Adds a new rpcv2json (+querycompatible) protocol test suite under our new corpus strategy.

Note that the models here do not build w/o https://github.com/smithy-lang/smithy/pull/3248.

Generated and ran these new tests against https://github.com/aws/smithy-go/pull/695.

### DIFF
--- a/smithy-aws-protocol-tests/model/v2/query-compat.smithy
+++ b/smithy-aws-protocol-tests/model/v2/query-compat.smithy
@@ -1,0 +1,38 @@
+$version: "2.0"
+
+namespace aws.protocoltests.corpus
+
+use aws.protocols#awsQueryError
+
+// Shapes for @awsQueryCompatible coverage: services that migrated off awsQuery and
+// have to keep exposing awsQuery error codes to old callers. The service itself is
+// declared per protocol, since the trait is service-level and only some protocols
+// accept it; this file holds the operation they share. Its two errors cover the code
+// being derived from the shape name and the code being overridden by @awsQueryError.
+//
+// Note @awsQueryError's httpResponseCode is NOT applied by the RPC v2 family: it
+// records what awsQuery itself would have returned, and the status still comes from
+// the @error trait.
+operation QueryCompatErrorOp {
+    input: QueryCompatErrorOpInput
+    output: QueryCompatErrorOpOutput
+    errors: [
+        QueryCompatError
+        QueryCompatCustomCodeError
+    ]
+}
+
+structure QueryCompatErrorOpInput {}
+
+structure QueryCompatErrorOpOutput {}
+
+@error("client")
+structure QueryCompatError {
+    message: String
+}
+
+@awsQueryError(code: "CustomCode", httpResponseCode: 402)
+@error("client")
+structure QueryCompatCustomCodeError {
+    message: String
+}
--- a/smithy-aws-protocol-tests/model/v2/rpcv2jsonquerycompatible/querycompat-test.smithy
+++ b/smithy-aws-protocol-tests/model/v2/rpcv2jsonquerycompatible/querycompat-test.smithy
@@ -1,0 +1,73 @@
+$version: "2.0"
+
+namespace aws.protocoltests.corpus
+
+use aws.protocoltests.config#ErrorCodeParams
+use smithy.protocols#rpcv2Json
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+
+apply QueryCompatErrorOp @httpRequestTests([
+    {
+        id: "RpcV2JsonQueryCompatSendsQueryModeHeader"
+        documentation: "Clients for query-compatible services MUST send the x-amzn-query-mode header"
+        protocol: rpcv2Json
+        method: "POST"
+        uri: "/service/RpcV2JsonQueryCompatCorpusTests/operation/QueryCompatErrorOp"
+        body: """
+            {}"""
+        bodyMediaType: "application/json"
+        headers: {
+            "smithy-protocol": "rpc-v2-json"
+            "Content-Type": "application/json"
+            Accept: "application/json"
+            "x-amzn-query-mode": "true"
+        }
+        forbidHeaders: ["X-Amz-Target"]
+        params: {}
+    }
+])
+
+apply QueryCompatError @httpResponseTests([
+    {
+        id: "RpcV2JsonQueryCompatNoCustomCodeError"
+        documentation: """
+            Parses an error that declares no @awsQueryError, so no
+            x-amzn-query-error header is sent and the code exposed to the caller
+            is the error shape's own name"""
+        protocol: rpcv2Json
+        code: 400
+        body: """
+            {
+                "__type": "aws.protocoltests.corpus#QueryCompatError",
+                "message": "query compatible failure"
+            }"""
+        bodyMediaType: "application/json"
+        headers: { "smithy-protocol": "rpc-v2-json", "Content-Type": "application/json" }
+        params: { message: "query compatible failure" }
+        vendorParamsShape: ErrorCodeParams
+        vendorParams: { code: "QueryCompatError" }
+    }
+])
+
+apply QueryCompatCustomCodeError @httpResponseTests([
+    {
+        id: "RpcV2JsonQueryCompatCustomCodeError"
+        documentation: """
+            Parses an error whose @awsQueryError code is surfaced through the
+            x-amzn-query-error header, and exposes that customized code rather
+            than the shape name"""
+        protocol: rpcv2Json
+        code: 400
+        body: """
+            {
+                "__type": "aws.protocoltests.corpus#QueryCompatCustomCodeError",
+                "message": "customized query error code"
+            }"""
+        bodyMediaType: "application/json"
+        headers: { "smithy-protocol": "rpc-v2-json", "Content-Type": "application/json", "x-amzn-query-error": "CustomCode;Sender" }
+        params: { message: "customized query error code" }
+        vendorParamsShape: ErrorCodeParams
+        vendorParams: { code: "CustomCode", type: "Sender" }
+    }
+])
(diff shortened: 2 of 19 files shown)


smithy-aws-protocol-tests/model/v2/query-compat.smithy
JordonPhillips: We should also have a test asserting that the compat header isn't sent if compat mode isn't enabled

smithy-protocol-tests/model/v2/rpcv2json/core-test.smithy
```diff
@@ -0,0 +1,3353 @@
...
+        code: 200
+        body: ""
+        bodyMediaType: "application/json"
+        headers: { "smithy-protocol": "rpc-v2-json", "Content-Type": "application/json" }
+        appliesTo: "client"
+        params: {}
+    }
+])
+
+apply NoInputOutput @httpRequestTests([
```
JordonPhillips, line 3296: We need a test for the output being a unit type too

smithy-aws-protocol-tests/model/v2/rpcv2jsonquerycompatible/querycompat-test.smithy
```diff
@@ -0,0 +1,73 @@
+$version: "2.0"
+
+namespace aws.protocoltests.corpus
+
+use aws.protocoltests.config#ErrorCodeParams
+use smithy.protocols#rpcv2Json
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+
+apply QueryCompatErrorOp @httpRequestTests([
```
JordonPhillips, line 10: We should also have a test asserting that the compat header isn't sent if compat mode isn't enabled

smithy-aws-protocol-tests/model/v2/rpcv2jsonquerycompatible/querycompat-test.smithy
```diff
@@ -0,0 +1,73 @@
...
+            "Content-Type": "application/json"
+            Accept: "application/json"
+            "x-amzn-query-mode": "true"
+        }
+        forbidHeaders: ["X-Amz-Target"]
+        params: {}
+    }
+])
+
+apply QueryCompatError @httpResponseTests([
```
JordonPhillips, line 31: This needs to use `forbidHeaders`

smithy-aws-protocol-tests/model/v2/query-compat.smithy
```diff
@@ -0,0 +1,38 @@
...
+// declared per protocol, since the trait is service-level and only some protocols
+// accept it; this file holds the operation they share. Its two errors cover the code
+// being derived from the shape name and the code being overridden by @awsQueryError.
+//
+// Note @awsQueryError's httpResponseCode is NOT applied by the RPC v2 family: it
+// records what awsQuery itself would have returned, and the status still comes from
+// the @error trait.
+operation QueryCompatErrorOp {
+    input: QueryCompatErrorOpInput
+    output: QueryCompatErrorOpOutput
```
JordonPhillips, line 18: You should be using inline IO everywhere unless there's a specific reason not to.

smithy-protocol-tests/model/v2/optionality.smithy
```diff
@@ -0,0 +1,188 @@
...
+    defaultByte: Byte = 0
+    defaultShort: Short = 0
+    defaultInteger: Integer = 0
+    defaultLong: Long = 0
+    defaultFloat: Float = 0
+    defaultDouble: Double = 0
+    defaultString: String = ""
+    defaultBlob: Blob = ""
+    defaultEnum: CorpusStringEnum = "Foo"
+    defaultIntEnum: CorpusIntEnum = 1
```
JordonPhillips, line 40: These defaults are identical to the zero values, and this very same structure has separate zero value members. Best to change them to something else.

smithy-protocol-tests/model/v2/rpcv2json/core-test.smithy
```diff
@@ -0,0 +1,3353 @@
+$version: "2.0"
+
+namespace smithy.protocoltests.corpus
+
+use smithy.protocols#rpcv2Json
+use smithy.test#httpRequestTests
+use smithy.test#httpResponseTests
+
+apply ScalarMembers @httpRequestTests([
```
JordonPhillips, line 9: These tests don't make assertions about bigDecimal / bigInteger. More generally, I would prefer that each test only tests one thing. It's fine to have the operation set up to test all the scalars, but I'd prefer to have one test for each type. From experience I can say it's really annoying to debug these kitchen sink style tests.

smithy-protocol-tests/model/v2/rpcv2json/core-test.smithy
```diff
@@ -0,0 +1,3353 @@
...
+        protocol: rpcv2Json
+        code: 200
+        body: """
+            {}"""
+        bodyMediaType: "application/json"
+        headers: { "smithy-protocol": "rpc-v2-json", "Content-Type": "application/json" }
+        params: {}
+    }
+    {
+        id: "RpcV2JsonEmptyInputOutputDeserializeEmptyBody"
```
JordonPhillips, line 3283: Should servers accept a body like this? If so, there needs to be a request version of it.

smithy-protocol-tests/model/v2/rpcv2json/optionality-test.smithy
```diff
@@ -0,0 +1,499 @@
...
+
+apply DefaultScalars @httpRequestTests([
+    {
+        id: "RpcV2JsonDefaultScalarsOmitsDefaults"
+        tags: ["top-level-default-omission"]
+        documentation: """
+            Client does not synthesize top-level input defaults. The members are
+            absent from params, so nothing is serialized: an explicitly provided
+            value is always sent, even when it equals the default, so this case
+            asserts the client does not fill them in on the caller's behalf."""
```
JordonPhillips, line 17: There isn't a test here that asserts that an explicitly provided value is always sent.

smithy-protocol-tests/model/v2/rpcv2json/optionality-test.smithy
```diff
@@ -0,0 +1,499 @@
...
+                "requiredBooleanWithDefault": true,
+                "requiredListWithDefault": ["c", "d"],
+                "requiredMapWithDefault": {"k1": "v1", "k2": "v2"}
+            }"""
+        bodyMediaType: "application/json"
+        headers: { "smithy-protocol": "rpc-v2-json", "Content-Type": "application/json", Accept: "application/json" }
+        requireHeaders: ["Content-Length"]
+        forbidHeaders: ["X-Amz-Target"]
+        params: {
+            requiredString: "hello"
```
JordonPhillips, line 326: By providing these here, you sorta miss the point of the test. Required values must be serialized, even with a default. Therefore you provide no value and then assert that the default is serialized.

smithy-protocol-tests/model/v2/rpcv2json/edges-test.smithy
```diff
@@ -0,0 +1,427 @@
...
+        documentation: "Serializes the minimum value of each integral type, including negatives"
+        protocol: rpcv2Json
+        method: "POST"
+        uri: "/service/RpcV2JsonCorpusTests/operation/ScalarMembers"
+        body: """
+            {
+                "byteMember": -128,
+                "shortMember": -32768,
+                "integerMember": -2147483648,
+                "longMember": -9007199254740993
```
JordonPhillips, line 69: You should add a comment about this in the doc string

smithy-protocol-tests/model/v2/event-stream.smithy
```diff
@@ -0,0 +1,292 @@
...
+    @required
+    events: HeaderEventStream
+}
+
+@streaming
+union HeaderEventStream {
+    headerEvent: HeaderEvent
+}
+
+structure HeaderEvent {
```
JordonPhillips, line 135: This is missing shorts and bytes

smithy-protocol-tests/model/v2/rpcv2json/eventstream-test.smithy
```diff
@@ -0,0 +1,645 @@
...
+                }
+                body: """
+                    {"content":"implicit payload content","count":33,"nested":{"stringMember":"nestedString","integerMember":44,"booleanMember":true}}"""
+                bodyMediaType: "application/json"
+            }
+        ]
+    }
+])
+
+apply EventStreamError @eventStreamTests([
```
JordonPhillips, line 309: These are all client tests, we also need server tests.

smithy-protocol-tests/model/v2/rpcv2json/eventstream-test.smithy
```diff
@@ -0,0 +1,645 @@
...
+apply EventStreamInitialResponse @eventStreamTests([
+    {
+        id: "RpcV2JsonEventStreamInitialResponse"
+        documentation: """
+            Non-stream output members are delivered ahead of the stream. Under
+            rpcv2Json the operation's @httpHeader bindings are ignored, so they
+            travel in the payload of an initial-response event rather than in
+            HTTP response headers."""
+        protocol: rpcv2Json
+        initialResponse: {
```
JordonPhillips, line 485: You should add `forbidHeaders` for the headers that shouldn't be here

smithy-protocol-tests/model/v2/rpcv2json/eventstream-test.smithy
JordonPhillips, line 1: These should all forbid `Content-Length` in the initial http request since it's forbidden for streaming operations.

smithy-protocol-tests/model/v2/event-stream.smithy
```diff
@@ -0,0 +1,292 @@
...
+    @required
+    events: BlobPayloadEventStream
+}
+
+@streaming
+union BlobPayloadEventStream {
+    blobEvent: BlobPayloadEvent
+}
+
+structure BlobPayloadEvent {
```
JordonPhillips, line 109: There's no structure / union `eventPayload`.

JordonPhillips (changes requested):
I think these all need a nice `smithy format` run too. Though please do that as a separate commit for now as it'll be impossible to review the diff otherwise.

PR #3266, opened 2026-08-27
Better gradle practices for adding Smithy generated content to the so…

swpalmer-cl:
…urceSets

This version is lazy and avoids the bad practice of using `dependsOn` and `afterEvaluate`

#### Background
* What do these changes do? 
 Use lazy evaluation and avoid Gradle bad practices.
* Why are they important?
 Improve compatibility with Gradle's configuration cache

#### Testing
 As these are recommendations for your project build scripts, these are what I use instead of the previous recommendations.
 
#### Links
https://docs.gradle.org/current/userguide/best_practices_general.html#avoid_after_evaluate
https://docs.gradle.org/current/userguide/best_practices_tasks.html#avoid_depends_on

---

### DIFF
--- a/docs/source-2.0/languages/java/client/generating-clients.rst
+++ b/docs/source-2.0/languages/java/client/generating-clients.rst
@@ -110,20 +110,12 @@ Gradle build script:
 
     // Add generated Java sources to the main sourceSet so they are compiled alongside
     // any other Java code in your package
-    afterEvaluate {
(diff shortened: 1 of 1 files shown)


docs/source-2.0/languages/java/client/generating-clients.rst
JordonPhillips, line 167: You should update here too

JordonPhillips (commented):
Looks good, thanks for the contribution! Just gotta carry that change down to the full example

PR #3267, opened 2026-08-27
Resolve endpoint rule-set components against the model's classloader

sugmanue:
#### Background

Endpoint rule-set components (functions like `aws.partition`, built-ins like `AWS::Region`, auth-scheme validators like `sigv4`) are contributed by `EndpointRuleSetExtension` providers found via `ServiceLoader`. The rules engine discovered them using `EndpointRuleSet.class.getClassLoader()` instead of the classloader passed to `Model.assembler(ClassLoader)`. When a model is assembled against a dependency-closure classloader (a child of the rules-engine loader) that holds the providers, a parent loader cannot see a child's providers, so assembly failed with `` `aws.partition` is not a valid function ``, `` `AWS::Region` built-in used is not registered ``, and spurious `sigv4` auth-scheme warnings.

This threads the model's discovery classloader into component resolution as a resolved `EndpointComponentFactory`:

* `TraitService` gains a defaulted `createTrait(ShapeId, Node, ClassLoader)` overload; `TraitFactory` captures the loader from `Model.assembler(ClassLoader)` and passes it to providers.
* `EndpointRuleSetTrait` and `EndpointBddTrait` build and memoize one factory from that loader and thread it through deserialization (`fromNode` into `Rule`/`Condition`/`Endpoint`/`Expression`/`FunctionNode`, including nested-argument functions and functions in URLs, headers, and errors) and the `RuleSetBuiltIn`/`RuleSetAuthSchemes` validators.
* The factory is carried on `EndpointRuleSet`, `Cfg`, and `EndpointBddTrait` so transforms that round-trip a rule-set or compile a BDD (`CfgBuilder`, `CoalesceTransform`, `IsSetBooleanCoalesceTransform`, the `TreeMapper` transforms, `EndpointBddTrait.from(Cfg)`) and `EndpointPathCollector` re-resolve against the same classloader.
* Resolution is cached per trait instance; a shared default factory is reused when no distinct classloader is present, so closure classloaders are not pinned.

Any consumer that assembles models with a custom classpath (build tools isolating a dependency closure, IDEs, application servers, runtime model-assembly services) can now validate models whose endpoint functions live only in that closure. All changes are additive overloads with defaults: no public signature changes, and `equals`/`hashCode`/`toNode`/serialization are unaffected. `adr/0001` records the decision, rejected alternatives, and the pattern for future traits.

#### Testing

* New integration test `smithy-aws-endpoints/src/it/EndpointRuleSetClassLoaderIntegrationTest`: assembles models using `aws.partition` (`@endpointRuleSet` and `@endpointBdd`) and nested `aws.parseArn`, under a classloader that sees the extension (0 errors) and one that hides it (reproduces the failure). The negative cases fail on the pre-fix code.
* `smithy-rules-engine:test` (421) and `smithy-model:test` (2672) pass on `--rerun-tasks`, including the BDD/CFG round-trip and equivalence suites.
* `compileJava`, `spotbugsMain`, `spotlessJavaCheck` pass for both modules.
* Not guarded by a failing-on-pre-fix test (documented in the ADR): the `Cfg`/`CfgBuilder` re-parse during BDD compilation and functions used directly in a URL, header, or error. Both are correct by construction; the fixtures cannot isolate them without a synthetic string-returning extension.

#### Links

* ADR: `adr/0001-endpoint-rule-set-classloader-resolution.md`
* Issue #:

### DIFF
--- a/smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
@@ -30,16 +30,24 @@ public interface TraitFactory {
     Optional<Trait> createTrait(ShapeId id, ShapeId target, Node value);
 
     /**
+     * @return the classloader used to discover trait providers, or null if unknown.
+     *
+     * <p>Factories created via {@link #createServiceFactory(ClassLoader)} return that classloader
+     * so it can be threaded into providers that need it (see
+     * {@link TraitService#createTrait(ShapeId, Node, ClassLoader)}). The default returns null.
+     */
+    default ClassLoader traitClassLoader() {
+        return null;
+    }
+
+    /**
      * Creates a TraitFactory that uses a List of TraitService provider instances.
      *
      * @param services List of TraitService provider instances.
      * @return Returns the created TraitFactory.
      */
     static TraitFactory createServiceFactory(Iterable<TraitService> services) {
-        Map<ShapeId, TraitService> serviceMap = new HashMap<>();
-        services.forEach(service -> serviceMap.put(service.getShapeId(), service));
-        return (name, target, value) -> Optional.ofNullable(serviceMap.get(name))
-                .map(provider -> provider.createTrait(target, value));
+        return new ServiceFactory(services, null);
     }
 
     /**
@@ -60,6 +68,34 @@ public interface TraitFactory {
      * @return Returns the created TraitFactory.
      */
     static TraitFactory createServiceFactory(ClassLoader classLoader) {
-        return createServiceFactory(ServiceLoader.load(TraitService.class, classLoader));
+        return new ServiceFactory(ServiceLoader.load(TraitService.class, classLoader), classLoader);
+    }
+
+    /**
+     * Default {@link TraitFactory} backed by a map of {@link TraitService} providers.
+     *
+     * <p>Retains the classloader used to discover the providers (when known) and passes it to
+     * {@link TraitService#createTrait(ShapeId, Node, ClassLoader)} so providers can resolve
+     * classloader-sensitive components from the same closure the model is assembled against.
+     */
+    final class ServiceFactory implements TraitFactory {
+        private final Map<ShapeId, TraitService> serviceMap = new HashMap<>();
+        private final ClassLoader classLoader;
+
+        private ServiceFactory(Iterable<TraitService> services, ClassLoader classLoader) {
+            services.forEach(service -> serviceMap.put(service.getShapeId(), service));
+            this.classLoader = classLoader;
+        }
+
+        @Override
+        public Optional<Trait> createTrait(ShapeId id, ShapeId target, Node value) {
+            return Optional.ofNullable(serviceMap.get(id))
+                    .map(provider -> provider.createTrait(target, value, classLoader));
+        }
+
+        @Override
+        public ClassLoader traitClassLoader() {
+            return classLoader;
+        }
     }
 }
--- a/smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
+++ b/smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
@@ -30,17 +30,6 @@ public interface TraitFactory {
     Optional<Trait> createTrait(ShapeId id, ShapeId target, Node value);
 
     /**
-     * @return the classloader used to discover trait providers, or null if unknown.
-     *
-     * <p>Factories created via {@link #createServiceFactory(ClassLoader)} return that classloader
-     * so it can be threaded into providers that need it (see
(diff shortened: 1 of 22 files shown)


smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/rule/Rule.java
```diff
@@ -80,23 +82,38 @@ public static Builder builder(FromSourceLocation sourceLocation) {
...
+     * Creates a {@link Rule} instance from the given Node, resolving any functions in conditions
+     * using the given {@link EndpointComponentFactory}. When {@code factory} is {@code null}, the
+     * default statically-discovered function factory is used.
+     *
+     * @param node the node to deserialize.
+     * @param factory the factory used to resolve functions, or null.
+     * @return the created Rule.
+     */
+    @SmithyInternalApi
+    public static Rule fromNode(Node node, EndpointComponentFactory factory) {
```
JordonPhillips, line 98: Can we have a settings or context object we pass around instead so we don't have to be changing all these signatures for the next thing?

smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
```diff
@@ -60,6 +68,34 @@ static TraitFactory createServiceFactory() {
...
+    }
+
+    /**
+     * Default {@link TraitFactory} backed by a map of {@link TraitService} providers.
+     *
+     * <p>Retains the classloader used to discover the providers (when known) and passes it to
+     * {@link TraitService#createTrait(ShapeId, Node, ClassLoader)} so providers can resolve
+     * classloader-sensitive components from the same closure the model is assembled against.
+     */
+    final class ServiceFactory implements TraitFactory {
```
JordonPhillips, line 70: I don't think this needs to be public

smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
```diff
@@ -29,17 +29,25 @@ public interface TraitFactory {
...
     Optional<Trait> createTrait(ShapeId id, ShapeId target, Node value);
 
+    /**
+     * @return the classloader used to discover trait providers, or null if unknown.
+     *
+     * <p>Factories created via {@link #createServiceFactory(ClassLoader)} return that classloader
+     * so it can be threaded into providers that need it (see
+     * {@link TraitService#createTrait(ShapeId, Node, ClassLoader)}). The default returns null.
+     */
+    default ClassLoader traitClassLoader() {
```
JordonPhillips: This doesn't seem to be used

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointBddTrait.java
```diff
@@ -57,6 +61,8 @@ public final class EndpointBddTrait extends AbstractTrait implements ToSmithyBui
     private final List<Condition> conditions;
     private final List<Rule> results;
     private final Bdd bdd;
+    private final ClassLoader classLoader;
+    private volatile EndpointComponentFactory componentFactory;
```
JordonPhillips, line 65: Why do we need to store these? Seems really odd

JordonPhillips (commented):
Brief first look stuff, I'll do a proper review tomorrow

smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
sugmanue: Indeed, I will remove it.

smithy-model/src/main/java/software/amazon/smithy/model/traits/TraitFactory.java
sugmanue: The class is not public, it's private package.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/language/syntax/rule/Rule.java
sugmanue: I don't feel like that's needed, this code is well encapsulated and I don't foresee it needing anything else, the factory is now passed instead of being used statically to keep its state.

smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointBddTrait.java
sugmanue: The factory use service loader to create extensions and it's used in the validation multiple times. We store it to reuse and avoid the doing the work everytime.

JordonPhillips (approved):

PR #3269, opened 2026-08-31
Fix property and identifier bindings for collection operations

### LINKED ISSUE #2968: ResourceOperationInputOutput errors if a nested resource only has a list operation
Consistently getting the following error for all resource properties when I attempt to create a nested resource with only a list operation that references all of the nested resource properties in a structure used by a list. If I add a get operation referencing the resource properties in an output structure, the errors clear. 

```console
──  ERROR  ──────────────────────────────────────── ResourceOperationInputOutput
Shape: com.example#NamespaceTaskResource
File:  [redacted]/model/namespace/task/namespace-task.smithy:8:1

7| @since("2025-10-09")
8| resource NamespaceTaskResource {
 | ^

Resource property `Name` is not used in the input or output of create or an
instance operation.
```

tadurisaikiran:
### Issue

Closes #2968

### Description of changes

Resources whose properties only appear in the output of a `list` operation are rejected by `ResourceOperationInputOutput` with "Resource property `X` is not used in the input or output of create or an instance operation", because neither `PropertyBindingIndex` nor `IdentifierBindingIndex` resolves bindings through the output's list member. Following the review discussion, element binding is now explicit and opt-in, with a narrow automatic case for the `list` lifecycle:

- Explicit element binding: `nestedProperties` can target a member that targets a list of structures in a collection-bound operation's input or output, marking each list element as carrying the state of one resource instance. `notProperty` now applies to members of such element structures. With the trait present, element members are validated strictly, using the same rules as top-level members of instance operations.
- `list` lifecycle auto-detection: without the trait, the output element is detected only when exactly one output member targets a list of structures. Findings there are warnings and unmatched members are ignored, since automatically detected element structures may be shared between resources or carry derived data. Ambiguous outputs keep current behavior.
- Resource-scoped bindings: element bindings are exposed through new accessors on both indexes, keyed by resource and operation. Element members never enter the member-keyed maps, so a shared element structure cannot leak bindings from one resource into another.
- No top-level changes: top-level input and output indexing behaves exactly as it did before this PR, so batch operations and all existing models are unaffected unless they opt in.
- Docs: the resource properties section of the spec gained a subsection on binding properties to collection operations, and the `nestedProperties` and `notProperty` trait references are updated.

New test cases include the exact example from the issue discussion, the shared element structure example from the review (asserting zero events), a strict unknown member (error), an auto-detected property mismatch (warning), a batch-style input binding, explicit disambiguation between two list members, an `@resourceIdentifier` rename inside an element, and an ambiguous two-list output falling back to current behavior.

Validation: full `./gradlew build` passes, the smithy-model suite is green including all pre-existing errorfiles, and `smithy-aws-traits` and `smithy-aws-cloudformation-traits` (the two `PropertyBindingIndex` consumers) pass.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
```diff
@@ -37,6 +38,7 @@ public final class PropertyBindingIndex implements KnowledgeIndex {
     private final Map<ShapeId, String> memberShapeToPropertyName = new HashMap<>();
```
JordonPhillips, line 40: Exposing collection properties exposes and amplifies a latent bug with these structures. The key is a shape id. Before that was fine because that would be a member of an input or output shape, which are almost never shared. But now it targets some other shape that very likely could be shared. For example:

```smithy
$version: "2"

namespace example.catalog

service CatalogService {
    version: "2026-09-02"
    resources: [Product, Supplier]
}

resource Product {
    identifiers: { productId: String }
    properties: { displayName: String }
    read: GetProduct
    list: ListProducts
}

resource Supplier {
    identifiers: { supplierId: String }
    properties: { contactEmail: String }
    read: GetSupplier
    list: ListSuppliers
}

@readonly
operation GetProduct {
    input := { @required productId: String }
}

@readonly
operation GetSupplier {
    input := { @required supplierId: String }
    output := { contactEmail: String }
}

@readonly
operation ListProducts {
    output := { items: SearchResults }
}

@readonly
operation ListSuppliers {
    output := { items: SearchResults }
}

list SearchResults {
    member: SearchResult
}

structure SearchResult {
    displayName: String
}
```

This would result in an error like this:

```
──  ERROR  ──────────────────────────────────────── ResourceOperationInputOutput
Shape: example.catalog#SearchResult$displayName
File:  shared-element.smithy:50:5
49| structure SearchResult {
50|     displayName: String
  |     ^
Member `displayName` does not target a property or identifier for resource
`example.catalog#Supplier`. If it is an identifier, apply the
`smithy.api#resourceIdentifier` trait. If it is a property, apply the
`smithy.api#property` trait. If it is neither, apply the
`smithy.api#notProperty` trait.
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
JordonPhillips, line 298: This needs updated

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
```diff
@@ -106,6 +113,65 @@ public static PropertyBindingIndex of(Model model) {
         return model.getKnowledge(PropertyBindingIndex.class, PropertyBindingIndex::new);
     }
 
+    /**
+     * Indexes an operation bound to a resource through the {@code list}
+     * lifecycle or the {@code collectionOperations} property.
+     *
+     * <p>Collection operation inputs carry filters and pagination details
+     * rather than resource state, so no input member binds to a property.
```
JordonPhillips: This is not true. A collection operation is an operation that operates on the collection, not necessarily an operation that returns a collection. For example, a batch put or delete would be collection operations that bind properties.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/CollectionElementResolver.java
```diff
@@ -0,0 +1,72 @@
...
+        if (itemsMember.isPresent()) {
+            return output.get().getMember(itemsMember.get()).flatMap(member -> resolveListElement(model, member));
+        }
+
+        List<StructureShape> candidates = new ArrayList<>();
+        for (MemberShape member : output.get().members()) {
+            resolveListElement(model, member).ifPresent(candidates::add);
+        }
+        return candidates.size() == 1 ? Optional.of(candidates.get(0)) : Optional.empty();
+    }
```
JordonPhillips, line 86: I'm not sure I like this. Mostly I'm wary of trying to guess with things like the paginated trait. That could point to a map or a separately-paginated list (it happens). See the top-level comment for more musings.

JordonPhillips (commented):
Good draft! Some answers and suggestions:

Leniency is generally the way to go. Introducing new error-level validators is intolerable, so we should avoid it as much as possible. This doesn't go all the way down that route though, since we're still validating surface that wasn't validated before.

What really concerns me is guessing output collections. A collection operation operates on the collection, but it does not necessarily return a collection. A single list in the output does not imply that the list is a collection of the resource for anything but the `list` lifecycle operation.

What if you updated `nestedProperties` to work on a list member for collection operations? You'd also have to update `notProperty` to apply to list members. That way it would be explicit and opt-in, so we could be strict. Maybe we could automatically detect the list for the `list` lifecycle. That'll probably have to be downgraded to warning without the nestedProperties trait present through.

Batch operations can't be ignored. See comments for details, but the changes made result in a regression for them.

And yes, docs should be updated.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
tadurisaikiran: Good catch, and I fixed it structurally rather than patching around it. Element members no longer go into the member-keyed maps at all. Element bindings are computed and exposed per resource and operation through new accessors (getOperationOutputElementShape and getOperationOutputElementProperties, plus input equivalents), and the validator reads those instead. Your exact example is now a test case (shared-list-element-structure.smithy) asserting zero events.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
tadurisaikiran: Agreed. That javadoc claim was wrong and the behavior behind it is gone. Top-level input and output indexing is back to exactly what it was before this PR, so batch operations bind properties the same way they always did. They can additionally bind through list elements by marking the carrying member with nestedProperties. There is a new test for a batch-style input (collection-operation-input-properties.smithy).

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/CollectionElementResolver.java
tadurisaikiran: Removed the paginated heuristic entirely. Resolution is now nestedProperties first, and the only automatic case left is the list lifecycle output when exactly one output member targets a list of structures. Anything ambiguous falls back to doing nothing.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Updated. That path no longer handles element structures at all. Element validation moved to a dedicated resource-scoped path that uses the new element binding accessors for identifier exemption. The unused-property message now also mentions list elements of collection operations.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/IdentifierBindingIndex.java
```diff
@@ -100,6 +104,49 @@ public Map<String, String> getOperationOutputBindings(ToShapeId resource, ToShap
                 .orElseGet(Collections::emptyMap);
     }
 
+    /**
```
joewyz: can you please format the doc for these two methods, this seems misplaced.

tadurisaikiran: Fixed in f5c5b00. The output accessor's javadoc had ended up stacked above the input accessor. Each method now has its own doc in the right place.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
```diff
@@ -114,62 +119,152 @@ public static PropertyBindingIndex of(Model model) {
...
+                ? explicitOutput
+                : CollectionElementResolver.isListLifecycle(resource, operation.getId())
+                        ? CollectionElementResolver.resolveAutoOutputElement(model, operation)
+                        : Optional.empty();
+        outputElement.ifPresent(element -> outputElementBindings.get(resource.getId())
+                .put(operation.getId(),
+                        createElementBinding(resource,
+                                element,
+                                explicit,
+                                identifierIndex.getOperationOutputElementBindings(resource, operation))));
```
JordonPhillips, line 163: `resolveExplicitElement` returns empty if `nestedProperties` targets something other than a list, and then this will go down an the auto path.

Realistically it's not likely that somebody would do this, but I'd rather give them an error

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
```diff
@@ -114,13 +116,122 @@ private List<ValidationEvent> validateResource(Model model, ResourceShape resour
...
+                        "This member must target `%s`. This member is used as part of the `%s` operation of the `%s` "
+                                + "resource and conflicts with its `%s` resource property.",
+                        expectedTarget,
+                        bindingName,
+                        resource.getId(),
+                        propertyName);
+                events.add(strict ? error(member, message) : warning(member, message));
+            }
+        }
+        validateConflictingProperties(events, element, propertyToMemberMappings);
```
JordonPhillips, line 232: This needs to pass along the strict property

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
```diff
@@ -114,13 +116,122 @@ private List<ValidationEvent> validateResource(Model model, ResourceShape resour
...
+                                    + "`%s`. If it is an identifier, apply the `%s` trait. If it is a property, apply "
+                                    + "the `%s` trait. If it is neither, apply the `%s` trait.",
+                                    member.getMemberName(),
+                                    resource.getId().toString(),
+                                    ResourceIdentifierTrait.ID,
+                                    PropertyTrait.ID,
+                                    NotPropertyTrait.ID)));
+                }
+                continue;
+            }
```
JordonPhillips, line 210: You remove `notProperty` members from this map, so this error will trigger even if the trait is being appropriately used.

JordonPhillips (changes requested):
Some missing test coverage:

* deleting isListLifecycle results in no test failures
* There's not a test asserting that auto-detection is for list only
* There's no tests for a nested `notProperty`

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Fixed. Strict mode now checks for notProperty, including traits that carry it (same meta-trait logic the index uses), before reporting an unbound element member. New test: list-element-notproperty.smithy has a notProperty member in a strict element and asserts zero events.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Done. validateConflictingProperties now takes the strict flag, so conflicts in auto-detected elements are warnings, while explicit elements and the existing top-level paths stay errors.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/PropertyBindingIndex.java
tadurisaikiran: Agreed. An unresolvable marker no longer falls back to auto-detection in either index, and the validator reports it as an error saying the trait must target a list of structures when applied to a collection operation member. Covered by list-element-marker-misuse.smithy. Instance operations are untouched, so the classic structure-targeting usage still works there.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
```diff
@@ -93,18 +96,180 @@ private List<ValidationEvent> validateResource(Model model, ResourceShape resour
...
+            String bindingName,
+            ShapeId elementId,
+            Map<String, String> properties,
+            Map<String, String> identifierBindings,
+            boolean strict,
+            List<ValidationEvent> events
+    ) {
+        StructureShape element = model.expectShape(elementId, StructureShape.class);
+        Set<String> identifierMembers = new HashSet<>(identifierBindings.values());
+        Set<ShapeId> notPropertyTraits = computeNotPropertyTraits(model);
```
sugmanue, line 219: This will be recomputed for each call, let's pull this up and get the computation made upstream and pass down instead.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
```diff
@@ -93,18 +96,180 @@ private List<ValidationEvent> validateResource(Model model, ResourceShape resour
...
+        }
+        validateConflictingProperties(events, element, propertyToMemberMappings, strict);
+    }
+
+    private Set<ShapeId> computeNotPropertyTraits(Model model) {
+        return model.getShapesWithTrait(NotPropertyTrait.class)
+                .stream()
+                .filter(shape -> shape.hasTrait(TraitDefinition.ID))
+                .map(Shape::toShapeId)
+                .collect(Collectors.toSet());
```
sugmanue, line 270: Here and elsewhere, prefer plain old loops over streams. Streams have a measurable performance impact and add pressure to the garbage collector. Also, this same information is already computed by `PropertyBindingIndex`, I feel like we could expose it from there instead of adding the duplicated logic here.
```suggestion
        Set<ShapeId> result = new HashSet<>();
        for (Shape shape : model.getShapesWithTrait(NotPropertyTrait.class)) {
            if (shape.hasTrait(TraitDefinition.ID)) {
                ShapeId shapeId = shape.toShapeId();
                result.add(shapeId);
            }
        }
        return result;
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
```diff
@@ -93,18 +96,180 @@ private List<ValidationEvent> validateResource(Model model, ResourceShape resour
...
+        Set<ShapeId> notPropertyTraits = computeNotPropertyTraits(model);
+        Map<String, Set<MemberShape>> propertyToMemberMappings = new TreeMap<>();
+
+        for (MemberShape member : element.members()) {
+            if (identifierMembers.contains(member.getMemberName())) {
+                continue;
+            }
+            String propertyName = properties.get(member.getMemberName());
+            if (propertyName == null) {
+                if (strict && notPropertyTraits.stream().noneMatch(member::hasTrait)) {
```
sugmanue, line 228: Avoid the use of streams, if needed introduce a new method to keep the code nice to read.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/CollectionElementResolver.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Returns true if any member of the given input or output shape is
+     * marked with the {@code @nestedProperties} trait.
+     */
+    static boolean hasElementMarker(Model model, ShapeId ioShapeId) {
+        return model.getShape(ioShapeId)
+                .flatMap(Shape::asStructureShape)
+                .map(shape -> shape.members().stream().anyMatch(m -> m.hasTrait(NestedPropertiesTrait.ID)))
```
sugmanue, line 58: Please avoid the use of streams, you can create a helper method if needed but use plain old loops.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Good call. The index already computes this once per model, so I exposed isNotPropertyMember from PropertyBindingIndex and deleted the validator's copy entirely. That removes both the per-call recomputation and the streams.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Addressed together with your comment below: the validator's computation is gone, replaced by a single call to the index, which computes the set once per model.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceOperationInputOutputValidator.java
tadurisaikiran: Gone with the same change. The check is now a single index call.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/CollectionElementResolver.java
tadurisaikiran: Replaced with a plain loop.

tadurisaikiran:
@JordonPhillips @sugmanue latest commit addresses all of the round-3 comments: the validator now uses a new isNotPropertyMember accessor on PropertyBindingIndex (computed once per model), its duplicated helper and streams are deleted, and the remaining new stream in the resolver is a plain loop. I left the pre-existing streams that this PR did not introduce untouched to keep the diff focused. Full smithy-model suite and repo build are green. Ready for another look whenever you have time.

JordonPhillips (approved):

PR #3273, opened 2026-09-03
Add resource lifecycle traits

sugmanue:
### What behavior changes?

Adds five `@unstable` `smithy.api` prelude traits that declare which resources an operation affects when the relationship is not expressed through a resource's standard lifecycle: `@createsResources`, `@putsResources`, `@deletesResources`, `@readsResources`, and `@updatesResources`.

Each trait is a list of bindings. A binding names a resource and, for its identifiers and properties, may be unspecified, explicit (a map of name to JMESPath locator), or inferred (an `identifiersFrom` / `propertiesFrom` pointer at a nested structure or projection whose members are matched by name). Sides are fixed per trait (create: identifiers out, properties in; put and update: both in; read: identifiers in, properties out; delete: identifiers in, no properties).

New validation resolves each locator against the operation input or output: identifier and property name matching, supported-JMESPath enforcement, exact leaf-type checking, a cardinality check that rejects identifiers projecting through different lists, and inference rules for the `...From` form. `ResourceLifecycleIndex` exposes operation-to-resource lookups, and a transform plugin drops bindings when a referenced resource is removed.

`MemberShouldReferenceResourceValidator` no longer warns about an operation's own identifier-named members when a lifecycle binding on that operation already accounts for them (by resolved locator or by naming the resource).

### Why is this change needed?

Supports non-standard and legacy operations that handle resources outside the standard 1:1 resource lifecycle, such as batch create and delete, operations that affect several resource types at once, and operations not bound to the resource they affect. Declaring the relationship in the model lets any model-driven tool consume it. The minimal form also overlaps with `aws.iam#@iamAction(createsResources:)` for the multi-resource-creation case, and the traits live in the prelude so they are available to any model.

### How was this validated?

`./gradlew :smithy-model:test` (full suite) and `:smithy-model:spotlessJavaCheck` pass. Added:
- Trait unit tests for all five traits (node round-trip including all binding members).
- `ResourceLifecycleResolverTest`: the JMESPath shape-walk, cardinality signature, flatten, nested projections, map traversal, and name inference.
- `ResourceLifecycleIndexTest` and `CleanResourceLifecycleReferencesTest`.
- Errorfiles covering valid models plus each error and warning path (name mismatch, unsupported and invalid JMESPath, structural resolution, cardinality, no-root, strict-unmatched, property type-mismatch, schema-level rejection of `properties` on delete), negative cases for puts, reads, and updates, and a case proving lifecycle bindings suppress the `@references` advisory.

### What should reviewers focus on?

- `loader/prelude.smithy`: the five trait definitions, `ResourceLifecycleBinding` / `ResourceDeletionBinding` / `ResourceMemberBinding`, `@unstable`, and `breakingChanges`.
- `validation/validators/ResourceLifecycleResolver.java`: the JMESPath-to-shape walk, cardinality signature, and name inference.
- `validation/validators/ResourceLifecycleTraitValidator.java`: per-side dispatch and the validation matrix.
- `validation/validators/MemberShouldReferenceResourceValidator.java`: the exemption for lifecycle-bound members.
- `docs/source-2.0/spec/resource-traits.rst`: the public spec, including the boundary with `@nestedProperties` collection binding.

### Additional Links

(none)

---

### DIFF
--- a/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
+++ b/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
@@ -1,0 +1,5 @@
+{
+  "type": "feature",
+  "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits, letting you declare which resources an operation affects when the relationship falls outside a resource's standard lifecycle.",
+  "pull_requests": []
+}
--- a/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
+++ b/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
@@ -1,5 +1,7 @@
 {
   "type": "feature",
   "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits, letting you declare which resources an operation affects when the relationship falls outside a resource's standard lifecycle.",
-  "pull_requests": []
+  "pull_requests": [
+    "[#3273](https://github.com/smithy-lang/smithy/pull/3273)"
+  ]
 }
--- a/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
+++ b/.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
@@ -1,6 +1,6 @@
 {
   "type": "feature",
-  "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits, letting you declare which resources an operation affects when the relationship falls outside a resource's standard lifecycle.",
+  "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits that declare which resources an operation affects for relationships that fall outside a resource's standard lifecycle.",
   "pull_requests": [
     "[#3273](https://github.com/smithy-lang/smithy/pull/3273)"
   ]
--- a/docs/source-2.0/spec/resource-traits.rst
+++ b/docs/source-2.0/spec/resource-traits.rst
@@ -448,5 +448,271 @@ identifier.
         name: String
     }
 
+.. _resource-lifecycle-association-traits:
+
+Resource lifecycle association traits
+=====================================
+
+The ``createsResources``, ``putsResources``, ``deletesResources``,
+``readsResources``, and ``updatesResources`` traits declare which resources an
+operation affects when that relationship is not expressed through a resource's
+standard lifecycle operations. They let tooling understand, from the model
+alone, that an operation creates, replaces, deletes, reads, or updates one or
+more resources, and where to find the affected resources' identifiers and
+properties.
+
+These traits are useful for operations that a 1:1 resource lifecycle cannot
+express, such as batch operations, operations that affect several resource
+types at once, or operations that are not bound to the resource they affect.
+
+For an operation that is bound to a resource's standard lifecycle, including a
+``list`` or other collection operation, resource properties and identifiers are
+bound through the operation's members, and the :ref:`nested-properties-trait` is
+used to bind them within a list element. These traits are for the operations
+that fall outside that standard lifecycle binding, and they locate identifiers
+and properties explicitly with :ref:`JMESPath <supported-jmespath>` rather than
+by member binding.
+
+Each trait is a ``list`` of bindings. A binding names a resource and,
+optionally, declares where the resource's identifiers and properties are
+located in the operation's input or output.
+
+Identifiers, properties, and sides
+----------------------------------
+
+A binding can locate two independent kinds of information about the resource:
+
+* **identifiers**: where the affected resource's identifiers are located.
+* **properties**: where the affected resource's properties are located.
+
+Identifiers and properties each resolve against a fixed side of the operation,
+determined by the trait:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 34 33 33
+
+    * - Trait
+      - Identifiers side
+      - Properties side
+    * - ``createsResources``
+      - output
+      - input
+    * - ``putsResources``
+      - input
+      - input
+    * - ``updatesResources``
+      - input
+      - input
+    * - ``readsResources``
+      - input
+      - output
+    * - ``deletesResources``
+      - input
+      - (none)
+
+The sides reflect where the data lives: a created resource's identifier is
+generated into the output while its properties are supplied in the input; a read
+resource is selected by an identifier in the input and its state is returned in
+the output; and so on. ``deletesResources`` does not locate properties, because
+a resource is deleted by its identifier alone.
+
+Specifying identifiers and properties
+-------------------------------------
+
+Identifiers and properties are each optional and are specified in one of three
+ways:
+
+* **Unspecified**: neither the explicit map nor the ``...From`` pointer is set.
+  The binding only asserts that the operation affects the resource; nothing is
+  inferred and nothing is validated for it. For example,
+  ``@createsResources([{resource: Forecast}])`` states only that the operation
+  creates a ``Forecast``.
+* **Explicit**: an ``identifiers`` or ``properties`` map associates each
+  resource identifier or property name with a :ref:`JMESPath <supported-jmespath>`
+  expression that locates its value on the appropriate side.
+* **Inferred**: an ``identifiersFrom`` or ``propertiesFrom`` expression points
+  at a nested structure or a projection whose members are matched by name to the
+  resource's identifiers or properties. Member-name differences are reconciled
+  with the :ref:`property-trait` and :ref:`resourceIdentifier-trait`, and the
+  :ref:`notProperty-trait` excludes a member.
+
+The explicit map and the ``...From`` pointer may be combined; the map overrides
+inference for a given name. The value located for an identifier must resolve to
+a string, and the value located for a property must match the resource
+property's target shape.
+
+.. _supported-jmespath:
+
+Supported JMESPath
+------------------
+
+Locators and ``...From`` pointers use a structural subset of JMESPath: field
+access, subexpressions, wildcard projections (``[*]``), flatten (``[]``), and
+the current node. Functions, filters, comparators, index and slice expressions,
+and multi-select expressions are not supported. A ``...From`` pointer must
+resolve to a nested structure or a projection of structures; it may not resolve
+to the whole input or output root, nor to a scalar.
+
+Binding structures
+------------------
+
+``createsResources``, ``putsResources``, ``updatesResources``, and
+``readsResources`` use the ``ResourceLifecycleBinding`` structure:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 20 33 47
+
+    * - Property
+      - Type
+      - Description
+    * - resource
+      - :ref:`shape-id`
+      - **Required**. The absolute shape ID of the affected resource. Must
+        target a :ref:`resource` shape.
+    * - identifiers
+      - ``map<string, ResourceMemberBinding>``
+      - Explicit map of resource identifier name to a locator for its value.
+    * - identifiersFrom
+      - ``string``
+      - Structural JMESPath pointing at a nested structure or projection from
+        whose members identifiers are inferred by name.
+    * - properties
+      - ``map<string, ResourceMemberBinding>``
+      - Explicit map of resource property name to a locator for its value.
+    * - propertiesFrom
+      - ``string``
+      - Structural JMESPath pointing at a nested structure or projection from
+        whose members properties are inferred by name.
+
+``deletesResources`` uses the ``ResourceDeletionBinding`` structure, which is
+identical but omits the ``properties`` and ``propertiesFrom`` members.
+
+Both structures locate values with the ``ResourceMemberBinding`` structure:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 20 33 47
+
+    * - Property
+      - Type
+      - Description
+    * - path
+      - ``string``
+      - **Required**. A structural JMESPath expression that locates the value in
+        the operation input or output.
+
+.. smithy-trait:: smithy.api#createsResources
+.. _createsResources-trait:
+
+``createsResources`` trait
+--------------------------
+
+Summary
+    Declares which resources an operation creates. Identifiers are located in
+    the output; properties are located in the input.
+Trait selector
+    ``operation``
+Value type
+    ``list`` of ``ResourceLifecycleBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is a breaking change.
+
+.. code-block:: smithy
+
+    @createsResources([
+        {
+            resource: Snapshot
+            identifiers: { snapshotId: { path: "snapshots[*].snapshotId" } }
+        }
+    ])
+    operation BatchCreateSnapshots {
+        input: BatchCreateSnapshotsInput
+        output: BatchCreateSnapshotsOutput
+    }
+
+.. smithy-trait:: smithy.api#putsResources
+.. _putsResources-trait:
+
+``putsResources`` trait
+-----------------------
+
+Summary
+    Declares which resources an operation creates or replaces with put
+    semantics (a caller-provided identifier). Identifiers and properties are
+    both located in the input.
+Trait selector
+    ``operation``
+Value type
+    ``list`` of ``ResourceLifecycleBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is a breaking change.
+
+.. smithy-trait:: smithy.api#deletesResources
+.. _deletesResources-trait:
+
+``deletesResources`` trait
+--------------------------
+
+Summary
+    Declares which resources an operation deletes. Identifiers are located in
+    the input. Delete does not locate properties.
+Trait selector
+    ``operation``
+Value type
+    ``list`` of ``ResourceDeletionBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is a breaking change.
+
+.. code-block:: smithy
+
+    @deletesResources([
+        {
+            resource: Volume
+            identifiers: { volumeId: { path: "volumeIds[*]" } }
+        }
+    ])
+    @idempotent
+    operation BatchDeleteVolumes {
+        input: BatchDeleteVolumesInput
+        output: BatchDeleteVolumesOutput
+    }
+
+.. smithy-trait:: smithy.api#readsResources
+.. _readsResources-trait:
+
+``readsResources`` trait
+------------------------
+
+Summary
+    Declares which resources an operation reads. Identifiers are located in the
+    input; properties are located in the output.
+Trait selector
+    ``operation``
+Value type
+    ``list`` of ``ResourceLifecycleBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is flagged as a possible
+breaking change.
+
+.. smithy-trait:: smithy.api#updatesResources
+.. _updatesResources-trait:
+
+``updatesResources`` trait
+--------------------------
+
+Summary
+    Declares which resources an operation updates. Identifiers and properties
+    are both located in the input.
+Trait selector
+    ``operation``
+Value type
+    ``list`` of ``ResourceLifecycleBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is flagged as a possible
+breaking change.
+
+
 .. _CreateTable: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html
 .. _standard link relation: https://www.iana.org/assignments/link-relations/link-relations.xhtml
--- a/docs/source-2.0/spec/resource-traits.rst
+++ b/docs/source-2.0/spec/resource-traits.rst
@@ -453,11 +453,11 @@ identifier.
 Resource lifecycle association traits
 =====================================
 
-The ``createsResources``, ``putsResources``, ``deletesResources``,
-``readsResources``, and ``updatesResources`` traits declare which resources an
+The ``createsResources``, ``putsResources``, ``readsResources``,
+``updatesResources``, and ``deletesResources`` traits declare which resources an
 operation affects when that relationship is not expressed through a resource's
 standard lifecycle operations. They let tooling understand, from the model
-alone, that an operation creates, replaces, deletes, reads, or updates one or
+alone, that an operation creates, replaces, reads, updates, or deletes one or
 more resources, and where to find the affected resources' identifiers and
 properties.
 
@@ -501,12 +501,12 @@ determined by the trait:
     * - ``putsResources``
       - input
       - input
-    * - ``updatesResources``
-      - input
-      - input
     * - ``readsResources``
       - input
       - output
+    * - ``updatesResources``
+      - input
+      - input
     * - ``deletesResources``
       - input
       - (none)
@@ -523,11 +523,6 @@ Specifying identifiers and properties
 Identifiers and properties are each optional and are specified in one of three
 ways:
 
-* **Unspecified**: neither the explicit map nor the ``...From`` pointer is set.
-  The binding only asserts that the operation affects the resource; nothing is
-  inferred and nothing is validated for it. For example,
-  ``@createsResources([{resource: Forecast}])`` states only that the operation
-  creates a ``Forecast``.
 * **Explicit**: an ``identifiers`` or ``properties`` map associates each
   resource identifier or property name with a :ref:`JMESPath <supported-jmespath>`
   expression that locates its value on the appropriate side.
@@ -536,11 +531,16 @@ ways:
   resource's identifiers or properties. Member-name differences are reconciled
   with the :ref:`property-trait` and :ref:`resourceIdentifier-trait`, and the
   :ref:`notProperty-trait` excludes a member.
+* **Unspecified**: neither the explicit map nor the ``...From`` pointer is set.
+  The binding only asserts that the operation affects the resource; nothing is
+  inferred and nothing is validated for it. For example,
+  ``@createsResources([{resource: Forecast}])`` states only that the operation
+  creates a ``Forecast``.
 
 The explicit map and the ``...From`` pointer may be combined; the map overrides
-inference for a given name. The value located for an identifier must resolve to
-a string, and the value located for a property must match the resource
-property's target shape.
+inference for a given name. The value located for an identifier must match the
+resource identifier's target shape, and the value located for a property must
+match the resource property's target shape.
 
 .. _supported-jmespath:
 
@@ -557,8 +557,8 @@ to the whole input or output root, nor to a list, map, or scalar.
 Binding structures
 ------------------
 
-``createsResources``, ``putsResources``, ``updatesResources``, and
-``readsResources`` use the ``ResourceLifecycleBinding`` structure,
+``createsResources``, ``putsResources``, ``readsResources``, and
+``updatesResources`` use the ``ResourceLifecycleBinding`` structure,
 which has the following members:
 
 .. list-table::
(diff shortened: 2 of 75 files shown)


.changes/next-release/feature-d4a59053a7b286e0d1d8e998531eccbb6266bdb1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "feature",
+  "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits, letting you declare which resources an operation affects when the relationship falls outside a resource's standard lifecycle.",
```
kstich:
```suggestion
  "description": "Added the @createsResources, @putsResources, @deletesResources, @readsResources, and @updatesResources traits that declare which resources an operation affects for relationships that fall outside a resource's standard lifecycle.",
```

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+generated into the output while its properties are supplied in the input; a read
+resource is selected by an identifier in the input and its state is returned in
+the output; and so on. ``deletesResources`` does not locate properties, because
+a resource is deleted by its identifier alone.
+
+Specifying identifiers and properties
+-------------------------------------
+
+Identifiers and properties are each optional and are specified in one of three
+ways:
```
kstich, line 524: In the list below, I'd move either "Unspecified" or "Explicit" to the bottom so the list is in descending or ascending specificity order.

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+  expression that locates its value on the appropriate side.
+* **Inferred**: an ``identifiersFrom`` or ``propertiesFrom`` expression points
+  at a nested structure or a projection whose members are matched by name to the
+  resource's identifiers or properties. Member-name differences are reconciled
+  with the :ref:`property-trait` and :ref:`resourceIdentifier-trait`, and the
+  :ref:`notProperty-trait` excludes a member.
+
+The explicit map and the ``...From`` pointer may be combined; the map overrides
+inference for a given name. The value located for an identifier must resolve to
+a string, and the value located for a property must match the resource
```
kstich: The `identifier` side of this sentence allows for a target to be a `String` while the resource's identifier may point to `ConstrainedString`. Is that intended?

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+
+Supported JMESPath
+------------------
+
+Locators and ``...From`` pointers use a structural subset of JMESPath: field
+access, subexpressions, wildcard projections (``[*]``), flatten (``[]``), and
+the current node. Functions, filters, comparators, index and slice expressions,
+and multi-select expressions are not supported. A ``...From`` pointer must
+resolve to a nested structure or a projection of structures; it may not resolve
+to the whole input or output root, nor to a scalar.
```
kstich:
```suggestion
and multi-select expressions are not supported. A ``...From`` pointer must
resolve to a nested structure or a projection of structures. It must not resolve
to the whole input or output root, nor to a list, map, or scalar.
```

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+
+Binding structures
+------------------
+
+``createsResources``, ``putsResources``, ``updatesResources``, and
+``readsResources`` use the ``ResourceLifecycleBinding`` structure:
```
kstich:
```suggestion
``readsResources`` use the ``ResourceLifecycleBinding`` structure,
which has the following members:
```

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+      - Explicit map of resource property name to a locator for its value.
+    * - propertiesFrom
+      - ``string``
+      - Structural JMESPath pointing at a nested structure or projection from
+        whose members properties are inferred by name.
+
+``deletesResources`` uses the ``ResourceDeletionBinding`` structure, which is
+identical but omits the ``properties`` and ``propertiesFrom`` members.
+
+Both structures locate values with the ``ResourceMemberBinding`` structure:
```
kstich:
```suggestion
Both structures locate values with the ``ResourceMemberBinding`` structure,
which has the following members:
```

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+``readsResources`` use the ``ResourceLifecycleBinding`` structure:
+
+.. list-table::
+    :header-rows: 1
+    :widths: 20 33 47
+
+    * - Property
+      - Type
+      - Description
+    * - resource
```
kstich, line 571: Does this need to carry any information about the service the targeted resource is bound to?

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+    ])
+    operation BatchCreateSnapshots {
+        input: BatchCreateSnapshotsInput
+        output: BatchCreateSnapshotsOutput
+    }
+
+.. smithy-trait:: smithy.api#putsResources
+.. _putsResources-trait:
+
+``putsResources`` trait
```
kstich, line 640: Doesn't have an example.

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+
+.. code-block:: smithy
+
+    @createsResources([
+        {
+            resource: Snapshot
+            identifiers: { snapshotId: { path: "snapshots[*].snapshotId" } }
+        }
+    ])
+    operation BatchCreateSnapshots {
```
kstich, line 632: `deletesResources` has a batch, could this be a simpler example like a `RestoreTable` operation that uses the `...From` member?

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+    * - ``createsResources``
+      - output
+      - input
+    * - ``putsResources``
+      - input
+      - input
+    * - ``updatesResources``
+      - input
+      - input
+    * - ``readsResources``
```
kstich: nit: Flip these two so it matches CRUD ordering.

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
         name: String
     }
 
+.. _resource-lifecycle-association-traits:
+
+Resource lifecycle association traits
+=====================================
+
+The ``createsResources``, ``putsResources``, ``deletesResources``,
+``readsResources``, and ``updatesResources`` traits declare which resources an
```
kstich: The order of these throughout the spec aren't consistent. This listing and the table below are different, then it switches to match this listing for the traits order.

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+    @idempotent
+    operation BatchDeleteVolumes {
+        input: BatchDeleteVolumesInput
+        output: BatchDeleteVolumesOutput
+    }
+
+.. smithy-trait:: smithy.api#readsResources
+.. _readsResources-trait:
+
+``readsResources`` trait
```
kstich, line 680: Doesn't have an example.

docs/source-2.0/spec/resource-traits.rst
```diff
@@ -448,5 +448,271 @@ identifier.
...
+Value type
+    ``list`` of ``ResourceLifecycleBinding`` structures.
+
+Removing a binding or changing a binding's ``resource`` is flagged as a possible
+breaking change.
+
+.. smithy-trait:: smithy.api#updatesResources
+.. _updatesResources-trait:
+
+``updatesResources`` trait
```
kstich, line 722: Doesn't have an example.

kstich (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanResourceLifecycleReferences.java
```diff
@@ -0,0 +1,130 @@
...
+            OperationShape.Builder builder,
+            Set<ShapeId> removedIds,
+            ShapeId traitId
+    ) {
+        Optional<Trait> traitOptional = operation.findTrait(traitId);
+        if (!traitOptional.isPresent()) {
+            return builder;
+        }
+
+        AbstractResourceLifecycleTrait lifecycleTrait = (AbstractResourceLifecycleTrait) traitOptional.get();
```
kstich: Use `hasTrait`/`expectTrait`

docs/source-2.0/spec/resource-traits.rst
sugmanue: No at the moment, as of now the only information is implicit in the relationship between the operation and the service.

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanResourceLifecycleReferences.java
```diff
@@ -0,0 +1,130 @@
...
+import software.amazon.smithy.model.traits.ResourceLifecycleBinding;
+import software.amazon.smithy.model.traits.Trait;
+import software.amazon.smithy.model.traits.UpdatesResourcesTrait;
+import software.amazon.smithy.model.transform.ModelTransformer;
+import software.amazon.smithy.model.transform.ModelTransformerPlugin;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+
+/**
+ * Removes references to resources that are removed from resource lifecycle
```
kstich, line 30: My read is that models are broken if an explicit identifier/property binding has the member it points at removed. Does that need another cleaner?

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ResourceLifecycleIndex.java
```diff
@@ -0,0 +1,148 @@
...
+        traitToLifecycle.put(PutsResourcesTrait.ID, Lifecycle.PUT);
+        traitToLifecycle.put(ReadsResourcesTrait.ID, Lifecycle.READ);
+        traitToLifecycle.put(UpdatesResourcesTrait.ID, Lifecycle.UPDATE);
+
+        for (OperationShape operation : model.getOperationShapes()) {
+            for (Map.Entry<ShapeId, Lifecycle> entry : traitToLifecycle.entrySet()) {
+                operation.findTrait(entry.getKey())
+                        .ifPresent(trait -> index(operation.getId(),
+                                entry.getValue(),
+                                ((AbstractResourceLifecycleTrait) trait).getBindings()));
```
kstich: `hasTrait`/`expectTrait` again

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ResourceLifecycleIndex.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * The lifecycle effect an operation has on a resource.
+     */
+    public enum Lifecycle {
+        CREATE,
+        DELETE,
+        PUT,
+        READ,
+        UPDATE
+    }
```
kstich, line 60: a) CPRUD
b) These are public interface and should have documentation on the values.

smithy-model/src/main/java/software/amazon/smithy/model/traits/DeletesResourcesTrait.java
```diff
@@ -0,0 +1,52 @@
...
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.utils.SmithyUnstableApi;
+import software.amazon.smithy.utils.ToSmithyBuilder;
+
+/**
+ * Declares which resources an operation deletes and where to find their
+ * identifiers in the operation input.
+ */
+@SmithyUnstableApi
+public final class DeletesResourcesTrait extends AbstractResourceLifecycleTrait
```
kstich, line 16: I think the way this is set up allows it to have the `properties` and `propertiesFrom` fields set.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+ *   <li>{@link #inferByName} matches the members of a structure to a resource's identifier or
+ *       property names, honoring {@code @resourceIdentifier}, {@code @property}, and
+ *       {@code @notProperty}, mirroring the standard lifecycle resolution.</li>
+ * </ul>
+ *
+ * <p>Only the structural JMESPath subset is supported (field, subexpression, projection
+ * {@code [*]}, flatten {@code []}, current node). Callers reject unsupported expressions
+ * before walking.
+ */
+final class ResourceLifecycleResolver {
```
kstich, line 62: There are a lot of comments in here that are `/**` style instead of `//` style.

kstich (changes requested):

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -1566,3 +1566,176 @@ structure requestCompression {
...
+}
+
+@private
+map ResourceMemberBindings {
+    key: NonEmptyString
+    value: ResourceMemberBinding
+}
+
+@private
+structure ResourceMemberBinding {
```
kstich, line 1738: I believe this structure needs documentation as well.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+    /** The result of inferring resource members from a structure by name. */
+    static final class InferenceResult {
+        /** Matched identifier or property name to the structure member that provides it. */
+        final Map<String, MemberShape> matched = new LinkedHashMap<>();
+        /** Members that matched no identifier or property name and are not {@code @notProperty}. */
+        final List<MemberShape> unmatched = new ArrayList<>();
+    }
+
+    // Internal descriptor: a value of "shape" wrapped in the listed arrays (outermost first).
+    private static final class Type {
```
kstich, line 252: This is just a `Pair`

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleTraitValidator.java
```diff
@@ -0,0 +1,764 @@
...
+                    trait,
+                    format("Binding for resource `%s` in `@%s` specifies properties, but `@%s` deletes a resource "
+                            + "by its identifiers alone and has no properties.",
+                            binding.getResource(),
+                            descriptor.name,
+                            descriptor.name)));
+        }
+    }
+
+    private void validateBindingKind(
```
kstich, line 244: There are some massive method signatures in this class, like this one, that cause this file to be huge line wise. Feels like an inner data class would help.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+            List<ShapeId> arrays = new ArrayList<>(inner.type.arrays);
+            if (arrays.size() >= 2) {
+                arrays.remove(1);
+            }
+            return Result.ok(new Type(inner.type.shape, arrays));
+        }
+        return Result.err("unsupported expression");
+    }
+
+    private static List<ShapeId> tail(List<ShapeId> arrays) {
```
kstich: This is used at one call site.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+ * <p>Only the structural JMESPath subset is supported (field, subexpression, projection
+ * {@code [*]}, flatten {@code []}, current node). Callers reject unsupported expressions
+ * before walking.
+ */
+final class ResourceLifecycleResolver {
+
+    private ResourceLifecycleResolver() {}
+
+    /** Whether a member set is being matched against a resource's identifiers or its properties. */
+    enum BindingKind {
```
kstich, line 154: The organization of methods and inner classes here could use some work. The entry point is buried behind several private/package private helper types.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+        IDENTIFIER, PROPERTY
+    }
+
+    /** The result of walking a structural path over a starting shape. */
+    static final class PathResult {
```
kstich, line 159: This is probably chunky enough to be its own file.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+            return new PathResult(leaf, root, arrays, null);
+        }
+
+        static PathResult error(String error) {
+            return new PathResult(null, false, null, error);
+        }
+    }
+
+    /** The result of inferring resource members from a structure by name. */
+    static final class InferenceResult {
```
kstich, line 191: Maybe also enough for its own class, given its package private?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+            if (matched) {
+                result.matched.put(boundName, member);
+            } else {
+                result.unmatched.add(member);
+            }
+        }
+        return result;
+    }
+
+    private static String boundName(MemberShape member, BindingKind kind) {
```
kstich: I think there are indexes that precompute these and are used in validation elsewhere.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+
+    /**
+     * Infers which members of a structure provide the resource's identifiers or properties.
+     *
+     * @param resource The resource being bound.
+     * @param element The structure to match members against.
+     * @param kind Whether to match identifier names or property names.
+     * @return The matched and unmatched members.
+     */
+    static InferenceResult inferByName(ResourceShape resource, StructureShape element, BindingKind kind) {
```
kstich: Can this be pulled from the IdentifierBindingIndex or PropertyBindingIndex?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
```diff
@@ -0,0 +1,322 @@
...
+        static Result ok(Type type) {
+            return new Result(type, null);
+        }
+
+        static Result err(String error) {
+            return new Result(null, error);
+        }
+    }
+
+    private static Result resolve(Model model, JmespathExpression expr, Type in) {
```
kstich: Should this instead be an `ExpressionVisitor` implementation?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleTraitValidator.java
```diff
@@ -0,0 +1,764 @@
...
+                break;
+            }
+        }
+        if (!anyApplied) {
+            return Collections.emptyList();
+        }
+
+        List<ValidationEvent> events = new ArrayList<>();
+        for (OperationShape operation : model.getOperationShapes()) {
+            for (Descriptor descriptor : DESCRIPTORS) {
```
kstich: Inverting these loops would allow for `getOperationShapesWithTrait` use, though maybe that's marginally slower? Either way this loop set should use the `has`+`expect` paradigm instead.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleTraitValidator.java
```diff
@@ -0,0 +1,764 @@
...
+    @Override
+    public List<ValidationEvent> validate(Model model) {
+        Set<ShapeId> applied = model.getAppliedTraits();
+        boolean anyApplied = false;
+        for (Descriptor descriptor : DESCRIPTORS) {
+            if (applied.contains(descriptor.traitId)) {
+                anyApplied = true;
+                break;
+            }
+        }
```
kstich: Is it worth making this a map of `trait => applied`?

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleTraitValidator.java
```diff
@@ -0,0 +1,764 @@
...
+    private static ShapeId unwrapBaseId(Model model, ShapeId id) {
+        Shape shape = model.expectShape(id);
+        while (shape instanceof ListShape) {
+            shape = model.expectShape(((ListShape) shape).getMember().getTarget());
+        }
+        return shape.getId();
+    }
+
+    private boolean isLifecycleBound(ResourceShape resource, ShapeId operationId, String traitName) {
+        switch (traitName) {
```
kstich: This should use the trait's IDs

kstich (changes requested):

docs/source-2.0/spec/resource-traits.rst
sugmanue: Addressed.

docs/source-2.0/spec/resource-traits.rst
sugmanue: Done

docs/source-2.0/spec/resource-traits.rst
sugmanue: I have adjusted.

docs/source-2.0/spec/resource-traits.rst
sugmanue: I will make it strict.

docs/source-2.0/spec/resource-traits.rst
sugmanue: Added.

docs/source-2.0/spec/resource-traits.rst
sugmanue: Added one.

docs/source-2.0/spec/resource-traits.rst
sugmanue: Added one.

docs/source-2.0/spec/resource-traits.rst
sugmanue: Added one.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
sugmanue: Yeah, this looks similar to these indexes but has a different input and contract, so reuse would be counter productive.

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanResourceLifecycleReferences.java
sugmanue: Yes, I think is best to let the assembler fail for that case, we have the same issue currently with `@references` where we don't clean up when the member is removed.

smithy-model/src/main/java/software/amazon/smithy/model/traits/DeletesResourcesTrait.java
sugmanue: Yeah, tried but it makes the type signature way too complicated or the logic has to account for it. I lean towards keeping this as-is.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ResourceLifecycleIndex.java
sugmanue: Fixed.

smithy-model/src/main/java/software/amazon/smithy/model/knowledge/ResourceLifecycleIndex.java
sugmanue: Changed.

smithy-model/src/main/java/software/amazon/smithy/model/traits/DeletesResourcesTrait.java
sugmanue: Let me know if you feel strong about it.

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanResourceLifecycleReferences.java
sugmanue: Changed.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
sugmanue: I do prefer to keep the specific type just for the sake of readability. Pair just has left and right which doesn't tell you much about what those are.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
sugmanue: It's a bit more complicated for the inner dependencies, I'm leaning towards keeping it as-is for now, let me know if you feel strong about it.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
sugmanue: Yes, changed.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
sugmanue: I didn't find any, do you remember offhand which ones do it?

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
sugmanue: Added.

smithy-model/src/main/java/software/amazon/smithy/model/traits/DeletesResourcesTrait.java
kstich: I do think this needs to be changed across the board, especially in the modeling for the trait. IDE completions are going to let developers try to populate the field that will then fail in validation.

smithy-model/src/main/java/software/amazon/smithy/model/transform/plugins/CleanResourceLifecycleReferences.java
kstich: I'd counter with I think both of them should be cleaned up (`@references` in another PR). This keeps models correct after transformation for other uses, like client generation, when something like an in-development resource property is removed.

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/ResourceLifecycleResolver.java
kstich: Not sure what makes it complicated, they're not cross dependent on each other or any of the other inner components. Same for `PathResult`.

docs/source-2.0/spec/resource-traits.rst
kstich: For this, I meant making the `createsResources` example a `RestoreTable` operation as that creates a table. That meant keeping `deletesResources`'s example as the batch.

sugmanue: done!

kstich (approved):

PR #3277, opened 2026-09-08
Expand unstable trait for preview usage

joewyz:
#### Background
This PR extends the current `@unstable` trait to allow customers to annotate shapes as preview features.
The change has three parts:

1. Modeling. Expands the existing `@unstable` trait with an optional `featureId` member (converting it from an annotation trait to a structure trait. Bare @unstable is unchanged and fully backward-compatible),and adds a companion `@unstableFeatures` trait on the service that maps each `featureId` to its `message` and `reason`. Storing the metadata once per feature and referencing it by featureId avoids duplicating message/reason across every shape in a feature.
2. Validation. Adds `UnstableFeaturesValidator` (every referenced featureId must be defined on each enclosing service; defined features should be referenced; preview features must not be nested) backed by a new `UnstableFeatureIndex` knowledge index that resolves which preview feature a shape belongs to, excluding shapes also reachable from stable (non-preview) usage. 
3. Smithy Diff. Introduces `DiffEventDecorator` (an SPI-loaded, diff-aware event transformer that receives both models via Differences) and `PreviewFeatureDiffEventDecorator`, which downgrades ERROR/DANGER diff events to WARNING for shapes in a PREVIEW closure. 

---


smithy-diff/src/main/java/software/amazon/smithy/diff/PreviewFeatureDiffEventDecorator.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private static Shape resolvePreviewOwner(Model model, ShapeId shapeId) {
+        UnstableFeatureIndex index = UnstableFeatureIndex.of(model);
+        if (!index.isInPreviewClosure(shapeId)) {
+            return null;
+        }
+
+        // isInPreviewClosure only resolves for a shape with exactly one owner, so this owner is unambiguous.
+        Set<ShapeId> owners = index.getFeatureOwners(shapeId);
+        return model.getShape(owners.iterator().next()).orElse(null);
```
JordonPhillips, line 97: Why does it matter if the feature is part of more than once service? As long as it says preview everywhere it should be good no?

JordonPhillips (approved):

smithy-diff/src/main/java/software/amazon/smithy/diff/PreviewFeatureDiffEventDecorator.java
joewyz: this is getting the shapes which contains unstable trait including the service shape, we prevent people using nested unstable trait with featureid member. And you are right that if one service is preview then everywhere else is preview too

Tongs2000: I think this is conflating two separate concerns, and the nesting should be handled as a separate validator rather than here.

The key question for a shape reachable from multiple preview owners is whether those owners resolve to the same feature, not how many owner shapes there are. Counting owner shapes rejects a valid case where multiple operations in the same feature share a downstream structure, e.g.

```smithy
@unstable(featureId: "F")
operation A { input := { x: Shared } }

@unstable(featureId: "F")
operation B { input := { y: Shared } }

structure Shared { z: String }
// owners = {A, B} -> size 2 -> not treated as preview
```
Maybe we should dedupe by resolved `featureId` (or `UnstableFeatureInfo`) instead of raw owner count.

smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```diff
@@ -1059,8 +1059,48 @@ structure sparse {}
...
-@trait
-structure unstable {}
+@trait(
+    breakingChanges: [
+        {
+            path: "/featureId"
+            change: "add"
+            message: "The `@unstable` trait's `featureId` cannot be added to an existing shape. A preview feature must be declared when the shape is first introduced; `@unstable` with a `featureId` cannot be applied retroactively to an existing shape."
+        }
+    ]
```
Tongs2000, line 1074: Adding `breakingChanges` will ignore the default strategies including `update` and `remove`. Please check if they need to add here as well.

smithy-diff/src/main/java/software/amazon/smithy/diff/PreviewFeatureDiffEventDecorator.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>An event is downgraded only when its shape is owned by a preview feature (resolved via
+ * {@link UnstableFeatureIndex}) in both models it appears in.
+ */
+public final class PreviewFeatureDiffEventDecorator implements DiffEventDecorator {
+
+    // Event-id prefixes for changes that make a member non-nullable will not be downgraded, these two can cause
+    private static final Set<String> NULLABILITY_EVENT_ID_PREFIXES =
+            SetUtils.of("ChangedNullability", "AddedRequiredMember");
+
+    //These misuses of the trait itself must stay blocking rather than be downgraded as a preview change.
```
sugmanue, line 30:
```suggestion
    // These misuses of the trait itself must stay blocking rather than be downgraded as a preview change.
```

smithy-diff/src/main/java/software/amazon/smithy/diff/PreviewFeatureDiffEventDecorator.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Downgrades backward-incompatible diff events from ERROR or DANGER to WARNING when they occur on a shape
+ * governed by a preview unstable feature.
+ *
+ * <p>An event is downgraded only when its shape is owned by a preview feature (resolved via
+ * {@link UnstableFeatureIndex}) in both models it appears in.
+ */
+public final class PreviewFeatureDiffEventDecorator implements DiffEventDecorator {
+
+    // Event-id prefixes for changes that make a member non-nullable will not be downgraded, these two can cause
```
sugmanue, line 26: is this comment truncated? The "the two can cause" seems so, what these two can cause?

smithy-model/src/main/java/software/amazon/smithy/model/traits/UnstableFeaturesTrait.java
```diff
@@ -0,0 +1,203 @@
...
+
+        /**
+         * Creates an {@code UnstableReason} from a node.
+         *
+         * @param node Node to parse.
+         * @return Returns the parsed reason.
+         */
+        public static UnstableReason fromNode(Node node) {
+            return UnstableReason.valueOf(node.expectStringNode().expectOneOf(NAMES));
+        }
```
sugmanue, line 92: The traverse is going to be done either way, you can avoid the static allocation + the array allocation of calling `expectOneOf(String...args)` by just inlining the logic here.
```suggestion
       public static UnstableReason fromNode(Node node) {
           StringNode string = node.expectStringNode();
           for (UnstableReason reason : values()) {
               if (reason.name().equals(string.getValue())) {
                   return reason;
               }
           }
           throw new ExpectationNotMetException(
                   "Expected one of " + Arrays.toString(values()) + "; got `" + string.getValue() + "`.",
                   string);
       }
```

smithy-model/src/main/java/software/amazon/smithy/model/traits/UnstableFeaturesTrait.java
```diff
@@ -0,0 +1,203 @@
...
+
+        @Override
+        public Node toNode() {
+            return Node.from(toString());
```
sugmanue, line 96: Explicitly use name instead of relying on the `toString()` default implementation.
```suggestion
            return Node.from(name());
```

smithy-model/src/main/java/software/amazon/smithy/model/validation/validators/UnstableFeaturesValidator.java
```diff
@@ -0,0 +1,151 @@
...
+                                    featureId),
+                            UNREFERENCED_FEATURE));
+                }
+            }
+        }
+    }
+
+    private static boolean definesFeature(Model model, ShapeId serviceId, String featureId) {
+        return model.getShape(serviceId)
+                .flatMap(service -> service.getTrait(UnstableFeaturesTrait.class))
```
sugmanue, line 147: Avoid using `getTrait(UnstableFeaturesTrait.class)` since it does a linear scan, Prefer `findTrait(ID)` that's just a map lookup. Same applies to the `getTrait(...)` calls in `UnstableFeatureIndex.resolveFeatures`.

sugmanue (approved):

PR #3290, opened 2026-09-17
Make inline collection synthetic names injective

sugmanue:
## Description

Inline collection declarations (`[String]`, `{String: String}`) desugar into synthetic `list`/`map` shapes whose names were derived from the target's *simple name* only. Because a synthetic shape is created in the declaring structure's namespace, two inline collections whose element targets share a simple name but resolve to different namespaces produced the same synthetic shape ID. For example, in one namespace `[audit.foo#Widget]` and `[audit.bar#Widget]` both generated `_SyntheticListOfWidget`; the loader deduplicated on that name alone, so the second declaration silently reused the first shape and its element type was replaced by a shape from a different namespace.

This is reachable for serviceless shape models: the only place Smithy enforces case-insensitively unique simple names is within a service closure, so a model with no shared service can legitimately contain two `Widget` shapes in different namespaces.

This change derives the synthetic name from each *resolved* target relative to the declaring namespace, and deduplicates on the resolved element types rather than the generated name.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Part of the in-progress Smithy IDL 2.1 work (targets the `idl-2.1` branch)

## Naming scheme

Each target is encoded into a token, decided purely from the resolved target's namespace and simple name (independent of what else exists in the model or prelude, so a name does not change if a shape is later added to the prelude):

- Simple name starting with `_`: flattened-namespace form `_seg1_seg2_Name`. This covers nested synthetic shapes (whose names always start with `_`) and underscore-prefixed user shapes, keeping them from fusing with the markers below.
- Prelude target (`smithy.api`): `__Name`.
- Same namespace as the declaring structure: `_Name`.
- Any other namespace: `_seg1_seg2_Name`.

Lists are `_SyntheticListOf<token>`; maps are `_SyntheticMapOf<keyToken>_To_<valueToken>`.

The names are deterministic and order-independent (derived from resolved content, not declaration order), and are stable across Smithy versions.

## Design decisions and tradeoffs

- Readability over provable injectivity. An earlier draft used underscore escaping and length prefixes to be provably injective, but the names were long and unreadable, especially when nested. Since synthetic names appear only in the AST and semantic model (never in authored IDL), a readable content-derived scheme was preferred.
- The scheme is injective for all realistic inputs but not by construction. A pathological combination of namespaces and underscores can still collide (for example `com.amazon#String` and `com#amazon_String` both flatten to `com_amazon_String`).
- To keep that residual safe, the loader deduplicates on the resolved element signature and emits a validation error if two different collections ever map to the same synthetic name, rather than silently reusing the first shape. We accept the risk of that rare collision because it surfaces as a clear error, not silent corruption.

## Corner cases

- Same-namespace shape named `_String` versus prelude `String`: resolved by the leading-underscore rule (the local one takes the flattened-namespace form), so they do not collide.
- Flatten collision (`com.amazon#String` versus `com#amazon_String`): reported as a validation error.
- Nested collections: the inner synthetic shape is same-namespace but its name starts with `_`, so it consistently uses the flattened-namespace form.

## Changes

- `LoaderUtils`: `listName`/`mapName` take the declaring namespace and resolved target shape IDs and apply the encoding above.
- `IdlModelLoader`: resolves inline targets before naming, deduplicates on the resolved element signature, and reports a validation error on a genuine name collision instead of silently reusing a shape.
- Updated the inline-collection loader fixtures, the `smithy-diff` synthetic event expectations, and `SyntheticShapePreludeTest` to the new names.
- Kept the spec (`idl.rst`) format-agnostic about the exact encoding; documented the precise scheme and the collision behavior in the design doc.

## Testing

- `SyntheticShapeNamingTest`: the prelude / same-namespace / foreign forms, prelude-vs-local divergence, leading-underscore falling back to the flattened form, and distinct namespaces with the same simple name producing distinct names.
- `InlineCollectionInjectiveNamingTest`: two same-simple-name targets from different namespaces produce two distinct shapes that each keep their own element type; identical targets share one shape; a clean load produces no errors; and a genuine name collision is reported rather than silently reused.
- `./gradlew :smithy-model:test :smithy-diff:test` -- BUILD SUCCESSFUL.


smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
JordonPhillips, line 901: registerSyntheticShape reports the collision, but we return the winning synthetic ID either way, so the losing member still targets the other collection's shape. ModelAssembler returns the model when error events exist and skips the remaining validators, so a caller reading getResult() without inspecting events gets a list whose element type came from the wrong namespace. That's probably fine since it's an error and you kinda get what you get when doing that.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
```diff
@@ -102,7 +102,10 @@ final class IdlModelLoader {
     private final String filename;
     private final IdlInternalTokenizer tokenizer;
     private final Map<String, ShapeId> useShapes = new HashMap<>();
-    private final Set<ShapeId> emittedSyntheticShapes = new HashSet<>();
+    // Maps a generated synthetic shape ID to the canonical signature of the collection it was
+    // created from, so an identical collection reuses the shape and a different collection that
+    // somehow maps to the same ID is reported rather than silently overwriting the first.
+    private final Map<ShapeId, String> emittedSyntheticShapes = new HashMap<>();
```
JordonPhillips, line 108: This map is an instance field and the loader is constructed per file, so the guard's scope is one file, not one namespace. Two collections in two files of the same namespace never meet here and fall through to onShapeConflict instead, giving a different error for the same mistake.

smithy-model/src/main/java/software/amazon/smithy/model/loader/LoaderUtils.java
```diff
@@ -170,26 +170,73 @@ static ModelSyntaxException idlSyntaxError(ShapeId shape, String message, Source
...
+     *   <li>Prelude target ({@code smithy.api}): {@code __Name}.</li>
+     *   <li>Same namespace as the declaring structure: {@code _Name}.</li>
+     *   <li>Any other namespace: {@code _seg1_seg2_Name}.</li>
+     * </ul>
+     *
+     * <p>The encoding is deterministic and injective for all but pathological namespace/underscore
+     * constructions (e.g. {@code com.amazon#String} versus {@code com#amazon_String}). Those are
+     * detected by the loader and reported as an error rather than silently reused.
+     */
+    private static String syntheticToken(ShapeId target, String containingNamespace) {
```
JordonPhillips: nit: the param order here is inconsistent with listName and mapName

smithy-model/src/test/java/software/amazon/smithy/model/loader/InlineCollectionInjectiveNamingTest.java
```diff
@@ -0,0 +1,143 @@
...
+        assertThat(fooTarget, is(not(barTarget)));
+
+        ShapeId fooElement = result.expectShape(fooTarget, ListShape.class).getMember().getTarget();
+        ShapeId barElement = result.expectShape(barTarget, ListShape.class).getMember().getTarget();
+        assertThat(fooElement, equalTo(ShapeId.from("audit.foo#Widget")));
+        assertThat(barElement, equalTo(ShapeId.from("audit.bar#Widget")));
+    }
+
+    @Test
+    public void identicalTargetsShareASingleSyntheticShape() {
```
JordonPhillips, line 72: This asserts only that the two member targets are equal, so it would still pass if every inline collection collapsed onto a single shape. Asserting the shape ID and its element target would make it meaningful.

JordonPhillips (changes requested):

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
sugmanue: You're absolutely right. I will remove this logic and instead let the existing loader validation mechanism kick, the error won't be as terse but it will be precise.

smithy-model/src/main/java/software/amazon/smithy/model/loader/IdlModelLoader.java
sugmanue: I think it's fine. Let's leave it as-is.

smithy-model/src/test/java/software/amazon/smithy/model/loader/InlineCollectionInjectiveNamingTest.java
sugmanue: Fixed.

JordonPhillips (approved):

